PR #14, opened 2017-06-26
Add option to ignore SSL certs

### LINKED ISSUE #13: Conditionally allow insecure SSL for local environments
Local environments use the snakeoil SSL cert which don't validate, so we need an option to turn this off locally.

rhysyngsun:
Fixes #13 

#### What's this PR do?
Adds a flag to ignore the SSL certificate validation for API requests

```python
from themes.api import Api()

api = Api()
api.list_themes().next()
api.user = True  # hacky way to fake a user
api.create_theme('test1', 'Test 1')
```


blarghmatey:
:+1:

PR #35, opened 2017-07-13
Fixed API tests

noisecapella:
None

#### What's this PR do?
Fixes asserts in `themes/api_test.py` and fixes a bug in `list_themes`. The `assert mock.called_once_with(...)` syntax isn't valid since `called_once_with` isn't a method, so it evaluates to a mock which is always true.

`list_themes` should return the right results for anonymous users


noisecapella:
Just pushed a commit to add a few extra asserts

Issue #29, opened 2017-07-06
Add Rest APIs for channels, posts


noisecapella:

noisecapella:
Here's the API outline for channels:

    GET api/v0/channels - returns list of channels the user is subscribed to. Just list of detail views
    GET api/v0/channels/{subreddit}/
    {
        "name": "subreddit", # this is a unique identifier, what comes after /r/
        "title": "Title of the subreddit",
        "theme_type": "private" or "public",
    }
    
    POST api/v0/channels
    {
        "name": "subreddit",
        "title": "Title",
        "theme_type": "private" or "public"
    }
    
    PATCH api/v0/channels/{subreddit}/
    {
        "title": "Title",
        "theme_type": "private" or "public",
    }
    
    # No DELETE, not sure if reddit offers this

noisecapella:
The API outline for posts:

    
    POST api/v0/channels/{subreddit}/posts/
    {
      "title": "post title",
      "text": "text",  # only one of these two should be present
      "url": "url",
    }
    
    Returns same as GET
    
    GET api/v0/channels/{subreddit}/posts/
    Just a list of posts, same format as detail view
    
    
    GET api/v0/channels/{subreddit}/posts/{post_id}/
    {
        "id": "post_id",
        "title": "post title",
        "upvotes": upvotes,
        "downvotes": downvotes,
        "author": "reddit username goes here",
        "text": "text",  # Only one of these should be present
        "url": "url",
    }
    
    PATCH api/v0/channels/{subreddit}/posts/{post_id}/
    Only `text` field allowed to replace text context. No URL editing allowed. Return value is `{ "text": "your text here" }`
    POST api/v0/channels/{subreddit}/posts/{post_id}/upvote
    Returns new upvotes, downvotes values in dict as { upvotes: 456, downvotes: 123 }
    POST api/v0/channels/{subreddit}/posts/{post_id}/downvote
    Returns new upvotes, downvotes values in dict as { upvotes: 456, downvotes: 123 }

    DELETE api/v0/channels/{subreddit}/posts/{post_id}/
    Deletes a post

aliceriot:
How do we deal with fetching the comments? I was just thinking about this and realized I hadn't seen anything about it yet.

PR #39, opened 2017-07-14
REST API for channels

noisecapella:
Part of #29 

#### What's this PR do?
Implements rest API for channels

You can go to `/api/v0/channels/` and play around with it using the DRF HTML view. You should also be able to go to `/api/v0/channels/<subreddit>/` to view a particular channel or to patch it

### DIFF
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -20,8 +20,7 @@ class ChannelSerializer(serializers.Serializer):
     )
 
     def create(self, validated_data):
-        api = Api()
-        api.user = True
+        api = Api(user=self.context['request'].user)
         return api.create_channel(
             name=validated_data['display_name'],
             title=validated_data['title'],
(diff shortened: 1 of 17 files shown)


channels/serializers.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+
+    def update(self, instance, validated_data):
+        api = Api()
+        api.user = True
+
+        name = instance.display_name
+        key_map = {
+            'title': 'title',
+            'channel_type': 'subreddit_type',
```
rhysyngsun: The arg for `update_channel` is actually also `channel_type` so this mapping is unnecessary and actually causing updates to `channel_type` to not work. `update_channel validates it's `kwargs` so you could probably just catch the `ValueError` and reraise as a drf `ValidationError`.

rhysyngsun: I tried fixing this locally and still didn't see the channel_type update so something else is going on here too.

noisecapella: I see a problem too, I'll look into it

noisecapella:
Should be ready for review again. I also rebased on #44

PR #40, opened 2017-07-18
Bumped deps and added homepage route

rhysyngsun:
N/A

#### What's this PR do?
Bumps JS dependencies. Specifically, it upgrades to react-router 4.x and implements a barebones home page.

Open http://localhost:8063/ and verify you see the text "Home page"


aliceriot:
functionality looks good! reading code now.

static/js/reducers/posts.js
```diff
@@ -0,0 +1,15 @@
+// @flow=
```
aliceriot: are we planning to use `redux-hammock` for hitting these APIs? If so, we wont need these reducers.

static/js/containers/App.js
```diff
@@ -1,38 +1,16 @@
...
   render() {
-    const { checked } = this.props.checkbox;
-    return <div>
-      <div id="app-body">
-        Click the checkbox:
-        <input type="checkbox" checked={checked} onClick={this.handleClick.bind(this)} />
+    const { match } = this.props;
+    return (
+      <div>
+        <Route exact path={match.url} component={HomePage}/>
```
aliceriot: I'm reading through the v4 documentation but I'm not sure I totally get the whole thing yet. That said, why do we use `match.url` for the path here, instead of `"/"`?

static/js/util/withTracker.js
```diff
@@ -0,0 +1,31 @@
...
+const withTracker = (WrappedComponent: Class<React.Component<*, *, *>>) => {
+  const debug = SETTINGS.reactGaDebug === "true";
+
+  if (SETTINGS.gaTrackingID) {
+    ga.initialize(SETTINGS.gaTrackingID, { debug: debug });
+  }
+
+  const trackPage = (page: string) => {
+    ga.pageview(page);
+  };
```
aliceriot: is this function necessary for some reason? why not just call `ga.pageview` below, on line 21?

static/js/reducers/posts.js
rhysyngsun: Yeah, but I'm not sure how to configure that. Should I just nix these here then? I'm currently using these as a hack for dummy data on my other branch.

static/js/containers/App.js
rhysyngsun: `match.url` allows for relative nesting without having to care what's above you. Maybe we don't need this for `App`, but it would allow for us to mount `App`  in MM like `<Route path="/discussions" component={discussions.App}/>` or something like that without having to modify `App`. I think that's generally a good approach for keeping these components reusable without assumptions of where they live relative to the root.

static/js/util/withTracker.js
rhysyngsun: Sure, that's a holdover from the original, which was doing a bit more in here.

static/js/containers/App.js
aliceriot: gotcha, makes sense.

static/js/reducers/posts.js
aliceriot: If they're just to provide dummy data I think that's fine for now, as the APIs get into place we can quickly switch over to using `hammock`.

also if you want to read about it there's [documentation](https://github.com/mitodl/redux-hammock) now!

rhysyngsun: ok, i'll leave them for now

PR #43, opened 2017-07-19
add a command for running JS tests in watch mode

aliceriot:

none

#### What's this PR do?

just adds tooling to run mocha in 'watch' mode, which re-runs the tests whenever a file is touched. very handy!

unfortunately mocha doesn't have the chops to figure out which files are relevant to re-run if a particular file is changed (we'd need to switch to a fancy, newer runner like AVA to get that) but this is still pretty nice :)

running tests in the normal way should still work, but you should also be able to do `npm run watch` instead of `npm run test`, and it should run the tests in watch mode.

### DIFF
--- a/scripts/test/js_test.sh
+++ b/scripts/test/js_test.sh
@@ -7,6 +7,9 @@ then
 elif [[ ! -z "$CODECOV" ]]
 then
     export CMD="node ./node_modules/nyc/bin/nyc.js --reporter=lcovonly -R spec mocha"
+elif [[ ! -z "$WATCH" ]]
+then
+    export CMD="node ./node_modules/mocha/bin/_mocha --watch"
 else
     export CMD="node ./node_modules/mocha/bin/_mocha"
 fi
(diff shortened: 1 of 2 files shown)


scripts/test/js_test.sh
```diff
@@ -7,6 +7,9 @@ then
 elif [[ ! -z "$CODECOV" ]]
 then
     export CMD="node ./node_modules/nyc/bin/nyc.js --reporter=lcovonly -R spec mocha"
+elif [[ ! -z "$WATCH" ]]
+then
+    export CMD="node ./node_modules/.bin/mocha --watch"
```
gsidebo: seems odd that the base command here is different from the one below (`node ./node_modules/.bin/mocha` vs. `node ./node_modules/.bin/mocha`). any reason for that difference?

aliceriot: `.bin` is just a directory for symlinks to executables, anything in `${package}/bin` will get symlinked there. if we want it to be consistent I can change it, I forgot that we usually specify the full path.

gsidebo: yeah i think those should be consistent and the presence of `$WATCH` should just tack on the `--watch` parameter at the end of the command

PR #47, opened 2017-07-19
Implement UI for channel page

### LINKED ISSUE #26: As a logged in user, I would like to view a channel page

rhysyngsun:
Part of #26 

#### What's this PR do?
Adds the channel listing page.

Go to /channels/something and verify you see the page as shown in the screenshot.
Also go to `/api/v0/channels` and `/api/v0/channels/pics` and verify you see `public_description` in there now. it will probably be empty strings unless you've set it on a sub.

selection_077

### DIFF
--- a/static/scss/sidebar.scss
+++ b/static/scss/sidebar.scss
@@ -1,0 +1,18 @@
+.sidebar {
+  margin: 20px;
+
+  .new-post {
+    width: 100%;
+    margin: 10px 0;
+    text-align: center;
+    font-size: 1.5em;
+  }
+
+  .user-count {
+    color: $font-grey;
+  }
+
+  .description {
+    padding-bottom: 10px;
+  }
+}
--- a/cassettes/channels.views_test.test_create_channel.json
+++ b/cassettes/channels.views_test.test_create_channel.json
@@ -87,7 +87,7 @@
       "request": {
         "body": {
           "encoding": "utf-8",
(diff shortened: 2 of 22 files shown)


aliceriot:
If I go to `/channels/foobar` or `/channels/test_channel` I just see a blank page. It looks like the `Route` component is rendering `null` for it's children somehow?

nullnullwhy

aliceriot:
I also don't seem to be able to visit the API pages (like `/api/v0/channels/pics`), I just get an empty React app at that URL.

rhysyngsun:
@aliceriot I've fixed those. One was a regression with how I dealt with urls.py, the other was the new `<Loading/>` I wrote in this PR was expecting the `posts` reducer to set `loaded = true` which it never did. I removed that from the list for now.

aliceriot:
API stuff appears to be working! although we should probably note down the quirk that new users have to take some sort of 'action' (subscribing to a subreddit worked for me) before their subscriptions show up in the API.

aliceriot:
UI shows up too

rhysyngsun:
@aliceriot I'll file a bug for the subreddit list not showing up right away. Might be some user initialization that we need to add to the refresh_token api. CC @noisecapella

static/scss/sidebar.scss
```diff
@@ -0,0 +1,20 @@
+
+
```
aliceriot: blank lines

rhysyngsun: :trollface:

Issue #15, opened 2017-06-30
As a logged in user, I would like to be able to see the front page


pdpinch:
Simple listing of posts in the front page

pdpinch:
What's the status of this?

rhysyngsun:
@pdpinch you can view the frontpage (only a list of hot posts) and navigation to individual posts. If that's enough for this ticket it can be closed.

PR #48, opened 2017-07-19
set up basic front-end data model

aliceriot:

none, I did this as initial work for starting on #15 and figured I might as well open as a separate PR.

#### What's this PR do?

This sets up reducers for posts, channels, posts-for-channels, and the frontpage. there are also random data generators for relevant things, for testing and development.

tests should pass, code should be reasonably clean.

### DIFF
--- a/static/js/factories/posts.js
+++ b/static/js/factories/posts.js
@@ -1,0 +1,21 @@
+// @flow
+import R from 'ramda';
+import casual from 'casual-browserify';
+
+import type { Post } from '../flow/discussionTypes';
+
+export const makePost = (isURLPost: boolean = false): Post => ({
+  id: String(Math.random()),
+  title: casual.sentence,
+  upvotes: Math.round(Math.random() * 15),
+  downvotes: 0,
+  upvoted: Math.random() < 0.5,
+  downvoted: Math.random() < 0.5,
+  author: `justareddituser${String(Math.random())}`,
+  text: isURLPost ? null : casual.text,
+  url: isURLPost ? casual.url : null,
+});
+
+export const makeChannelPostList = () => (
+  R.range(0, 20).map(() => makePost(Math.random() > .5))
+);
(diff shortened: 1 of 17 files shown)


static/js/factories/posts_test.js
```diff
@@ -0,0 +1,62 @@
...
+      let upvoteScores = R.range(1,20)
+        .map(makePost)
+        .map(R.prop('upvotes'));
+      assert.isAbove(new Set(upvoteScores).size, 1);
+    });
+
+    it('should randomly generate id', () => {
+      let firstPost = makePost();
+      let secondPost = makePost();
+      assert.notEqual(firstPost.id, secondPost.id);
```
rhysyngsun: This has a probability of failing, and I'm not certain there's a value in asserting the uniqueness of this from a UI perspective, is there a particular case you're concerned about here?

static/js/factories/posts.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+import R from 'ramda';
+
+import type { Post } from '../flow/discussionTypes';
+
+export const makePost = (isURLPost: boolean = false): Post => ({
```
rhysyngsun: Curious if you think using something like [casual](https://github.com/boo1ean/casual) would be useful here since we do a similar thing with `factoryboy` in python?

aliceriot: would probably be handy! I wrote a `randomString` function which is probably not the best thing ever, haha.

static/js/factories/posts_test.js
aliceriot: just more of a smoke test, I could remove the test if it's not useful.

rhysyngsun: My concern more is that it could fail at an inopportune time (e.g. merge commit in `release` branch). We've been bit a few times by this type of test.

aliceriot: ok I will remove

aliceriot:
@rhysyngsun comments addressed

Issue #27, opened 2017-07-05
As a logged in user I would like to view all comments of a particular post


Ferdi:

pdpinch:
So the API is done for this but @aliceriot is still working on the front-end?

aliceriot:
yeah that's right

pdpinch:
Is this done now that #79 has merged?

gsidebo:
@pdpinch That PR is showing generated comments. I'm guessing this issue covers the work to fetch comments via API and render them in the same way

rhysyngsun:
@aliceriot merged her PR for the API integration last night, so this should be considered done

PR #60, opened 2017-07-25
add post detail display

aliceriot:

this is part of #27 

#### What's this PR do?

this PR basically implements the front-end UI for a 'detail' view of a `Post`. This includes a container and new route for that, a little bit of logic around when to fetch the relevant `Channel` and `Post` objects, and so on. I also updated the `Post` factory to reflect the latest thinking about what the API will look like.

From any thing which shows a list of posts you should be able to click on the `comments` link to get to a detail view for the post. If it is a text post, you should see something like this:

textpost

If it's a URL post you should see basically the same thing, just without the blue text box.

You should be able to click around to the frontpage, to the subreddit, etc, and go to other posts (without a full-page refresh) with no problem.

Visiting a post URL directly (e.g. http://localhost:8063/channel/voluptatum/0.1173598155720883) should work as well.


noisecapella:
If I go to `/` and click one of the posts, I see a page that says `Loading` but nothing more. If I then refresh the page I can see the detail view however

aliceriot:
wuuuut it definitely wasn't doing that before. looking into it.

static/js/components/Loading.js
```diff
@@ -4,8 +4,8 @@ import R from 'ramda';
 
 import type { RestState } from '../flow/restTypes';
 
-const anyProcessing = R.any(R.propEq('processing', true));
-const allLoaded = R.all(R.propEq('loaded', true));
+export const anyProcessing = R.any(R.propEq('processing', true));
```
noisecapella: Should these be moved to a utility module instead?

aliceriot:
Ah, found the problem. When you hit hte frontpage or a subreddit the `frontpage` and `postsForChannel` functions return posts, which we set into the `posts` reducer using the `setPost` custom action. The problem is that this action doesn't set `loaded` to be `true` - I was not seeing that behavior because I was starting my navigation on a post detail view, in which case the `actions.posts.get` function has been called, and `loaded` has already been set to `true` in the posts reducer.

anyway, I'll change the action to set that.

aliceriot:
ok that should be fixed

static/js/containers/PostPage.js
```diff
@@ -0,0 +1,89 @@
...
+  getMatchParams = () => {
+    const { match: { params }} = this.props;
+    return [params.postID, params.channelName];
+  }
+
+  updateRequirements = () => {
+    const { dispatch, channels, posts } = this.props;
+    const [postID, channelName] = this.getMatchParams();
+
+    if (R.isNil(posts.data.get(postID))) {
```
noisecapella: Do we need these checks here? It seems like we only get to this function if the request already completed successfully

static/js/containers/PostPage.js
```diff
@@ -0,0 +1,89 @@
...
+    if (!anyProcessing(restStates) || allLoaded(restStates)) {
+      this.updateRequirements();
+    }
+  }
+
+  renderContents(postID, posts, channelName, channels) {
+    const post = posts.get(postID);
+    const channel = channels.get(channelName);
+
+    if (R.isNil(channel) || R.isNil(post)) {
```
noisecapella: Is this check necessary? It seems like `renderContents` only gets called if all requests have finished

static/js/containers/PostPage.js
aliceriot: We will only get to this function if a request for the `Post` API has already been made. The trick thing is that the request in question could have been for a *different* post than the one we're currently interested in. Consider this case:

- navigate to `/channel/foobar/postid`
- navigate back to `/channel/foobar`
- navigate to `/channel/foobar/anotherpostid`

at this point we have already hit the posts API once (to fetch `postid`). when we load the `PostPage` component, to try to view `anotherpostid`, we will therefore already have `loaded === true`, but we can't quite trust it. It seems, in my testing, that there is typically 1 render call on the changed component before the new request (to get `anotherpostid`) kicks off, when we will have the case that `processing` is false, `loaded` is true, and the data for `anotherpostid` hasn't yet been fetch. So I added this check just to make sure the data is actually there before we try to render anything with it.

Does that make sense? I ran into the same issue with navigating around between subreddits when I was working on the frontpage UI, so I added an equivalent check to the `ChannelPage` container then.

static/js/containers/PostPage.js
aliceriot: See the comment above

static/js/components/Loading.js
aliceriot: yeah, that makes sense. `lib/rest.js` or something.

aliceriot:
@noisecapella comments addressed

static/js/containers/PostPage.js
noisecapella: Maybe we should clear the data when the post or channel unmounts, but I'm fine with this solution for now

PR #71, opened 2017-07-27
use prettier to format JavaScript code

### LINKED ISSUE #64: consider using Prettier for JavaScript formatting
[prettier](https://prettier.io/) is an automated code-formatting tool for JS that looks pretty nice, basically it parses the JavaScript to an AST and then re-writes it all, allowing it to do more complicated types of automated transformations that ESLint, for instance, can't do in an automatic way. It seems like there's support to run it alongside ESLint, since prettier doesn't catch certain kinds of semantic errors in the code (which ESLint does).

anyway, I think it looks interesting and would like to try it out! @noisecapella @rhysyngsun thoughts?

aliceriot:

closes #64 

#### What's this PR do?

This installs [prettier-eslint](https://github.com/prettier/prettier-eslint-cli), which is a little wrapper than basically does

`prettier` ➡️`eslint --fix`

I added a command to `package.json` which runs the formatter (`npm run fmt`). It's also handy to configure your editor to run it, either on save or with a key command.

If all the tests are passing and all that I think we're good.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,6 +6,16 @@
       "plugin:react/recommended"
   ],
   "rules": {
+    "key-spacing": [2, {
+      "beforeColon": false,
+      "afterColon": true,
+      "align": {
+        "beforeColon": false,
+        "afterColon": true,
+        "mode": "minimum",
+        "on": "value"
+      }
+    }],
     "quotes": [0], // no opinion on ' vs "
     "object-curly-spacing": [0],
     "new-cap": [0], // allows function calls like Immutable.Map(...)
@@ -25,7 +35,7 @@
     }],
     "comma-dangle": [0],
     "no-unreachable": [2],
-    "semi": [2, "always"],
+    "semi": [2, "never"],
     "eqeqeq": [2],
     "no-var": [2],
     "camelcase": [2, {
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "nyc": "^11.0.3",
     "object.entries": "^1.0.4",
     "postcss-loader": "^2.0.6",
+    "prettier-eslint-cli": "^4.1.1",
     "prop-types": "^15.5.10",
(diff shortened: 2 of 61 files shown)


aliceriot:
it's possible to exclude things with a comment, so we could disable it for multi-line readability and things like that. I looked into the visual index for (flow definitions, props, etc) and I didn't find an easy way to do that :(

aliceriot:
bah so there's an `eslint` rule (called `key-spacing`) which allows you to tell eslint to consistently space object literals, but flow types are not technically object literals, so it doesn't apply there. I've looked around a little, but I can't seem to find a way to require it to consistently space flow object types :(

I like having consistently spaced flow types, but it's not a make or break thing for me. @rhysyngsun what do you think?

rhysyngsun:
@aliceriot I think the visual indent is a bit of a productivity enhancer for me, but I'm not married to it.

aliceriot:
we can still do it with `prettier-eslint`, we just have to put `// prettier-ignore` on the line right above the type each time - I feel like I should be able to pass an 'ignore AST nodes matching this regex' kind of option but it doesn't seem to support that

rhysyngsun:
does `// prettier-ignore` work at the file level or can we configure the plugin to ignore our `flow` directory and be strict about putting typedefs in there?

aliceriot:
`// prettier-ignore` works at the AST node level, so we can do something like this:

```js
// @flow

// prettier-ignore
export type Post = {
  nicely:     "aligned",
  properties: "here",
}
```

aliceriot:
we do pass a blob to the `prettier` CLI to figure out which files to reformat, so we could definitely just have it not worry about the flow definitions

noisecapella:
Only `.eslintrc` and `package.json` changed, right? The rest of the lines are just reformatting?

aliceriot:
yep!

PR #73, opened 2017-07-27
Add page to create channel

### LINKED ISSUE #16: As a logged in user I would like to be able to create a channel
The most minimal form to create a channel.  

- [x] Implement backend API (PR #39)
- [ ] Implement frontend UI
- [ ] Integrate frontend UI with API

rhysyngsun:
Fixes #16 

#### What's this PR do?
Adds a page to create a channel

Go to http://localhost:8063/manage/channel/new and fill in the form. Verify you're redirected to your new channel.

No validation yet

### DIFF
--- a/static/js/components/admin/ChannelEditForm.js
+++ b/static/js/components/admin/ChannelEditForm.js
@@ -1,0 +1,74 @@
+// @flow
+import React from "react"
+
+import { CHANNEL_TYPE_PUBLIC, CHANNEL_TYPE_PRIVATE } from "../../lib/channels"
+
+import type { Channel } from "../../flow/discussionTypes"
+
+export default class ChannelEditForm extends React.Component {
+  props: {
+    onSubmit: Function,
+    onUpdate: Function,
+    channel: ?Channel
+  }
+
+  render() {
+    const { onSubmit, onUpdate, channel } = this.props
+    if (channel === null || channel === undefined) {
+      return null
+    }
+    return (
+      <form onSubmit={onSubmit} className="form">
+        <div className="form-item">
+          <label htmlFor="title" className="label">
+            Title
+          </label>
+          <input name="title" type="text" className="input" value={channel.title} onChange={onUpdate} />
+        </div>
+        <div className="form-item">
+          <label htmlFor="name" className="label">
+            Name
+          </label>
+          <input name="name" type="text" className="input" value={channel.name} onChange={onUpdate} />
+        </div>
+        <div className="form-item">
+          <label htmlFor="public_description" className="label">
+            Description
+          </label>
+          <textarea
+            name="public_description"
+            type="text"
+            className="input"
+            value={channel.public_description}
+            onChange={onUpdate}
+          />
+        </div>
+        <div className="form-item">
+          <div className="label">Channel Type</div>
+          <div className="input">
+            <input
+              id="channel_public"
+              type="radio"
+              name="channel_type"
+              value={CHANNEL_TYPE_PUBLIC}
+              checked={channel.channel_type === CHANNEL_TYPE_PUBLIC}
+              onChange={onUpdate}
+            />
+            <label htmlFor="channel_public">Public</label>
+            <input
+              id="channel_private"
+              type="radio"
+              name="channel_type"
+              value={CHANNEL_TYPE_PRIVATE}
+              checked={channel.channel_type === CHANNEL_TYPE_PRIVATE}
+              onChange={onUpdate}
+            />
+            <label htmlFor="channel_private">Private</label>
+          </div>
+        </div>
+
+        <button type="submit">Save</button>
+      </form>
+    )
+  }
+}
--- a/static/js/components/admin/ChannelEditForm_test.js
+++ b/static/js/components/admin/ChannelEditForm_test.js
@@ -1,0 +1,80 @@
+// @flow
+import React from "react"
+import { assert } from "chai"
+import sinon from "sinon"
+import { shallow } from "enzyme"
+
+import ChannelEditForm from "./ChannelEditForm"
+import { CHANNEL_TYPE_PUBLIC, CHANNEL_TYPE_PRIVATE, newChannel } from "../../lib/channels"
+
+import type { Channel } from "../../flow/discussionTypes"
+
+describe("ChannelEditForm", () => {
+  const renderForm = (channel: ?Channel, { onSubmit, onUpdate } = { onSubmit: () => {}, onUpdate: () => {} }) =>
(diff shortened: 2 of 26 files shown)


static/js/containers/admin/AdminPage.js
```diff
@@ -0,0 +1,26 @@
...
+  props: {
+    match: Match,
+  }
+
+  render() {
+    const { match } = this.props;
+    return (
+      <div className="admin double-column">
+        <Card>
+          <h1>Admin</h1>
```
aliceriot: should we maybe add a `title` prop to the `Card` component? I'm thinking we may want to style that in a consistent way across the app (just occurred to me, not necessary something to do in this PR)

rhysyngsun: Yeah, I think something like that would be good eventually.

noisecapella:
After I successfully create a channel I'm redirected to the channel page but I get an error page:

screenshot from 2017-07-31 14-31-08

static/js/components/admin/ChannelEditForm.js
```diff
@@ -0,0 +1,76 @@
+// @flow
+import React from 'react';
+
+export default class ChannelEditForm extends React.Component {
+  props: {
+    onSubmit: Function,
+    onUpdate: Function,
+    channel:  ?Object,
```
noisecapella: Can you use the `Channel` type here?

static/js/components/admin/ChannelEditForm_test.js
```diff
@@ -0,0 +1,88 @@
...
+        assert.isNotOk(onSubmit.called);
+        wrapper.find('form').simulate('submit');
+        assert.isOk(onSubmit.called);
+        assert.isNotOk(onUpdate.called);
+      });
+    });
+
+    describe('onUpdate', () => {
+      for (let inputName of ['title', 'name', 'public_description']) {
+        it(`should be called when ${inputName} input is modifed`, () => {
```
noisecapella: Should be `modified`

static/js/containers/admin/AdminPage.js
```diff
@@ -0,0 +1,26 @@
...
+          <Route path={`${match.url}/channel/new`} component={CreateChannelPage}/>
```
noisecapella: Should we use the functions in `lib/url.js` instead for this kind of thing?

static/js/containers/admin/CreateChannelPage.js
```diff
@@ -0,0 +1,61 @@
...
+    }));
+  }
+
+  onSubmit = (e: Object) => {
+    const { dispatch, history, channelForm } = this.props;
+
+    e.preventDefault();
+
+    dispatch(actions.channels.post(channelForm.value)).then((channel) => {
+      history.push(`/channel/${channel.name}`);
```
noisecapella: Can we make a function in `lib/url.js` to create this URL?

static/js/lib/api.js
```diff
@@ -17,6 +18,15 @@ export function getChannel(channelName: string): Promise<Channel> {
   return fetchJSONWithCSRF(`/api/v0/channels/${channelName}/`);
 }
 
+export function createChannel(channel: Channel): Promise<Channel> {
+  return fetchJSONWithCSRF(`/api/v0/channels/`, {
+    method: POST,
+    body: JSON.stringify({
+      ...channel
```
noisecapella: Why are we using the spread syntax here?

static/js/lib/channels_test.js
```diff
@@ -0,0 +1,15 @@
...
+
+import { newChannel } from './channels';
+
+describe('Channel utils', () => {
+  it('should return a new channel with empty values and public type', () => {
+    assert.deepEqual(newChannel(), {
+      'title': '',
+      'name': '',
+      'public_description': '',
+      'channel_type': 'public',
```
noisecapella: Can we make constants for `public` and `private`?

static/js/lib/form_actions.js
```diff
@@ -0,0 +1,46 @@
...
+};
+
+export const deriveReducer = (name: string, newDefaults: () => Object): FormReducer => {
+  const actionTypes = deriveActionTypes(name);
+  return (state: FormValue = INITIAL_STATE, action: Action<any, null>): FormValue => {
+    switch(action.type) {
+    case actionTypes.reset:
+    case actionTypes.create:
+      return {
+        ...state,
```
noisecapella: What about `errors`? It seems like we should clear it here

static/js/lib/form_actions_test.js
```diff
@@ -0,0 +1,103 @@
...
+      sandbox = sinon.sandbox.create();
+      store = configureTestStore(combineReducers({ channels }));
+      dispatchThen = store.createDispatchThen(state => state.channels);
+    });
+
+    afterEach(() => {
+      sandbox.restore();
+    });
+
+    //
```
noisecapella: Can you remove this line?

static/js/lib/form_actions.js
rhysyngsun: `errors` is covered by the spread, should I get rid of the spread and just be explicit about it?

noisecapella: It's not the explicitness, I'm wondering about if we should replace `errors` here with `{}` to clear the errors

static/js/containers/admin/AdminPage.js
rhysyngsun: Not in this case because it's the route config, not a link to a page.

static/js/lib/api.js
rhysyngsun: Yeah, that's less than useful, it also isn't guarding against the other channel properties which could cause a 4xx.

static/js/lib/form_actions.js
rhysyngsun: Good point.

rhysyngsun: I won't have a good way of testing that errors gets reset until I add support for that, which will be in a separate PR, but worth fixing this issue now.

rhysyngsun:
@noisecapella the diff is probably hosed now since I rebased. Look at this commit to see what I fixed relative your feedback: https://github.com/mitodl/open-discussions/pull/73/commits/ce770eccb011bf87b3e1785b24cef3d08e837155

rhysyngsun:
@noisecapella didn't see the error until just now. This would've been caused by an error from the backend so you might have to check your network tab or backend logs for more details

noisecapella:
I just checked out the newest commit, it looks like the error is gone now

PR #79, opened 2017-07-31
basic comment display

aliceriot:

part of #27 

#### What's this PR do?

this adds a very bare-bones display for showing the comments on a post

right now the comments reducer is not yet hooked up to the API, so every post will have a randomly generated comment tree. so you should be able to just navigate to a post and see a bunch of replies of varying depth. The first comment should always have two levels of replies.

### DIFF
--- a/static/js/components/CommentTree.js
+++ b/static/js/components/CommentTree.js
@@ -1,0 +1,45 @@
+// @flow
+import React from "react"
+import R from "ramda"
+import moment from "moment"
+
+import Card from "../components/Card"
+
+import type { Comment } from "../flow/discussionTypes"
+
+const renderTopLevelComment = (comment: Comment, idx: number) =>
+  <Card key={idx}>
+    <div className="top-level-comment">
+      {renderComment(0, comment)}
+    </div>
+  </Card>
+
+const renderComment = R.curry((depth: number, comment: Comment) =>
+  <div className="comment" key={`${depth}${comment.created}`}>
+    <div className="author-info">
+      <a className="author-name">
+        {comment.author_id}
+      </a>
+      <div>
+        {moment(comment.created).fromNow()}
+      </div>
+    </div>
+    <div>
+      {comment.text}
+    </div>
+    <div className="replies">
+      {R.map(renderComment(depth + 1), comment.replies)}
+    </div>
+  </div>
+)
+
+type CommentTreeProps = {
+  comments: Array<Comment>
+}
+
+const CommentTree = ({ comments }: CommentTreeProps) =>
+  <div className="comments">
+    {R.addIndex(R.map)(renderTopLevelComment, comments)}
+  </div>
+
+export default CommentTree
--- a/static/js/containers/PostPage.js
+++ b/static/js/containers/PostPage.js
@@ -7,6 +7,7 @@ import Card from "../components/Card"
 import Loading from "../components/Loading"
 import ChannelBreadcrumbs from "../components/ChannelBreadcrumbs"
 import PostDisplay from "../components/PostDisplay"
+import CommentTree from "../components/CommentTree"
 
 import { actions } from "../actions"
 import { anyProcessing, allLoaded } from "../util/rest"
@@ -19,7 +20,8 @@ class PostPage extends React.Component {
     match: Match,
     dispatch: Dispatch,
     posts: Object,
-    channels: Object
+    channels: Object,
+    comments: Object
   }
 
   getMatchParams = () => {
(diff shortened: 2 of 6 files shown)


gsidebo:
It looks like this is rendering all comments at the same level in the HTML and indenting with different widths to represent comment hierarchy/depth. Reddit seems to render comments in a hierarchical way (i.e.: a single comment's HTML element has a sibling element that contains _all_ of its descendant comments). 

Top-level comment:
ss 2017-07-31 at 13 39 08

Descendant comments:
ss 2017-07-31 at 13 39 21

The most obvious benefit from the approach that reddit uses is that a comment can be collapsed pretty easily. Have we discussed the benefits/drawbacks for flat comment rendering vs. hierarchical?

aliceriot:
We haven't discussed this yet (to my knowledge). Flattening the comment tree into an array seemed to be the most straightforward approach to me, but I wasn't really considering that we may need to implement folding at some point. Curious what @rhysyngsun @roberthouse54 and @noisecapella think as well

gsidebo:
My instinct is that we are going to need it, and sooner rather than later. Could be wrong about that. With hierarchical rendering, we would just need to set/unset a class on the comment element being folded/unfolded. 

It looks like Disqus also renders comments hierarchically with the same basic approach (a comment element and the element containing its reply tree are siblings). I think if there's even a chance that we would want to fold comments in the future, we should go with something like this

Top-level disqus comment:
ss 2017-07-31 at 14 25 31

Disqus replies (descendant comments):
ss 2017-07-31 at 14 25 15

aliceriot:
Our current approach would allow for easily doing top-level comment folding, but if we need to fold at arbitrary points we would need to translate the comment tree more directly into HTML. I'm not sure what the requirements are for this, but I can take a stab at rendering a tree right now.

static/js/containers/PostPage.js
```diff
@@ -1,39 +1,48 @@
...
 
-import Card from '../components/Card';
-import Loading from '../components/Loading';
-import ChannelBreadcrumbs from '../components/ChannelBreadcrumbs';
-import PostDisplay from '../components/PostDisplay';
+import Card from "../components/Card";
+import Loading from "../components/Loading";
+import ChannelBreadcrumbs from "../components/ChannelBreadcrumbs";
+import PostDisplay from "../components/PostDisplay";
+import CommentTree from '../components/CommentTree';
```
gsidebo: Double-quotes seems to be the convention now

static/js/containers/PostPage.js
```diff
@@ -44,19 +53,20 @@ class PostPage extends React.Component {
...
     if (!anyProcessing(restStates) || allLoaded(restStates)) {
       this.updateRequirements();
     }
   }
 
-  renderContents(postID, posts, channelName, channels) {
+  renderContents(postID, posts, channelName, channels, comments) {
     const post = posts.get(postID);
     const channel = channels.get(channelName);
+    const commentTree = comments.get(postID);
```
gsidebo: Nitpick: Since `CommentTree` is a component, there could be some confusion with a variable named `commentTree`. I think `commentData` or `commentTreeData` would work here

aliceriot:
I updated the code to render replies as children

aliceriot:
@gsidebo feedback addressed

static/js/components/CommentTree.js
```diff
@@ -0,0 +1,47 @@
...
+        {comment.author_id}
+      </a>
+      <div>
+        {moment(comment.created).fromNow()}
+      </div>
+    </div>
+    <div>
+      {comment.text}
+    </div>
+    { R.map(renderComment(depth + 1), comment.replies) }
```
gsidebo: With the current approach, every comment element apart from the top-level comments has this 'indented' class added to it. If we wrap all of a comments replies into some parent element (instead of rendering them in succession as we're doing here) we can take care of the indent for all reply elements by adding the indent on that parent element. So we could do something like...

```javascript
<div className="replies">
  { R.map(renderComment(depth + 1), comment.replies) }
</div>
```

... and the indentation could be defined for `.replies`. We wouldn't need the 'indented' class at all, so we wouldn't need to calculate the class name or repeat it for a large number of elements

aliceriot: I prefer the current approach I think, because I'd prefer for a comment which does not have any replies to not have an empty `.replies` div in it - in order to get around that I'd have to add another helper function or something which adds more complexity than what I have now, which is just a very simple recursive function.

aliceriot: nvm, thought about it a bit more and I'm fine with the empty `.replies`, just implemented that.

aliceriot:
@gsidebo comments addressed!

Issue #18, opened 2017-06-30
As a logged in user,  I would like to be able to comment on a post


pdpinch:

pdpinch:
Is this done now that #81 and #62 have merged?

PR #81, opened 2017-07-31
connect comments reducer to API

aliceriot:

part of #18 

#### What's this PR do?

This adds support for creating comments (replies to a post) and replies to comments. 

You could open up `PostPage.js` and add a line like `actions = window.actions`, and then try calling `actions.comments.post(postId, commentText, ?commentId)` to try things out.

### DIFF
--- a/static/js/lib/api_test.js
+++ b/static/js/lib/api_test.js
@@ -1,12 +1,13 @@
 /* global SETTINGS: false */
 import { assert } from "chai"
 import sinon from "sinon"
-
-import { createChannel, getChannel, getFrontpage, getPost, getPostsForChannel } from "./api"
-import { makeChannel } from "../factories/channels"
-import { makeChannelPostList, makePost } from "../factories/posts"
 import { POST } from "redux-hammock/constants"
 import * as fetchFuncs from "redux-hammock/django_csrf_fetch"
+
+import { createChannel, getChannel, getFrontpage, getPost, getPostsForChannel, getComments, createComment } from "./api"
+import { makeChannel } from "../factories/channels"
+import { makeChannelPostList, makePost } from "../factories/posts"
+import { makeCommentTree } from "../factories/comments"
 
 describe("api", function() {
   this.timeout(5000) // eslint-disable-line no-invalid-this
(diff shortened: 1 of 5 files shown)


static/js/lib/api_test.js
```diff
@@ -68,5 +71,45 @@ describe("api", function() {
...
       })
     })
+
+    it("gets comments for a post", () => {
+      const post = makePost()
+      const tree = makeCommentTree(post)
+      fetchStub.returns(Promise.resolve(tree))
+
+      return getComments(post.id).then(resp => {
+        assert.deepEqual(tree, resp)
```
rhysyngsun: Could you reverse the args? `expected` is the second.

aliceriot: haha, I never can keep that straight

aliceriot: I look up the argument order for `assert.include` literally every time I use it

static/js/reducers/comments.js
```diff
@@ -1,18 +1,19 @@
...
 export const commentsEndpoint = {
   name:              "comments",
-  verbs:             [GET],
-  getFunc:           async (post: Post) => makeCommentTree(post),
+  verbs:             [GET, POST],
+  getFunc:           (post: Post) => api.getComments(post.id),
   initialState:      { ...INITIAL_STATE, data: new Map() },
   getSuccessHandler: (comments: Array<Comment>, data: Map<string, Array<Comment>>) => {
     let update = new Map(data)
     update.set(comments[0].post_id, comments)
```
rhysyngsun: This line errors if there are no comments on the post.

static/js/reducers/comments.js
```diff
@@ -1,18 +1,19 @@
...
     return update
-  }
+  },
+  postFunc: (post: Post, comment: string, commentId: ?string) => api.createComment(post.id, comment, commentId)
```
rhysyngsun: Figured out why I was getting the 403, this code accepts a `Post` object, but the test code in your PR description says to pass just the id. Code works with the `Post`, but just want to confirm which one is the preferred api here. This code is also subject to the other error I posted.

rhysyngsun:
Both `dispatch(actions.comments.post({id:'1'}, 'comment'))` and `dispatch(actions.comments.post({id:'1'}, 'comment', '2'))` give me the above error.

aliceriot:
ok! that error should be addressed now.

rhysyngsun:
Should the new comment from the post to the comments api be reflected in the UI afterwards? I'm not seeing it if so.

aliceriot:
oh that would make sense :woman_facepalming:

aliceriot:
OK I think all the issues should be fixed now. I'm going to leave off for now handling the merging-in of updated data since it's somewhat nontrivial - I think for now it may be easier to just refresh the whole comment tree instead of trying to correctly merge the updated comment into the data we already have.

aliceriot:
(updated data after you issue a `POST` request, that is)

PR #90, opened 2017-08-02
Added UI to list channels

### LINKED ISSUE #83: As a logged in user, I'd like to be able to view a list of channels that I'm subscribed to
In the current screen designs, this is implemented as a left-hand sidebar that collapses to a menu in mobile. 

### acceptance criteria
- [x] REST API for list of subscribed channels
- [x] simple HTML rendering of list
- [ ] CSS will be addressed in #85 

Desktop:
screenshot 2017-08-01 09 29 33

Mobile:

screenshot 2017-08-01 09 30 26

noisecapella:
Part of #83

#### What's this PR do?
Shows the list of channels the user is subscribed to

View the front page. You should see a new sidebar on the left with your channels. Click on a channel and you should be directed to the page for that channel. Press the back button then click on another channel. You should see the posts for that channel.

### DIFF
--- a/static/js/actions/channel.js
+++ b/static/js/actions/channel.js
@@ -1,0 +1,5 @@
+// @flow
+import { createAction } from "redux-actions"
+
+export const SET_CHANNEL_DATA = "SET_CHANNEL_DATA"
+export const setChannelData = createAction(SET_CHANNEL_DATA)
--- a/static/js/containers/ChannelPage.js
+++ b/static/js/containers/ChannelPage.js
@@ -1,8 +1,8 @@
 // @flow
 import React from "react"
 import { connect } from "react-redux"
-import R from "ramda"
 
+import SubscriptionsSidebar from "../components/SubscriptionsSidebar"
 import Card from "../components/Card"
 import ChannelSidebar from "../components/ChannelSidebar"
 import PostList from "../components/PostList"
@@ -11,53 +11,87 @@ import ChannelBreadcrumbs from "../components/ChannelBreadcrumbs"
 
 import { actions } from "../actions"
 import { setPostData } from "../actions/post"
+import { setChannelData } from "../actions/channel"
 import { safeBulkGet } from "../lib/maps"
 import { getChannelName } from "../lib/util"
 import { toggleUpvote } from "../util/api_actions"
 
 import type { Dispatch } from "redux"
 import type { Match } from "react-router"
+import type { RestState } from "../flow/restTypes"
+import type { Channel, Post } from "../flow/discussionTypes"
 
 class ChannelPage extends React.Component {
   props: {
     match: Match,
     dispatch: Dispatch,
-    channels: Object,
-    postsForChannel: Object,
-    posts: Object
+    channels: RestState<Map<string, Channel>>,
+    postsForChannel: RestState<Map<string, Array<string>>>,
+    posts: RestState<Map<string, Post>>,
+    subscribedChannels: RestState<Array<string>>,
   }
 
   componentWillMount() {
-    const { dispatch } = this.props
+    this.updateRequirements()
+  }
+
+  componentDidUpdate(prevProps) {
+    if (getChannelName(this.props) !== getChannelName(prevProps)) {
+      // if the channel was changed, fetch the new information for this channel
+      this.updateRequirements()
+    }
+  }
+
+  updateRequirements = () => {
+    const { dispatch, subscribedChannels } = this.props
     const channelName = getChannelName(this.props)
     dispatch(actions.channels.get(channelName))
     dispatch(actions.postsForChannel.get(channelName)).then(({ posts }) => {
       dispatch(setPostData(posts))
     })
+
+    if (!subscribedChannels.loaded && !subscribedChannels.processing) {
+      dispatch(actions.subscribedChannels.get()).then(channels => {
+        dispatch(setChannelData(channels))
+      })
+    }
   }
 
   renderContents = () => {
-    const { channels, postsForChannel, posts, dispatch } = this.props
+    const { channels, postsForChannel, posts, subscribedChannels, dispatch } = this.props
     const channelName = getChannelName(this.props)
+    if (!channels.data) {
+      return
+    }
     const channel = channels.data.get(channelName)
+    if (!postsForChannel.data) {
+      return
+    }
     const postIds = postsForChannel.data.get(channelName)
 
-    if (R.isNil(channel) || R.isNil(postIds)) {
+    if (!channel || !postIds || !posts.data) {
       return null
     } else {
       return (
-        <div className="double-column">
+        <div className="triple-column">
           <ChannelBreadcrumbs channel={channel} />
           <div className="first-column">
+            <SubscriptionsSidebar
+              // $FlowFixMe: flow doesn't know that we already checked if these are undefined
+              subscribedChannels={safeBulkGet(subscribedChannels.data, channels.data)}
+            />
+          </div>
+          <div className="second-column">
             <Card title={channel.title}>
               <PostList
                 channel={channel}
+                // $FlowFixMe: flow doesn't know that the safeBulkGet above doesn't affect this one
                 posts={safeBulkGet(postIds, posts.data)}
                 toggleUpvote={toggleUpvote(dispatch)}
               />
             </Card>
           </div>
-          <div className="second-column">
+          <div className="third-column">
             <Card>
               <ChannelSidebar channel={channel} />
             </Card>
@@ -76,9 +110,10 @@ class ChannelPage extends React.Component {
 
 const mapStateToProps = state => {
   return {
-    channels:        state.channels,
-    postsForChannel: state.postsForChannel,
-    posts:           state.posts
+    channels:           state.channels,
+    postsForChannel:    state.postsForChannel,
+    posts:              state.posts,
+    subscribedChannels:   state.subscribedChannels,
   }
 }
 
--- a/static/js/containers/ChannelPage_test.js
+++ b/static/js/containers/ChannelPage_test.js
@@ -1,25 +1,31 @@
 // @flow
 import { assert } from "chai"
+import sinon from 'sinon'
 
 import PostList from "../components/PostList"
 
-import { makeChannel } from "../factories/channels"
+import { makeChannel, makeChannelList } from "../factories/channels"
 import { makeChannelPostList } from "../factories/posts"
 import { actions } from "../actions"
 import { SET_POST_DATA } from "../actions/post"
+import { SET_CHANNEL_DATA } from "../actions/channel"
 import IntegrationTestHelper from "../util/integration_test_helper"
 import { channelURL } from "../lib/url"
 
 describe("ChannelPage", () => {
-  let helper, renderComponent, channel, postList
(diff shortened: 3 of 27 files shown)


static/js/containers/ChannelPage.js
```diff
@@ -36,21 +49,30 @@ class ChannelPage extends React.Component {
...
+      return
+    }
     const channel = channels.data.get(channelName)
+    if (!postsForChannel.data) {
+      return
+    }
     const postIds = postsForChannel.data.get(channelName)
 
-    if (R.isNil(channel) || R.isNil(postIds)) {
+    if (!channel || !postIds || !posts.data) {
```
gsidebo: Seems like we could get some use out of a utility function that returns true if any nil args are passed in:

```javascript
function anyNull() {
  return R.any(R.isNil, arguments)
}
```

You can take or leave that. Not too important

static/js/containers/ChannelPage_test.js
```diff
@@ -0,0 +1,74 @@
...
+import { makeChannelPostList } from '../factories/posts'
+import IntegrationTestHelper from '../util/integration_test_helper'
+import { actions } from '../actions'
+import { channelURL } from '../lib/url'
+
+describe('ChannelPage', () => {
+  let helper, renderComponent, listenForActions, currentChannel, otherChannel, channels, posts
+
+  beforeEach(() => {
+    channels = makeChannelList()
```
gsidebo: I think it would be good to add an argument to `makeChannelList` for the number of channels to create, default it to its current value, and call `makeChannelList(5)` et al to make it explicit that we're creating 5 or more and that it's safe to reference index 4 in the resulting array

static/js/containers/CreatePostPage_test.js
```diff
@@ -2,20 +2,22 @@
...
+  let helper, listenForActions, renderComponent, currentChannel, channels
 
   beforeEach(() => {
-    channel = makeChannel()
+    channels = makeChannelList()
+    currentChannel = channels[5]
     helper = new IntegrationTestHelper()
-    helper.getChannelStub.returns(Promise.resolve(channel))
+    helper.getChannelStub.returns(Promise.resolve(currentChannel))
+    //helper.getChannelsStub.returns(Promise.resolve(channels))
```
gsidebo: comment

static/scss/subscriptions.scss
```diff
@@ -0,0 +1,3 @@
+.subscriptions {
+  margin: 20px;
+}
```
gsidebo: blank space needed?

static/js/containers/CreatePostPage_test.js
```diff
@@ -2,20 +2,22 @@
...
 import type { CreatePostPayload } from "../flow/discussionTypes"
 
 describe("CreatePostPage", () => {
-  let helper, listenForActions, renderComponent, channel
+  let helper, listenForActions, renderComponent, currentChannel, channels
 
   beforeEach(() => {
-    channel = makeChannel()
+    channels = makeChannelList()
+    currentChannel = channels[5]
```
gsidebo: Same comment as above re: `makeChannelList`

static/js/containers/SubscriptionsSidebar.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const mapStateToProps = state => {
+  return {
+    channels:      state.channels,
+    subscriptions: state.subscriptions
+  }
+}
+
+export default connect(mapStateToProps)(SubscriptionsSidebar)
```
gsidebo: I'm confused about a few things regarding this container. 

1. `state.channels` and `state.subscriptions` - Shouldn't `state.channels` be `state.channelDetail` or something like that? It looks like it only represents the endpoint to get/create a single channel. `channelsEndpoint` and other similar variables should also be renamed. I also think `state.subscriptions` should probably be `state.channels` even though at the moment I guess it's only pulling subscribed channels. It's using the `/channels` endpoint, right? I could be convinced otherwise.
1. I'm not sure we've ever followed a pattern where we render a container from another container. I think this should be a component with the subscribed channels passed in as a prop. It seems perfectly reasonable to do the work of fetching subscribed channels in `ChannelsPage`

static/js/lib/url.js
```diff
@@ -5,3 +5,5 @@ export const channelURL = (channelName: string) => `/channel/${channelName}`
 export const postDetailURL = (channelName: string, postID: string) => `/channel/${channelName}/${postID}`
 
 export const newPostURL = (channelName: string) => `/create_post/${channelName}`
```
noisecapella: `create_post` was chosen because `/channel/${channelName}/posts` would have collided with the `${postID}`, which was already a URL when I started working on it. I'm not opposed to rethinking the URLs but it is outside the scope here. It's not a `POST` request because `/create_post/${channelName}` is for the page that shows the form to create the post, not the endpoint for creating the post itself. That is at `/api/v0/channels/{channelName}/posts/`

static/js/containers/ChannelPage.js
noisecapella: This is here to appease flow which doesn't really understand `R.isNil`. When I made the types more specific flow had some warnings about these variables.

static/js/containers/ChannelPage_test.js
noisecapella: Makes sense

static/js/containers/SubscriptionsSidebar.js
noisecapella: The channels from `/api/v0/channels/` are the channels you're subscribed to, but not necessarily all the ones you have permission to see, so there is a need for a channel detail view. That's also why I named it `subscriptions`. Regarding the container rendering in another container, I wanted to do it this way since the sidebar is used in both `HomePage` and `ChannelPage`. I can go either way though, maybe we should talk with @aliceriot and @rhysyngsun when they're back in the office

gsidebo: OK that sounds reasonable. I think I'm comfortable with the individual API calls to channels and channel detail for now, and if we decide on something cleaner than `SubscriptionsSidebar` as a container, I think that would be pretty low effort. 

I do think we should change the `channels` naming convention, though. `subscriptions` refers to an API call for multiple channels, and `channels` refers to an API call for a single channel. I really think we should change that and other similarly named variables. Also f we want to keep the notion of subscriptions in the variable names, `subscribedChannels` might be more helpful than `subscriptions`

noisecapella:
Comments addressed

static/js/containers/ChannelPage.js
```diff
@@ -16,41 +16,68 @@ import { getChannelName } from "../lib/util"
...
   }
 
   componentWillMount() {
-    const { dispatch } = this.props
+    this.updateRequirements()
+  }
+
+  componentDidUpdate(prevProps) {
+    if (getChannelName(this.props) !== getChannelName(prevProps)) {
+      // if the page was changed, fetch the new information for this channel
```
gsidebo: "If the ~~page~~ channel was changed" (unless I'm missing something)

static/js/containers/ChannelPage.js
```diff
@@ -16,41 +16,68 @@ import { getChannelName } from "../lib/util"
...
   props: {
     match: Match,
     dispatch: Dispatch,
-    channels: Object,
-    postsForChannel: Object,
-    posts: Object
+    channels: RestState<Map<string, Channel>>,
+    postsForChannel: RestState<Map<string, Array<string>>>,
+    posts: RestState<Map<string, Post>>,
+    subscriptions: RestState<Array<Channel>>,
```
gsidebo: Any thoughts about renaming `channels` to `channel`/`channelDetail` et al, and `subscriptions` to `subscribedChannels`? I think either or both of those would be a lot clearer, and I think it's confusing enough to be annoying about it in this review 😁

static/js/containers/ChannelPage.js
```diff
@@ -16,41 +16,68 @@ import { getChannelName } from "../lib/util"
...
 
-  renderContents(channelName, channels, postsForChannel, posts) {
+  renderContents(channelName, channels, postsForChannel, posts, subscriptions) {
+    if (!channels.data || !posts.data || !subscriptions.data) {
+      return
+    }
     const channel = channels.data.get(channelName)
+    if (!postsForChannel.data) {
+      return
+    }
```
gsidebo: Any reason why we shouldn't just add `!postsForChannel.data` to the series of conditionals at line 57?

static/js/containers/ChannelPage.js
noisecapella: I'll rename `subscribedChannels` but I'm not sure about `channel` or `channelDetail`. The redux state is the canonical collection of channels so `channels` seems to fit best there. Maybe `subscribedChannels` should contain the names for the channels and the channel objects should be stored in `channels`, similar to how the post objects are stored in `posts` and the ids for each post are stored in `postsForChannel`

static/js/containers/ChannelPage.js
noisecapella: It's a flow quirk, it doesn't understand that `get()` won't affect `postsForChannel`

static/js/containers/ChannelPage.js
noisecapella: Sure

static/js/containers/ChannelPage.js
gsidebo: Sounds good. Just renaming `subscriptions` is certainly good enough for now, but I'll put it another way: if we had the need to fetch all channels (not just the subscribed ones), what would we name that in the state if not `channels`? Just something to think about. I'm still getting used to redux hammock so maybe it's just lack of familiarity

noisecapella:
Ready for review again

static/js/actions/channel.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+import { createAction } from "redux-actions"
+
+export const SET_CHANNEL_DATA = "SET_CHANNEL_DATA"
+export const setChannelData = createAction(SET_CHANNEL_DATA)
```
gsidebo: Any reason these shouldn't just be defined in `static/js/actions/channels.js`? The reducer is defined in `static/js/reducers/channels.js` after all

static/js/containers/ChannelPage.js
```diff
@@ -62,23 +104,26 @@ class ChannelPage extends React.Component {
...
-    const { channels, postsForChannel, posts } = this.props
+    const { channels, postsForChannel, posts, subscribedChannels } = this.props
     const channelName = getChannelName(this.props)
 
     return (
       <Loading
         restStates={[channels, postsForChannel]}
-        renderContents={() => this.renderContents(channelName, channels, postsForChannel, posts)}
+        renderContents={() => this.renderContents(
+          channelName, channels, postsForChannel, posts, subscribedChannels
```
gsidebo: We should do away with the bulky parameter passing here and use `this` in renderContents. Then we would just define the prop as `renderContents=this.renderContents` or `renderContents=this.renderContents.bind(this)`. I made a similar comment on Nathan's PR here: https://github.com/mitodl/open-discussions/pull/93#discussion_r131759798

static/js/actions/channel.js
noisecapella: I wanted to match `actions/post.js` which is singular

PR #91, opened 2017-08-02
Refactored abstract form code to work over multiple forms of the same type

rhysyngsun:
Part of #18

#### What's this PR do?
Refactors form code so it's more flexible for multiple comment forms

Ensure the create channel and create post forms still function correctly


static/js/containers/CreatePostPage.js
```diff
@@ -32,24 +37,41 @@ class CreatePostPage extends React.Component {
     if (!channels.loaded && !channels.processing) {
       dispatch(actions.channels.get(channelName))
     }
-    dispatch(actions.forms.post.create())
+    dispatch(
+      actions.forms.formBeginEdit(
+        R.merge(CREATE_POST_PAYLOAD, {
+          value: newPostForm()
+        })
```
gsidebo: I think this is a friendlier syntax for the action payloads:

```javascript
actions.forms.formBeginEdit({
  ...CREATE_POST_PAYLOAD,
  value: newPostForm()
})
```

static/js/containers/CreatePostPage.js
```diff
@@ -17,12 +18,16 @@ import type { Match } from "react-router"
...
-  postForm: FormValue,
+  postForm: ?FormValue,
   channel: Channel,
   channels: RestState<Array<Channel>>,
   history: Object
 }
 
+const CREATE_POST_KEY = "post:new"
+const CREATE_POST_PAYLOAD = { key: CREATE_POST_KEY }
+const getForm = R.prop(CREATE_POST_KEY)
```
gsidebo: A few nitpicks that apply equally to this container and the channel container: 

1. Should we use `"post:new"` for the key name? If I'm understanding correctly, the key is just being used to identify the correct form in the state, which isn't necessarily 'new'. Maybe just `"post_form"` is more on the mark.
1. I think `CREATE_POST_KEY` should be changed to `POST_FORM_KEY`, and `POST_FORM_KEY_PAYLOAD` or `POST_FORM_IDENTIFIER` might be a little clearer (IMO) than `CREATE_POST_PAYLOAD` (`_PAYLOAD` makes me think that this is the entire payload you will be sending to the action)

Also, would it make sense to stick these variables in `static/js/lib/posts.js`? You could also create a `POST_FORM_INITIAL_STATE` object there and use that as the payload to `formBeginEdit` in `componentWillMount`. I'll leave that to you

rhysyngsun: We'll eventually have a different form to edit a post (editable fields are limited to `text`), so I want to push back a tad on the namespacing because I think it's  important to have deterministic keys that represent distinct form data structures. For the same reason I think the defs of those values should stay here because they're only pertinent to the `CreatePostPage`.

static/js/containers/CreatePostPage.js
rhysyngsun: I should know that :man_facepalming:

PR #92, opened 2017-08-03
Added commenting on posts and comments

### LINKED ISSUE #18: As a logged in user,  I would like to be able to comment on a post

rhysyngsun:
Fixes #18 

#### What's this PR do?
This adds UI to add comments  in reply to posts and other comments.

Verify you can comment on a post and comment.

### DIFF
--- a/static/js/components/CreateCommentForm_test.js
+++ b/static/js/components/CreateCommentForm_test.js
@@ -1,0 +1,159 @@
+// @flow
+import React from "react"
+import R from "ramda"
+import { Provider } from "react-redux"
+import { assert } from "chai"
+import { mount } from "enzyme"
+
+import { ReplyToCommentForm, ReplyToPostForm } from "./CreateCommentForm"
+import * as forms from "../actions/forms"
+import { actions } from "../actions"
+import { SET_POST_DATA, setPostData } from "../actions/post"
+import { makePost } from "../factories/posts"
+import { makeComment } from "../factories/comments"
+import IntegrationTestHelper from "../util/integration_test_helper"
+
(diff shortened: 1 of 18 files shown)


static/js/reducers/comments_test.js
```diff
@@ -0,0 +1,99 @@
...
+  })
+
+  it("should let you reply to a post", () => {
+    const { requestType, successType } = actions.comments.post
+    return listenForActions(
+      [actions.comments.get.successType, actions.comments.get.requestType, requestType, successType],
+      () => {
+        store.dispatch(actions.comments.get(post.id))
+        store.dispatch(actions.comments.post(post.id, "comment text"))
+      },
```
noisecapella: The below code isn't getting run, did you mean to do `).then(state => ...` here?

static/js/components/CreateCommentForm_test.js
```diff
@@ -0,0 +1,129 @@
...
+        assert.equal(wrapper.find("a").text(), "Reply")
+      })
+
+      it("should show and empty when reply is clicked", () => {
+        return helper.listenForActions([
+          forms.FORM_BEGIN_EDIT
+        ], () => {
+          wrapper.find("a").simulate("click")
+        }, () => {
+          assert.equal(wrapper.find("input[text]").value, "")
```
noisecapella: This isn't getting run, you need to put this in the `then` instead

noisecapella:
It looks like the number of comments isn't getting updated after a comment is made:

screenshot from 2017-08-04 11-01-14

rhysyngsun:
@noisecapella I've addressed your feedback and the bug you found

PR #93, opened 2017-08-04
Integrate UI for upvoting posts

### LINKED ISSUE #70: As a logged in user, I want to be able to upvote posts
On:

- Frontpage
- Channel page
- Individual post page

rhysyngsun:
Fixes #70 

#### What's this PR do?
Adds upvoting for posts

You should be able to add and remove an upvote on the front, channel, and indidvual post pages.

### DIFF
--- a/static/js/components/PostDisplay.js
+++ b/static/js/components/PostDisplay.js
@@ -1,6 +1,7 @@
 // @flow
 import React from "react"
 import moment from "moment"
+import { connect } from "react-redux"
 import { Link } from "react-router-dom"
 
 import { postDetailURL } from "../lib/url"
@@ -21,11 +22,23 @@ const postTitle = (post: Post) =>
       {post.title}
     </a>
 
-export default class PostDisplay extends React.Component {
+class PostDisplay extends React.Component {
   props: {
     post: Post,
     showChannelLink?: boolean,
-    expanded?: boolean
+    expanded?: boolean,
+    toggleUpvote: Post => void
+  }
+
+  state: {
+    upvoting: boolean
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      upvoting: false
+    }
   }
 
   showChannelLink = () => {
@@ -38,13 +51,28 @@ export default class PostDisplay extends React.Component {
       : null
   }
 
+  onToggleUpvote = async () => {
+    const { toggleUpvote, post } = this.props
+    this.setState({
+      upvoting: true
+    })
+    await toggleUpvote(post)
+    this.setState({
+      upvoting: false
+    })
+  }
+
   render() {
     const { post, expanded } = this.props
+    const { upvoting } = this.state
     const formattedDate = moment(post.created).fromNow()
+    const upvoteClass = post.upvoted ? "upvoted" : ""
     return (
       <div className="post-summary">
-        <div className="upvotes">
-          <button> &uArr;</button>
+        <div className={`upvotes ${upvoteClass}`}>
+          <button className="upvote-button" onClick={this.onToggleUpvote} disabled={upvoting}>
+            {" "}&uArr;
+          </button>
           <span className="votes">
             {post.score}
           </span>
@@ -67,3 +95,9 @@ export default class PostDisplay extends React.Component {
     )
   }
 }
+
+const mapStateToProps = state => ({
+  posts: state.posts
+})
+
+export default connect(mapStateToProps)(PostDisplay)
--- a/static/js/containers/PostPage.js
+++ b/static/js/containers/PostPage.js
@@ -11,6 +11,7 @@ import CommentTree from "../components/CommentTree"
 import { ReplyToPostForm } from "../components/CreateCommentForm"
 
 import { actions } from "../actions"
+import { toggleUpvote } from "../util/api_actions"
 import { anyProcessing, allLoaded } from "../util/rest"
 
 import type { Dispatch } from "redux"
@@ -60,10 +61,12 @@ class PostPage extends React.Component {
     }
   }
 
-  renderContents(postID, posts, channelName, channels, comments, forms) {
-    const post = posts.get(postID)
-    const channel = channels.get(channelName)
-    const commentTreeData = comments.get(postID)
+  renderContents = () => {
+    const { posts, channels, comments, forms, dispatch } = this.props
+    const [postId, channelName] = this.getMatchParams()
+    const post = posts.data.get(postId)
+    const channel = channels.data.get(channelName)
+    const commentTreeData = comments.data.get(postId)
 
     if (R.isNil(channel) || R.isNil(post) || R.isNil(commentTreeData)) {
       return null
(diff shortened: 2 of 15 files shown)


gsidebo:
Rapid-fire clicks on the upvote button results in exactly as many upvote API calls. I think at minimum we should disable the click handler (or have the click handler do nothing) while the upvote API is processing. 

I also noticed that it takes some time for the API call to complete, resulting in a delay between the click and the visual cue. This isn't urgent, and I know that the local environment makes things slower, but I wonder if it's worth it to think about updating the state of the arrow first, then executing the API call. Again, not urgent. Just a thought.

static/js/components/PostDisplay.js
```diff
@@ -23,30 +27,41 @@ const postTitle = (post: Post) =>
...
       </Link>
-      : null
+      : null)
+  }
+
+  toggleUpvote = () => {
+    const { dispatch, post } = this.props
+    dispatch(actions.postUpvotes.post(post.id, !post.upvoted)).then(post => {
+      dispatch(setPostData(post))
+    })
```
gsidebo: In MM we typically defined functions like these (click handlers that make an API call) in containers, then passed them down to the components that need them. We don't necessarily need to stick to that - I know we're rendering this component from 2 different places. Maybe @aliceriot could chime in.

rhysyngsun: I'm fine with that approach, but it's obviously going to touch a lot more code. We can discuss Monday.

rhysyngsun:
I think we can possibly look into some kind of optimistic ui implementation for the responsiveness issue, but that's probably best addressed in a separate issues as it's more of a usability issue and possibly not MVP.

I'll definitely address the multi-click issue in this PR though.

static/js/components/PostDisplay.js
aliceriot: I think to me it would make the most sense in this case to stick with the page level container defining the API functions, and then handing them down to the presentational components. If we want to make the `PostDisplay` a container component (as we're doing here with the `connect()` call) I think it would probably make sense to centralize the relevant API functions in that container so that they aren't scattered between e.g. `PostPage` and `PostDisplay`. However, to me the overall pattern that makes the most sense is to have only a few, top-level containers, and for those containers to sort of 'feed' the presentational components data and callbacks. Does that all make sense?

If there's a concern with code duplication we could do something like the following:

```js
toggleUpvote = R.curry(async (dispatch, post) => {
  let post = await dispatch(actions.postUpvotes.post(post.id, !post.upvoted))
  dispatch(setPostData(post))
})
```

then in our container components we would do:

```js
import { toggleUpvote } from 'post-api-utils' // or something
...
const { dispatch } = this.props
return <PostList toggleUpvote={toggleUpvote(dispatch)} />
```

gsidebo: Passing `dispatch` around is somewhat confusing, IMO. Might make more sense to define `toggleUpvote` in the `PostPage` container with a parameter for the ID, then for each rendering of `PostDisplay`, use `R.partial` or something to satisfy that ID

rhysyngsun: @gsidebo we need `toggleUpvote` in `Postpage`, `ChannelPage`, and `Homepage` so that's why we probably want to pull the impl out into a util.

gsidebo: Ah, indeed we do

gsidebo: This is something we could (should?) solve with a higher-order component like we do for email dialogs in MM, but I'm OK with holding off on that if it's too unwieldy for the time being.

static/js/containers/PostPage.js
```diff
@@ -83,13 +84,23 @@ class PostPage extends React.Component {
...
+        renderContents={() =>
+          this.renderContents(
+            postId,
+            posts.data,
+            channelName,
+            channels.data,
+            comments.data,
+            forms,
+            dispatchableToggleUpvote
+          )}
```
gsidebo: The parameter passing here is really bulky. Let's just bind `this` to the handler function. Rewritten, this would like like: 

```javascript
renderContents() {
    const { posts, channels, comments, forms, dispatch } = this.props
    const [postId, channelName] = this.getMatchParams()
    const dispatchableToggleUpvote = toggleUpvote(dispatch)

    const post = posts.data.get(postId)
    const channel = channels.data.get(channelName)
    const commentTreeData = comments.data.get(postId)

    if (R.isNil(channel) || R.isNil(post) || R.isNil(commentTreeData)) {
      return null
    }
    return (
      <div className="double-column">
        <ChannelBreadcrumbs channel={channel} />
        <div className="first-column">
          <Card>
            <PostDisplay post={post} toggleUpvote={dispatchableToggleUpvote} expanded />
            <ReplyToPostForm forms={forms} post={post} />
          </Card>
          <CommentTree comments={commentTreeData} forms={forms} />
        </div>
      </div>
    )
  }

  render() {
    const { posts, channels, comments } = this.props
    return (
      <Loading
        restStates={[posts, channels, comments]}
        renderContents={this.renderContents.bind(this)}
      />
    )
  }
```

The same idea can also be applied to `ChannelPage.js`.

rhysyngsun:
@gsidebo this is ready for review again, regarding the regression on the `PostPage`, I reverted the removal of `postUpvotesEndpoint` as the quickest patch to getting that fixed.

static/js/reducers/posts.js
```diff
@@ -22,7 +22,7 @@ const mergeMultiplePosts = (posts: Array<Post>, data: Map<string, Post>): Map<st
 
 export const postsEndpoint = {
   name:               "posts",
-  verbs:              [GET, POST],
+  verbs:              [GET, POST, PATCH],
```
gsidebo: Guessing we don't still need this `PATCH` verb for this endpoint definition?

gsidebo:
A few UI notes that can be implemented in separate issues:

1. I don't think the arrow should be greyed out while the click is disabled and the API function is executing. IMO it's kind of jarring to see the button go grey after I click the button to upvote. I think it's better to leave the styling unchanged until the upvote succeeds
1. On reddit, the count updates after you click upvote/downvote. I _think_ it's an optimistic update, i.e.: they just increment/decrement the number as soon as you click. I think it might be bad UX if a user click the upvote button and doesn't see the count change.
1. Just wondering: do we have a plan to use prettier arrows?

static/js/reducers/posts.js
rhysyngsun: Nope, I'll remove it

rhysyngsun:
1. I'll defer to @roberthouse54 on this, you had mentioned earlier in the review to disable it, so that's why I did it that way.
2. The count is a fuzzy number, @pdpinch has emphasize d we just want to use the value from the reddit back verbatim so I would hesitate to do any optimistic UI around that value.
3. Rob started to introduce some better looking buttons in #95, but those changes conflicted with this PR, so Alice reverted back to the button. We can figure out how to style those later on now that the code is more settled.

Issue #85, opened 2017-08-01
CSS Styling 


pdpinch:
Based on the designs in Invision:

- [x] Basic fonts, colors, and styling for Channel pages, post summaries, and post content. (Addressed in #95
- [x] Styling for the sidebar with the channels and the "Submit Post" button.
- [x] Styles for the Create Post page ~~and the Create Channel page.~~
- [x] HTML and CSS for a navbar, assuming we are going to have one.
- [x] Styling for the comments UI.
- [x] #115 Responsive CSS for mobile for all of the above.

roberthouse54:
Work that needs to be done:
- Basic fonts, colors, and styling for Channel pages, post summaries, and post content. (Addressed in https://github.com/mitodl/open-discussions/pull/95
- Styling for the sidebar with the channels and the "Submit Post" button.
- Styles for the Create Post page and the Create Channel page. 
- HTML and CSS for a navbar, assuming we are going to have one.
- Styling for the comments UI. 
- Responsive CSS for mobile for all of the above.

aliceriot:
I'm going to work on #115, just basically getting the spacing, width, etc set-up so the site will work on a phone.

aliceriot:
Opened a few issues (which I'll work on) for small issues that Rob has noticed:

- #123 
- #125
- #126
- #127

pdpinch:
"Create Channel" isn't part of the open-discussions app. There's an issue for it at https://github.com/mitodl/micromasters/issues/3473. I've updated the description of this issue.

aliceriot:
There is a 'create channel' page in the OD admin (`/manage/channel/new`)

PR #95, opened 2017-08-04
Font, color and margins style for Discussions UI

roberthouse54:
This is the First round of style changes for the discussions. I am going to submit small incremental PRs as I go along with this.

#### What's this PR do?
Sets some of the basic colors, fonts, etc for discussions.

#### What Issue is this related to
https://github.com/mitodl/open-discussions/issues/85


static/js/components/ChannelBreadcrumbs.js
```diff
@@ -12,10 +12,10 @@ const ChannelBreadcrumbs = (props: BreadCrumbProps) => {
   const { channel } = props
 
   return (
-    <div>
+    <div className="breadcrumbs">
       <Link to="/">Discussions</Link>&nbsp;
       <span>&gt;</span>&nbsp;
-      <Link to={`/channel/${channel.name}`}>{channel.title}</Link>
+      {channel.title}
```
aliceriot: This should probably remain a `<Link />` - we show the `ChannelBreadcrumbs` component on the page for each post, so it supplies a link back to the subreddit the post is from.

static/js/containers/ChannelPage.js
```diff
@@ -55,6 +56,7 @@ class ChannelPage extends React.Component {
               <ChannelSidebar channel={channel} />
             </Card>
           </div>
+          <br className="clear"/>
```
aliceriot: is there any way we can avoid using `<br />`? I just find it yucky, haha

roberthouse54: I am open to any solutions you can come up with, although this is in my opinion, a basic failure with the CSS model and contained floats inside divs. But I am open to any way you'd prefer to do this. I do want those divs actually contained inside the container div though.

static/js/components/ChannelBreadcrumbs.js
roberthouse54: OK, that is fine.

roberthouse54:
I will be out today and tomorrow. @aliceriot , can you make the required changes and merge the PR?

roberthouse54:
One further comment:

The channel page UI, and the Home page both use a "Post Summary" component, that displays the Post title, post author, number of comments, and upvoting. The Post Details page is currently reusing this same component. But we should really create a different component for the Post Details page as the design is really not the same as it is on the channel/home page.

aliceriot:
On the `PostSummary` component, it basically has two modes. On a post detail view it shows more information (like the text if its a text post), displays slightly differently, etc. Do you think that we could just continue to build out the two different display modes? If we want to style them differently we could have a css class which only appears on the 'expanded' view, for instance. There is a fair amount of display which is more or less identical looking in the designs (i.e. we show the link title, the author, etc) which is why I re-used the component for the 'channel' view.

aliceriot:
the post display is a little messed up:

post_display

I'm going to make a few quick changes to restore the spacing

static/js/containers/ChannelPage.js
aliceriot: Maybe I just don't understand what it's doing, but if I delete this element I don't see any changes in the display on the page. I'll leave it in for now if it's necessary.

roberthouse54:
The post display component is identical on the channel page and the home page, and yse I did see that there is an "expanded" version. But I think the layout on the actual post page is significantly different enough to warrant a different component and different html. The # of comments is down below, next to the comments UI, the upvoting is in a different location, etc

roberthouse54:
About using the break with class="clear"... yes it is a workaround. And I think you might not see any visual changes on the page. But the two floated divs are not actually contained inside the container div. I do want to have that working,as I think, otherwise, we will not be able to apply box property rules to the container etc.

aliceriot:
it seems like setting `display: flex` on the container div (`.double-column` or `.single-column`) has the same effect, I think?

aliceriot:
I'll leave it as-is for now and we can discuss when you're in

static/js/components/PostDisplay.js
```diff
@@ -46,19 +44,23 @@ export default class PostDisplay extends React.Component {
     return (
       <div className="post-summary">
         <div className="upvotes">
-          <button> &uArr;</button>
+          <img className="upvote-arrow" src="/static/images/upvote_arrow.png" />
```
rhysyngsun: Is there a reason for the move away from a `<button>`? This has accessibility issues and has no native `disabled` behavior (which is something I'm doing in #93, which will merge soon too).

aliceriot: let's defer to what you have in #93 for now

PR #97, opened 2017-08-07
added very basic tests for the container components

aliceriot:

none

#### What's this PR do?

I noticed we had started adding container tests, but we don't yet have them everywhere. this just adds some very basic tests that the container loads, hits the APIs, etc.

tests should pass, code should make sense.

### DIFF
--- a/static/js/containers/ChannelPage_test.js
+++ b/static/js/containers/ChannelPage_test.js
@@ -1,0 +1,48 @@
+// @flow
+import { assert } from "chai"
+
+import PostList from "../components/PostList"
+
+import { makeChannel } from "../factories/channels"
+import { makeChannelPostList } from "../factories/posts"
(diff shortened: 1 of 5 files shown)


static/js/containers/ChannelPage_test.js
```diff
@@ -0,0 +1,47 @@
...
+    helper.getPostsForChannelStub.returns(Promise.resolve(postList))
+    renderComponent = helper.renderComponent.bind(helper)
+  })
+
+  afterEach(() => {
+    helper.cleanup()
+  })
+
+  const renderPage = channel =>
+    renderComponent(`/channel/${channel.name}`, [
```
noisecapella: Can you use `channelURL` here?

aliceriot: oh, didn't know that existed!

PR #98, opened 2017-08-07
Updated README with more accurate setup details

gsidebo:
No ticket

#### What's this PR do?
Updated README with more accurate setup details (discovered while getting this running locally for Rob)

N/A

Accompanying reddit-config PR: https://github.com/mitodl/reddit-config/pull/12

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,14 +50,14 @@ installation steps should be completed before the following steps, which are spe
 
 #### Create your docker container:
 
-The following commands create a Docker machine named ``open_discussions``, start the
+The following commands create a Docker machine named ``opendiscussions``, start the
 container, and configure environment variables to facilitate communication
 with the edX instance.
 
-    docker-machine create --driver virtualbox open_discussions
-    docker-machine start open_discussions
+    docker-machine create --driver virtualbox opendiscussions
+    docker-machine start opendiscussions
     # 'docker-machine env (machine)' prints export commands for important environment variables
-    eval "$(docker-machine env open_discussions)"
+    eval "$(docker-machine env opendiscussions)"
 
 
 ## Docker Container Configuration and Start-up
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,14 +50,14 @@ installation steps should be completed before the following steps, which are spe
...
 container, and configure environment variables to facilitate communication
 with the edX instance.
 
-    docker-machine create --driver virtualbox open_discussions
-    docker-machine start open_discussions
+    docker-machine create --driver virtualbox opendiscussions
+    docker-machine start opendiscussions
     # 'docker-machine env (machine)' prints export commands for important environment variables
-    eval "$(docker-machine env open_discussions)"
+    eval "$(docker-machine env opendiscussions)"
```
gsidebo: docker-machine does not allow underscores in the machine name

noisecapella: This is probably out of scope but is there any reason not to put everything in the default docker-machine instance, instead of one per app?

gsidebo: To be honest I don't really know enough about docker/docker-machine to say whether or not this is a good idea

noisecapella: I don't think this will cause a problem, it's mostly an issue if you wanted to destroy one VM for some reason without affecting the others. Anyway let's leave this as is, I'll see about fixing it in cookiecutter

README.md
```diff
@@ -119,6 +119,31 @@ and add a superuser in the now-running Docker container:
...
+superuser you created):
+ 
+ ```bash
+ # Create a channel
+ curl -X POST -u USERNAME:PASSWORD "http://<open_discussions_url>:8063/api/v0/channels/" \
+   -H "Content-Type: application/json" \
+   -d '{"title": "Test Channel", "name": "test_channel", "public_description": "This is a test channel", "channel_type": "public"}'
+ 
+ # Create a post for the channel
+ curl -X POST -u USERNAME:PASSWORD "http://<open_discussions_url>:8063/api/v0/channels/test_channel/posts/" \
```
noisecapella: This line and the next one end with spaces, but the `\` needs to go right before the newline

README.md
gsidebo: I think being able to destroy an individual machine might be reason enough in itself

PR #109, opened 2017-08-14
Refactored loading component

### LINKED ISSUE #102: remove flash when updating the page
Seen in our meeting Wednesday afternoon

rhysyngsun:
Fixes #102 

#### What's this PR do?
Updates `<Loading />`

- Exercise the channel and post pages and verify they load correctly including over a full page refresh.
- Verify the page doesn't flash when adding a comment to a post

I reimplemented `<Loading/>` as the HOC `withLoading`. This new design has a few benefits:

- The wrapped component is simpler and implements only a `render()` method instead of the previous design with both a `render()` and `renderContent()` implementation.
- The wrapped component now determines if it is loaded or not in its `mapStateToProps` implementation

Some design tradeoffs to achieve this:

- We need to pass two functions to `withLoading`: 
  - `shouldUpdateRequirements`: this function is pass the previous and current props and returns true if the difference in values indicates the requirements should be updated
  - `updateRequirements`: this function dispatches all actions required to update the requirements

It was necessary to move this type logic out of where it previously lived (the wrapped component's `componentWillMount` and `componentDidUpdate`) because those methods don't get called until after the component's data has been loaded. Instead, I chose to pass them to `withLoading` so they could be called in the HOC's `componentWillMount` and `componentDidUpdate` independent of whether the wrapped component was mounted/mounting yet or not.

The only piece I'm not completely satisfied with is the `mapStateToProps` implementations themselves, I feel they could be a little cleaner with some more time and effort, but they're functional and I didn't want to go down a rabbit hole of trying to make those DRY/simpler.

Loading

### DIFF
--- a/static/js/components/Loading.js
+++ b/static/js/components/Loading.js
@@ -1,27 +1,33 @@
 // @flow
 import React from "react"
 
-import { anyProcessing, anyError, allLoaded } from "../util/rest"
+type LoadingProps = {
+  loaded: boolean,
+  errored: boolean
+}
 
-import type { RestState } from "../flow/restTypes"
+const withLoading = (LoadedComponent: Class<React.Component<*, *, *>>) => {
+  return class extends LoadedComponent {
+    props: LoadingProps
 
-export default class Loading extends React.Component {
-  props: {
-    restStates: Array<RestState<*>>,
-    renderContents: Function
-  }
+    render() {
+      const { loaded, errored } = this.props
 
-  render() {
-    const { restStates, renderContents } = this.props
+      if (errored) {
+        return <div className="errored">Error loading page</div>
+      }
 
-    if (anyProcessing(restStates) || !allLoaded(restStates)) {
-      return <div>Loading</div>
+      if (!loaded) {
+        return <div className="loading">Loading</div>
+      }
+
+      return (
+        <div className="loaded">
+          {super.render()}
+        </div>
+      )
     }
-
-    if (anyError(restStates)) {
-      return <div>Error loading page</div>
-    }
-
-    return renderContents()
   }
 }
+
+export default withLoading
--- a/static/js/components/Loading_test.js
+++ b/static/js/components/Loading_test.js
@@ -1,135 +1,40 @@
 // @flow
 import React from "react"
 import { assert } from "chai"
-import { shallow } from "enzyme"
+import { mount } from "enzyme"
 
-import Loading from "./Loading"
+import withLoading from "./Loading"
+
+class Content extends React.Component {
+  render() {
+    return <div>CONTENT</div>
+  }
+}
+
+const LoadingContent = withLoading(Content)
 
 describe("Loading", () => {
-  const renderLoading = restStates =>
-    shallow(<Loading restStates={restStates} renderContents={() => <div>CONTENT</div>} />)
+  const renderLoading = (loaded: boolean, errored: boolean) => {
+    return mount(<LoadingContent loaded={loaded} errored={errored} />)
+  }
 
-  describe("should render loading correctly", () => {
-    it("single state with loaded:false", () => {
-      const wrapper = renderLoading([
-        {
-          loaded:     false,
-          processing: false
-        }
-      ])
-      assert.equal(wrapper.text(), "Loading")
-    })
-
-    it("multiple states with one loaded:false", () => {
-      const wrapper = renderLoading([
-        {
-          loaded:     true,
-          processing: false
-        },
-        {
-          loaded:     false,
-          processing: false
-        }
-      ])
-      assert.equal(wrapper.text(), "Loading")
-    })
-
-    it("single state with processing:true", () => {
-      const wrapper = renderLoading([
-        {
-          loaded:     false,
-          processing: true
-        }
-      ])
-      assert.equal(wrapper.text(), "Loading")
-    })
-
-    it("multiple states with one processing:true", () => {
-      const wrapper = renderLoading([
-        {
-          loaded:     false,
-          processing: false
-        },
-        {
-          loaded:     false,
-          processing: true
-        }
-      ])
-      assert.equal(wrapper.text(), "Loading")
-    })
+  it("render loading if not loaded and not errored", () => {
+    const wrapper = renderLoading(false, false)
+    assert.equal(wrapper.text(), "Loading")
   })
 
-  describe("should render errors correctly", () => {
-    it("single state with error", () => {
-      const wrapper = renderLoading([
-        {
-          loaded:     true,
-          processing: false,
-          error:      "bad things"
-        }
-      ])
-      assert.equal(wrapper.text(), "Error loading page")
-    })
-
-    it("multiple states all with errors", () => {
-      const wrapper = renderLoading([
-        {
-          loaded:     true,
-          processing: false,
-          error:      "bad things"
-        },
-        {
-          loaded:     true,
-          processing: false,
-          error:      "bad things"
-        }
-      ])
-      assert.equal(wrapper.text(), "Error loading page")
-    })
-
-    it("multiple states with one error", () => {
-      const wrapper = renderLoading([
-        {
-          loaded:     true,
-          processing: false,
-          error:      undefined
-        },
-        {
-          loaded:     true,
-          processing: false,
-          error:      "bad things"
-        }
-      ])
-      assert.equal(wrapper.text(), "Error loading page")
-    })
+  it("should render errors correctly if not loaded", () => {
+    const wrapper = renderLoading(false, true)
+    assert.equal(wrapper.text(), "Error loading page")
   })
 
-  describe("should the contents if no errors", () => {
-    it("single state", () => {
-      const wrapper = renderLoading([
-        {
-          loaded:     true,
-          processing: false,
-          error:      undefined
-        }
-      ])
-      assert.equal(wrapper.text(), "CONTENT")
-    })
+  it("should render errors correctly if loaded", () => {
+    const wrapper = renderLoading(true, true)
+    assert.equal(wrapper.text(), "Error loading page")
+  })
 
-    it("multiple states", () => {
-      const wrapper = renderLoading([
-        {
-          loaded:     true,
-          processing: false,
-          error:      undefined
-        },
-        {
-          loaded:     true,
-          processing: false,
-          error:      undefined
-        }
-      ])
-      assert.equal(wrapper.text(), "CONTENT")
-    })
+  it("should the contents if no errors and loaded", () => {
+    const wrapper = renderLoading(true, false)
+    assert.equal(wrapper.text(), "CONTENT")
   })
 })
--- a/static/js/containers/ChannelPage.js
+++ b/static/js/containers/ChannelPage.js
@@ -1,12 +1,13 @@
 // @flow
 import React from "react"
+import R from "ramda"
 import { connect } from "react-redux"
 
 import SubscriptionsSidebar from "../components/SubscriptionsSidebar"
 import Card from "../components/Card"
 import ChannelSidebar from "../components/ChannelSidebar"
 import PostList from "../components/PostList"
-import Loading from "../components/Loading"
+import withLoading from "../components/Loading"
 import ChannelBreadcrumbs from "../components/ChannelBreadcrumbs"
 
 import { actions } from "../actions"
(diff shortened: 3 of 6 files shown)


aliceriot:
functionality looks good! reading code now

static/js/components/Loading_test.js
```diff
@@ -1,134 +1,44 @@
...
-        }
-      ])
+    it("if loaded", () => {
+      const wrapper = renderLoading(true, true)
       assert.equal(wrapper.text(), "Error loading page")
     })
   })
 
-  describe("should the contents if no errors", () => {
+  describe("should the contents if no errors and loaded", () => {
```
aliceriot: could this `describe` block just be an `it` block? the message I think is a bit strangely worded too, I would say `it('should show the contents if no errors and loaded', () => {})` or something

static/js/components/Loading_test.js
```diff
@@ -1,134 +1,44 @@
...
+const LoadingContent = withLoading(Content, () => true, () => {})
 
 describe("Loading", () => {
-  const renderLoading = restStates =>
-    shallow(<Loading restStates={restStates} renderContents={() => <div>CONTENT</div>} />)
+  const renderLoading = (loaded: boolean, errored: boolean) => {
+    return mount(<LoadingContent loaded={loaded} errored={errored} />)
+  }
 
   describe("should render loading correctly", () => {
```
aliceriot: if you wouldn't mind I think we should just have `it` blocks here, I don't think we have so many tests that we need nested `describe` blocks

static/js/components/Loading_test.js
```diff
@@ -1,134 +1,44 @@
...
-        {
-          loaded:     true,
-          processing: false,
-          error:      undefined
-        }
-      ])
+      const wrapper = renderLoading(true, false)
       assert.equal(wrapper.text(), "CONTENT")
     })
   })
```
aliceriot: Is there a test that exercises the `updateRequirements` callback? if not I think it would be good to add a simple one that just ensures that a `sinon.stub` gets called.

static/js/containers/ChannelPage.js
```diff
@@ -101,20 +61,37 @@ class ChannelPage extends React.Component {
...
-    postsForChannel:    state.postsForChannel,
-    posts:              state.posts,
-    subscribedChannels: state.subscribedChannels
+    channelName,
+    channel,
+    posts:              safeBulkGet(postIds, state.posts.data),
+    subscribedChannels: state.subscribedChannels.loaded
+      ? safeBulkGet(state.subscribedChannels.data, state.channels.data)
+      : null,
+    loaded:  R.none(R.isNil, [channel]),
```
aliceriot: could this just be `!R.isNil(channel)` ?

static/js/containers/PostPage.js
```diff
@@ -107,23 +66,45 @@ class PostPage extends React.Component {
...
 
-  render() {
-    const { posts, channels } = this.props
-    return <Loading restStates={[posts, channels]} renderContents={this.renderContents} />
+const shouldUpdateRequirements = R.complement(R.allPass([R.eqProps("postID"), R.eqProps("channelName")]))
+
+const updateRequirements = ({ dispatch, channelName, postID, channel }) => {
+  if (!postID || !channelName) {
+    // should not happen, this should be guaranteed by react-router
+    throw Error("Match error")
```
aliceriot: should we redirect the user to `/` or something in case they hit this (unlikely) error?

static/js/lib/util.js
```diff
@@ -2,6 +2,7 @@
 import type { Match } from "react-router"
 
 export const getChannelName = (props: { match: Match }): string => props.match.params.channelName || ""
+export const getPostID = (props: { match: Match }): string => props.match.params.postID || ""
```
aliceriot: would `R.pathOr` be a little safer here?

gsidebo:
High-level comment: We would still have use for a `<Loader>` component that can just be used like a regular component, right? Example: 2 components on a page, each one depends on unrelated API calls and we want there to be an individual spinner that shows/hides for each one. It looks like we wouldn't be able to do that with the changes introduced in this PR. Am I right about that? Basically I think it would be useful to be able to have several individual loader components on a page rather than limiting us to a loader at the container level. If that's possible and I'm just misreading this lmk

static/js/containers/ChannelPage.js
```diff
@@ -101,20 +61,37 @@ class ChannelPage extends React.Component {
...
+    errored: anyError([state.channels, state.posts, state.subscribedChannels])
   }
 }
 
-export default connect(mapStateToProps)(ChannelPage)
+export default connect(mapStateToProps)(withLoading(ChannelPage, shouldUpdateRequirements, updateRequirements))
```
gsidebo: Should be written using `R.compose` to avoid the nesting:

```javascript
export default R.compose(
  withLoading(shouldUpdateRequirements, updateRequirements),
  connect(mapStateToProps)
)(ChannelPage);
```

`withLoading` can be rewritten with `R.curry` and accept the component as it's last argument. Check out [DashboardPage.js][email/hoc.js](https://github.com/mitodl/micromasters/blob/master/static/js/components/email/hoc.js) and (https://github.com/mitodl/micromasters/blob/master/static/js/containers/DashboardPage.js) in MM for implementation of the hoc and the client container.

static/js/containers/PostPage.js
```diff
@@ -107,23 +66,45 @@ class PostPage extends React.Component {
...
+    post,
+    channel,
+    commentsTree,
+    loaded:  R.none(R.isNil, [post, channel, commentsTree]),
+    errored: anyError([posts, channels, comments])
+  }
+}
 
-export default connect(mapStateToProps)(PostPage)
+export default connect(mapStateToProps)(withLoading(PostPage, shouldUpdateRequirements, updateRequirements))
```
gsidebo: Same comment as above re: `R.compose`

aliceriot:
@gsidebo yeah I think that makes sense, as I see it this change is addressing a different sort of use-case from e.g. showing a spinner on a button while a request is in-flight or something similar to that. this is more about having a clean and easy way to defer rendering a piece of data-driven UI until the relevant data is in place, so our UI code can be written in a more straightforward way.

rhysyngsun:
@gsidebo I agree with that. Fair to say that the part of the logic that displays "Loading" would use that component? We already have such a component over in MM which could be ported over to here: https://github.com/mitodl/micromasters/blob/master/static/js/components/Loader.js

I'd prefer to do that in another PR if that's ok.

static/js/components/Loading.js
```diff
@@ -1,27 +1,47 @@
...
+    componentWillMount() {
+      updateRequirements(this.props)
     }
 
-    if (anyError(restStates)) {
-      return <div>Error loading page</div>
+    componentDidUpdate(prevProps: Object) {
+      if (shouldUpdateRequirements(prevProps, this.props)) {
+        updateRequirements(this.props)
+      }
```
gsidebo: I know this was the convention before, I think `updateRequirements` and `shouldUpdateRequirements` are kind of awkward names here. What about `loadData` and `shouldLoadData`? To me that makes more sense in this context

static/js/containers/PostPage.js
rhysyngsun: I think we'd really want to know when this happens. I'm open to logging it to sentry and then redirecting, but I think we'd be causing trouble for ourselves to have logic that just quietly redirects. I don't think we have Raven setup for this repo yet though.

aliceriot: Oh yeah, I was suggesting we redirect in addition to throwing the error, just to try to improve the user experience.

static/js/containers/ChannelPage.js
```diff
@@ -101,20 +61,37 @@ class ChannelPage extends React.Component {
       )
     }
   }
+}
 
-  render() {
-    const { channels, postsForChannel } = this.props
-    return <Loading restStates={[channels, postsForChannel]} renderContents={this.renderContents} />
-  }
+const shouldUpdateRequirements = R.complement(R.eqProps("channelName"))
```
gsidebo: For readability's sake I'd strongly prefer `R.not` to `R.complement` here. Also a brief comment here explaining the logic might help. It takes a few mental leaps (at least for me) to determine that this is going to test if the `channelName` prop has changed after a component update

aliceriot: Just to clarify - `R.complement` and `R.not` aren't equivalent,

```
R.not :: Boolean -> Boolean
R.complement :: (Any -> Boolean) -> Boolean
```

`R.complement` takes a function as an argument, where `R.not` takes a boolean value. To rewrite with `R.not` you'd need to add a wrapping call to `R.compose`.

gsidebo: Ah, right. If we prefer the brevity, I'm OK with keeping it, but then I think a comment is even more important

static/js/components/Loading.js
rhysyngsun: it's shorter and clearer so yeah

static/js/containers/PostPage.js
rhysyngsun: well, if I throw the error then I can't redirect and vice versa. I'm inclined to leave it as-is for now and file a separate issue for Sentry integration linking to this line as one to fix.

static/js/lib/util.js
rhysyngsun: safer in what regard? I don't think we need to be quite that safe here, because if we don't have that data here it's a programmer error and we'd want it to fail hard. The `|| ""` is only here to quiet flow since it doesn't know better.

static/js/containers/ChannelPage.js
rhysyngsun: I actually ended up adjusting to include `postIds` as well, which it should.

static/js/containers/ChannelPage.js
rhysyngsun: That's much cleaner, although the args to `R.compose` needed to be the reverse of what you suggested since `withLoading` needs the props from `connect`.

rhysyngsun:
@gsidebo @aliceriot feedback addressed

static/js/containers/PostPage.js
aliceriot: oh true, haha. yeah, we can leave it as-is for now I think.

static/js/lib/util.js
aliceriot: fair enough, I meant safe in the sense that if `props.match` is `undefined` then calling `.params` on it will throw an error, but I think you're right that in that case we would want it to fail loudly.

static/js/components/Loading.js
```diff
@@ -1,27 +1,48 @@
...
+        <div className="loaded">
+          <LoadedComponent {...props} />
+        </div>
+      )
+    }
   }
-}
+})
+
+export default withLoading
```
gsidebo: if we define the HOC like this, I believe it's functionally equivalent and we could move those lifecycle methods back into the original containers:

```javascript
const withLoading = (
  LoadedComponent: Class<React.Component<*, *, *>>,
) => {
  return class extends LoadedComponent {
    props: LoadingProps

    render() {
      const { loaded, errored, ...props } = this.props

      if (errored) {
        return <div className="errored">Error loading page</div>
      }

      if (!loaded) {
        return <div className="loading">Loading</div>
      }

      return (
        <div className="loaded">
          { super.render(props) }
        </div>
      )
    }
  }
}
```

@aliceriot, any thoughts on that? Might be some drawbacks in terms of using inheritance, but I really strongly prefer this. The lifecycle methods exist in a place that you expect them to exist, and the original container still functions on its own with or without the loading HOC (not so with the current approach)

aliceriot: Personally I like the `updateData` and `shouldUpdateData` methods being defined as pure functions outside of the classes, this would make it easier to test them separately if they need to have any really intense logic in them, if we needed to do that.

I think you're right that that could work basically just as well, but IMO (and this is subjective) the approach @rhysyngsun has put together so far strikes me as more readable and with less indirection.

rhysyngsun: Neither `LoadedComponent.componentWillMount` nor `LoadedComponent.componentDidUpdate` will get called until the component mounts when `errored === false && loaded === true`, hence why I moved those implementations outside of the container. If we don't like this, I don't see the HOC approach being viable unless we hide the `LoadedComponent` with CSS or something hacky like that.

PR #110, opened 2017-08-14
render markdown posts and comments

### LINKED ISSUE #105: Format comments and post text

aliceriot:

this closes #105 

#### What's this PR do?

this adds the [react-markdown](https://github.com/rexxars/react-markdown) package and uses it to render comment and post text.

make a comment or text post containing some markdown stuff, make sure it works! you should be able to do fun stuff like this:

markdown

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "react-dom": "^15.6.1",
     "react-ga": "^2.2.0",
     "react-hot-loader": "next",
+    "react-markdown": "^2.5.0",
     "react-redux": "^5.0.5",
     "react-router": "^4.1.1",
     "react-router-dom": "^4.1.1",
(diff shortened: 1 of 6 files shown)


noisecapella:
I like the picture though I feel like we should disable image embeds since they could be used to track people. Is there a way to configure that with react-markdown?

aliceriot:
Yeah, we can set a list of node types which it should skip rendering - I already set it to skip rendering inline HTML for similar reasons. This seems like a product concern though, @roberthouse54 do you think it's important that users be able to embed images, or is it not important?

noisecapella:
I feel like there's a significant security risk there since random users can get IP addresses from other people. If we wanted to let users upload images we would need to host them ourselves. Can we ban images for now and figure out how to do this in a future PR?

PR #116, opened 2017-08-17
Added user APIs

### LINKED ISSUE #113: user API
A privileged application like MicroMasters needs a basic user API to create open-discussions users (among other things), so that they can be automatically subscribed to channels. 

Should allow storing of name and profile image

### Questions:
- [x] What do we use as the username on open-discussions, a UUID or the MM username?

rhysyngsun:
Part of #113 

#### What's this PR do?
Adds a user api and related models

- Via the DRf web interface, go to /api/v0/users/, verify you see a list of users.
- Verify you can create a new user, a username should be generated for you.
- Verify you can patch the nested profile record

Added here is a user `Profile` model, where we store the full name and profile images. For simplicity, the profile is patching via a nested serializer.

**Note:** not dealing with any authentication/authorization in this PR, it is solely the data model and accompanying APIs.


pdpinch:
@rhysyngsun do we need to open an issue for adding authentication/authorization?

rhysyngsun:
@pdpinch yeah, I think it would make sense to do the MM->OD auth seperately

Issue #125, opened 2017-08-29
Tweaks to post page display


aliceriot:
- [x] `PostDisplay` component should have slightly different layout on `PostPage` vs on a channel or the frontpage
- [x] the top-level comment reply (reply to the post) should always be open, unlike the comment reply text boxes
- [x] the comment 'reply' button should be to the left of the upvote / downvote buttons
- [x] the comment count should live in the gutter between the post card and the first comment card
- [x] the post page should have the navigation sidebar on it

PR #131, opened 2017-08-31
post page tweaks

aliceriot:

part of #125 

this does everything except moving the 'reply' button, which is a little more complicated (so I'm putting it off for now).

#### What's this PR do?

- fixes the PostDisplay layout, so that it shows the voting buttons in the right place and doesn't show comments when on the post detail view page
- show the comment count on the post page in the gutter between the post and the first comment
- add the navigation sidebar to the post page
- some small CSS layout cleanups / tweaks, moving a few things that were manually positioned to use flexbox instead.

#### testing

Just make sure everything looks good and is cleanly laid out, on a phone and on the desktop. We should be getting closer to the mocks for the post page. There should be no style regressions on the frontpage or the channel page display.

### DIFF
--- a/static/js/components/CommentVoteForm.js
+++ b/static/js/components/CommentVoteForm.js
@@ -58,30 +58,22 @@ export default class CommentVoteForm extends React.Component {
         <div className="score">
           {comment.score}
         </div>
-        <div className="votes">
-          <button
-            className={`vote upvote-button ${comment.upvoted ? "upvoted" : ""}`}
-            onClick={this.upvote}
-            disabled={voting}
-          >
-            <img
-              className="upvote-arrow"
-              src="/static/images/upvote_arrow.png"
-            />
-          </button>
-          <button
-            className={`vote downvote-button ${comment.downvoted
-              ? "downvoted"
-              : ""}`}
-            onClick={this.downvote}
-            disabled={voting}
-          >
-            <img
-              className="upvote-arrow"
-              src="/static/images/downvote_arrow.png"
-            />
-          </button>
-        </div>
+        <button
+          className={`vote upvote-button ${comment.upvoted ? "upvoted" : ""}`}
+          onClick={this.upvote}
+          disabled={voting}
+        >
+          <img className="vote-arrow" src="/static/images/upvote_arrow.png" />
+        </button>
+        <button
+          className={`vote downvote-button ${comment.downvoted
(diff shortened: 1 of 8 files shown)


noisecapella:
The padding looks a little off in the voting buttons:

screenshot from 2017-08-31 14-53-01

static/js/components/CommentVoteForm.js
```diff
@@ -58,30 +58,25 @@ export default class CommentVoteForm extends React.Component {
...
+        </button>
+        <button
+          className={`vote downvote-button ${comment.downvoted
+            ? "downvoted"
+            : ""}`}
+          onClick={this.downvote}
+          disabled={voting}
+        >
+          <img
+            className="upvote-arrow"
```
noisecapella: This was here before but should we change this to `downvote-arrow` for clarity? Or just `vote-arrow`?

aliceriot:
I removed the `border-right` that snuck in - how do you think the padding should be changed? as far as I can tell they're lined up currently:

buttons_lined_up

noisecapella:
No, the `border-right` is what I meant, the rest of the padding looks fine

static/js/components/PostDisplay.js
```diff
@@ -69,28 +70,33 @@ class PostDisplay extends React.Component {
...
-            <img
-              className="upvote-arrow"
-              src="/static/images/upvote_arrow.png"
-            />
-          </button>
-          <span className="votes">
-            {post.score}
-          </span>
-        </div>
+      <div className={`upvotes ${upvoteClass} ${expanded ? "expanded" : ""}`}>
```
noisecapella: What does the CSS class `expanded` do here?

aliceriot: basically I'm just using it so I can choose between `flex-direction: row` and `flex-direction: column`

aliceriot: b/c when we're showing a post in a `PostList` (i.e. in an index list of posts) the voting buttons appear in a column, but on the post page they appear in a row.

static/js/components/CommentVoteForm.js
aliceriot: yeah that's probably a good idea

aliceriot: I went with `vote-arrow` for now

Issue #127, opened 2017-08-29
tweaks to PostList component


aliceriot:
- if there are no posts to display, it should display something like `Hey there are no posts here ok'
- tweak the loading spinner, it should be in the middle and should have a spinner or something.

PR #135, opened 2017-08-31
add spinner to loader component

aliceriot:

part of #127 

#### What's this PR do?

this adds a spinner component to the `Loader` component that we have, so we show something a little nicer to look at than just the `Loading` text. I also added the `withLoading` HOC to the home page component, where it hadn't yet been added.

Visit a post page, a channel page, and the front page. you should see a spinner-ey thing there (when the requests are in flight) instead of some text that just says 'Loading'.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "sass-lint": "^1.10.2",
     "sass-loader": "^6.0.6",
     "sinon": "^2.3.8",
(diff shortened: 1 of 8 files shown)


noisecapella:
Looks like the test failure is just linting

aliceriot:
oh oops, haha

PR #138, opened 2017-09-01
move reply button

### LINKED ISSUE #125: Tweaks to post page display

aliceriot:

closes #125 

#### What's this PR do?

This makes the changes necessary to move the 'Reply' button to be next to the voting buttons for comments.

Replying to comments, posts, etc should work as normal.

### DIFF
--- a/static/js/components/CommentTree.js
+++ b/static/js/components/CommentTree.js
@@ -9,6 +9,11 @@ import Card from "./Card"
 import { ReplyToCommentForm } from "./CreateCommentForm"
 import CommentVoteForm from "./CommentVoteForm"
 
+import {
+  replyToCommentKey,
+  getCommentReplyInitialValue
+} from "../components/CreateCommentForm"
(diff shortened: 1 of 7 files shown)


noisecapella:
It looks like the reply button isn't calling `preventDefault` here. I added a print statement to `beginReply` and it looks like `e` is undefined

aliceriot:
oh yeah I had to add that because in the enzyme test it wasn't passing an `Event` to the `onClick` handler, I'll add it back real quick.

PR #140, opened 2017-09-01
Added Rest APIs to add/remove contributors in a channel

### LINKED ISSUE #132: Add/remove channel contributor API
Users need to be added as contributors to see the content of a private channel.

The `remove` part must not remove administrators (moderators in the reddit terminology)

giocalitri:
closes #132 

#### What's this PR do?
Adds a REST API to add and remove contributors in a channel

if you disable the `JwtIsStaffOrReadonlyPermission` in the views, you can use the django rest framework UI to perform the requests and you can go to `http://reddit.local/r/<channel_name>/about/contributors/` and verify that the action you perform via rest API correspond to what you see in the UI

### DIFF
--- a/channels/api.py
+++ b/channels/api.py
@@ -2,10 +2,14 @@
 from urllib.parse import urljoin
 
 import requests
+from django.conf import settings
+from django.contrib.auth import get_user_model
 import praw
 from praw.models.reddit import more
+from praw.models.reddit.redditor import Redditor
+from rest_framework.exceptions import NotFound
 
-from django.conf import settings
+from channels import exceptions
 
 CHANNEL_TYPE_PUBLIC = 'public'
 CHANNEL_TYPE_PRIVATE = 'private'
@@ -25,6 +29,7 @@ CHANNEL_SETTINGS = (
     'submit_text',
     'submit_text_label',
 )
+User = get_user_model()
 
 
 def _get_user_credentials(user):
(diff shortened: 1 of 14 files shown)


channels/api.py
```diff
@@ -361,3 +366,52 @@ def more_comments(self, comment_fullname, parent_fullname, count, children=None)
...
+
+        """
+        try:
+            user = User.objects.get(username=contributor_name)
+        except User.DoesNotExist:
+            raise NotFound("User {} does not exist".format(contributor_name))
+        channel = self.get_channel(channel_name)
+        if contributor_name in channel.moderator():
+            raise exceptions.RemoveUserException(
+                "User {} is a moderator of the channel{}".format(contributor_name, channel_name))
```
noisecapella: Can you add a space between `channel` and `{}`?

noisecapella:
Functionality looks great

channels/urls.py
```diff
@@ -20,6 +22,16 @@
         name='post-list',
     ),
     url(
+        r'^api/v0/channels/(?P<channel_name>[A-Za-z0-9_]+)/contributors/$',
+        ContributorListView.as_view(),
+        name='contributors-list',
```
noisecapella: Can you change this to `contributor-list` and the below to `contributor-detail`, for consistency with the other URL names?

channels/api_test.py
```diff
@@ -261,3 +266,56 @@ def test_frontpage(mock_client):
...
+    assert mock_client.subreddit.return_value.contributor.remove.call_count == 0
+
+
+def test_remove_contributor(mock_client):
+    """Test remove contributor"""
+    client_user = UserFactory.create()
+    contributor = UserFactory.create()
+    client = api.Api(client_user)
+    client.remove_contributor(contributor.username, 'foo_channel_name')
+    mock_client.subreddit.return_value.contributor.remove.assert_called_once_with(contributor)
```
annagav: this only tests if the method was called, but it does not test if the user was removed from the list of contributors.

giocalitri: well... the assumption is that if the library is called and it does not return an error, the user has been removed.

Actually the problem is even more complicated, given that `praw` will not return errors if the user is not a contributor, for example.

PR #148, opened 2017-09-07
Added Rest APIs to add/remove moderators in a channel

### LINKED ISSUE #128: Add channel_admins API
A privileged application like MicroMasters needs to be able to add and remove admins (contributors and moderators) on channels/subreddits. 

We'll need a REST API wrapper for the reddit API, that has similar authentication and authorization to the create_channel API.

annagav:
Fix #128

#### What's this PR do?
Adds REST API to add and remove moderators in a channel.

Login as a moderator, create a user and make sure that the user exists on reddit, try adding and removing this user as a moderator.


channels/api.py
```diff
@@ -361,3 +366,50 @@ def more_comments(self, comment_fullname, parent_fullname, count, children=None)
...
+            moderator_name(str): username of the user
+            channel_name(str): name of the channel
+
+        Returns:
+            praw.models.Redditor: the reddit representation of the user
+        """
+        try:
+            user = User.objects.get(username=moderator_name)
+        except User.DoesNotExist:
+            raise Exception("User {} does not exist".format(moderator_name))
```
noisecapella: Can you change this to use `NotFound` similar to how it's used in https://github.com/mitodl/open-discussions/pull/140/files?

channels/urls.py
```diff
@@ -20,6 +22,16 @@
         name='post-list',
     ),
     url(
+        r'^api/v0/channels/(?P<channel_name>[A-Za-z0-9_]+)/moderators/$',
+        ModeratorListView.as_view(),
+        name='moderators-list',
```
noisecapella: Can you change this to `moderator-list` and the below line to `moderator-detail`?

channels/api.py
```diff
@@ -434,9 +434,20 @@ def add_moderator(self, moderator_name, channel_name):
             raise NotFound("User {} does not exist".format(moderator_name))
 
         self.get_channel(channel_name).moderator.add(user)
-
+        new_moderator = Api(user)
+        new_moderator.accept_invite(channel_name)
+        del new_moderator
```
noisecapella: FYI You generally should never need to do this, it will get deleted when the local variable goes out of scope (when the reference count is 0)

channels/api.py
```diff
@@ -434,9 +434,20 @@ def add_moderator(self, moderator_name, channel_name):
             raise NotFound("User {} does not exist".format(moderator_name))
 
         self.get_channel(channel_name).moderator.add(user)
-
+        new_moderator = Api(user)
+        new_moderator.accept_invite(channel_name)
```
noisecapella: Did you talk to @giocalitri about this approach?

channels/serializers.py
```diff
@@ -312,3 +311,26 @@ def create(self, validated_data):
...
+    moderator_name = WriteableSerializerMethodField()
+
+    def get_moderator_name(self, instance):
+        """Returns the name for the moderator"""
+        return instance.name
+
+    def validate_moderator_name(self, value):
+        """Validates the moderator name"""
+        if not isinstance(value, str):
+            raise ValidationError("contributor name must be a string")
```
noisecapella: Can you change this to `moderator name must be a string`? And also modify the next validation error below

channels/api.py
```diff
@@ -1,4 +1,5 @@
 """Channels APIs"""
+# pylint: disable=R0904
```
noisecapella: Can you change this to the written out version for clarity?

noisecapella: I mean change to `disable=too-many-public-methods`

channels/api.py
```diff
@@ -432,9 +432,10 @@ def add_moderator(self, moderator_name, channel_name):
             user = User.objects.get(username=moderator_name)
         except User.DoesNotExist:
             raise NotFound("User {} does not exist".format(moderator_name))
-
-        self.get_channel(channel_name).moderator.add(user)
-        Api(user).accept_invite(channel_name)
+        channel = self.get_channel(channel_name)
+        if moderator_name not in channel.moderator():
+            channel.moderator.add(user)
+            Api(user).accept_invite(channel_name)
```
noisecapella: It looks like an error is raised if there is no invite already. Can you wrap this in a `try: except APIException:` block and log the exception if one happens?

PR #154, opened 2017-09-08
add profile image to comments serializer

### LINKED ISSUE #123: set up default profile image if user does not have one
We should do something similar to what we did in MM, so that a default profile image is used if the user profile hasn't had profile image data synced. This will be useful for development especially.

aliceriot:

closes #123 

#### What's this PR do?

This adds profile images to the comments serializer and sets up the comments views to deal with that efficiently. Also adds profile image to the post serializer.

If you look at a post, the comments for any users which have `image_small` set on their profile should include the image. Any other users should have a default 'person' icon thing shown instead. Doing any UI actions which modify comments (adding a reply, upvoting, etc) should retain the profile images. There should also be a profile image shown for the post author, next to the post at the top.


channels/serializers.py
```diff
@@ -229,6 +231,14 @@ def get_upvoted(self, instance):
         """Is a comment upvoted?"""
         return instance.likes is True
 
+    def get_profile_image(self, instance):
+        """Find the Profile for the comment author"""
+        profile = self.context['request'].user.profile
+        if (profile and profile.image_small):
+            return profile.image_small
+        else:
+            return "/static/images/avatar_default.png"
```
noisecapella: We should get the static URL from Django so we can use cloudfront to serve this file. I found this code which is used by the Django static templatetag, it might work here:

    from django.contrib.staticfiles.storage import staticfiles_storage
    return staticfiles_storage.url(path)

channels/serializers.py
```diff
@@ -229,6 +231,14 @@ def get_upvoted(self, instance):
         """Is a comment upvoted?"""
         return instance.likes is True
 
+    def get_profile_image(self, instance):
+        """Find the Profile for the comment author"""
+        profile = self.context['request'].user.profile
+        if (profile and profile.image_small):
```
noisecapella: The parentheses around the if expression are unnecessary in Python. Also `profile` should exist for all users since it is created by a signal when the user is created

channels/serializers.py
```diff
@@ -229,6 +231,14 @@ def get_upvoted(self, instance):
         """Is a comment upvoted?"""
         return instance.likes is True
 
+    def get_profile_image(self, instance):
+        """Find the Profile for the comment author"""
+        profile = self.context['request'].user.profile
```
noisecapella: This `user` is the user viewing the API, not the user for the profile

aliceriot: yeah I just fixed that

noisecapella:
Regarding your comment on Slack on the database queries, you can query the users using `DiscussionUser.objects.filter(author__in=set_of_comment_users)`, put it in a dict, then pass the dict to the serializer in its context for use as a cache. There's a `get_serializer_context(...)` method you can override on the view where you can do this

noisecapella:
One other case we might need to handle is deleted comments, which will have `null` for the author

aliceriot:
cool, I'll add some code to set the serializer context correctly.

aliceriot:
I added some code in the view to set the serializer context to a dict mapping `username` to `profile_image`, so in the serializer we just have to do `self.context["profile_images"].get(username)`. With the caching setup on the `get_queryset` function I'm pretty sure it will do what we want overall.

aliceriot:
@noisecapella I think this is ready for review!

channels/serializers.py
```diff
@@ -121,11 +122,23 @@ class PostSerializer(serializers.Serializer):
...
+    def get_profile_image(self, instance):
+        """Find the Profile for the comment author"""
+        if instance.author is None:
+            return default_profile_image
+
+        profile = User.objects.get(username=instance.author.name).profile
+        if profile.image_small:
+            return profile.image_small
+        else:
+            return default_profile_image
```
noisecapella: Can you add a test for this case?

channels/serializers.py
```diff
@@ -233,6 +247,21 @@ def get_upvoted(self, instance):
...
+        # so we can get the profile image from instance.author
+        if self.context.get("profile_images"):
+            image = self.context["profile_images"].get(instance.author.name)
+        else:
+            image = User.objects.get(username=instance.author.name).profile.image_small
+
+        if image:
+            return image
+        else:
+            return default_profile_image
```
noisecapella: Can you add a test for this case?

channels/serializers.py
```diff
@@ -233,6 +247,21 @@ def get_upvoted(self, instance):
         """Is a comment upvoted?"""
         return instance.likes is True
 
+    def get_profile_image(self, instance):
+        """Find the Profile for the comment author"""
+        # context["profile_images"] is set only when doing GET
+        # on comment list view. else we're dealing with one record,
+        # so we can get the profile image from instance.author
+        if self.context.get("profile_images"):
```
noisecapella: Can you change this to `if "profile_images" in self.context`?

channels/views.py
```diff
@@ -80,10 +84,29 @@ class CommentListView(ListCreateAPIView):
...
+    def get_serializer_context(self):
+        context = {
+            "request": self.request,
+            "view": self,
+            "format": self.format_kwarg,
+        }
+
+        if self.request.method == 'GET':
+            comments = self.get_queryset()
+            commentUsers = User.objects.filter(
```
noisecapella: This should be `comment_users` to fit with Python naming conventions

channels/views.py
```diff
@@ -80,10 +84,29 @@ class CommentListView(ListCreateAPIView):
     """
     serializer_class = CommentSerializer
 
+    def get_serializer_context(self):
+        context = {
```
noisecapella: Can you get this from the super class instead: `context = super().get_serializer_context()`

channels/views_test.py
```diff
@@ -4,12 +4,23 @@
...
 from open_discussions.factories import UserFactory
+from profiles.factories import ProfileFactory
+from channels.serializers import default_profile_image
 
 
 # pylint: disable=redefined-outer-name, unused-argument
 pytestmark = pytest.mark.django_db
 
 
+def create_profile_and_login(client):
```
noisecapella: Can you change this to a fixture? Maybe something like `logged_in_profile`

channels/views_test.py
```diff
@@ -4,12 +4,23 @@
...
+    """little helper function"""
+    profile = ProfileFactory.create()
```
noisecapella: You can set attributes via the dunder syntax with factories, like `ProfileFactory.create(user__username='george')`

channels/views_test.py
```diff
@@ -691,10 +721,14 @@ def test_add_contributor(client, use_betamax, praw_settings, staff_jwt_header):
     """
     Adds a contributor to a channel
     """
-    client.force_login(UserFactory.create(username='fooadmin'))
-    contributor = UserFactory.create(username='othercontributor')
+    admin = ProfileFactory.create()
+    admin.user.username = 'fooadmin'
```
noisecapella: This is missing a `.save()`, but you can also change this to `ProfileFactory.create(user__username='fooadmin')`

channels/views_test.py
```diff
@@ -703,10 +737,13 @@ def test_detail_contributor_error(client, use_betamax, praw_settings):
     """
     Detail of a contributor in a channel in case the user is not a contributor
     """
-    client.force_login(UserFactory.create(username='fooadmin'))
-    nocontributor = UserFactory.create(username='nocontributor')
+    admin = ProfileFactory.create()
+    admin.user.username = 'fooadmin'
```
noisecapella: This and below need a `.save()`

channels/serializers.py
aliceriot: I believe the `test_get_post` test (in `channels/views_test.py`) covers this case, unless I'm mistaken

channels/serializers.py
aliceriot: ah, didn't know I could do that! snazzy

channels/views.py
aliceriot: oh good idea

channels/serializers.py
noisecapella: It looks like `create_profile_and_login` uses `ProfileFactory` to create the profile, which fills in values for the images here, so `profile.image_small` would always be truthy

channels/serializers.py
aliceriot: sure

channels/views_test.py
aliceriot: sure

channels/serializers.py
aliceriot: yes, but there's a test where `post.author` is `None`, which tests that path.

aliceriot: oh nvm yeah I do need to add another one :woman_facepalming:

noisecapella:
Functionality looks good, found an unrelated issue #161

channels/serializers.py
aliceriot: A D D E D what an odyssey that was

aliceriot:
@noisecapella I think all the issues should be fixed!

channels/test_utils.py
```diff
@@ -0,0 +1,18 @@
+"""utils for dealing with channel tests"""
+from functools import partialmethod
+import contextlib
+import warnings
+import requests
+
+
+@contextlib.contextmanager
```
noisecapella: Did you mean to keep this around? I think it would be preferable just to remove the `filterwarnings` from `pytest.ini` when we create these cassettes

aliceriot: There is still the exception raised in Praw, so even after removing the `filterwarnings` I needed to use this to make the requests for the cassette. I figured I might as well check it in so others can use it to record cassettes in the future.

PR #156, opened 2017-09-11
Add user on reddit when user is added via REST API

### LINKED ISSUE #155: Add user on reddit when user is created via REST API

noisecapella:
Fixes #155

#### What's this PR do?
Adds the user using the refresh token plugin when the user is created using the `/api/v0/users/` REST API

Create a user using `/api/v0/users/`, then go to reddit and verify that the user exists there (go to `/u/username`

### DIFF
--- a/channels/api.py
+++ b/channels/api.py
@@ -33,6 +33,23 @@ CHANNEL_SETTINGS = (
 User = get_user_model()
 
 
+def get_or_create_user(username):
+    """
+    Get or create a user on reddit using our refresh_token plugin
+
+    Args:
+        username (str): The reddit username
+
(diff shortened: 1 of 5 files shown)


channels/api.py
```diff
@@ -32,6 +32,23 @@
 User = get_user_model()
 
 
+def get_or_create_user(username):
+    """
+    Get or create a user on reddit using our refresh_token plugin
```
giocalitri: nit: I would add some comments about the fact that this relies on our modified version of the reddit code

noisecapella: done

giocalitri:
👍

PR #158, opened 2017-09-11
Added JWT session renewal

### LINKED ISSUE #153: Handle expired MM tokens
Part of https://github.com/mitodl/micromasters/issues/3479

- [ ] perform a background request browserside to MM to get a new token.
- [ ] If the MM session expired or they don't have a token (and thus we can't get a new token), a hard redirect to MM is performed to login and issue new token. User is then redirected back to OD for the page they were trying to access.

rhysyngsun:
Fixes #153 
Fixes mitodl/micromasters#3479

#### What's this PR do?
Adds code to renew JWT session and redirect to MM if logged out

Follow configuration steps in mitodl/micromasters#3523 and verify in the chrome inspector that when your JWT token expires it queries MM for a new token at `/api/v0/discussion_token` and retries the failed API request(s).

Use this setting in MM  `.env` to lower your JWT lifespan to 30s:

```
OPEN_DISCUSSIONS_JWT_EXPIRES_DELTA=30
```

### DIFF
--- a/open_discussions/views.py
+++ b/open_discussions/views.py
@@ -2,9 +2,12 @@
 open_discussions views
 """
 import json
+from datetime import timedelta
 
+import jwt
 from django.conf import settings
 from django.shortcuts import render
+from rest_framework_jwt.settings import api_settings
 
 from open_discussions.templatetags.render_bundle import public_path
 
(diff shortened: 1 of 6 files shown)


rhysyngsun:
Need to solve the issue where the JWT expires and we can no longer validate it to get the `session_url` and `auth_url`. Reproduced with a browser refresh after the token expires.

rhysyngsun:
@aliceriot fixed the issue on browser refresh

aliceriot:
I set the MM token to expire in 30 seconds and can confirm that new tokens are now successfully fetched upon a full-browser refresh :confetti_ball: :tada: 

one question: when this happens there is a rejected promise that shows up in the console:

unfjfjffj

do you think that's ok? might be better than the user can't see such an error.

rhysyngsun:
@aliceriot I don't think we have a choice: https://stackoverflow.com/a/42986081

aliceriot:
@rhysyngsun yeah realized I was confusing myself >_<

PR #162, opened 2017-09-12
Increased max length of image fields

### LINKED ISSUE #161: Increase max length of profile image fields
The max length used makes sense for relative URLs used for images hosted on the server, but it won't work for S3 URLs and external URLs

noisecapella:
Fixes #161 

#### What's this PR do?
Increases max length of image char fields for the profile to 1024 characters

Find an image URL which is long (for example one of the avatars from micromasters) and set `profile.image_small` to it. You should not get an error.


profiles/models.py
```diff
@@ -2,8 +2,7 @@
 from django.db import models
 from django.conf import settings
 
-# This matches ImageField.max_length
-MAX_IMAGE_FIELD_LENGTH = 100
+MAX_IMAGE_FIELD_LENGTH = 1024
```
giocalitri: Why this other size?

noisecapella: It's less than the max size which is about 2k, and I think it's large enough to handle the URLs we'll probably use

giocalitri:
👍

PR #167, opened 2017-09-12
Added add/remove subscriber REST API

### LINKED ISSUE #112: subscribe API
A privileged application like MicroMasters needs to be able to subscribe (and eventually unsubscribe) users from a channel. 

The scope of this issue is limited to a simple subscription API. It should be implemented similar to the APIs for adding contributors (https://github.com/mitodl/open-discussions/pull/140) and moderators (https://github.com/mitodl/open-discussions/pull/148).

giocalitri:
closes #112

#### What's this PR do?
Adds add/remove subscriber REST API

Use the API and check the results on the reddit UI

### DIFF
--- a/channels/views_test.py
+++ b/channels/views_test.py
@@ -698,6 +698,14 @@ def test_list_moderators(client, use_betamax, praw_settings):
     assert resp.json() == [{'moderator_name': 'fooadmin'}]
 
 
+def test_list_subscribers_not_allowed(client, staff_jwt_header):
+    """
+    Get method not allowed on the list of subscribers
+    """
+    url = reverse('subscriber-list', kwargs={'channel_name': 'test_channel'})
+    assert client.get(url, **staff_jwt_header).status_code == status.HTTP_405_METHOD_NOT_ALLOWED
+
+
 def test_add_contributor(client, use_betamax, praw_settings, staff_jwt_header):
     """
     Adds a contributor to a channel
@@ -720,6 +728,17 @@ def test_add_moderator(client, use_betamax, praw_settings, staff_jwt_header):
     resp = client.post(url, data={'moderator_name': moderator.username}, format='json', **staff_jwt_header)
     assert resp.status_code == status.HTTP_201_CREATED
     assert resp.json() == {'moderator_name': 'othermoderator'}
+
+
+def test_add_subscriber(client, use_betamax, praw_settings, staff_jwt_header):
+    """
+    Adds a subscriber to a channel
+    """
+    subscriber = UserFactory.create(username='othersubscriber')
+    url = reverse('subscriber-list', kwargs={'channel_name': 'test_channel'})
+    resp = client.post(url, data={'subscriber_name': subscriber.username}, format='json', **staff_jwt_header)
+    assert resp.status_code == status.HTTP_201_CREATED
+    assert resp.json() == {'subscriber_name': 'othersubscriber'}
 
 
 def test_detail_contributor_error(client, use_betamax, praw_settings):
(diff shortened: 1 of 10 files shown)


channels/views_test.py
```diff
@@ -722,6 +731,18 @@ def test_add_moderator(client, use_betamax, praw_settings, staff_jwt_header):
...
+def test_add_subscriber(client, use_betamax, praw_settings, staff_jwt_header):
+    """
+    Adds a subscriber to a channel
+    """
+    client.force_login(UserFactory.create(username='fooadmin'))
+    subscriber = UserFactory.create(username='othersubscriber')
+    url = reverse('subscriber-list', kwargs={'channel_name': 'test_channel'})
+    resp = client.post(url, data={'subscriber_name': subscriber.username}, format='json', **staff_jwt_header)
+    assert resp.status_code == status.HTTP_201_CREATED
+    assert resp.json() == {'subscriber_name': 'othersubscriber'}
```
rhysyngsun: Could you add a test to verify that this fails for a nonstaff user trying to modify a second user? Same for the remove.

channels/views_test.py
```diff
@@ -722,6 +731,18 @@ def test_add_moderator(client, use_betamax, praw_settings, staff_jwt_header):
     assert resp.json() == {'moderator_name': 'othermoderator'}
 
 
+def test_add_subscriber(client, use_betamax, praw_settings, staff_jwt_header):
+    """
+    Adds a subscriber to a channel
+    """
+    client.force_login(UserFactory.create(username='fooadmin'))
```
rhysyngsun: Shouldn't need to force a login since the `staff_jwt_header` is what we'll end up using to authenticate the user in normal cases. This is basically providing 2 auth mechanisms in this test. The JWT should take precedence anyway, but I wouldn't want us to have a false positive in case something with the JWT breaks.

channels/views_test.py
```diff
@@ -747,6 +768,19 @@ def test_detail_contributor(client, use_betamax, praw_settings):
     assert resp.json() == {'contributor_name': 'othercontributor'}
 
 
+def test_detail_subscriber(client, use_betamax, praw_settings):
+    """
+    Detail of a subscriber in a channel
+    """
+    client.force_login(UserFactory.create(username='fooadmin'))
```
rhysyngsun: Use `jwt_header` instead.

channels/views_test.py
```diff
@@ -781,3 +815,15 @@ def test_remove_contributor_moderator(client, use_betamax, praw_settings, staff_
         'contributor-detail', kwargs={'channel_name': 'test_channel', 'contributor_name': contributor.username})
     resp = client.delete(url, **staff_jwt_header)
     assert resp.status_code == status.HTTP_409_CONFLICT
+
+
+def test_remove_subscriber(client, use_betamax, praw_settings, staff_jwt_header):
+    """
+    Removes a subscriber from a channel
+    """
+    client.force_login(UserFactory.create(username='fooadmin'))
```
rhysyngsun: See my comment above.

channels/views_test.py
```diff
@@ -698,6 +698,15 @@ def test_list_moderators(client, use_betamax, praw_settings):
     assert resp.json() == [{'moderator_name': 'fooadmin'}]
 
 
+def test_list_subscribers_not_allowed(client):
+    """
+    Get method not allowed on the list of subscribers
+    """
+    client.force_login(UserFactory.create(username='fooadmin'))
```
rhysyngsun: You should use `jwt_header` fixture instead (see my comment on the next test).  That one is for a nonstaff user.

rhysyngsun:
Functionality looks good

channels/views_test.py
giocalitri: I am not sure why I should do this: isn't that tested for the `JwtIsStaffOrReadonlyPermission` permission class?

rhysyngsun: Yeah, that class is tested, my thought was we'd want to know if that breaks against an api for whatever reason. That could break independently of the behavior of `JwtIsStaffOrReadonlyPermission`. I guess we don't test it elsewhere so i'm fine leaving it.

PR #170, opened 2017-09-13
add preventDefault for cancel button on create post page

### LINKED ISSUE #151: BUG: Error after clicking cancel button on Submit New Post Page
I receive an error after clicking the cancel button on the Submit New Post Page

aliceriot:

closes #151 

#### What's this PR do?

the cancel button in the create post form didn't have `e.preventDefault()`, so the form was issuing an unintentional POST.

go to the create post page, and click cancel. there shouldn't be any errors in the console, and it should instead just take you back to the subreddit. Also it's probably a good idea to confirm that you can reproduce the error (on `master`) by following the same steps.

note: I had trouble writing a regression test for this in enzyme, I think because of React's event system vs the browser one or something like that. Not sure exactly, but if the reviewer has ideas about strategies for that that would be :100:

### DIFF
--- a/static/js/components/CreatePostForm.js
+++ b/static/js/components/CreatePostForm.js
@@ -1,5 +1,6 @@
 // @flow
 import React from "react"
+import R from "ramda"
 
 import Card from "../components/Card"
 import ChannelBreadcrumbs from "../components/ChannelBreadcrumbs"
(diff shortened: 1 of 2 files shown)


noisecapella:
Functionality works fine. Does the test work if you just call `onClick` directly via `props().onClick`?

aliceriot:
oh yeah that's a good idea! I will add such a test now.

aliceriot:
ok I added a basic regression test, which fails on master but passes on this branch.

PR #171, opened 2017-09-13
disable post submit while it is in progress

### LINKED ISSUE #150: Disable submit button after click
On the Submit Post page, I am able to click the Submit Button multiple times... this will cause multiple posts to be created. We should disable the button after it has been clicked. We should also do this for the comment submit buttons.

### LINKED ISSUE #175: clicking 'submit' with a blank post-reply comment causes an error
I think it's the same thing as #151

aliceriot:

closes #150 
closes #175

#### What's this PR do?

this disables the 'submit' button on the create post form when the request is in-flight, and it also adds a `.disabled` class to the button so we can grey it out a little bit.

If you go to the create post form you should be able to create a post as normal. when you click the button it should get greyed out, and you should not be able to click multiple times and create duplicate posts.

### DIFF
--- a/static/js/components/CommentTree.js
+++ b/static/js/components/CommentTree.js
@@ -24,12 +24,21 @@ const renderTopLevelComment = R.curry(
     upvote: CommentVoteFunc,
     downvote: CommentVoteFunc,
     beginReply,
+    processing: boolean,
     comment: Comment,
     idx: number
   ) =>
     <Card key={idx}>
       <div className="top-level-comment">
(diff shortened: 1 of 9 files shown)


noisecapella:
The issue says this should also be done for the comment submit button. Can you do that in this PR too, or remove the `closes ` part of the PR description?

aliceriot:
oh I totally missed that, haha. I'll do that as well

aliceriot:
I took care of that and I also fixed another bug I noticed ( #175 )

PR #182, opened 2017-09-19
Only redirect to auth on a 401 response

### LINKED ISSUE #181: Redirect to auth behavior can cause a redirect loop
The auth behavior should only redirect on a 401 response, which should help prevent redirect issues related to other non-auth errors

noisecapella:
Fixes #181

#### What's this PR do?
Changes behavior to redirect only on a 401 status code

Add a `raise Exception()` to the frontpage API to simulate an error case. Add another `raise Exception()` to the `discussions_token` function on Micromasters. Check out master on open-discussions and go to `/discussions/` on micromasters. You should see a redirect loop. Check out this PR and the loop should not happen anymore.


aliceriot:
functionality works!

Issue #184, opened 2017-09-20
Figure out error cases when users don't exist in all servers


noisecapella:
For example, if a user exists on reddit and not in open-discussions or micromasters, if they have posted in a thread it will cause an exception. There shouldn't be any extra users on reddit, but we may want to add code to guard against this anyway, and for a missing reddit user when the open-discussion user exists

noisecapella:
- User exists on MM and reddit, not on OD: we should show post or comment as deleted (for now maybe something like `[deleted]` in the title and username output). Without the username mapping we don't have any idea who the user is, and the text won't make sense in that context.
 - User exists only on reddit: show post or comment as deleted (see previous)
 - User exists on MM and OD, not on reddit. This necessarily means they have no posts, comments or subreddits. When the user tries to access an OD api we will attempt to create the user via generate_refresh_token plugin. If that fails we should pass an error to the user, but they can try again later.
 - User exists on OD and reddit, not on MM. If the user goes straight to open-discussions they should not be able to login without first creating a user on MM.
- User exists only on OD: user should not be able to login, and their posts will show up as deleted
 - User exists only on MM: Next time their profile is saved it will get created on OD. Otherwise they should get an error in the recent posts UI (currently the API call just fails, leaving the list blank) and calls synchronizing information on the backend should error similarly.

PR #198, opened 2017-09-26
Use default profile image if the user does not exist in open-discussions

noisecapella:
Part of #184 

#### What's this PR do?
Works around exception when user does not exist in open-discussions:
 - in a list of posts (frontpage or channel view) the ones missing a user should be removed from the list as if they were deleted
 - in a list of comments the comments missing a user should show up as '[deleted]'
 - if a post or comment has a missing user it should return a 404 in the detail view

Note that the text or url will still show up for comments and posts even if the author is `[deleted]`.

Make a post in open-discussions and delete that user from the database. View a post by that user. You should see the default avatar image.

Make a comment in open-discussions. Delete the open-discussions user who made the comment. View the comment in `/api/v0/posts/<post_id>/comments/` and it should show up with `[deleted]` as the author name.

One more case: make a comment in open-discussions and make a child comment of that comment. In reddit delete the parent comment. View the comments in the REST API and you should still see the child comment, with the parent comment showing up as `[deleted]` and text `[deleted]`


rhysyngsun:
Code looks good

PR #202, opened 2017-09-27
Added access token header and settings (#164)

### LINKED ISSUE #164: Add custom shared secret header for nginx verification of requests to reddit
We want to be able to provide an shared secret via a header so that nginx in front of reddit can filter out unauthorized requests. This should be configuration via an environment variable. We should be able to add this to the existing custom session [here](https://github.com/mitodl/open-discussions/blob/master/channels/api.py#L67).

rhysyngsun:
Fixes #164 

#### What's this PR do?
Adds the `Access-Token` header to all requests to Reddit backend.

Set `OPEN_DISCUSSIONS_REDDIT_ACCESS_TOKEN` to some value in `.env`

Open a shell and:
```python
from channels.api import _get_session
from urllib.parse import urljoin
refresh_token_url = urljoin(settings.OPEN_DISCUSSIONS_REDDIT_URL, '/api/v1/generate_refresh_token')

session = _get_session()
response = session.get(refresh_token_url, params={'username': 'thatuser'})
```

Assert that the header is in `response.request.headers` with the correct value


blarghmatey:
:+1:

Issue #188, opened 2017-09-25
Display user’s Profile name instead of username when viewing posts


pdpinch:
- [x] Post API
- [x] tests
- [x] Post UI
- [x] tests
- [x] comments API
- [x] tests
- [x] comments UI
- [x] tests

PR #206, opened 2017-09-28
add profile name to post and comment APIs

aliceriot:

#188 

#### What's this PR do?

this adds the `.name` from the user's profile to the results returned by both the comment and post APIs, and makes the changes in the UI to display that information.

If you look at a post or a channel you should see names instead of usernames, same with comments on a post.

### DIFF
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -124,6 +124,7 @@ class PostSerializer(serializers.Serializer):
     num_comments = serializers.IntegerField(read_only=True)
     channel_name = serializers.SerializerMethodField()
     profile_image = serializers.SerializerMethodField()
+    author_name = serializers.SerializerMethodField()
 
     def _get_user(self, instance):
         """
@@ -143,6 +144,13 @@ class PostSerializer(serializers.Serializer):
     def get_url(self, instance):
         """Returns a url or null depending on if it's a self post"""
         return instance.url if not instance.is_self else None
+
+    def get_author_name(self, instance):
+        """get the authors name"""
+        user = self._get_user(instance)
+        if user and user.profile.name:
+            return user.profile.name
+        return "[deleted]"
 
     def get_profile_image(self, instance):
         """Find the Profile for the comment author"""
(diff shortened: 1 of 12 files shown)


rhysyngsun:
I think this PR needs to be rebased/refactored off @noisecapella's PR #198 once that merges so that we can accoutn for cases where `User` doesn't exist.

aliceriot:
@rhysyngsun ok I have rebased over George's changes and incorporated the newer pattern for including profile information on the responses. I also added `.select_related('profile')` when we grab the users, to avoid making another query on each call to the serializer (for the post and comment list views) to grab the profile information.

channels/serializers.py
```diff
@@ -144,6 +145,13 @@ def get_url(self, instance):
         """Returns a url or null depending on if it's a self post"""
         return instance.url if not instance.is_self else None
 
+    def get_author_name(self, instance):
+        """get the authors name"""
+        user = self._get_user(instance)
+        if user and user.profile.name:
+            return user.profile.name
+        return "Someone"
```
rhysyngsun: Could you change this to be "[deleted]" to preserve the original default value?

channels/serializers.py
```diff
@@ -290,6 +299,15 @@ def get_profile_image(self, instance):
...
         return default_profile_image
 
+    def get_author_name(self, instance):
+        """get the author name"""
+        # this is a similar setup to the get_profile_image thingy above
+        user = self._get_user(instance)
+
+        if user and user.profile.name:
+            return user.profile.name
+        return "Someone"
```
rhysyngsun: Same comment as above

static/js/components/PostDisplay.js
```diff
@@ -115,7 +115,7 @@ class PostDisplay extends React.Component {
             {postTitle(post)}
           </div>
           <div className="authored-by">
-            <a>{post.author_id || "someone"}</a> {formattedDate}{" "}
+            <a>{post.author_name || "someone"}</a> {formattedDate}{" "}
```
rhysyngsun: This will never be null now, so we don't need the default.

rhysyngsun:
Functionality looks good.

channels/serializers.py
aliceriot: sure

aliceriot:
@rhysyngsun I believe I address all your comments

PR #225, opened 2017-09-29
Added page for users who aren't logged in

### LINKED ISSUE #200: UI for users who are not logged in
We should redirect anonymous users to micromasters to login, or at least show a message saying they need to log in to see content. Currently it just looks like an empty set of posts

noisecapella:
Fixes #200 

#### What's this PR do?
Redirects users without a JWT to `/auth_required/` to see a message about logging in. There is a link which will redirect the user to micromasters.

There is a new environment variable `OPEN_DISCUSSIONS_EXTERNAL_LOGIN_URL` which is a link to micromasters.

Open the browser in incognito mode and verify that you are redirected to `/auth_required/` no matter what path you choose. If you click on the link you should be directed to micromasters.


aliceriot:
should `OPEN_DISCUSSIONS_LOGIN_URL` be set to `${micromasters_base_url}/discussions` ?

aliceriot:
I tried that and it doesn't seem to work correctly

aliceriot:
ah nevermind! I put `""` around the URL by accident, without those it works as advertised.

open_discussions/settings.py
```diff
@@ -407,6 +407,10 @@
     'JWT_AUTH_HEADER_PREFIX': 'Bearer',
 }
 
+EXTERNAL_LOGIN_URL = get_string('OPEN_DISCUSSIONS_EXTERNAL_LOGIN_URL', None)
```
aliceriot: would it make sense to call this `MM_EXTERNAL_LOGIN_URL`? I'm just thinking that if, at some point, users can authenticate through another service we may have multiple external login URLs, so it might make sense to go ahead and name this one specifically now and thereby avoid changing the name in the future.

static/js/containers/App.js
```diff
@@ -36,7 +38,12 @@ class App extends React.Component {
   }
 
   loadData = async () => {
-    const { dispatch } = this.props
+    const { dispatch, location: { pathname } } = this.props
+
+    if (pathname === "/auth_required/") {
+      return
+    }
```
aliceriot: what do you think about putting this check into `componentDidMount` (where we call `loadData`) instead of here? To me its just a little bit more clear to keep `loadData` a function which just loads some particular data, and decide whether or not we want to call it, instead of having logic inside of the `loadData` function to make that decision.

static/js/containers/App_test.js
```diff
@@ -0,0 +1,47 @@
...
+  it("doesn't load requirements for auth_required", async () => {
+    helper.browserHistory.push("/auth_required/")
+    mount(
+      <div>
+        <Router history={helper.browserHistory} store={helper.store}>
+          {routes}
+        </Router>
+      </div>
+    )
+    await wait(200)
```
aliceriot: is there no way to write this test with `renderComponent`? maybe using `assert.isRejected`?

static/js/containers/AuthRequiredPage.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+/* global SETTINGS: false */
+import React from "react"
+import DocumentTitle from "react-document-title"
+
+import Card from "../components/Card"
+import { formatTitle } from "../lib/title"
```
aliceriot: can you put a blank line in between the component import and the function import?

static/js/entry/root.js
```diff
@@ -20,23 +20,32 @@ const store = configureStore()
...
-    rootEl
-  )
-}
+const authRequiredEndpoint = "/auth_required/"
+if (
+  !SETTINGS.session_url &&
+  window.location.pathname !== authRequiredEndpoint
+) {
+  // user does not have the jwt cookie, they must go through login workflow first
+  window.location = authRequiredEndpoint
```
aliceriot: do you think there's an easy way to do this redirect with react-router instead? this might be a helpful pattern: https://reacttraining.com/react-router/web/example/auth-workflow

noisecapella:
> should `OPEN_DISCUSSIONS_LOGIN_URL` be set to `${micromasters_base_url}/discussions` ?
I'm not sure what would be preferable here but it doesn't really affect the open-discussions login workflow in either case

noisecapella:
Added 'blocked' label waiting for an update in text from @Ferdi

noisecapella:
I need to add handling for the case where the user has a JWT but it can't be refreshed because the user isn't logged in on micromasters. In that case we should redirect the user to the `/auth_required/` page from the failed fetch API call.

open_discussions/settings.py
noisecapella: Sure

noisecapella:
I think that's everything

open_discussions/settings.py
```diff
@@ -407,6 +407,10 @@
     'JWT_AUTH_HEADER_PREFIX': 'Bearer',
 }
 
+EXTERNAL_LOGIN_URL = get_string('MICROMASTERS_EXTERNAL_LOGIN_URL', None)
```
aliceriot: sorry, should have been more specific! I was suggesting we call the setting something like `MM_EXTERNAL_LOGIN_URL` as well as the environment variable

static/js/lib/url.js
```diff
@@ -19,3 +19,5 @@ export const getChannelNameFromPathname = R.compose(
   R.view(R.lensIndex(1)),
   R.match(/^\/channel\/([^/]+)\/?/)
 )
+
+export const authRequiredURL = () => "/auth_required/"
```
aliceriot: any particular reason this isn't just a string constant?

static/js/lib/auth.js
```diff
@@ -15,11 +17,8 @@ const renewSession = async () => {
 
 const redirectAndReject = async (reason: string) => {
   // redirect to the authenticating app
-  if (SETTINGS.auth_url) {
-    window.location = SETTINGS.auth_url
-    return Promise.reject(reason)
-  }
-  return Promise.reject("Unable to redirect to authenticating app")
+  window.location = authRequiredURL()
+  return Promise.reject(reason)
```
aliceriot: do we still want to check for `SETTINGS.auth_url` and redirect there if it's present?

static/js/containers/App_test.js
```diff
@@ -0,0 +1,41 @@
...
+      actions.subscribedChannels.get.requestType,
+      actions.subscribedChannels.get.successType
+    ])
+    sinon.assert.calledWith(helper.getChannelsStub)
+  })
+
+  it("doesn't load requirements for auth_required", () => {
+    helper.browserHistory.push("/auth_required/")
+    assert.isRejected(
+      renderComponent("/missing", [
```
aliceriot: should this be `renderComponent("/auth_required/")` ?

static/js/lib/auth.js
noisecapella: I don't think so. At this point the code tried to get a new token but failed for some reason. The `auth_url` set in micromasters should be `/discussions/` on its site. I don't think there's a circumstance where the user fails to refresh the JWT but is also still logged into micromasters.

@rhysyngsun can you confirm?

static/js/containers/App_test.js
noisecapella: Yes

noisecapella:
I think that should be everything. Waiting on @Ferdi's input on the text for the page and waiting for @rhysyngsun to confirm on the comment I posted

static/js/lib/auth.js
aliceriot: That makes sense, the only thing that occurs to me is if MM drops the JWT refresh request for some reason, but in that case we would probably just want to go to the `AUTH_REQUIRED_URL` anyway.

rhysyngsun: There's not a normal use case a user is logged in but can't refresh the token, but there could be some abnormal ones (e.g. `DiscussionUser` gets deleted or `DiscussionUser.username` gets nulled out).

noisecapella: In either of those cases we lose the connection between MM and OD for that user. When they log in again they will get a new user account. I think it's not worth handling this beyond redirecting the user to the auth_required page

aliceriot: Makes sense!

noisecapella:
I updated the text, removing the first sentence per @Ferdi recommendation:

screenshot from 2017-10-03 10-42-38

noisecapella:
@aliceriot should be everything

roberthouse54:
If it is not too much trouble, can you center that block of text with the icon?

Issue #212, opened 2017-09-29
Clicking the comment upvote button highlights the submit button 


pdpinch:
See https://share.vidyard.com/watch/fkpH6VAgbMrbnjPFzN28KT

pdpinch:
@aliceriot can you take a look into this?

aliceriot:
it's disabled because we have no way (currently) to tell whether the `PATCH` operation in progress for a `Post` object was initiated to upvote a comment or to make a comment. We could try to add some state to keep track of which one it is.

PR #240, opened 2017-10-03
submit button disable

aliceriot:

#212 

#### What's this PR do?

This fixes a little UI quirk, where upvoting a comment would prevent the user from submitting a new top-level comment or replying to an existing comment. This uses the same pattern we're using on the upvote / downvote buttons to get around that (using `this.state` to track whether that particular instance was clicked on or not).

If you upvote or downvote a comment the `submit` button for both post replies (top-level comments) and comment replies should not be grayed out. you should be able to submit a comment just after upvoting or downvoting something.

### DIFF
--- a/static/js/components/CreateCommentForm.js
+++ b/static/js/components/CreateCommentForm.js
@@ -17,9 +17,9 @@ type CreateCommentFormProps = {
   post: Post,
   initialValue: CommentForm,
   formKey: string,
-  beginReply: (initialValue: Object, e: ?Object) => void,
-  onSubmit: (p: string, t: string, c: string, p: Post) => () => void,
-  onUpdate: (e: Object) => void,
+  beginReply: (initialValue: Object, e: ?Event) => void,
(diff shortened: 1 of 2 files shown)


static/js/components/CreateCommentForm.js
```diff
@@ -18,7 +18,7 @@ type CreateCommentFormProps = {
   initialValue: CommentForm,
   formKey: string,
   beginReply: (initialValue: Object, e: ?Object) => void,
-  onSubmit: (p: string, t: string, c: string, p: Post) => () => void,
+  onSubmit: (p: string, t: string, c: string, p: Post, e: Object) => Promise<*>,
```
noisecapella: Can you use `e: Event` here and in the calling functions?

noisecapella:
Functionality looks good

static/js/components/CreateCommentForm.js
aliceriot: sure

aliceriot:
@noisecapella comment addressed

Issue #80, opened 2017-07-31
Google analytics


noisecapella:

pdpinch:
moved to "In Progress"

aliceriot:
discovered it had already been set up :laughing:

PR #248, opened 2017-10-04
set up google analytics

aliceriot:

#80 

#### What's this PR do?

This sets up very basic google analytics for logging page views.

Set up a google analytics account and make a new account / app combo. Call it whatever you like, but get that google analytics code and set it as the value of the `GA_TRACKING_ID` env var (in your `.env`). Then just do a few page-loads and you should see some pageviews show up in the GA 'realtime' web dashboard. 

NOTE: make sure that you have any extensions which might interfere with GA disabled. Or, better yet, use either incognito mode or a different browser than your day-to-day browser to test.

### DIFF
--- a/static/js/components/Analytics.js
+++ b/static/js/components/Analytics.js
@@ -3,12 +3,21 @@ import React from "react"
 import { withRouter } from "react-router"
 import ReactGA from 'react-ga'
 
+if (SETTINGS.gaTrackingID) {
+  console.log(SETTINGS.gaTrackingID);
+  ReactGA.initialize(SETTINGS.gaTrackingID, { debug: true })
+  console.log('initializing');
+}
+
 class Analytics extends React.Component {
   componentDidUpdate(prevProps) {
     const { location } = this.props
 
+    console.log('potato');
+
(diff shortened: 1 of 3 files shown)


static/js/components/Analytics.js
```diff
@@ -0,0 +1,26 @@
...
+if (SETTINGS.gaTrackingID) {
+  ReactGA.initialize(SETTINGS.gaTrackingID, { debug: debug })
+}
+
+class Analytics extends React.Component {
+  componentDidUpdate(prevProps) {
+    const { location } = this.props
+
+    if (location !== prevProps.location) {
+      ReactGA.pageview(window.location.pathname)
```
noisecapella: Should we use `location.pathname` here?

noisecapella:
Functionality works great

static/js/components/Analytics.js
aliceriot: I went with `window.location.pathname` simply because that's what we were using in MM - do you think there's a reason to go with one or the other? I suppose they _should_ be equal all the time

pdpinch:
So, should this be closed now?

PR #254, opened 2017-10-04
Make public_description not required when creating a channel

annagav:
Related to https://github.com/mitodl/micromasters/issues/3581

#### What's this PR do?
Make `public_description` not required

Should be able to create a channel with empty description.


noisecapella:
Can you add or update a test where `public_description` is omitted?

annagav:
The tests for channel serializer doesn't test for data validation, so different values for 'public_description' would not affect the result of the test.

PR #277, opened 2017-10-10
Decrease the size of avatars in comments

roberthouse54:
none

#### What's this PR do?
This makes the user avatars in the comments section smaller. It also makes the replies indentation bigger, that that replies align better with the parent comment above them.

(See that it looks correct and the css is valid)


aliceriot:
reply indentation and spacing stuff looks nice!

PR #285, opened 2017-10-11
Added a setting for the JWT cookie name

rhysyngsun:
N/A

#### What's this PR do?
Adds a setting for the JWT cookie name

Set `OPEN_DISCUSSIONS_COOKIE_NAME=boop` in both MM and OD `.env`
Verify you can still hit /discussions/ on MM, login to OD, and that you see the cookie with the new name set.


open_discussions/settings.py
```diff
@@ -404,7 +404,7 @@
     'JWT_EXPIRATION_DELTA': datetime.timedelta(seconds=60*60),
     'JWT_ALLOW_REFRESH': True,
     'JWT_REFRESH_EXPIRATION_DELTA': datetime.timedelta(days=7),
-    'JWT_AUTH_COOKIE': 'open_discussions_jwt',
+    'JWT_AUTH_COOKIE': get_string('OPEN_DISCUSSIONS_COOKIE_NAME', 'open_discussions_jwt'),
```
noisecapella: Should we make this and the micromasters environment variable required in `app.json`?

rhysyngsun: Sure, I can do that.

rhysyngsun:
@noisecapella addressed your feedback

PR #289, opened 2017-10-13
Added docs with gh-pages style.

rhysyngsun:
N/A

#### What's this PR do?
Add docs and sets them up so they'll get compiled with gh-pages for easy access/navigation. Note: needs to get merged to master before that piece works though.

N/A

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -1,0 +1,12 @@
+Operations
+---
+
+## Common Operations
+
+#### Resetting Data
+ - If you clear data on the Reddit backend:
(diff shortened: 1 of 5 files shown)


docs/operations.md
```diff
@@ -0,0 +1,12 @@
+Integration
```
aliceriot: should this say 'Operations'?

rhysyngsun: yeah, renamed it like 5 times...

aliceriot: hahaha

PR #291, opened 2017-10-16
set focus on comment reply things, add key combo to submit form

### LINKED ISSUE #237: When clicking on "reply" for a comment, the focus should go in the text area
It does not go automatically there and it is confusing.

screen shot 2017-10-02 at 3 10 19 pm

aliceriot:

closes #237 

#### What's this PR do?

this sets the `autoFocus` prop on the textarea for replying to comments, so that we you click the 'reply' button the focus automatically goes into the text box (and you can compose your comment without having to click or tab into it first).

I also added an `onKeyDown` to the reply form, which listens for the ctrl-enter key combo and submits the form when you press that (for that power user life).

go to a post. the focus should not be automatically into the 'top level' reply box. click 'reply' to a comment, the focus should automatically go to that text box. in both a comment and a post reply you should be able to submit your comment by pressing ctrl+enter.

### DIFF
--- a/static/js/components/CreateCommentForm.js
+++ b/static/js/components/CreateCommentForm.js
@@ -47,10 +47,19 @@ const commentForm = (
   onUpdate: (e: any) => void,
   cancelReply: () => void,
   isComment: boolean,
-  disabled: boolean
+  disabled: boolean,
+  autoFocus: boolean
 ) =>
   <div className="reply-form">
-    <form onSubmit={onSubmit} className="form">
+    <form
+      onSubmit={onSubmit}
+      className="form"
+      onKeyDown={e => {
+        if (e.key === "Enter" && e.ctrlKey && !disabled && !isEmptyText(text)) {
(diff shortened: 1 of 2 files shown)


static/js/components/CreateCommentForm.js
```diff
@@ -47,10 +47,19 @@ const commentForm = (
...
+  autoFocus: boolean
 ) =>
   <div className="reply-form">
-    <form onSubmit={onSubmit} className="form">
+    <form
+      onSubmit={onSubmit}
+      className="form"
+      onKeyDown={e => {
+        if (e.key === "Enter" && e.ctrlKey) {
+          onSubmit(e)
```
noisecapella: Can you add a check that this doesn't happen if the button is disabled?

aliceriot: I added that functionality and added a few tests for it as well

noisecapella:
merge freeze lifted!

Issue #270, opened 2017-10-09
For link posts, show domain name next to post title


pdpinch:
As a user, I'd like to know what site I will be going to when I click on a link post.

pdpinch:
e.g https://www.dropbox.com/s/yd4ciax2ryb33ty/Screenshot%202017-10-09%2009.12.05.png?dl=0

PR #294, opened 2017-10-17
show link post domain after the link

aliceriot:

#270 

#### What's this PR do?

this adds a little display next to the title link for a URL post which just shows what domain the link is from.

if you look at the frontpage or a channel page with link posts you should see something like:

fjfjasdf

the same thing should show up if you look at the comments for the post.

### DIFF
--- a/static/js/components/PostDisplay.js
+++ b/static/js/components/PostDisplay.js
@@ -5,7 +5,7 @@ import { connect } from "react-redux"
 import { Link } from "react-router-dom"
 import ReactMarkdown from "react-markdown"
 
-import { channelURL, postDetailURL } from "../lib/url"
+import { channelURL, postDetailURL, urlHostname } from "../lib/url"
 import { formatCommentsCount } from "../lib/posts"
 
 import type { Post } from "../flow/discussionTypes"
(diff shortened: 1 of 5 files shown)


static/js/components/PostDisplay.js
```diff
@@ -26,9 +26,14 @@ const postTitle = (post: Post) =>
     >
       {post.title}
     </Link>
-    : <a className="post-title" href={post.url} target="_blank">
-      {post.title}
-    </a>
+    : <div>
+      <a className="post-title" href={post.url} target="_blank">
```
noisecapella: Can you add `rel="noopener noreferrer"` to this?

aliceriot: weird that the linter didn't catch that...

aliceriot:
@noisecapella addressed

aliceriot:
ahaha I accidentally added it to the wrong branch https://github.com/mitodl/open-discussions/pull/296/commits/54fa169bff040b5424011ae5af32429a6a6d6cfe

:woman_facepalming:

static/js/components/PostDisplay.js
noisecapella: Would be a good use case for a custom eslint plugin

PR #298, opened 2017-10-18
Comments pagination

### LINKED ISSUE #78: Paginate Comments
- [ ] pagination in the API
- [ ] pagination in the front-end UI

giocalitri:
Fixes #78 

#### What's this PR do?
Implements the comment pagination

Add many comments to a post, such that when you see it in reddit you are not shown all of them at once. In open-discussions click on the 'Load more comments' link. You should see the link disappear and be replaced with some extra comments


noisecapella:
I'll take over the front end piece of this PR

giocalitri:
@noisecapella it needs to go together with the backend in they PR: the changes are not backwards compatible

noisecapella:
Rebasing has caused some problems, I'll let you know when it's ready to look at

noisecapella:
I fixed the issue, should be ready now

channels/api.py
```diff
@@ -476,32 +489,93 @@ def list_comments(self, post_id):
...
+            'children': children,
+            'count': len(children),
+            'parent_id': qualified_parent_id
+        })
+        more_comments.submission = self.reddit.submission(post_id)
+        more_comments.submission.comment_limit = settings.OPEN_DISCUSSIONS_REDDIT_COMMENTS_LIMIT
+
+        original_load_comment = more_comments._load_comment  # pylint: disable=protected-access
+
+        def replacement_load_comment(*args, **kwargs):
```
rhysyngsun: I'd suggest refactoring this into a separate method that returns a replacement method like:

```python
def replace_load_comment(load_comment):
    def replacement_load_comment(*args, **kwargs):
         # impl
    return replacement_load_comment
```

And then the replacement can become a one-liner and we don't have a inlined function anymore:

```python
more_comments._load_comment = replace_load_comment(more_comments._load_comment)
```

static/js/components/CommentTree.js
```diff
@@ -50,104 +57,123 @@ const renderComment = R.curry(
...
     beginEditing: (fk: string, iv: Object, e: ?Object) => void,
     processing: boolean,
     depth: number,
-    comment: Comment
+    comment: GenericComment
   ) => {
-    const formKey = replyToCommentKey(comment)
-    const editFormKey = editCommentKey(comment)
-    const initialValue = getCommentReplyInitialValue(comment)
+    if (comment.comment_type === "comment") {
```
rhysyngsun: Could you refactor this to separate functions for rendering each of the comment types?

rhysyngsun: `"comment"` should also be refactored to a `const`

channels/views.py
```diff
@@ -312,9 +311,54 @@ def post(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+
+    def get(self, request, *args, **kwargs):  # pylint: disable=unused-argument
+        """Get list for comments and attach User objects to them"""
+        with translate_praw_exceptions():
+            children = request.query_params.getlist('children')
+
+            # validate the request parameters: each are required
+            try:
+                post_id = request.query_params['post_id']
+                parent_id = request.query_params['parent_id'] or None
```
rhysyngsun: Is an empty string expected here or is the `or None` extraneous?

static/js/flow/discussionTypes.js
```diff
@@ -76,22 +78,42 @@ export type PostListResponse = {
...
+
 export type CommentForm = {
   post_id:      string,
   comment_id?:  string,
   text:         string,
 }
+
+// If you're looking for a 'Comment' type this is probably what you want.
+// It represents any element in the comment tree stored in the reducer
+export type GenericComment = CommentInTree|MoreCommentsInTree
```
rhysyngsun: Nit: could you put spaces around the `|` here and on line 116 for readability?

static/js/lib/api.js
```diff
@@ -76,22 +78,21 @@ export function getPost(postId: string): Promise<Post> {
...
-  comment: string,
+  text: string,
   commentId: ?string
-) {
+): Promise<CommentFromAPI> => {
   const body =
     commentId === undefined
-      ? { text: comment }
-      : { text: comment, comment_id: commentId }
+      ? { text: text }
```
rhysyngsun: Nit: could just be `{ text }`, same for the else branch.

static/js/lib/comments.js
```diff
@@ -1,37 +1,40 @@
...
 ): Object | null => {
   for (const [index, comment] of enumerate(commentTree)) {
-    if (comment.id === commentId) {
-      const path = []
-      for (const parentIndex of parentIndexes) {
-        path.push(parentIndex)
-        path.push("replies")
-      }
-      path.push(index)
+    if (comment.comment_type === "comment") {
```
rhysyngsun: Suggestion: use `!==` and `continue` to avoid the extra indentation.

static/js/reducers/comments.js
```diff
@@ -2,100 +2,281 @@
...
+  // first we need to remove the existing more_comments instances. There should be exactly one at the level we are at
+  if (parentId) {
+    const commentLens = findComment(oldTree, parentId)
+    if (!commentLens) {
+      throw new Error("Unable to find parent id for comment to update")
+    }
+
+    const comment = R.view(commentLens, oldTree)
+    const newReplies = comment.replies.filter(
+      comment => comment.comment_type === "comment"
```
rhysyngsun: Make sure to update these usages to the constant value suggested above.

static/js/reducers/comments.js
```diff
@@ -2,100 +2,281 @@
...
+}
+
+export const moreCommentsEndpoint = {
+  name:    "morecomments",
+  verbs:   [GET],
+  getFunc: async (
+    postId: string,
+    parentId: string | null,
+    children: Array<string>
+  ) => await api.getMoreComments(postId, parentId, children),
```
rhysyngsun: Doesn't need to be `async`/`await` since api methods already return a `Promise`

noisecapella:
FYI I force pushed and squashed but that only covers rebasing changes, I'll put review comments in separate commits

static/js/components/CommentTree.js
noisecapella: `"comment"` actually needs to be a literal here for flow to derive the appropriate type of `comment`. I could move this into a const within the same file, but flow assumes consts exported from other modules are not really const. What do you think?

channels/views.py
noisecapella: There's no use case for an empty string, I'll see about removing it from the frontend if it's null

static/js/components/CommentTree.js
rhysyngsun: ok, fair enough, I think the tradeoff of having flow be able to assert correctness here is more important than having `const`s

noisecapella:
All comments addressed

static/js/components/CommentTree.js
```diff
@@ -53,122 +55,138 @@ const renderTopLevelComment = R.curry(
...
     }
   }
 )
 
+const renderComment = (
+  forms: FormsState,
+  upvote: CommentVoteFunc,
+  downvote: CommentVoteFunc,
+  loadMoreComments: LoadMoreCommentsFunc,
+  beginEditing: (fk: string, iv: Object, e: ?Object) => void,
```
rhysyngsun: This could use the `BeginEditingFunc` you defined above.

rhysyngsun:
Functionality works fine, although I think the "Load More Comments" link could use more padding, particularly for when they bunch up like this:

selection_104

static/scss/comment.scss
```diff
@@ -121,12 +128,24 @@
       }
     }
 
-    .replies > .comment {
-      padding-left: 8px;
+    .replies > & {
```
rhysyngsun: This change broke some of the layout, if you look at the profile images, they're now touching the vertical dividers:

Your branch:
selection_106


Master:
selection_105

PR #300, opened 2017-10-18
Use ExtractTextPlugin to split CSS into separate file

noisecapella:
Related to https://github.com/mitodl/micromasters/pull/3637

#### What's this PR do?
Splits off CSS for production instances into separate file. This lets the browser cache the CSS and load it in parallel to the JS

Go to PR build's page, verify that CSS loads correctly. You should also see no changes locally.

### DIFF
--- a/app.json
+++ b/app.json
@@ -62,7 +62,7 @@
       "description": "E-mail to send 500 reports to."
     },
     "OPEN_DISCUSSIONS_CHANNEL_POST_LIMIT": {
-      "description": "Number of posts to display on the forntpage and channels",
+      "description": "Number of posts to display on the frontpage and channels",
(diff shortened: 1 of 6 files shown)


aliceriot:
the PR environment seems to be broken, I tried to add the required settings in the heroku UI but that doesn't seem to work. I think reviewing this is blocked until that is fixed, unless i'm missing something.

noisecapella:
Yeah, the PR environment should be getting its env vars from the CI environment. They are missing from the CI environment which explains the PR build error message, but it doesn't explain why the CI deployment succeeded anyway. I'll talk to devops when I get a chance

noisecapella:
PR build is ready!

PR #331, opened 2017-11-08
split post display into two components

### LINKED ISSUE #327: split post display into two components
Currently we use one component (`PostDisplay`) to display posts in a list view (e.g. on the frontpage, on a channel) and also to display a post on a post detail page. For #231 I'm going to have to add a bunch of functionality to that component, and I think with that added complexity it's just going to be too much to live in one component (i.e. a bunch of editing functionality that only matters on the post detail page). So! I'm going to split the `PostDisplay` into two separate components first, before I start on the #231.

aliceriot:

closes #327 

#### What's this PR do?

basically we currently have the `PostDisplay` component have two different modes, one which is used in a list-view (frontpage, channel view) and another which is used on a post detail view. To work on adding post editing I need to modify the post detail version but not the list-view version, and I think it's just too much to have it all in the same component. So, this splits it out into two different components, each of which is only responsible for one thing. They still share a lot of styling, and code which is common to the two I pulled out into helper functions in `lib/posts.js`.

All functionality should be exactly as before.


static/js/components/CompactPostDisplay.js
```diff
@@ -0,0 +1,89 @@
...
+  formatPostTitle,
+  PostVotingButtons
+} from "../lib/posts"
+
+import type { Post } from "../flow/discussionTypes"
+
+class CompactPostDisplay extends React.Component {
+  props: {
+    post: Post,
+    showChannelLink?: boolean,
```
noisecapella: It looks like `showChannelLink` is always present, can you make it required here?

static/js/components/CompactPostDisplay.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  state: {
+    upvoting: boolean
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      upvoting: false
```
noisecapella: Should this be part of `PostVotingButtons`?

aliceriot: yeah that's a good idea

aliceriot:
@noisecapella comments addressed

PR #336, opened 2017-11-14
Switched factories to class-based model and added created field

rhysyngsun:
N/A

#### What's this PR do?
Refactors the factory models into classes so we can move the creation of posts into a `post_generation` hook and posthumously set both `id` and `created`. It also modifies the `FactoryStore` to only write the `factory_data` 

Delete `cassettes/channels.views_test.test_list_channels.json` and rerun tests. Ensure the cassette is recreated and the `factory_data` file updated.

Also, if you run this you should see an ISO formatted `created` date:

```python
from channels.factories import *
from channels.api import Api
u = User.objects.first()
api = Api(u)
p = TextPostFactory.create(api=api)
p.created  # -> '2017-11-14T16:20:14+00:00'
```

### DIFF
--- a/channels/factories.py
+++ b/channels/factories.py
@@ -1,7 +1,7 @@
 """Factories for making test data"""
 import json
 import os
-from collections import namedtuple
+from datetime import datetime
 
 import pytz
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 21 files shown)


aliceriot:
functionality is good!

channels/factories.py
```diff
@@ -17,13 +17,37 @@
...
+
+class Post:
+    """Simple factory representation for a post"""
+    def __init__(self, **kwargs):
+        self.id = kwargs.get('id', None)
+        self.title = kwargs.get('title', None)
+        self.text = kwargs.get('text', None)
+        self.url = kwargs.get('url', None)
+        self.channel = kwargs.get('channel', None)
+        self.api = kwargs.get('api', None)
```
aliceriot: should it have `created` here as well?

aliceriot: Or does it now always get filled in with a value from the API request to reddit? Just wondering why `id` is here but `created` isn't, since they both are values that we don't set (i.e. they come back from reddit).

rhysyngsun: Good catch, it should for consistency so the property is always present.

rhysyngsun:
@aliceriot done https://github.com/mitodl/open-discussions/pull/336/files#diff-518ea9fe9d13be234e6bf2e804ac8dacR39

PR #338, opened 2017-11-15
add subreddit title to subreddit API and frontend

### LINKED ISSUE #257: Show channel title in byline instead of channel name
The channel names are more obscure than the channel titles. Also, in the channel create dialog, we say that the channel names are only visible in URLs.

aliceriot:

closes #257 

#### What's this PR do?

Basically we want to show the channel title in the little byline, not the channel name (which basically a url component). To do this I made a small change to the channel serializer. This basically added a new request to a ton of tests in `channels/views_tests.py` so I had to regenerate a whole bunch of cassettes. As a nice side effect, several tests are now converted to use the new, reproducible factory strategy for creating reddit objects that Nathan came up with. So this should be much less painful for the next person who has to fiddle with it, haha.

When you go to look at the frontpage in the UI (at `/`) you should see the channel title instead of the channel name in the byline. Clicking on the channel title should take you to the post listing for the channel itself. Other than that the code should be clean looking and all the tests should be passing.


rhysyngsun:
Code looks good.

Ferdi:
@aliceriot @rhysyngsun I hope it's not that hard to get extra field from these APIs.  I noticed that we are doing some work here https://github.com/mitodl/open-discussions/pull/294 to extract the domain , when we could have used the one returned by reddit api.

rhysyngsun:
@Ferdi sorry, I'm not sure I follow what you're asking?

Ferdi:
Sorry @rhysyngsun , what is the reason we do this :  
https://github.com/mitodl/open-discussions/blob/master/static/js/lib/url.js#L33

Instead of using the "domain" attribute of the post from reddit api 

https://www.reddit.com/.json

aliceriot:
I did that at the time because I didn't know that the reddit APIs returned that value.

PR #339, opened 2017-11-15
Remove iflow-lodash, add flow-typed

noisecapella:
Part of #298 

#### What's this PR do?
Updates flow, migrates from iflow to flow-typed

Nothing should break


static/js/components/CreateCommentForm_test.js
```diff
@@ -98,7 +98,8 @@ describe("CreateCommentForm", () => {
     })
 
     it("should begin form editing on load", async () => {
-      wrapper = await renderPostForm()
+      wrapper = renderPostForm()
+      // $FlowFixMe
```
rhysyngsun: Could you add a description to this so we know what flow error is underlying this?

static/js/components/CreateCommentForm_test.js
```diff
@@ -255,6 +256,7 @@ describe("CreateCommentForm", () => {
     })
 
     it("should show an empty form when reply has been started", async () => {
+      // $FlowFixMe
```
rhysyngsun: Same as above

noisecapella:
Comments addressed

PR #340, opened 2017-11-16
Updated posts API to handle remove moderation

### LINKED ISSUE #195: API for moderator to remove post / submission
As a moderator, I need to be able to remove posts / submissions that don't mean the channel guidelines. 

[PRAW docs](http://praw.readthedocs.io/en/latest/code_overview/other/submissionmoderation.html?highlight=remove#praw.models.reddit.submission.SubmissionModeration.remove)

### LINKED ISSUE #333: Fix factory_data serialization
The factory data we write out sometimes gets output as other types of data (e.g. kwargs turn into lists). Unsure of the cause be we should probably just write a straightforward mapping to output format.

rhysyngsun:
Fixes #195 
Fixes #333 

#### What's this PR do?
Adds a flag for removing posts for moderators.

- Create a post as a normal contributor on a a channel (not a moderator).
- As a moderator go to `/api/v0/post/:id` and verify you see `"removed": false` in the response.
- `PATCH` with `{"removed": true}` and verify the response reflects the change
- Verify the post no longer shows up on the frontpage or channel listing
- You should still be able to view the post as it's content if you are a moderator or the OP, if you view as a third contributor, you should see `"text": "[removed]"`
- `PATCH` with `{"removed": false}` and verify the response reflects the change
- Verify the post shows up on the frontpage or channel listing again

### DIFF
--- a/channels/api.py
+++ b/channels/api.py
@@ -408,6 +408,26 @@ class Api:
 
         return post.edit(text)
 
+    def remove_post(self, post_id):
+        """
+        Removes the post, opposite of approve_post
+
+        Args:
+            post_id(str): the base36 id for the post
+        """
+        post = self.get_post(post_id)
+        post.mod.remove()
+
+    def approve_post(self, post_id):
+        """
+        Approves the post, oppsite of remove_post
+
+        Args:
+            post_id(str): the base36 id for the post
+        """
+        post = self.get_post(post_id)
+        post.mod.approve()
+
     def create_comment(self, text, post_id=None, comment_id=None):
         """
         Create a new comment in reply to a post or comment
--- a/channels/views_test.py
+++ b/channels/views_test.py
@@ -1,13 +1,11 @@
 """Tests for views for REST APIs for channels"""
-# pylint: disable=too-many-lines
-import time
 import pytest
 from betamax.fixtures.pytest import _casette_name
 from django.core.urlresolvers import reverse
 from praw.exceptions import APIException
 from rest_framework import status
 
-from channels.api import Api, CHANNEL_TYPE_PRIVATE, get_or_create_auth_tokens
+from channels.api import Api, CHANNEL_TYPE_PRIVATE
 from channels.factories import RedditFactories, FactoryStore, STRATEGY_BUILD
 from channels.serializers import default_profile_image
 from open_discussions.factories import UserFactory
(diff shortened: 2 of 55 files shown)


channels/api.py
```diff
@@ -606,17 +626,18 @@ def remove_moderator(self, moderator_name, channel_name):
...
 
         Args:
             channel_name(str): name of the channel
+            moderator_name(str): username of the user
 
         Returns:
             praw.models.listing.generator.ListingGenerator: a generator representing the contributors in the channel
         """
-        return self.get_channel(channel_name).moderator()
+        return self.get_channel(channel_name).moderator(redditor=moderator_name)
```
noisecapella: According to the code comment `redditor` here will restrict the list of moderators to only `moderator_name`. Do we want to do that here?

noisecapella: Looking at the rest of the code it looks like we don't use `moderator_name` anywhere. Did you mean to use it in `ModeratorDetailView`?

Can you also update the docstring to say that `moderator_name` will do filtering if specified?

channels/views_test.py
```diff
@@ -395,21 +398,13 @@ def test_get_post_not_found(client, logged_in_profile, use_betamax, praw_setting
...
-):
+def test_list_posts(client, missing_user, private_channel_and_contributor, reddit_factories):
     """List posts in a channel"""
     channel, user = private_channel_and_contributor
-
-    first_post = reddit_factories.text_post('my post', user, channel=channel)
-    second_post = reddit_factories.text_post('my 2nd post', user, channel=channel)
-
-    time.sleep(3)
+    posts = list(reversed([
```
noisecapella: What's the `reversed` here for?

rhysyngsun: It's because the api returns the posts in the reverse order that they're made, so this is a inline form of:

```python
post1 = reddit_factories.post(...)
post2 = reddit_factories.post(...)

posts = [post2, post1]
```

If you feel that's clearer, I can change it, but they're functionally equivalent.

channels/api.py
rhysyngsun: I was using this at one point, and this is possibly useful in other places where we're just trying to check for a moderator's existence. I can back this out for now though since I'd prefer to keep the delta down for this PR.

rhysyngsun:

rhysyngsun:
Going to wait for #344 to merge first since they both affect the cassettes and that one is more time-sensitive.

Issue #232, opened 2017-10-02
User can Edit their own comment


roberthouse54:
Design [below](https://github.com/mitodl/open-discussions/issues/232#issuecomment-334805536)

If a user has edited a comment, we will show a little text that says [edited by author].

roberthouse54:
We will show an edit link under the comment. This will only be shown to the author of the post. 

![image](https://user-images.githubusercontent.com/20047260/31288277-37fe3bc2-aa92-11e7-8640-e45c5ef6c23d.png)

Editing is done inline. 

![image](https://user-images.githubusercontent.com/20047260/31288285-3f38feae-aa92-11e7-88a3-53cd36988402.png)

After the comment has been edited we will add "[edited by author]" at the end of the text.

![image](https://user-images.githubusercontent.com/20047260/31288307-5a4ebb20-aa92-11e7-8409-d86b91d3044f.png)

pdpinch:
@aliceriot Can this be closed?

pdpinch:
Need to add flag to indicate that the comment has been edited.

PR #341, opened 2017-11-16
edit comments

aliceriot:

#232 

#### What's this PR do?

This adds UI to be able to edit a comment. 

If you look at comments that you wrote you should see an 'Edit' button right next to where the 'Reply' button currently is. Clicking on this button should replace the comment text display with a form containing the current text for the comment. Saving this form should update the comment with your changes. Canceling should return you to the normal UI.

### DIFF
--- a/static/js/components/CommentTree_test.js
+++ b/static/js/components/CommentTree_test.js
@@ -10,14 +10,14 @@ import ReactMarkdown from "react-markdown"
 
 import Card from "../components/Card"
 import CommentTree from "./CommentTree"
-import { ReplyToCommentForm } from "./CreateCommentForm"
+import { ReplyToCommentForm } from "./CommentForms"
 
 import { makeCommentTree } from "../factories/comments"
 import { makePost } from "../factories/posts"
-import { replyToCommentKey } from "../components/CreateCommentForm"
+import { replyToCommentKey, editCommentKey } from "../components/CommentForms"
 
 describe("CommentTree", () => {
-  let comments, post, sandbox, upvoteStub, downvoteStub, beginReplyStub
+  let comments, post, sandbox, upvoteStub, downvoteStub, beginEditingStub
 
   beforeEach(() => {
     post = makePost()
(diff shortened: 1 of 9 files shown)


rhysyngsun:
@aliceriot could you resolve the conflict before I review, please?

aliceriot:
@rhysyngsun  should be fixed now

static/js/components/CommentTree_test.js
```diff
@@ -39,8 +42,8 @@ describe("CommentTree", () => {
         forms={{}}
         upvote={upvoteStub}
         downvote={downvoteStub}
-        beginReply={R.curry((formKey, initialValue, e) => {
-          beginReplyStub(formKey, initialValue, e)
+        beginEditing={R.curry((formKey, initialValue, e) => {
```
rhysyngsun: Could this just be `beginEditing={R.curry(beginEditingStub)}`?

static/js/components/CreateCommentForm.js
```diff
@@ -201,6 +220,49 @@ export const ReplyToCommentForm = connect(mapStateToProps, mapDispatchToProps)(
...
+
+      const updatedComment = { ...comment, text: text }
+      this.setState({ patching: true })
+      await patchComment(updatedComment)
+      this.setState({ patching: false })
+    }
+
+    render() {
+      const { forms, formKey, onUpdate, cancelReply } = this.props
+      const { patching } = this.props
```
rhysyngsun: This should be`this.state`, the submit buttons are not currently disabling because of that.

static/js/components/CreateCommentForm.js
```diff
@@ -201,6 +220,49 @@ export const ReplyToCommentForm = connect(mapStateToProps, mapDispatchToProps)(
   }
 )
 
+export const EditCommentForm = connect(mapStateToProps, mapDispatchToProps)(
```
rhysyngsun: Should we rename this file to `CommentForm` since we now have a second component other than the one the module is named after?

rhysyngsun:
Functionality generally looks fine, except for the button disabling issue I mentioned.

static/js/components/CreateCommentForm.js
aliceriot: yeah that's a good idea

static/js/components/CreateCommentForm.js
aliceriot: :woman_facepalming:

aliceriot:
ok I think comments are all addressed

Issue #263, opened 2017-10-05
Moderation: About this channel 


pdpinch:
As a channel moderator, I'd like to display some information about the channel, such as rules for posting, along side the list of posts. 

### acceptance criteria
- [ ] Store description in reddit, if possible <- let's discuss before implementing
- [ ] Use existing PRAW APIs, if possible

Be sure to discuss the approach on this ticket before implementing.

roberthouse54:
The "About" card will appear in the right column. The about text comes from the "Description" field in the Create New Channel modal. [Nice-to-have]: If possible, display the # of users that have access to the channel, and the # of users currently active in the channel.

TBD: Not sure what to do with this card in the mobile view.

![image](https://user-images.githubusercontent.com/20047260/31397720-c2f9dd7e-adb4-11e7-8641-4bd796427537.png)

rhysyngsun:
@roberthouse54 your mockups don't show the channel list on the left side, which is currently there on larger browsers. Also, how is this handled for mobile?

roberthouse54:
Yes, that will be ok. We will have a three column layout for large browsers.

roberthouse54:
OK, for mobile, let's just move it underneath the posts card. We can revisit this later if we want to.

rhysyngsun:
We decided in planning yesterday that for the sake of delivering this sooner we'll implement the description edit on a full page so we don't have to deal with pulling in a new dialog component.

PR #344, opened 2017-11-20
Added channel description to API and UI

rhysyngsun:
Part of #263 

#### What's this PR do?
Adds the channel description to the API and renders it as markdown on the channel page.

- Go to /api/v0/channels/:channel_name/ and verify you see the description field returned.
- Update the channel description with some markdown like (not all these markdown examples will work, I was lazy and copy/pasted an excessive example):
```python
from channels.api import Api
user = User.objects.get(username="mitodl")
api = Api(user)
api.update_channel("channel_name", description="""
An h1 header
============

Paragraphs are separated by a blank line.

2nd paragraph. *Italic*, **bold**, and `monospace`. Itemized lists
look like:

  * this one
  * that one
  * the other one

Note that --- not considering the asterisk --- the actual text
content starts at 4-columns in.

> Block quotes are
> written like so.
>
> They can span multiple paragraphs,
> if you like.

Use 3 dashes for an em-dash. Use 2 dashes for ranges (ex., "it's all
in chapters 12--14"). Three dots ... will be converted to an ellipsis.
Unicode is supported. ☺

An h2 header
------------

Here's a numbered list:

 1. first item
 2. second item
 3. third item

### An h3 header ###

Now a nested list:

 1. First, get these ingredients:

      * carrots
      * celery
      * lentils
2. hiiiii

Notice again how text always lines up on 4-space indents (including
that last line which continues item 3 above).

Here's a link to [a website](http://foo.bar), to a [local
doc](local-doc.html), and to a [section heading in the current
doc](#an-h2-header). 

A horizontal rule follows.

***

Here's a definition list:

apples
  : Good for making applesauce.
oranges
  : Citrus!
tomatoes
  : There's no "e" in tomatoe.

Again, text is indented 4 spaces. (Put a blank line between each
term/definition pair to spread things out more.)

Here's a "line block":

| Line one
|   Line too
| Line tree

""")
````
- Verify this content and/or the default value is correctly rendered in the sidebar.
- Ensure no adverse affects on other pages in the app (e.g. pages can load and layout is unaffected by changes to support right sidebar).

#### Screenshots

localhost_8063_channel_odl

### DIFF
--- a/channels/factories.py
+++ b/channels/factories.py
@@ -17,6 +17,9 @@ from channels.models import RedditAccessToken, RedditRefreshToken
 
 FAKE = faker.Factory.create()
 
+STRATEGY_CREATE = 'create'
+STRATEGY_BUILD = 'build'
+
 
 class Channel:
     """Simple factory representation for a channel"""
@@ -24,6 +27,7 @@ class Channel:
         self.name = kwargs.get('name', None)
         self.title = kwargs.get('title', None)
         self.channel_type = kwargs.get('channel_type', None)
+        self.description = kwargs.get('description', None)
         self.public_description = kwargs.get('public_description', None)
         self.api = kwargs.get('api', None)
 
@@ -95,6 +99,7 @@ def serialize_factory_result(obj):
             "name": obj.name,
             "title": obj.title,
             "channel_type": obj.channel_type,
+            "description": obj.description,
             "public_description": obj.public_description,
         }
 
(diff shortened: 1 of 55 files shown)

### CODE REVIEW

**Comment by aliceriot** [comment]
functionality looks good, reading code now.

File: static/js/components/ChannelSidebar_test.js [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+
+describe("ChannelSidebar", () => {
+  const renderSidebar = channel => shallow(<ChannelSidebar channel={channel} />)
+
+  it("should render sidebar", () => {
+    const channel = makeChannel()
+    const wrapper = renderSidebar(channel)
+    const [description] = wrapper.find(ReactMarkdown)
+    assert.equal(description.props.source, channel.description)
+  })
```
aliceriot: a basic test that "There is no description..." shows up when `channel.description` is empty might be worthwhile here

static/js/containers/HomePage.js
```diff
@@ -94,6 +94,8 @@ const mapStateToProps = state => {
   }
 }
 
-export default R.compose(connect(mapStateToProps), withNavSidebar, withLoading)(
-  HomePage
-)
+export default R.compose(
+  connect(mapStateToProps),
+  withOnlyNavSidebar,
```
aliceriot: suggestion you can take or leave: to me it makes a little more semantic sense to have `withNavSidebar` and `withNavAndChannelSidebars` or similar, those names seem more accurate to me

static/js/hoc/withNavSidebar.js
```diff
@@ -1,30 +1,8 @@
...
-            <WrappedComponent {...this.props} />
-          </div>
-        </div>
-      )
-    }
-  }
-  return WithNavSidebar
-}
-
-export default withNavSidebar
```
aliceriot: I might suggest we simplify this a little bit and just declare two HOCs (`withNavSidebar` and `withNavAndChannelSidebars`) that just take a `WrappedComponent` and render stuff directly - as is there's kind of a lot of indirection going on that I don't quite see the need for. Figuring out what `withNavSidebar = withSidebars(NavSidebar)` means, for instance, requires quite a bit of poking around. My thinking is that right now we only have two variations (nav and nav + channel description) and I don't think we're likely to have more anytime soon.

channels/factories.py
```diff
@@ -152,27 +157,47 @@ def get_instances(self, factory_type):
...
+            strategy (str): either a create or build strategy for the factory
+            kwargs (dict): the kwargs to pass to the factory
+
+        Returns:
+            object: the created object
+        """
+        if strategy == STRATEGY_CREATE:
+            return factory_cls.create(**kwargs)
+        elif strategy == STRATEGY_BUILD:
+            return factory_cls.build(**kwargs)
```
aliceriot: what's the difference between `.create` and `.build` here?

rhysyngsun: `factoryboy` explains the difference here: http://factoryboy.readthedocs.io/en/latest/reference.html#strategies

TLDR: `create` means persist, `build` means don't

rhysyngsun:
@aliceriot feedback addressed

channels/factories.py
aliceriot: Ok I figured that was what was going on, sounds good.

aliceriot:
Looking good now, the only thing I just noticed is that the spacing between the navigation sidebar and the list of posts is now slightly different on the homepage than it is on the subreddit page.

homepage:

homepagespacing

subredditspacing

these images are different sizes so it's maybe a little hard to tell but on the subreddit there's more space between them.

rhysyngsun:
@aliceriot I fixed that with a fixed padding around main-content. Note: the layout *may* appear to jump if your homepage scrolls but your channel page doesn't or vice versa.

Issue #231, opened 2017-10-02
User can edit their own discussion post


roberthouse54:
Design [below](https://github.com/mitodl/open-discussions/issues/231#issuecomment-334804533)

If a user has edited a post, we will show a little text that says [edited by author].

roberthouse54:
We will show an edit link under the post. This will only be shown to the author of the post. Only text posts can be edited. Link posts cannot be edited.

![image](https://user-images.githubusercontent.com/20047260/31288168-c6b7163c-aa91-11e7-8115-3a635b14b451.png)

Editing is done inline. Only the post text can be edited. The post title cannot be edited.

![image](https://user-images.githubusercontent.com/20047260/31288107-982cc622-aa91-11e7-93c9-f4298467da10.png)

After the post has been edited we will add "[edited by author]" at the end of the text.

![image](https://user-images.githubusercontent.com/20047260/31288157-bdf6fa80-aa91-11e7-849c-2045331c7bc5.png)

PR #348, opened 2017-11-21
Add post editing

aliceriot:

#231 

#### What's this PR do?

This adds UI for editing discussions posts (text posts). If the user is viewing the detail page for their own post there should be a an 'edit' button that opens a text box for editing the post.

You should be able to edit your own post, and you shouldn't be able to edit others! The edit button shouldn't show up on a url post.

### DIFF
--- a/static/js/components/CommentForms.js
+++ b/static/js/components/CommentForms.js
@@ -24,6 +24,7 @@ type CommentFormProps = {
   formDataLens: (s: string) => Object,
   processing: boolean,
   patchComment: (c: Comment) => void,
+  patchPost: (p: Post) => void,
   comment: Comment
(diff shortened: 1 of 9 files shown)


noisecapella:
I think according to the design the edit button should be lowercase

noisecapella:
Also it looks like the voting and edit buttons are not visible when the edit box is open

noisecapella:
I can edit text successfully but I see this warning in the console:
screenshot from 2017-11-27 12-32-12

aliceriot:
comments should be addressed! I added a check to hide the button row if the post is currently being edited, and I fixed the `setState` issue as well.

Issue #279, opened 2017-10-10
Moderator can remove posts


roberthouse54:
Remove is similar to delete, but slightly different. Only moderators can remove posts. Moderators can remove any posts, including their own. Moderators can also delete their own posts, hence moderators will see both "remove" and "delete" links on their own posts. On all other posts, moderators only have the remove option.

When a post is removed, it is no longer listed on the channel page. Users with the link can still go there if they paste the url in the browser—there, users will see a message that says "[Post removed]". They will still be able to read the comments on the page. 

Moderators will still be able to see the text of removed posts (although it will be grayed out). Removed posts can be un-removed by moderators by clicking "Approve". 

(See mocks below)

The API issue for this is #195.

![image](https://user-images.githubusercontent.com/20047260/31553425-778a8b1c-b008-11e7-90d4-18b19d21202b.png)

![image](https://user-images.githubusercontent.com/20047260/31553567-e7f02b14-b008-11e7-9c68-f31bd5ea6bc8.png)

**_After the post is removed, we will show a standard "Toast" message saying "Post Removed" (no mock for this.)_**

Post page as seen by moderator (comments will still be shown below this):
![image](https://user-images.githubusercontent.com/20047260/31553591-f6a0b606-b008-11e7-8bf5-80c796ec389d.png)

Post page as seen by user (comments will still be shown below this):
![image](https://user-images.githubusercontent.com/20047260/31553599-ff86a24e-b008-11e7-8586-3d16491965e0.png)

rhysyngsun:
@roberthouse54 @pdpinch One thing I discovered in testing the backend API is that for the user that made the post it can't be known if the post was removed. 

For other contributors, Reddit by default overrides the text post with the value `[removed]`. For url posts, nothing changes from the viewpoint of a non-moderator perspective.

Implementation-wise this means that the easiest thing is to just render the posts as-is (e.g. for removed text posts still display the title), otherwise it complicates the logic slightly for only a portion of use cases.

PR #355, opened 2017-11-27
Added requests for channel moderators

rhysyngsun:
Prerequisite to all moderator-related UI, e.g. #279, #263, etc

#### What's this PR do?
Adds a request to the channel moderators endpoint and adds some utils around it.

Go to a channel and post page and verify the pages load correctly and load the moderator list.

To support moderator tasks in the UI, we need to have an idea if the current user is a moderator on the channel. This PR adds the api calls to retrieve that list as well as a util to see if a user is a moderator.

### DIFF
--- a/static/js/factories/channels.js
+++ b/static/js/factories/channels.js
@@ -19,3 +19,12 @@ export const makeChannel = (privateChannel: boolean = false) => ({
 export const makeChannelList = (numChannels: number = 20) => {
   return R.range(0, numChannels).map(() => makeChannel(Math.random() > 0.5))
 }
+
+export const makeModerators = (username: ?string = null) =>
+  username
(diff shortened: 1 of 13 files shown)


aliceriot:
functionality looks good! reading code now.

static/js/factories/channels.js
```diff
@@ -19,3 +19,12 @@ export const makeChannel = (privateChannel: boolean = false) => ({
...
 }
+
+export const makeModerators = (username: ?string = null) =>
+  username
+    ? [
+      {
+        moderator_name: username
+      }
+    ]
+    : []
```
aliceriot: this formatting is goofy, haha. prettier does some weird things sometimes I guess

PR #366, opened 2017-11-29
Fix field name for channel description

### LINKED ISSUE #365: Channel description is not showing up in about panel

noisecapella:
Fixes #365 

#### What's this PR do?
Fixes the field used for the about page

Verify that the text shows up in the about panel

### DIFF
--- a/static/js/components/ChannelSidebar.js
+++ b/static/js/components/ChannelSidebar.js
@@ -19,8 +19,7 @@ const ChannelSidebar = ({ channel }: ChannelSidebarProps) => {
       <ReactMarkdown
         disallowedTypes={["Image"]}
         source={
-          channel.public_description ||
-          "(There is no description of this channel)"
+          channel.description || "(There is no description of this channel)"
         }
         escapeHtml
         className="description"
--- a/static/js/components/ChannelSidebar.js
+++ b/static/js/components/ChannelSidebar.js
@@ -1,3 +1,4 @@
+// @flow
 import React from "react"
 import ReactMarkdown from "react-markdown"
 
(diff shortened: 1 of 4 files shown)


noisecapella:
@aliceriot I reverted the change, this PR now just makes some minor tweaks to tests and the factory

aliceriot:
ok, did we decide that we want the `.description` field actually?

static/js/components/ChannelSidebar.js
```diff
@@ -1,3 +1,4 @@
+// @flow
```
aliceriot: :scream:

static/js/components/ChannelSidebar_test.js
```diff
@@ -13,17 +14,17 @@ describe("ChannelSidebar", () => {
   it("should render sidebar", () => {
     const channel = makeChannel()
     const wrapper = renderSidebar(channel)
-    const [description] = wrapper.find(ReactMarkdown)
-    assert.equal(description.props.source, channel.description)
+    const description = wrapper.find(ReactMarkdown)
+    assert.equal(description.props().source, channel.description)
```
aliceriot: what's the reason for this change?

noisecapella: Flow is unhappy with the `const [description] = wrapper.find(ReactMarkdown)` line since it doesn't think the `Wrapper` is indexable. https://github.com/facebook/flow/issues/1323

aliceriot: hmmm well I suppose we can go with this, although personally I think if the code if valid then flow is wrong and the code shouldn't have to change (especially given that this is a test file, and flow is of somewhat limited value when writing tests)

PR #367, opened 2017-11-29
s/self/text/ on guidelines

### LINKED ISSUE #354: Explain what a self post is
From the discussion guidelines: `Comments and the content of self posts can be edited after being submitted`

I think we want to change this to `text post` since that's the term we use on the create post page, or add additional language to explain a self post. It's a reddit term so we shouldn't assume people know what it means

aliceriot:

closes #354

#### What's this PR do?

changes `self` to `text` so that the guidelines match up with the 'create post' UI.

you can visit the guidelines and make sure it doesn't say 'self post' anywhere.


aliceriot:
phew, glad this huge and crucial PR got approved!

PR #370, opened 2017-11-29
add pin post api

### LINKED ISSUE #194: API for making posts "sticky"
As a moderator, I'd like to be able to specify that certain posts are pinned to the top of the list. 

This can be done using [PRAW's `sticky()` API](https://github.com/praw-dev/praw/blob/bc2eb71572344327afe5663136f07f1df6357ffc/praw/models/reddit/submission.py#L381)

aliceriot:

closes #194 

#### What's this PR do?

this adds a field ('stickied') to the Post API which can be toggled to make a post 'pinned' / 'sticky'.

You should be a moderator on your channels, so you should be able to toggle the boolean using the DRF web interface (just type `{"stickied": true}` and make sure it saves).

### DIFF
--- a/channels/api.py
+++ b/channels/api.py
@@ -443,6 +443,17 @@ class Api:
 
         return post.edit(text)
 
+    def pin_post(self, post_id, pinned):
+        """
+        Pin the Post!
+
+        Args:
+            post_id(str): the base36 id for the post
+            pinned(bool): the value for the 'stickied' field
+        """
+        post = self.get_post(post_id)
+        post.mod.sticky(pinned)
+
     def remove_post(self, post_id):
         """
         Removes the post, opposite of approve_post
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -124,6 +124,7 @@ class PostSerializer(serializers.Serializer):
     title = serializers.CharField()
     upvoted = WriteableSerializerMethodField()
     removed = WriteableSerializerMethodField()
+    stickied = serializers.BooleanField()
     score = serializers.IntegerField(source='ups', read_only=True)
     author_id = serializers.CharField(read_only=True, source='author')
     id = serializers.CharField(read_only=True)
@@ -263,6 +264,10 @@ class PostSerializer(serializers.Serializer):
 
         if "text" in validated_data:
             instance = api.update_post(post_id=post_id, text=validated_data['text'])
+
+        if "stickied" in validated_data:
+            sticky = validated_data["stickied"]
+            api.pin_post(post_id, sticky)
 
         _apply_vote(instance, validated_data, False)
         return api.get_post(post_id=post_id)
--- a/channels/views_test.py
+++ b/channels/views_test.py
@@ -74,7 +74,7 @@ def test_list_channels(client, jwt_header, private_channel_and_contributor):
     channel, _ = private_channel_and_contributor
     url = reverse('channel-list')
     resp = client.get(url, **jwt_header)
-    assert resp.status_code == 200
+    assert resp.status_code == status.HTTP_200_OK
     assert resp.json() == [
         {
             'title': channel.title,
@@ -100,7 +100,7 @@ def test_create_channel(client, staff_user, staff_jwt_header, reddit_factories):
         'public_description': channel.public_description,
     }
     resp = client.post(url, data=payload, **staff_jwt_header)
-    assert resp.status_code == 201
+    assert resp.status_code == status.HTTP_201_CREATED
     assert resp.json() == payload
 
 
@@ -134,7 +134,7 @@ def test_create_channel_nonstaff(client, jwt_header):
         'public_description': 'public',
     }
     resp = client.post(url, data=payload, **jwt_header)
-    assert resp.status_code == 403
+    assert resp.status_code == status.HTTP_403_FORBIDDEN
 
 
 def test_create_channel_noauth(client):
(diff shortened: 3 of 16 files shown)


channels/serializers.py
```diff
@@ -124,6 +124,7 @@ class PostSerializer(serializers.Serializer):
     title = serializers.CharField()
     upvoted = WriteableSerializerMethodField()
     removed = WriteableSerializerMethodField()
+    stickied = WriteableSerializerMethodField()
```
rhysyngsun: This could just be a `serializers.CharField()` to avoid the method.

channels/views_test.py
```diff
@@ -568,7 +572,36 @@ def test_update_post_text(client, private_channel_and_contributor, reddit_factor
...
+
+
+def test_update_post_stickied(client, private_channel_and_contributor, reddit_factories, staff_user):
+    """Test updating just the stickied boolean on a post"""
+    channel, user = private_channel_and_contributor
+    post = reddit_factories.text_post('just a post', user, channel=channel)
+    client.force_login(staff_user)
+    url = reverse('post-detail', kwargs={'post_id': post.id})
+    resp = client.patch(url, format='json', data={"stickied": True})
+    assert resp.status_code == 200
```
rhysyngsun: Could you change this to use `status.HTTP_200_OK` instead of the magic number?

channels/api.py
```diff
@@ -443,6 +443,12 @@ def update_post(self, post_id, text):
 
         return post.edit(text)
 
+    def pin_post(self, post_id, pinned):
+        """make the post sticky"""
```
rhysyngsun: Can you document the args here please?

channels/api.py
```diff
@@ -443,6 +443,12 @@ def update_post(self, post_id, text):
...
+        post = self.get_post(post_id)
+        post.mod.sticky(pinned)
+        return post
```
rhysyngsun: You can probably drop the return statement here since we don't need it

channels/views_test.py
aliceriot: Oh sure, I had just copied this from an existing test so I'll change those too

rhysyngsun:
@aliceriot could you also add tests for the following:

- [ ] Unstickying a post
- [ ] API returns `"stickied": true` for a stickied post

rhysyngsun:
Functionality looks good.

channels/serializers.py
aliceriot: I'm not exactly sure what you mean  - what method will this let me avoid?

aliceriot: I tried that real quick and I get `"False"` instead of `false` back from the API, so it seems like it's coercing the boolean to a string - is there a way around that?

channels/api.py
aliceriot: right, that makes sense, will do

channels/views_test.py
rhysyngsun: :woman_facepalming: I meant to write `serializers.BooleanField`

channels/serializers.py
rhysyngsun: :woman_facepalming: I meant to write serializers.BooleanField

aliceriot: cool, the BooleanField works great

aliceriot:
@rhysyngsun I added those tests and also added an additional test to make sure that a stickied post shows up first in a channel post listing

channels/api.py
```diff
@@ -443,6 +443,17 @@ def update_post(self, post_id, text):
 
         return post.edit(text)
 
+    def pin_post(self, post_id, pinned):
+        """
+        Removes the post, opposite of approve_post
+
+        Args:
+            post_id(str): the base36 id for the post
+            pinned(boolean): the value for the 'stickied' field
```
rhysyngsun: type should be `bool`

channels/views_test.py
```diff
@@ -411,6 +414,38 @@ def test_get_post_not_found(client, logged_in_profile):
     assert resp.status_code == status.HTTP_404_NOT_FOUND
 
 
+def test_get_post_stickied(client, private_channel_and_contributor, reddit_factories, staff_user):
+    """test that stickied posts come back that way"""
+    channel, user = private_channel_and_contributor
+    post = reddit_factories.text_post('just a post', user, channel=channel)
+    client.force_login(staff_user)
+    url = reverse('post-detail', kwargs={'post_id': post.id})
```
rhysyngsun: suggest using `staff_api.sticky()` instead of an API request because we're interested only in testing the `GET` here.

channels/views_test.py
```diff
@@ -455,12 +490,51 @@ def test_list_posts(client, missing_user, private_channel_and_contributor, reddi
...
 
+def test_list_posts_stickied(client, private_channel_and_contributor, reddit_factories, staff_user):
+    """test that the stickied post is first"""
+    channel, user = private_channel_and_contributor
+    posts = [
+        reddit_factories.text_post("great post!{}".format(i), user, channel=channel)
+        for i in range(4)
+    ]
+
+    client.force_login(staff_user)
```
rhysyngsun: Same suggestion as above to use `staff_api`

channels/views_test.py
```diff
@@ -568,18 +642,71 @@ def test_update_post_text(client, private_channel_and_contributor, reddit_factor
...
+    }
+
+
+def test_update_post_unsticky(client, private_channel_and_contributor, reddit_factories, staff_user):
+    """Test updating just the stickied boolean on a post"""
+    channel, user = private_channel_and_contributor
+    post = reddit_factories.text_post('just a post', user, channel=channel)
+    client.force_login(staff_user)
+    url = reverse('post-detail', kwargs={'post_id': post.id})
+    resp = client.patch(url, format='json', data={"stickied": True})
```
rhysyngsun: Same suggestion as above to use `staff_api`

aliceriot:
@rhysyngsun ok comments should be addressed

rhysyngsun:
@aliceriot looks good!

PR #375, opened 2017-11-30
Added UI to remove posts as a moderator

### LINKED ISSUE #279: Moderator can remove posts

rhysyngsun:
Fixes #279 

#### What's this PR do?
Adds the UI for a moderator to remove posts

- As a regular user, verify you see no option to remove posts
- As a moderator verify you see the option to remove posts
- Remove a post, it should no longer show up in the listing
- As a moderator, you should be able to see the post content
- As a regular user you should see a message that the post has been removed
- As a moderator, you should be able to re-approve the post and it should return to a non-removed state  (shows up in channel listing and regular users can see the content)

### DIFF
--- a/static/js/components/ExpandedPostDisplay_test.js
+++ b/static/js/components/ExpandedPostDisplay_test.js
@@ -6,7 +6,6 @@ import { mount } from "enzyme"
 import { Link } from "react-router-dom"
 import ReactMarkdown from "react-markdown"
 import R from "ramda"
-import sinon from "sinon"
 
 import ExpandedPostDisplay from "./ExpandedPostDisplay"
 import Router from "../Router"
@@ -19,7 +18,7 @@ import { actions } from "../actions"
 import { editPostKey } from "../components/CommentForms"
 
 describe("ExpandedPostDisplay", () => {
-  let helper, post, beginEditingStub
+  let helper, post, beginEditingStub, approvePostStub, removePostStub
 
   const renderPostDisplay = props => {
     props = {
@@ -27,7 +26,9 @@ describe("ExpandedPostDisplay", () => {
       beginEditing: R.curry((key, post, e) => {
         beginEditingStub(key, post, e)
       }),
-      forms: {},
+      approvePost: approvePostStub,
+      removePost:  removePostStub,
+      forms:       {},
       ...props
     }
     return mount(
@@ -40,7 +41,9 @@ describe("ExpandedPostDisplay", () => {
   beforeEach(() => {
     helper = new IntegrationTestHelper()
     post = makePost()
-    beginEditingStub = sinon.stub()
+    beginEditingStub = helper.sandbox.stub()
+    approvePostStub = helper.sandbox.stub()
+    removePostStub = helper.sandbox.stub()
   })
 
   afterEach(() => {
@@ -127,7 +130,7 @@ describe("ExpandedPostDisplay", () => {
     const post = makePost(false)
     SETTINGS.username = post.author_id
     const wrapper = renderPostDisplay({ post })
-    wrapper.find(".comment-action-button").simulate("click")
+    wrapper.find(".edit-post").simulate("click")
     assert.ok(beginEditingStub.called)
   })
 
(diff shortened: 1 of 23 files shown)


rhysyngsun:
I don't know what's up with Codecov today, but the coverage is pretty solid.

noisecapella:
When I click the 'Remove' button I see a '#' appear in the URL... maybe there's a missing `preventDefault`?

noisecapella:
I'm unable to see the post I removed in my own post list, though if I'm still on the page I can click to approve it. Not sure if we can do anything about it in this PR, I assume reddit has some special listing of removed posts for moderators

rhysyngsun:
Yeah, posts are completely removed from listings for all users, nothing we can do short of modifying reddit upstream to fix that. Not sure on whether they're visible elsewhere.

static/js/components/ExpandedPostDisplay_test.js
```diff
@@ -41,6 +43,8 @@ describe("ExpandedPostDisplay", () => {
     helper = new IntegrationTestHelper()
     post = makePost()
     beginEditingStub = sinon.stub()
```
noisecapella: Can you refactor this to use a sandbox instead?

static/js/components/material/Snackbar.js
```diff
@@ -0,0 +1,52 @@
...
+  snackbarRoot = null
+
+  props: SnackbarProps
+
+  componentDidMount() {
+    this.snackbar = new MDCSnackbar(this.snackbarRoot)
+  }
+
+  componentWillReceiveProps(nextProps: SnackbarProps) {
+    if (this.props.snackbar !== nextProps.snackbar) {
```
noisecapella: This will compare identity, not value. Is there something in ramda for deep equality?

noisecapella:
I'm noticing if you click 'Remove' then go to another tab (at least in Firefox) and then go back, the snackbar stays up forever

Looks like this is in the MDC component: https://github.com/material-components/material-components-web/issues/1398

static/js/reducers/index_test.js
```diff
@@ -256,6 +261,16 @@ describe("reducers", () => {
...
     })
+
+    it("should set snackbar message", async () => {
+      const payload = {
+        message: "hey there!"
+      }
+      const state = await dispatchThen(setSnackbarMessage(payload), [
+        SET_SNACKBAR_MESSAGE
+      ])
+      assert.equal(state.snackbar, payload)
```
noisecapella: This should be `deepEqual`

static/js/reducers/index_test.js
```diff
@@ -256,6 +261,16 @@ describe("reducers", () => {
...
+    })
   })
 
   describe("subsribed channels (getChannels) reducer", () => {
```
noisecapella: Would you mind changing this to `subscribed` too?

static/js/components/material/Snackbar.js
rhysyngsun: Ok, I'll need to add in something to guarantee object uniqueness per dispatch of `setSnackbarMessage` too.

rhysyngsun: I added a sequential id to the state, this should guarantee the above.

rhysyngsun:

Issue #278, opened 2017-10-10
Moderator can remove comments


roberthouse54:
Remove is similar to delete, but slightly different. Only moderators can remove comments. Moderators can remove any comments, including their own. Moderators can also delete their own comments, hence moderators will see both "remove" and "delete" links on their own comments. On all other comments, moderators only have the remove option.

When a comment is removed, it still appears in the comment tree, with a  message saying "[comment removed]". Removed comments can be un-removed by moderators by clicking "Approve". Moderators will still be able to see the text of removed comments, although they will be grayed out.

(See mocks below)

The API issue for this is #196

![image](https://user-images.githubusercontent.com/20047260/31553718-5f0a3230-b009-11e7-8f02-54dd428eba25.png)

![image](https://user-images.githubusercontent.com/20047260/31553733-6d13ec54-b009-11e7-8f1a-8c9436eb558b.png)

**_NOTE: After the comment is removed, we will show a standard "Toast" message saying "Post Removed" (no mock for this.)_**

Removed comment as seen by moderator:
![image](https://user-images.githubusercontent.com/20047260/31553761-7e06d562-b009-11e7-8360-41b3711a94f0.png)

Removed comment as seen by user:
![image](https://user-images.githubusercontent.com/20047260/31553774-883edf5c-b009-11e7-9926-1721ae7a391b.png)

rhysyngsun:
@pdpinch @roberthouse54 is there a confirmation for approving a comment? if yes, what message should that display?

roberthouse54:
I assume you mean for removing a comment. If that is right, let's go with "Are you sure you want to remove this comment?" Yes | Cancel

roberthouse54:
Sorry, scratch that..... I forgot we had "Approve". I don't think we need a confirmation for this.

rhysyngsun:
Note: removed comments don't come back from the reddit API at all for non-moderator users, so they just won't show up for them (as opposed to the mockup above where you seen an indicator it was removed).

PR #383, opened 2017-12-05
Added comment removal API

rhysyngsun:
Part of #278 

#### What's this PR do?
Updates the comments api to allow removal of comments by moderators.

Also reworked the `CommentFactory` because it hadn't been really used in a test yet and the previous iteration was impractical.

Go to `http://localhost:8063/api/v0/comments/:comment_id/` and `PATCH` with `{"removed": true}` and `{"removed": false}` and verify you see the expected state changes.


aliceriot:
OK I can remove a comment, but it looks like it still comes back in the list of comments for a post. Is this the expected behavior?

rhysyngsun:
@aliceriot if you're moderator, yes, because you need something to be able to re-approve.

aliceriot:
aha that makes sense!

channels/serializers.py
```diff
@@ -401,10 +410,18 @@ def create(self, validated_data):
     def update(self, instance, validated_data):
         if validated_data.get('comment_id'):
             raise ValidationError("comment_id must be provided via URL")
+
+        # import pdb ; pdb.set_trace()
```
aliceriot: :thinking:

channels/utils.py
```diff
@@ -87,6 +87,7 @@ def translate_praw_exceptions():
         # but I haven't seen any other causes for redirects
         raise NotFound() from exc
     except APIException as exc:
+        print(exc)
```
aliceriot: stray print?

aliceriot:
ok cool, I also confirmed that, if I remove a comment as a moderator, as a non-moderator I see this:

removed_thing

pdpinch:
Can we make it look like the design?

<image src="https://user-images.githubusercontent.com/20047260/31553774-883edf5c-b009-11e7-9926-1721ae7a391b.png">

aliceriot:
As I understand it this is just covering the API work

rhysyngsun:
@aliceriot feedback addressed

PR #384, opened 2017-12-06
add pinning ui

### LINKED ISSUE #236: UI for Pinning posts
There is a separate API issue for this: #194

Moderators will see a "pin" link under every post. Clicking the link will pin posts to the top of the channel. We will display pinned posts with a blue background, and the pin icon. If multiple posts are pinned, we will show the latest pinned post at the top. There will be an "unpin" link for pinned posts, which will return the post to the regular list of posts.

image

aliceriot:

closes #236 

#### What's this PR do?

This adds UI for pinning posts on a channel. There's a special display for the 'pinned' post, and then also a little button for pinning / unpinning a post. This button should only show up for moderators.

The frontpage should look the same as it does on master. On a subreddit, if your user is a moderator you should see an option (on text posts only) to 'pin' a post. When you select this option the pinned post should change it's display, and after a quick refresh of the post data it should appear at the top of the list. Unpinning it should reverse this.

These buttons shouldn't show up for a non-moderator, but a non-moderator should still see pinned posts on the channel page. A post which is pinned on a subreddit shouldn't look any different on the homepage.

### DIFF
--- a/static/js/components/CompactPostDisplay.js
+++ b/static/js/components/CompactPostDisplay.js
@@ -17,7 +17,10 @@ class CompactPostDisplay extends React.Component<*, void> {
   props: {
     post: Post,
     showChannelLink: boolean,
-    toggleUpvote: Post => void
+    toggleUpvote: Post => void,
+    togglePinPost: Post => void,
+    showPinUI: boolean,
+    isModerator: boolean
   }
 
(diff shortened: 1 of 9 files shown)


rhysyngsun:
Functionality looks good.

PR #387, opened 2017-12-08
Add comment removal UI

### LINKED ISSUE #278: Moderator can remove comments

rhysyngsun:
Closes #278 

#### What's this PR do?
Adds links and confirmation dialog (on remove only) for removing/adding comments.

- Be sure you don't see these options as a non-moderator user
- Verify that clicking remove and approve does what you would expect

remove...approve...remove...approve...

### DIFF
--- a/static/js/containers/PostPage.js
+++ b/static/js/containers/PostPage.js
@@ -4,6 +4,7 @@ import React from "react"
 import { connect } from "react-redux"
 import R from "ramda"
 import DocumentTitle from "react-document-title"
+import { Dialog } from "@mitodl/mdl-react-components"
 
 import Card from "../components/Card"
 import withLoading from "../components/Loading"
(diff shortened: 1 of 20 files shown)


aliceriot:
The functionality all seems to be good, but I see this stack trace in the console when I 'remove' a comment:

ffhfhfhasdfasdflkjfnfnfnfn

the request seems to still go through ok so I'm not sure exactly what's up

aliceriot:
that bug is fixed!

static/js/containers/PostPage.js
```diff
@@ -140,6 +155,48 @@ class PostPage extends React.Component<*, void> {
...
+  }
+
+  hideRemoveCommentDialog = () => {
+    const { dispatch } = this.props
+    dispatch(clearModeratingComment())
+    dispatch(hideDialog(DIALOG_REMOVE_COMMENT))
+  }
+
+  removeComment = async () => {
+    const { dispatch, removingComment } = this.props
```
aliceriot: `removingComment` is sort of like the focused comment, right? I might suggest rename to something like `commentToRemove` or something along those lines, the `removingComment` name to me sounds a bit more like something telling us about the status (currently removing vs no).

static/js/reducers/moderation.js
```diff
@@ -0,0 +1,29 @@
...
+} from "../actions/moderation"
+
+import type { Action } from "../flow/reduxTypes"
+import type { CommentInTree } from "../flow/discussionTypes"
+
+export type ModerationState = {
+  comment: ?CommentInTree
+}
+
+export const INITIAL_UI_STATE: ModerationState = {
```
aliceriot: should this be called something like `INITIAL_MODERATION_STATE`?

rhysyngsun: Ooops you caught my copy/paste!

rhysyngsun:
@aliceriot feedback addressed

PR #401, opened 2017-12-13
add user post deletion

### LINKED ISSUE #234: Post owner can delete posts
There is a separate issue, #273 for the API work necessary for this, which includes deleting a post and retrieving the deleted post from a channel. 

We will show a delete link under a post. We will only show this under the user's own posts. 

image

We will show a delete confirmation modal:

image

After the post has been deleted, we will delete the content and the title of the post, and show the text "[post deleted by author]" or "[post deleted by moderator]" in the post title. The user name and avatar will no longer be shown, nor will the upvoting arrows or the score. The post will no longer be displayed in the channel page or on the home page. But if a user has a link to the post, they will be able to link to the page and see the comments tree, but the post itself will be displayed as shown.

image

aliceriot:

closes #234 

#### What's this PR do?

This adds UI so that a user can delete their own posts.

If you look at the detail view for one of your own posts you should see a little 'delete' button.

- clicking this button should open a confirmation dialog
- canceling the dialog, hitting 'escape', or clicking outside should close it
- confirming the dialog should redirect you to the channel page.

If you delete your post it's, like, _really_ deleted (forever...) so don't delete a post that is special, I guess?

### DIFF
--- a/static/js/containers/PostPage.js
+++ b/static/js/containers/PostPage.js
@@ -35,6 +35,7 @@ import { anyError } from "../util/rest"
 import { getSubscribedChannels } from "../lib/redux_selectors"
 import { beginEditing } from "../components/CommentForms"
 import { formatTitle } from "../lib/title"
+import { channelURL } from "../lib/url"
 
 import type { Dispatch } from "redux"
 import type { Match } from "react-router"
@@ -62,7 +63,9 @@ type PostPageProps = {
   commentInFlight: boolean,
   // from the router match
   channelName: string,
-  postID: string
+  postID: string,
+  postDeleteDialogVisible: boolean,
+  history: Object
 }
 
 const DIALOG_REMOVE_COMMENT = "DIALOG_REMOVE_COMMENT"
@@ -71,6 +74,8 @@ const DIALOG_REMOVE_COMMENT = "DIALOG_REMOVE_COMMENT"
 const shouldLoadData = R.complement(
   R.allPass([R.eqProps("postID"), R.eqProps("channelName")])
 )
+
+const DELETE_POST_DIALOG = "DELETE_POST_DIALOG"
 
 class PostPage extends React.Component<*, void> {
   props: PostPageProps
@@ -197,6 +202,27 @@ class PostPage extends React.Component<*, void> {
     )
   }
 
+  setShowDeletePostDialog = (visible: boolean) => {
+    const { dispatch } = this.props
+    if (visible) {
+      dispatch(showDialog(DELETE_POST_DIALOG))
+    } else {
+      dispatch(hideDialog(DELETE_POST_DIALOG))
+    }
+  }
+
+  deletePost = async (post: Post) => {
+    // ⚠️  this is a destructive action! ⚠️
+    const { dispatch, history, channelName } = this.props
+    await dispatch(actions.posts["delete"](post.id))
+    history.push(channelURL(channelName))
+    dispatch(
+      setSnackbarMessage({
+        message: "Post has been deleted"
+      })
+    )
+  }
+
   render() {
     const {
       dispatch,
@@ -206,7 +232,8 @@ class PostPage extends React.Component<*, void> {
       forms,
       commentInFlight,
       showRemoveCommentDialog,
-      isModerator
+      isModerator,
+      postDeleteDialogVisible
     } = this.props
     if (!channel || !post || !commentsTree) {
       return null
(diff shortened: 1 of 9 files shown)


rhysyngsun:
Functionality looks good, but could you change the affirmation button label to match the mockups? (e.g. "Yes, Delete". Also, I noticed that the label for the delete link on the post is capitalized but the others are lowercase.

aliceriot:
I should look at mockups more closely :woman_facepalming: 

comments should be addressed

static/js/containers/PostPage.js
```diff
@@ -49,14 +51,18 @@ type PostPageProps = {
   commentInFlight: boolean,
   // from the router match
   channelName: string,
-  postID: string
+  postID: string,
+  showDialog: boolean,
```
rhysyngsun: Should this be refactored this to a name that's a little more specific to which dialog it is for?

static/js/containers/PostPage.js
```diff
@@ -158,6 +187,15 @@ class PostPage extends React.Component<*, void> {
       <div>
         <ChannelBreadcrumbs channel={channel} />
         <DocumentTitle title={formatTitle(post.title)} />
+        <Dialog
+          open={showDialog}
+          onCancel={() => this.setShowDeletePostDialog(false)}
```
rhysyngsun: You can change this to `hideDialog={...}` and then you don't need to also hide the dialog in `deletePost`

static/js/lib/auth.js
```diff
@@ -44,3 +47,27 @@ export const fetchWithAuthFailure = async (...args: any) => {
     return fetchJSONWithCSRF(...args)
   }
 }
+
+export const fetchWithAuthFailure = async (...args: any) => {
```
rhysyngsun: Suggestion: refactor this and the above to something like this:

```js
const withAuthFailure = (fetchFunc: Function) = {
  return async (...args: any) => {
    try {
      return await fetchFunc(...args)
     } catch (fetchError) {
       if (!fetchError || fetchError.errorStatusCode !== 401) {
         // not an authentication failure, rethrow
         throw fetchError
       }
     
       try {
         // renew the session
         const session = await renewSession()
         const json = await session.json()
         if (!json.has_token) {
           return redirectAndReject("New token was not created")
         }
       } catch (_) {
         return redirectAndReject("Could not renew session")
       }
         // try again now that we have a proper auth
         return fetchFunc(...args)
       }
     }
  }
}

const fetchJSONWithAuthFailure = withAuthFailure(fetchJSONWithCSRF)
const fetchWithAuthFailure = withAuthFailure(fetchWithCSRF)
```

static/js/containers/PostPage.js
aliceriot: oh nice! I will do that.

static/js/containers/PostPage.js
aliceriot: yeah that's a good idea

static/js/lib/auth.js
aliceriot: That was my first instinct as well, I ended up just going with two separate functions because I ran into an issue trying to mock out things for tests with the above approach. Basically, because if we declare

```js
const fetchJSONWithAuthFailure = withAuthFailure(fetchJSONWithCSRF)
```

Then `fetchJSONWithAuthFailure` has a reference to `fetchJSONWithCSRF` which resolves at module-load time. When we're running the tests we use sinon to stub out `fetchJSONWithCSRF`, but because `fetchJSONWithAuthFailure` is already holding on to the un-stubbed function the stub doesn't do anything.

One way to do something to DRY up the code would be to try to resolve the `fetchFunc` at call-time instead of module-load time. I considered doing something like this:

```js
const withAuthFailure = (fetchJSON: boolean) = {
  return async (...args: any) => {
    const fetchFunc = fetchJSON
      ? fetchJSONWithCSRF
      : fetchWithCSRF
      
    try {
      return await fetchFunc(...args)
     } catch (fetchError) {
       if (!fetchError || fetchError.errorStatusCode !== 401) {
         // not an authentication failure, rethrow
         throw fetchError
       }
     
       try {
         // renew the session
         const session = await renewSession()
         const json = await session.json()
         if (!json.has_token) {
           return redirectAndReject("New token was not created")
         }
       } catch (_) {
         return redirectAndReject("Could not renew session")
       }
         // try again now that we have a proper auth
         return fetchFunc(...args)
       }
     }
  }
}

const fetchJSONWithAuthFailure = withAuthFailure(true)
const fetchWithAuthFailure = withAuthFailure(false)
```

and that works just fine, but idk, it struck me as a big inelegant, so I thought the simplest and most clear thing would be to just go with the two functions. What do you think?

rhysyngsun: Hmm, I agree that's a bit obtuse. An alternative using my original code that should still work with sinon is: 

```js
export const fetchJSONWithAuthFailure = withAuthFailure((...args: any) => fetchJSONWithCSRF(...args))
export const fetchWithAuthFailure = withAuthFailure((...args: any) => fetchWithCSRF(...args))
```

But I'm also ok with things as-is if you feel that extra anonymous func is too much.

aliceriot: no I think that makes sense, I'll go with that

aliceriot:
@rhysyngsun feedback addressed

PR #404, opened 2017-12-18
add user comment deletion

### LINKED ISSUE #233: User can delete their own comments
API issue is #280

We will show a delete link under all of a user's own comments. 

image

We will show a delete confirmation modal:

image

After the comment has been deleted, we will delete the content of the comment, and show the text "[comment deleted by author]" or "[comment deleted by moderator]". The comment will stay in the comment tree. The user name and avatar will no longer be shown, nor will the upvoting arrows or the score. There is no way to un-delete a comment.

image

image

aliceriot:

closes #233 

#### What's this PR do?

this adds deletion functionality for the user's own comments.

Basically, if you're looking at your own comments you should see a 'delete' button. Clicking the button should open a confirmation dialog. If you click 'yes' the comment should update to look like this in the UI:

fjfjfjasdfasdfasdf

If you refresh the page you should see one of two things: if the comment you deleted did not have any replies it should just disappear, and if the comment in question had at least one reply you should see something like the above picture, with the replies sticking around under a 'deleted' placeholder.

If you're looking at someone else's comments you shouldn't see the 'delete' button.


static/js/components/CommentTree_test.js
```diff
@@ -111,15 +114,35 @@ describe("CommentTree", () => {
   })
 
   it('should include an "Edit" button, if the user wrote the comment', () => {
-    comments[0].author_id = "username"
-    SETTINGS.username = "username"
+    // comments[0].author_id = "username"
```
noisecapella: Can you remove these lines?

static/js/containers/PostPage.js
```diff
@@ -202,6 +202,19 @@ class PostPage extends React.Component<*, void> {
     )
   }
 
+  deleteComment = async () => {
+    // ⚠️  this is a destructive action! ⚠️
```
noisecapella: ⚠️

noisecapella:
After the comment is deleted I see a 'delete' and 'edit' link on my comment:

screenshot from 2017-12-19 11-16-41

noisecapella:
When I refresh the browser the comment doesn't have those two links anymore. However the 'remove' should probably also not be visible for a deleted comment

Other than this everything looks good

aliceriot:
ok I made a small change to clear the `author_id` so that the 'delete' and 'edit' actions don't show up anymore.

The 'remove' action will only show up for moderators - do you think it's a big deal if they see that for a second? We don't have a clean way to tell whether a comment has been previously deleted or not, so I'm not sure of a good way to tell whether we want to show the 'remove' button or not. It's a small edge case anyway, since moderators likely will not be making lots of comments, I imagine.

static/js/components/CommentTree_test.js
aliceriot: oopsie

noisecapella:
That's fine with me

PR #405, opened 2017-12-18
some dependency upgrades

### LINKED ISSUE #106: Upgrade sinon

aliceriot:

closes #106 
upgrades a bunch of stuff besides sinon too

#### What's this PR do?

Does some of the easier dependency upgrades. We should circle back sometimes soon to bump out build / dev tools (webpack + babel + react-hot-loader and all that) to the latest versions, and we should also circle back to upgrade React soon as well. Upgrading flow is probs a good idea too.

everything should work as before, tests should pass, etc.


webpack.config.shared.js
```diff
@@ -33,7 +33,6 @@ module.exports = {
     query: {
       "presets": [
         ["env", { "modules": false }],
-        "latest",
```
noisecapella: What's the significance of this change?

aliceriot: the `latest` preset is deprecated, see https://babeljs.io/docs/plugins/preset-latest/

not sure why we were using it at all, it has been deprecated for quite some time (I think since before we started this project iirc). the `env` preset takes its place, so using the `latest` preset here was pretty redundant.

PR #407, opened 2017-12-19
upgrade node version

### LINKED ISSUE #321: Upgrade to node 8.x

aliceriot:

closes #321 

#### What's this PR do?

This upgrades our node.js version to 9.3 everywhere, and also does a healthy bump to our yarn version.

It seems that in the time since we last upgraded node it started throwing warnings for unhandled promise rejections, so I went through the tests to fix all of the examples that I could find (mostly things related to our API stub functions). Most of the code changes (other than just changing node.js version numbers) have to do with that.

I also refactored some of the node.js Dockerfiles to be more inline with how we have them in MM now.

Everything should work normally. You should be able to check out this branch, build the docker container, and run things in development mode. The tests should pass on travis, and the build should work alright on the CI.

### DIFF
--- a/static/js/components/SpinnerButton_test.js
+++ b/static/js/components/SpinnerButton_test.js
@@ -17,32 +17,39 @@ describe("SpinnerButton", () => {
   })
 
   for (const promiseState of ["not_clicked", "rejected", "resolved"]) {
-    it(`passes through all props when promise state is ${promiseState}`, async () => {
-      const promise =
-        promiseState === "rejected" ? Promise.reject() : Promise.resolve()
-      const onClickPromise = () => promise
-      const props = {
-        "data-x":       "y",
-        onClickPromise: onClickPromise,
-        className:      "class1 class2"
-      }
-      const wrapper = shallow(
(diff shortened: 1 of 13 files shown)


static/js/components/SpinnerButton_test.js
```diff
@@ -17,32 +17,39 @@ describe("SpinnerButton", () => {
...
-        "data-x":       "y",
-        onClickPromise: onClickPromise,
-        className:      "class1 class2"
-      }
-      const wrapper = shallow(
-        <SpinnerButton {...props}>childText</SpinnerButton>
-      )
-      const button = wrapper.find("button")
-      const buttonProps = button.props()
+    it(`passes through all props when promise state is ${promiseState}`, () => {
```
noisecapella: Was the async function not working here?

aliceriot: the rejected promise was causing an issue because we weren't handling the rejection, I rewrite it to return a `new Promise` just because it made it a little easier

PR #417, opened 2017-12-29
Added api to report posts and comments (#197)

### LINKED ISSUE #197: API for reporting posts
As a user, I need to be able to report offensive or spammy posts or comments

[PRAW docs](http://praw.readthedocs.io/en/latest/code_overview/models/submission.html?highlight=report()#praw.models.Submission.report)

rhysyngsun:
Fixes #197

#### What's this PR do?
Adds an api for reporting posts and comments to moderators

Go to /api/v0/reports/ and POST a new report (e.g. `{'comment_id': 'ID', 'reason': 'spam'} and verify in the reddit UI that this shows up for your moderator user under the reports.

### DIFF
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -536,3 +536,34 @@ class SubscriberSerializer(serializers.Serializer):
         api = Api(user=self.context['request'].user)
         channel_name = self.context['view'].kwargs['channel_name']
         return api.add_subscriber(validated_data['subscriber_name'], channel_name)
+
+
+class ReportSerializer(serializers.Serializer):
+    """Serializer for reporting posts and comments"""
+    post_id = serializers.CharField(required=False)
+    comment_id = serializers.CharField(required=False)
+    reason = serializers.CharField(max_length=100)
+
+    def validate(self, attrs):
+        """Validate data"""
+        if 'post_id' not in attrs and 'comment_id' not in attrs:
+            raise ValidationError("You must provide one of either 'post_id' or 'comment_id'")
+        elif 'post_id' in attrs and 'comment_id' in attrs:
+            raise ValidationError("You must provide only one of either 'post_id' or 'comment_id', not both")
+
+        return attrs
+
+    def create(self, validated_data):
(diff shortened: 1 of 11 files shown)


aliceriot:
ok this works! I can report a comment, and if I do it shows up in the moderation UI in reddit. Reading code now.

channels/serializers.py
```diff
@@ -536,3 +536,34 @@ def create(self, validated_data):
...
+            raise ValidationError("You must provide only one of either 'post_id' or 'comment_id', not both")
+
+        return attrs
+
+    def create(self, validated_data):
+        """Create a new report"""
+        api = Api(user=self.context['request'].user)
+        post_id = validated_data.get('post_id', None)
+        comment_id = validated_data.get('comment_id', None)
+        reason = validated_data['reason']
```
aliceriot: should we validate that `reason` is included as well?

channels/views/reports.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    View to report a comment or post
+    """
+    permission_classes = (IsAuthenticated,)
+
+    def get_serializer_context(self):
+        """Context for the request and view"""
+        return {
+            'request': self.request,
+            'view': self,
```
aliceriot: it doesn't look like we're using this `view` property in the serializer anyway - do we need to add it for some behind-the-scenes thing?

channels/serializers.py
rhysyngsun: serializer fields are required by default in DRF so we shouldn't need to

channels/views/reports.py
rhysyngsun: It was just from copy/paste of other views, I'll give removing it a try

rhysyngsun: Removed it.

channels/serializers.py
aliceriot: oh right, I learned that before but forgot! haha

rhysyngsun:
@aliceriot feedback addressed

PR #419, opened 2018-01-03
Add check_pip.sh

noisecapella:
None

#### What's this PR do?
Add script to check whether the python docker image is up to date, and have it run with tox

 - Add a package to `requirements.txt`, for example `tornado`
 - `rm -rf .tox`, then run `docker-compose run web tox`
 - You should see a diff listing the new `tornado` package as a missing requirement

### DIFF
--- a/scripts/test/check_pip.sh
+++ b/scripts/test/check_pip.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -e -o pipefail
+
+PIP_OLD=$(mktemp)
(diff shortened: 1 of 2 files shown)


noisecapella:
There was another change recently to `update-docker-hub.sh` where each docker image has its own hash, so each PR which makes a change to dependencies can have its own docker image

rhysyngsun:
Ok, sounds like you have things in hand to avoid that issue, 👍

Issue #215, opened 2017-09-29
Implement a 404 page for posts


rhysyngsun:
You currently get a "Error loading page" message if you try to go to a nonexistent or deleted post

See also #408

pdpinch:
@roberthouse54 is there a design for this?

Issue #408, opened 2017-12-20
Requesting a non-existant channel raises an error to Sentry instead of returning 404


pdpinch:
I thought we had addressed this in #319/#293 but I saw the error below in Sentry:

Redirect: Redirect to /subreddits/search

https://sentry.io/mit-office-of-digital-learning/discussions/issues/421462250/

```
Redirect: Redirect to /subreddits/search
(15 additional frame(s) were not displayed)
...
  File "rest_framework/views.py", line 474, in dispatch
    response = handler(request, *args, **kwargs)
  File "rest_framework/generics.py", line 240, in get
    return self.list(request, *args, **kwargs)
  File "rest_framework/mixins.py", line 40, in list
    queryset = self.filter_queryset(self.get_queryset())
  File "channels/views.py", line 88, in get_queryset
    return api.list_moderators(channel_name)
  File "channels/api.py", line 756, in list_moderators
    return self.get_channel(channel_name).moderator(redditor=moderator_name)

Redirect: Redirect to /subreddits/search
```

The user experience needs work too, similar to #215. The user simply gets the message "Error loading page"

https://www.dropbox.com/s/0s85ay9u5y7ra7u/Screenshot%202017-12-20%2008.03.11.png?dl=0

PR #421, opened 2018-01-05
add a 404 page to the post

aliceriot:

#215 

#### What's this PR do?

This implements a 'not found' component and makes some changes so we can show it on the post page, when we get a 404 back from Reddit.

Go to a post that doesn't exist, and you should see a nice 404 page instead of just a generic error.

You should be able to do anything else you would normally be able to do with a post

### DIFF
--- a/static/js/containers/PostPage.js
+++ b/static/js/containers/PostPage.js
@@ -13,6 +13,7 @@ import ExpandedPostDisplay from "../components/ExpandedPostDisplay"
 import CommentTree from "../components/CommentTree"
 import { ReplyToPostForm } from "../components/CommentForms"
 import withNavSidebar from "../hoc/withNavSidebar"
+import NotFound from "../components/404"
 
 import { formatCommentsCount } from "../lib/posts"
 import { actions } from "../actions"
@@ -28,7 +29,7 @@ import {
 } from "../util/api_actions"
 import { getChannelName, getPostID } from "../lib/util"
 import { isModerator } from "../lib/channels"
-import { anyError } from "../util/rest"
+import { anyErrorExcept404 } from "../util/rest"
 import { getSubscribedChannels } from "../lib/redux_selectors"
 import { beginEditing } from "../components/CommentForms"
 import { formatTitle } from "../lib/title"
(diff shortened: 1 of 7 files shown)


noisecapella:
If I click one of the channel links in the sidebar I get redirected to an error page:

screenshot from 2018-01-05 16-23-32

static/js/containers/PostPage.js
```diff
@@ -286,9 +296,10 @@ class PostPage extends React.Component<*, void> {
           title="Delete Post"
           submitText="Yes, Delete"
         >
-          Are you sure you want to delete this post?
+            Are you sure you want to delete this post?
         </Dialog>
         <Card>
+          <div>NO</div>
```
noisecapella: Haha I think this was meant for the dialog

static/js/util/rest_test.js
```diff
@@ -0,0 +1,86 @@
+// @flow
+import { assert } from "chai"
+
+import { anyProcessing, allLoaded, anyError, anyErrorExcept404 } from "./rest"
+
+describe("rest utils", () => {
+  describe("anyProcessing", () => {
+    it("should return false if nothing has processing true", () => {
```
noisecapella: Can you add a test where `processing: false` and another object has `processing:true`?

static/scss/not-found.scss
```diff
@@ -0,0 +1,15 @@
+.not-found {
```
noisecapella: Is there a design for this?

aliceriot: image_1024_720

aliceriot:
that issue with the channel page is also an issue on master, I was planning to address it when I did #408

static/js/util/rest_test.js
aliceriot: sure

static/js/containers/PostPage.js
aliceriot: oh oops haha

static/scss/not-found.scss
aliceriot: (I don't think this mock is posted on invision yet)

aliceriot:
@noisecapella comments should be addressed!

PR #422, opened 2018-01-08
add a 404 message to the channel page

### LINKED ISSUE #393: User requesting a channel without permission raises a 403 error on moderator API
This doesn't affect user experience, but leads to an error in Sentry.

### LINKED ISSUE #408: Requesting a non-existant channel raises an error to Sentry instead of returning 404

aliceriot:

closes #408 
closes #393 

#### What's this PR do?

This adds a 404 page to the channel, which you should see if you try to visit a channel that doesn't exist. I also moved stuff around so we no longer raise an unhandled exception when the request comes back with a 404.

Additional, there was previously an issue where if you visited a post that doesn't exist and then tried to navigate to a channel, the 404 error from the post would prevent the channel data from showing itself. To fix this I just added an action to the post reducer to clear out the error state, which we can call on the channel page if we need to.

Go to something like `/channel/flflasdfajsdf` (well unless you have a channel called that :P). You should see a 404 message, like what we have now on the post page. Then if you visit another channel all should be well. Additionally, the error situation I described above (visiting a 'not found' post and then going to a channel page) should no longer raise any errors.

### DIFF
--- a/cassettes/channels.views.channels_test.test_get_short_channel.json
+++ b/cassettes/channels.views.channels_test.test_get_short_channel.json
@@ -1,0 +1,74 @@
+{
+  "http_interactions": [
+    {
+      "recorded_at": "2018-01-09T14:38:53",
+      "request": {
+        "body": {
+          "encoding": "utf-8",
+          "string": ""
+        },
+        "headers": {
+          "Accept": [
+            "*/*"
+          ],
+          "Accept-Encoding": [
+            "gzip, deflate"
+          ],
+          "Connection": [
+            "keep-alive"
+          ],
+          "User-Agent": [
+            "python-requests/2.18.1"
+          ]
+        },
+        "method": "GET",
+        "uri": "https://reddit.local/api/v1/generate_refresh_token?username=01C3DNR1Z20EVR3AWNKMCY3N25"
+      },
+      "response": {
+        "body": {
+          "base64_string": "H4sIAAAAAAAAA6tWSkxOTi0uji/Jz07NU7JSUDI0NtItTAmNcPYyLwzwyzfz9UjJMQvOy4zKcUkPNQhV0lFQSq0oyCxKLY7PBGkwNjMwAIqB9ceXVBakggxJSk0sSi0CqS1OzocIaYF4RalpQI0ZqLaVuyS56Hr5uvqFFxZmpESlpeUYeDglhZsbJHmC9KSklmUmp8ZnpoAMhnCUagFXKKq/uAAAAA==",
+          "encoding": "UTF-8"
+        },
+        "headers": {
+          "Connection": [
+            "keep-alive"
+          ],
+          "Content-Encoding": [
+            "gzip"
+          ],
+          "Content-Type": [
+            "text/html; charset=UTF-8"
+          ],
+          "Date": [
+            "Tue, 09 Jan 2018 14:38:53 GMT"
+          ],
+          "Server": [
+            "PasteWSGIServer/0.5 Python/2.7.6"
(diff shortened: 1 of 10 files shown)


noisecapella:
If I go to a channel name with one letter I still see an error page. I'm not sure what reddit's doing there. Can you investigate that?

noisecapella:
It looks like the error page after navigating from a 404 post page to a channel page is fixed, but I still see it when I navigate from a 404 channel page to another channel page via sidebar

noisecapella:
One more case: if I go to a 404 post page, then navigate to a channel page, then navigate to another post which does exist, I see a 404 post page for that page

aliceriot:
with the one-letter thing, it seems (by doing some experimenting in the console) that Reddit returns a 400 from a one-letter subreddit, not sure why exactly

aliceriot:
I get a 404 when I try https://www.reddit.com/r/f, so maybe it's something praw is doing? idk

noisecapella:
@rhysyngsun do you know what the regex is that reddit uses to enforce channel names? How do we do it when the user creates a new channel?

noisecapella:
Maybe we can just change the regex in `urls.py` to require at least two characters for the channel names for the API. That will automatically return a 404 if the regex doesn't match

aliceriot:
I'm fine with that approach, it seems like reddit does not allow one-letter subreddit names (based on a quick check) so we should be ok doing that.

aliceriot:
Ok I think issues should be addressed. For the navigation error I just made the PostPage clean up after itself more (it checks if it has an error and, if so, clears it on `ComponentWillUnmount`). I also added a thingy to clear the post error on the channel page, which should help with navigating between channels.

For the 404 issue, I tried changing the regex in urls.py, but I ran into an issue. Basically if django doesn't match a url it returns a 404, but it doesn't return JSON, which the frontend expects. Because of this, I ended up adding a little thingy to the channel detail view, which just checks for a one-letter channel name and returns a 404 if it finds one (so we don't hit reddit and blow things up, haha).

aliceriot:
@noisecapella I think this should also address #393, so if you visit a channel that exists but you do not have permission for you should not have unhandled promise rejections

noisecapella:
I'm still seeing #393:
screenshot from 2018-01-09 15-51-36

noisecapella:
Everything else looks good

aliceriot:
Ok I think I addressed that, I just made it so that we only kick off the channel moderators request if the get channel request comes back without an error

PR #428, opened 2018-01-11
fixed spacing

roberthouse54:
(Required)

#### What's this PR do?
This fixes a spacing issue with the upvote downvote arrows being too close together.

See that it looks correct.


aliceriot:
only thing I'm seeing is that it now forces the 'report' button onto a second line when on a phone (this is chrome's iphone 7 dimension simulator):

phonefjfjf

vs master

phonefjfjff

not sure how much we care about that...

roberthouse54:
Let's not worry about that.

PR #439, opened 2018-01-19
add post sorting UI

### LINKED ISSUE #228: Sorting Posts Front End UI
We will have sorting for posts on the Discussions front page, and for each channel. Sorting will be by a dropdown. 

Sorting options 
- Most Active - this is what reddit calls "hot" - this is the default 
- New
- Top

image

API issue is #192

aliceriot:

closes #228 

#### What's this PR do?

This adds a UI for sorting posts for both the frontpage and the channel page. There will be a little drop-down in the upper-right corner which will let you choose from 'hot', 'top', and 'new'. Selecting from these will probably only change the output for you if you have a lot of posts and users, with some posts commented and upvoted a lot and others not.

Make sure that you can pick the sorting method. It should add a `sort` url query param when you pick it (it defaults to 'hot'). If you have pagination params in the url the `sort` param should be cleanly added to the existing params. This should all work on the frontpage and any channel page.

If the post listing isn't changing when you pick the different sorting methods it is probably enough to just confirm that the param gets sent to the backend in the post list request.

### DIFF
--- a/static/js/components/Card.js
+++ b/static/js/components/Card.js
@@ -5,7 +5,7 @@ import React from "react"
 type CardProps = {
   children: any,
   className?: string,
-  title?: string
+  title?: any
 }
 
 const Card = ({ children, className, title }: CardProps) =>
--- a/static/js/components/PostSortPicker.js
+++ b/static/js/components/PostSortPicker.js
@@ -1,0 +1,23 @@
+// @flow
+import React from "react"
+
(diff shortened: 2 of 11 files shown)


Ferdi:
Can we rename "hot"?   Maybe "active"?

aliceriot:
Sure, we can do that. @roberthouse54 what do you think? should we rename the other ones too?

PR #453, opened 2018-01-26
Added email and email_optin fields to user API (#447)

### LINKED ISSUE #447: Update create user API to take the user's email address
Having an e-mail address will allow the app to send notifications and channel digests to users.

rhysyngsun:
Resolves #447

#### What's this PR do?
Adds `email` and `email_optin` as optional fields to preserve backwards compatibility

- Verify you don't see any errors in MM profile sync after updating your profile
- Verify you can create a new user account via the API with and without the above values
- Verify you can update an existing user with/without the above values passed


profiles/serializers.py
```diff
@@ -39,17 +42,18 @@ def create(self, validated_data):
     def update(self, instance, validated_data):
         profile_data = validated_data.pop('profile', None)
 
+        instance.email = validated_data.pop('email', instance.email)
+        instance.save()
```
noisecapella: Might be better to do this along with `profile.save` below just in case there's an exception in setting one of the fields

rhysyngsun: Yeah, I'll wrap this in a transaction so it all either succeeds or fails.

rhysyngsun:

PR #454, opened 2018-01-30
add moderator view of posts and comments

### LINKED ISSUE #388: Moderator view of reported posts and comments
As a channel moderator, I need to see a list of all posts and comments that been reported by learners for violating the guidelines, so I can review and act on them.

aliceriot:

closes #388 

#### What's this PR do?

this adds a moderation view for looking at all the reported comments and posts for a subreddit, and it also adds some moderation UI. Here's a list:

- adds 'report count' indicators, which, if a post or comment has been reported, lets you know how many reports have been received for it
- 'reports' page (at `/moderation/channel/${channelName}`) which shows all of the reports for a channel
- functions for getting the reports for a channel and sticking that into the reducer in a sensible way
- fetch the reports on the channel and post-pages as well, so we can show the report count and moderation UI inline
- pulled post and comment moderation code out into HOCs so as not to duplicate code between some of these pages
- factories for report records

I would say just poke around and see if anything seems broken. Here are some general things:

- normal users shouldn't see a link to the moderation page from the channel page, moderators should
- on the moderation page itself normal users should just be redirected to the channel page
- moderators should be able to visit the moderation page and see a list of all the posts and comments that have been reported.
- moderators should see the report counts, and the buttons for moderation functions should do what it expected

I think that's about it? The UI should all work as before for normal users, and there should be no weird regressions with things.


static/js/containers/ChannelModerationPage.js
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  render() {
+    const { channel, reports, isModerator } = this.props
+
+    return isModerator
+      ? <div className="channel-moderation">
+        <DocumentTitle title={formatTitle(`${channel.title} moderation`)} />
+        <ChannelModerationBreadcrumbs channel={channel} />
+        <Card title="MODERATE!">
```
rhysyngsun: This could probably use a less imperative title ;)

aliceriot: Oops, I thought I had changed that! haha

aliceriot:
@rhysyngsun I addressed the redirect issue by changing a little bit how the data is fetched on the channel moderation page and also changing how the `loaded` prop is calculated, it should be working now.

I made a change to hide the 'report' button when in the moderation view as well, and also added the report count to the post detail page.

rhysyngsun:
Looks good

PR #457, opened 2018-02-05
add comment sorting UI

### LINKED ISSUE #229: Sorting Comments Front End UI
We will have sorting for comments on each post details page. Sorting will be by a dropdown. 

Sorting options 
- Best - this is the default
- New
- Old

image

API issue is #193

aliceriot:

closes #229 

#### What's this PR do?

this adds some UI to set the 'sort' param when you're fetching the comments for a post. Basically we add a little dropdown that has the sorting options, and we you pick one we re-issue the request for comments to the backend with the relevant param.

You should be able to load comments and posts normally, and you should be able to make comments, and all that should work as before.

The comment sort dropdown should have a default value of 'Best'. Selecting 'New' or 'Old' should re-fetch the comments, and the sort method should be passed to the backend as a URL query param.

I found that the 'Old' sort often changed the order from 'Best' and 'New', so switching between those two on a post with a lot of comments can be a way to confirm that the order is changing.


aliceriot:
hm, that's an unfortunate bug :/

Issue #444, opened 2018-01-22
Store site-specific data (e.g. redirect urls, site name, etc)


rhysyngsun:
Currently we have a certain amount of data either a) hardcoded in this app or b) stored in the shared cookie. This data doesn't differ user-to-user so we should instead store a key in the cookie and move this data into the DB under a `Site`/`Application` record that can be managed in django-admin. We also can't add anything else to the cookie because it's already too big, so this will in general clean things up. It also lines us up nicely to support multiple sites/applications down the road.

Values we need to port (examples not actual values):

- [x] Site name (e.g. "MicroMasters", currently hardcoded)
- [x] Site homepage (e.g. "micromasters.mit.edu", currently in settings.py)
- [x] Site auth/external login url (e.g. "micromasters.mit.edu/discussions", currently in the cookie and settings.py)
- [x] Site token session url (e.g. "micromasters.mit.edu/discussions_token", currently in the cookie)
- [x] TOS link (currently hardcoded)

#### Deployment

This work should probably be done in a few phases to avoid breaking things:

- [x] Add this model and django-admin UI (PR #444)
- [x] Migrate all environments
- [x] In the following envs create a record for MM and set `OPEN_DISCUSSIONS_DEFAULT_SITE_KEY`:
  - [x] CI
  - [x] RC
  - [x] Prod
- [x] Deploy update to open-discussions to take the value from the DB based on cookie key or default to MM (i.e. no longer use the token) (PR #492)
- [ ] Deploy update to MM to set only the key in new sessions (PR mitodl/micromasters#3834)

After this work is complete, we can remove all the extraneous hardcoded data and data stored in the cookie.

pdpinch:
The size of this cookie has continued to cause problems for me, especially on *.odl.mit.edu domains.

PR #484, opened 2018-02-08
Added Site models (#444)

rhysyngsun:
Part of #444

#### What's this PR do?
Adds a `Site` model and corresponding django-admin ui

Verify you can create a new site in django-admin


open_discussions/views.py
```diff
@@ -30,19 +31,25 @@ def index(request):
...
         "gaTrackingID": settings.GA_TRACKING_ID,
         "public_path": public_path(request),
         "auth_url": auth_url,
-        "session_url": session_url,
+        "session_url": session_url if site is None else site.session_url,
         "max_comment_depth": settings.OPEN_DISCUSSIONS_MAX_COMMENT_DEPTH,
-        "micromasters_external_login_url": settings.MICROMASTERS_EXTERNAL_LOGIN_URL,
-        "micromasters_base_url": settings.MICROMASTERS_BASE_URL,
+        "micromasters_external_login_url":
+            settings.MICROMASTERS_EXTERNAL_LOGIN_URL if site is None else site.login_url,
```
noisecapella: Would these settings essentially be deprecated after this PR?

noisecapella:
Have you looked at https://docs.djangoproject.com/en/2.0/ref/contrib/sites/?

open_discussions/views.py
rhysyngsun: Yeah, actually this slipped through the cracks. I meant to not update these as I don't want complicated code here. And the record for MM won't exist until we create it in django-admin so I want to phase in these changes as outlined in #444. I'll revert this bit of code.

rhysyngsun:
@noisecapella my understanding was django-sites was to support sites that were actually run in the django app itself. Although it brings up an import point, I should rename this model to something like `ExternalSite` to disambiguate.

rhysyngsun:
@noisecapella refactored now and removed the changes to the view. Once we have the model record created in production we can safely just update the view code to either use MM as a default or pull via a key in the JWT token.

pdpinch:
It's not a big deal, but I wonder if this is an issue of semantics. 

Some of the settings here will affect how Discussions functions, so we're customizing the Discussions site. 

But if `ExternalSite` makes more sense to everyone, I'm fine with it.

noisecapella:
Functionality looks great. Just one more thing, is it possible to update the admin site text to say 'MicroMasters login url' and 'MicroMasters base url' instead of just 'login url' and 'base url'?

rhysyngsun:
@noisecapella not sure that makes sense, would it fill in those values from `title`? That's an editable field here.

noisecapella:
It looks like `name` is used for the database field actually... maybe `verbose_name` would work though?

rhysyngsun:
@noisecapella all set

PR #488, opened 2018-02-09
fix non moderator post edit issue

### LINKED ISSUE #482: editing a post causes an error for non-moderator users
Same issue as #476

aliceriot:

closes #482 

#### What's this PR do?

Similar to the comment editing issue. We were posting the whole `Post` object, but for non-moderator users there are a few read-only fields on the object. So, when they tried to edit a post, they would get a 403 error. Unlike the comment issue, the edit wouldn't even go through (the comment edit would go through but then show an error page).

Anyway, this applies the same fix I did for the comment issue.

You should be able to edit posts successfully, without any network or UI errors, as both a moderator and non-moderator user.

### DIFF
--- a/static/js/components/CommentForms.js
+++ b/static/js/components/CommentForms.js
@@ -25,7 +25,7 @@ type CommentFormProps = {
   formDataLens: (s: string) => Object,
   processing: boolean,
   patchComment: (c: Object) => void,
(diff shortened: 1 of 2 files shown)


noisecapella:
Code looks good, just waiting for a free moment to do a functionality test

PR #496, opened 2018-02-14
Added report counts to frontpage

### LINKED ISSUE #495: Add report counts to the frontpage
This data is present in the APIs but conditional on the frontend on moderator status. The report coutn is null if you're not a moderator so we can safely show this UI.

rhysyngsun:
Fixes #495 

#### What's this PR do?
Adds the report counts to the frontpage and removes no longer needed calls to fetch the reports.

Verify that the counts still show up and you don't see report counts for channels you're not a moderator of.

### DIFF
--- a/static/js/components/CommentTree.js
+++ b/static/js/components/CommentTree.js
@@ -22,8 +22,7 @@ import {
 import type {
   GenericComment,
   CommentInTree,
-  MoreCommentsInTree,
-  CommentReportRecord
+  MoreCommentsInTree
 } from "../flow/discussionTypes"
 import type { FormsState } from "../flow/formTypes"
 import type { CommentRemoveFunc } from "./CommentRemovalForm"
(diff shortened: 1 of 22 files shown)


aliceriot:
works as advertised! reading code now

static/js/components/CommentTree.js
```diff
@@ -130,9 +125,9 @@ export default class CommentTree extends React.Component<*, *> {
               >
                 <a href="#">reply</a>
               </div>}
-            {isModerator && report !== undefined
+            {comment.num_reports
```
aliceriot: we may want to add a quick test just checking that `Reports: #` doesn't show up when `comment.num_reports` is null

static/js/components/CompactPostDisplay_test.js
```diff
@@ -182,11 +181,12 @@ describe("CompactPostDisplay", () => {
   })
 
   it("should render a report count, if passed a report and moderator", () => {
```
aliceriot: we might want to fix the test name here

static/js/containers/PostPage.js
```diff
@@ -183,10 +179,6 @@ class PostPage extends React.Component<*, void> {
         )
         moderators = response
       }
```
aliceriot: we're no longer doing anything with `moderators` here, so I think we can just change this to be

```js
if (!moderators) {
  await dispatch(actions.channelModerators.get(channelName))
}
```

rhysyngsun: Kinda odd that reassigning a `const` worked in the first place 🤔

Issue #464, opened 2018-02-06
Allow user to follow posts (but not comments)


pdpinch:
As a user, I'd like to be able to "follow" a post so that I will be notified (by email) whenever there is a new comment on it. 

### acceptance criteria
- [x] model for tracking post/comment subscriptions #465 
- [x] Mail: API for follow / unfollow #468 
- [x] modify comment method to send updates when there are subscribers #467 
- [x] email template for new comment notifications #470 
- [x] button for follow / unfollow (or subscribe / unsubscribe) #469 
- [x] automatically subscribe users to their own posts and comments #466 
- [x] model and API updates for setting automatic subscriptions on/off #475

rhysyngsun:
Closing this out as we're feature complete in prod.

Issue #450, opened 2018-01-24
Mail: periodic front-page digest emails


pdpinch:
As a learner, I'd like to (optionally) get email updates of recent changes to the Discussion's front page. 

### acceptance criteria
- [x] Learner receives regular e-mails with a list of top posts from their feed
  - [x] By default all members in OD get the daily e-mail, except for those who have already opted out of MM mail 
  - [x] #460 weekly subscribers (top)
  - [x] #461 daily subscribers (hot)
- [x] we should not send e-mail to a particular learner if there is not enough new posts since last mailing
- [x] #462 e-mail template should be responsive
- [x] E-mail template has a link to the settings for the user to unsubscribe or change the frequency 
  - [x] #451 UI
  - [x] #459 Model and API 
- [x] #534 Ensure email addresses have been backpopulated
- [x] #536 Add mail tasks to cron once everything is ready
- [x] #541 Ensure new users get NotificationSettings
- [x] #535 Ensure `NotificationSettings` have been backpopulated

pdpinch:
Blocked by #451 -- we don't want to start sending these emails without a way to opt out.

roberthouse54:
![image](https://user-images.githubusercontent.com/20047260/35813886-c0c0ad02-0a62-11e8-9201-e5797ddfee63.png)

rhysyngsun:
Everything is done for this

Issue #461, opened 2018-02-06
daily task for digest email


pdpinch:
### acceptance criteria
For daily subscribers, we will:
- [x] take the user's front-page feed
- [x] sort by hot posts
- [ ] filter out any posts more than 24 hours old
- [ ] take the first `n` posts, where n is set in a feature flag
- [ ] For our convenience, we can start this task at 10am every morning. 

### out of scope
- ideally, the mail would be sent an appropriate time for the user's time zone (i.e. local morning).

rhysyngsun:
For this, is it reasonable to only search through items on the first page of the user's frontpage?

pdpinch:
How big is a page? 25 posts? 50 posts?

I don't think it's likely we will get more than a page's worth of new posts each day, so limit the search to the first page seems OK. 

If there are more than a page's worth, or if hot posts are pushing new posts to the 2nd page, then I think those are the posts less worth notifying users about. 

> is it reasonable to only search through items on the first page of the user's frontpage?

rhysyngsun:
I think it's 25 right now.

Issue #460, opened 2018-02-06
weekly task for digest email


pdpinch:
### acceptance criteria
For weekly subscribers, we will:
- [ ] take the user's front-page feed
- [ ] sort by top posts
- [ ] filter out any posts more than a week old
- [ ] take the first `n` posts, where n is set in a feature flag

### out of scope
- ideally, the mail would be sent an appropriate time for the user's time zone (i.e. local morning)
- in the future, we might like to do some A/B testing about when this task is run. Currently, it looks like Tuesdays at 10am is recommended as the default

rhysyngsun:
Which day of the week do we want to send these?

pdpinch:
IIRC, someone found some research that said that Tuesday is the best day (at least, until we allow users to set their own preferred day)

PR #502, opened 2018-02-16
Proposed design for email notifications

rhysyngsun:
[Rendered](https://github.com/mitodl/open-discussions/blob/507961a85f944ce20e2d6153ed77d297e1625912/docs/design/email_notification.md)

Design for #450 and #464

#### What's this PR do?
Covers common design requirements of both of these tickets.

### DIFF
--- a/docs/design/email_notification.md
+++ b/docs/design/email_notification.md
@@ -1,0 +1,47 @@
+Email
+---
+
+#### Considerations
+
+- Be able to send frontpage email notifications on a daily or weekly basis
+  - Frequency is user-configurable
+  - If we don't have any new content since the last send, defer until next interval
+
+
+- Be able to notify users on specific events, controlled by a subscription to the entity
+  - Examples:
+    - Comment replies to posts
+    - Comment replies to other comments
+  - Capability to notify at the following frequencies:
+    - Near term we want to support:
+      - Never
+      - Immediate
+    - Long term we may want to support:
+      - Daily (digest/aggregation of new events)
+      - Weekly (digest/aggregation of new events)
+
+
+#### Proposal
+
+For frontpage digests the workflow would be:
+
+ - Daily and weekly tasks to:
+  - Walk `NotificationSettings` matching the corresponding `notification_type`
+  - For each of these settings, get frontpage data for that user
+  - Filter the list of frontpage posts to those that were made since the last frontpage email we sent this user
+    - A record needs to be stored at send time of when that happened
+  - If we have at least 1 post to send, render and send the emails
+  - Long a record of the sent email
+
+For event-specific notifications:
+
+ - When an event happens, trigger a django signal (so we can tie multiple handlers into it down the road)
+ - On that event, fire a celery task to check subscriptions
+ - Walk all subscriptions to that object and for each one create an `EmailNotification` record
+   - If their `NotificationSettings` record for that type has `frequency_type` of `never`, then we don't create this record. This has the effect of muting subscriptions temporarily with necessitating the removal of the subscriptions or effecting a massive catch-up of notifications if they turn it back on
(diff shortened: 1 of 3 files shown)


docs/design/email_notification.md
```diff
@@ -0,0 +1,47 @@
...
+#### Proposal
+
+For frontpage digests the workflow would be:
+
+ - Daily and weekly tasks to:
+  - Walk `NotificationSettings` matching the corresponding `notification_type`
+  - For each of these settings, get frontpage data for that user
+  - Filter the list of frontpage posts to those that were made since the last frontpage email we sent this user
+    - A record needs to be stored at send time of when that happened
+  - If we have at least 1 post to send, render and send the emails
```
aliceriot: If the user turns off the daily notification, doesn't use OD for three months, and then comes back and turns it back on, do they receive an email with all posts from the last three months? If not, do we just have an upper limit (the email always includes 10 posts max, for instance) or do we treat turning the notification back on as the timestamp we use for cutting off posts for the new email?

docs/design/email_notification.md
```diff
@@ -0,0 +1,47 @@
...
+  - Long a record of the sent email
+
+For event-specific notifications:
+
+ - When an event happens, trigger a django signal (so we can tie multiple handlers into it down the road)
+ - On that event, fire a celery task to check subscriptions
+ - Walk all subscriptions to that object and for each one create a `Notification` record
+   - If their `NotificationSettings` record for that type has `frequency_type` of `never`, then we don't create this record. This has the effect of muting subscriptions temporarily with necessitating the removal of the subscriptions or effecting a massive catch-up of notifications if they turn it back on
```
aliceriot: Just a thought, but we may want to create a `Notification` even if the user has `frequency_type == 'never'` so that we don't create difficulties for ourselves later-on if we want to start showing these things in the UI. `Notification` could have a `send_mail` or `muted` boolean that would skip sending mail, but keep the notification itself around so that we could build a REST api to expose a list of notifications, their 'read' status, etc, if we want to go that route later.

Alternatively, we could go with the simpler thing now and add this complication later, it doesn't seem like something that would be terrible hard to add.

docs/design/email_notification.md
rhysyngsun: The tickets for [daily](https://github.com/mitodl/open-discussions/issues/461) and [weekly](https://github.com/mitodl/open-discussions/issues/460) have a requirement for it to be the top `n` posts. So I would expect turning it back on to have it send the top n posts since the last email we sent the user (could be never, days, months, years, etc ago). But for the frontpage, those items are computed at send-time because it's a snapshot of the frontpage at that particular moment, it's not an aggregation of discrete events like comment notifications.

aliceriot: Cool, that makes sense to me then

docs/design/email_notification.md
rhysyngsun: That would definitely make sense for in-app notifications, but I'm trying to keep this design doc specific to the nuances of email notifications since in-app is out of scope for this initial implementation (although I'm trying hard not to design into a corner in that regard). We could always specify this to be an `EmailNotification` and then add a `InAppNotification` later and handle that differently.

That said, I think we ***really*** need some UI design around in-app notifications before proceeding with backend design though, because how I would design for in-app notifications hinges a lot on what the user experience is like, otherwise there's risk of designing something that's clunky or needs to be rewritten for our UX use cases.

aliceriot: Cool, I think that makes sense, just wanted to make sure we weren't going to make a decision now that boxes us in later

rhysyngsun:
Any objection to merging this?

aliceriot:
none here

Issue #463, opened 2018-02-06
link email to mail settings page (no login required)


pdpinch:

rhysyngsun:
The API support for this has been done in #506, but the actual work of adding the link to the emails still needs doing since we don't have those emails built yet.

rhysyngsun:
This will be supported in the UI in #511

PR #506, opened 2018-02-20
Added stateless token auth to notification settings api

rhysyngsun:
Part of #463 

#### What's this PR do?
Adds a utility to generate a stateless auth token and adds authentication using that to the notification settings API

- Run this to generate settings for your user: `./manage.py populate_notification_Settings --username USERNAME`
- Run this code in a python shell to get a token:

```python
from open_discussions.authentication import get_encoded_and_signed_subscription_token
user = User.objects.get(username='USERNAME')
get_encoded_and_signed_subscription_token(user)
```

Take the printed token value and run this (replacing `$TOKEN` with that value):

`curl -H "Authorization: Token $TOKEN" http://localhost:8063/api/v0/notification_settings/`


open_discussions/authentication.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Args:
+        user (User): user to generate the token for
+
+    Returns:
+        str: a signed and encoded token for subscription authentication_classes
+    """
+    signer = signing.TimestampSigner()
+    signed_value = signer.sign(user.username)
+    return b64encode(signed_value.encode('utf-8')).decode('ascii')
```
aliceriot: are we doing this `.encode` / `.decode` juggling because the username can have UTF-8 characters in it?

aliceriot: (that is to say, UTF-8 characters which are not also ASCII characters)

open_discussions/authentication.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # we base64 encode this so it is urlsafe
+    try:
+        signed_value = b64decode(token)
+    except:  # pylint: disable=bare-except
+        return None
+
+    signer = signing.TimestampSigner()
+    try:
+        print(signed_value)
```
aliceriot: stray print?

open_discussions/authentication.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Attempts to authenticate using a stateless token
+        """
+        if 'HTTP_AUTHORIZATION' in request.META:
+            header_value = request.META['HTTP_AUTHORIZATION']
+
+            if not header_value.startswith(HEADER_PREFIX):
+                return None
+
+            token = header_value[HEADER_PREFIX_LENGTH:]
```
aliceriot: this is just to slice off the `Token ` right?

open_discussions/authentication.py
```diff
@@ -0,0 +1,92 @@
...
+        signed_value = b64decode(token)
+    except:  # pylint: disable=bare-except
+        return None
+
+    signer = signing.TimestampSigner()
+    try:
+        print(signed_value)
+        return signer.unsign(
+            signed_value,
+            max_age=settings.OPEN_DISCUSSIONS_UNSUBSCRIBE_TOKEN_MAX_AGE_SECONDS
```
aliceriot: just a thought, but we may want to set this to either this `MAX_AGE` variable or the length of the subscription period (daily, weekly, etc) whichever is shorter, so that a daily subscriber would be getting a new token every day (and the one from yesterday would be invalidated). I guess that's a bit tricky though, because then we'd need to unsign the value, figure out the username, check what their email notification length is...yeah, lets not do that, nvm, haha.

open_discussions/authentication.py
```diff
@@ -0,0 +1,92 @@
...
+                return None
+
+            try:
+                user = User.objects.get(username=username)
+            except User.DoesNotExist:
+                return None
+
+            return (user, None)
+
+        return None
```
aliceriot: what will happen when I hit an API with an outdated token? will I get a 401 back?

open_discussions/authentication_test.py
```diff
@@ -0,0 +1,57 @@
...
+    token = get_encoded_and_signed_subscription_token(user)
+    assert decode_and_verify_username_from_token(token) == user.username
+
+
+def test_stateless_token_authentication_expired(rf):
+    """Tests that StatelessTokenAuthentication returns None if token is expired"""
+    token = 'MDFDNEZBV1dFOTc1VDFLRk4wQ0RZNllLUFo6MWVvRTlrOi05V2VacjYxWm9aR25wdTdVUlY2RGFQUXRRSQ=='
+    request = rf.get('api/v0/notification_settings', HTTP_AUTHORIZATION='Token {}'.format(token))
+    authentication = StatelessTokenAuthentication()
+    assert authentication.authenticate(request) is None
```
aliceriot: maybe add a similar, sanity-check type test that a nonsense token value (e.g. `mary had a little lamb`) also returns None?

open_discussions/settings.py
```diff
@@ -426,6 +426,11 @@
 
 OPEN_DISCUSSIONS_DEFAULT_SITE_KEY = get_string("OPEN_DISCUSSIONS_DEFAULT_SITE_KEY", None)
 
+OPEN_DISCUSSIONS_UNSUBSCRIBE_TOKEN_MAX_AGE_SECONDS = get_int(
+    'OPEN_DISCUSSIONS_UNSUBSCRIBE_TOKEN_MAX_AGE_SECONDS',
+    60 * 60 * 24 * 5  # 5 days
```
aliceriot: should this be 7 days, so the latest weekly email that the user received will always have a valid token?

open_discussions/authentication.py
rhysyngsun: In Python 3, `b64encode` requires a bytes-like object instead of a string and returns a bytes-like object so we have to then decode it. I don't believe django signing returns any values outside of the ascii range but utf-8 is certainly safer. I'll update the code so the encode/decode is symmetrical though.

open_discussions/authentication.py
rhysyngsun: Correct

open_discussions/authentication.py
rhysyngsun: Yes, at that point we would probably want to direct them to sign in.

open_discussions/settings.py
rhysyngsun: yeah, we can do that.

open_discussions/authentication_test.py
rhysyngsun: I had to update the exception handling in `decode_and_verify_username_from_token` to a bare except since a garbase token still base64 decodes but blows up with a unicode error during unsigning.

rhysyngsun:
@aliceriot feedback addressed

open_discussions/authentication.py
aliceriot: Cool, that should be easy to handle on the frontend

PR #511, opened 2018-03-01
add email settings page

### LINKED ISSUE #451: Mail: Manage Open Discussion Email Preferences
As a learner, I want an quick and easy way to unsubscribe from unwanted emails. Or edit my preferences as to how often I want to receive emails.

This shouldn't require logging in, so the URL embedded in the email needs to include a enctryped token containing enough information for unsubscribing, i.e. the email address and a reference ot the list for unsubscribing.

Since the settings component will be accessible with and without authentication, it may require two different templates and two different endpoints.

aliceriot:

closes #451 

#### What's this PR do?

This adds a notifications page at `/settings` where the user can adjust their notification preferences.

If the user is logged in, it works as any other page in the app. However, it also supports authenticating to the API with a stateless token, which we can use to authenticate only with the notification settings API.

Basically, if the user visits `/settings/${TOKEN]/` the application will include the `TOKEN` value in the headers for the HTTP requests it issues, so that the user can fetch their current notification preference values and also issue a `PATCH` request. If the `TOKEN` supplied is not a valid token (i.e. if we get a 401 back) it will redirect the user to the 'sign in' page.

First you'll need to create a `NotificationSettings` for yourself for the `frontpage` notification type. There are instructions for doing that on Nathan's PR, here: https://github.com/mitodl/open-discussions/pull/497

Normal auth:

- go to `/settings` and confirm you see the current value of the frontpage notification setting selected in the Radio
- confirm you can use the form to change the value

Token auth:

- follow the instructions here to generate a token for yourself: https://github.com/mitodl/open-discussions/pull/506
- in an incognito window (or a different browser where you are not signed in) go to `/settings/${token}/`, where `${token}` is the base64 encoded token you just generated
- you should not see any errors when fetching the current value for the notification setting or editing and saving it.
- if you do the same with a nonsense token (like `asdfasdf`) you should be redirected to `/auth_required/` after the first request.

I think that's it...

### DIFF
--- a/static/js/reducers/settings.js
+++ b/static/js/reducers/settings.js
@@ -1,0 +1,26 @@
+// @flow
+import * as api from "../lib/api"
+import { GET, PATCH } from "redux-hammock/constants"
+
+import type { NotificationSetting } from "../flow/settingsTypes"
+
+export const FREQUENCY_IMMEDIATE = "immediate"
+export const FREQUENCY_DAILY = "daily"
+export const FREQUENCY_WEEKLY = "weekly"
+export const FREQUENCY_NEVER = "never"
+
+export const FRONTPAGE_NOTIFICATION = "frontpage"
+
+export const FRONTPAGE_FREQUENCY_CHOICES = [
+  { value: FREQUENCY_NEVER, label: "Never" },
+  { value: FREQUENCY_DAILY, label: "Daily" },
+  { value: FREQUENCY_WEEKLY, label: "Weekly" }
+]
+
+export const settingsEndpoint = {
+  name:      "settings",
+  verbs:     [GET, PATCH],
+  getFunc:   (token: ?string) => api.getSettings(token),
+  patchFunc: (object: NotificationSetting, token: ?string) =>
+    api.patchFrontpageSetting(object, token)
+}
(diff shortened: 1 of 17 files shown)


static/js/reducers/settings.js
```diff
@@ -0,0 +1,27 @@
...
+export const FREQUENCY_IMMEDIATE = "immediate"
+export const FREQUENCY_DAILY = "daily"
+export const FREQUENCY_WEEKLY = "weekly"
+export const FREQUENCY_NEVER = "never"
+
+export const FRONTPAGE_NOTIFICATION = "frontpage"
+
+export const FREQUENCY_CHOICES = [
+  { value: FREQUENCY_NEVER, label: "Never" },
+  { value: FREQUENCY_IMMEDIATE, label: "Immediate" },
```
rhysyngsun: We don't want to let the user select the "Immediate" setting for frontpage digest emails.

aliceriot: are we going to want it for other notification types? if so I'll exclude it for just the current UI, if not I'll remove the option entirely

rhysyngsun: Yeah, and in the case of comment notifications we will only want to allow selection of Never / Immediate in the near-term so that makes sense.

aliceriot:
@rhysyngsun ok! I just added a similar setup to what we do for the comment submission button

aliceriot:
@rhysyngsun I added a little snackbar thing just to confirm that the user's selection has been saved

PR #516, opened 2018-03-05
Track user activity

### LINKED ISSUE #473: Track user activity
We should add a way of tracking the user's activity. We don't want to do this on every API request so the best way might be whenever we see a token with a newer `iat` (https://en.wikipedia.org/wiki/JSON_Web_Token#Standard_fields) value, we update with that value. This value could then be roughly interpreted as "last login". We should probably store this on the `Profile`.

rhysyngsun:
Fixes #473 

Adds user tracking middleware and refactors and upgrades existing middleware to use `MIDDLEWARES` style

Navigate around open discussions. You should see the JWT cookie updated on the first page/API request made to the server (subject to thundering herd, but shouldn't cause errors). You set see no further attempts to `Set-Cookie` unless you get a new JWT token.


open_discussions/middleware/user_activity.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        payload = self._track_activity(request)
+
+        response = self.get_response(request)
+
+        if payload is not None:
+            response.set_cookie(
+                settings.OPEN_DISCUSSIONS_COOKIE_NAME,
+                self.jwt_encode_handler(payload),
+                domain=settings.OPEN_DISCUSSIONS_COOKIE_DOMAIN,
```
mbertrand: Seems to work with or without this, but with it there is one cookie instead of two.

PR #520, opened 2018-03-07
Release 0.17.2

odlbot:
## Alice Pote
  - [x] Add settings page for adjusting notification prefs ([eda64502](../commit/eda64502d948d328c06f74e08db60e1335e966bd))

## Nathan Levesque
  - [x] Added frontpage digest email tasks (#460, #461) ([ce13d718](../commit/ce13d718b8e5cc331b8e6745986fab6ef09fdebc))


aliceriot:
@rhysyngsun for my commit the user will need to have a `NotificationSetting` for the `frontpage` in order to use it - do we want to get that figured out before we merge this code?

rhysyngsun:
@aliceriot does the page error if that's not present? I wasn't planning on populating those values for the majority of users for another few days. Creating one just for yourself shouldn't be an issue though.

aliceriot:
yeah it errors if there isn't one present

aliceriot:
we could add a feature flag redirect or something for now

rhysyngsun:
@aliceriot we're not linking to it anywhere though, right? so we could at least add that feature flag on another release tomorrow

aliceriot:
yeah I suppose so, the user would have to guess that the settings page exists, and for now they would just see a blank page basically

rhysyngsun:
Logged https://github.com/mitodl/open-discussions/issues/521 to deal with emails not sending in RC, should not affect stability of this RC.

PR #525, opened 2018-03-09
Added cancelation and better error handling to email sending

### LINKED ISSUE #521: Mark emails unable to be sent as skipped
Some emails may not be able to be sent because the user is marked as having emails turned off (`email_optin is not True`). We should mark the notification accordingly.

rhysyngsun:
Fixes #521 

#### What's this PR do?
(Required)

(Required)

(Optional)

(Optional)

(Optional)

(Optional)


notifications/utils.py
```diff
@@ -1,35 +1,48 @@
...
+            notification.save()
+        except CancelNotificationError:
+            log.debug("EmailNotification canceled: %s", notification.id)
+            notification.state = EmailNotification.STATE_CANCELED
+            notification.save()
+        except:  # pylint: disable=bare-except
+            # if any other error happens, roll back to pending so that the crontask picks it up again
+            log.exception("EmailNotification rolled back to pending: %s", notification.id)
+            notification.state = EmailNotification.STATE_PENDING
+            notification.save()
```
mbertrand: should the exception be raised here to alert Sentry?

notifications/notifiers/email.py
```diff
@@ -54,22 +55,30 @@ def send_notification(self, email_notification):
...
-                log.exception("Error sending email")
+            data = self._get_notification_data(notification_settings, last_notification)
+
+            # generate the message (there's only 1)
+            messages = list(api.messages_for_recipients([
+                (recipient, user, data) for recipient, user in api.safe_format_recipients([user])
+            ], self._template_name))
+
+            if not messages:
+                raise CancelNotificationError()
```
mbertrand: Maybe add a unit test to verify `CancelNotificationError` is raised in this case.

notifications/api.py
```diff
@@ -98,5 +101,8 @@ def send_email_notification_batch(notification_ids):
         notification_ids (list of int): notification ids to send
     """
     for notification in EmailNotification.objects.filter(id__in=notification_ids):
-        notifier = _get_notifier_for_notification(notification)
-        notifier.send_notification(notification)
+        try:
+            notifier = _get_notifier_for_notification(notification)
+            notifier.send_notification(notification)
+        except:  # pylint: disable=bare-except
+            log.exception('Error sending notification')
```
mbertrand: Should any exceptions that might occur here trigger a Sentry alert?  If so maybe these exceptions could be tracked over the iterations of the loop, and when the loop is complete raise an exception if errors do occur.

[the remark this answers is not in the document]
rhysyngsun: `log.exception` will do that.

notifications/utils.py
rhysyngsun: See my other comment above

notifications/api.py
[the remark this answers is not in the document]
mbertrand: That's what I always assumed, but after talking with Sar this morning about an [OVS issue](https://github.com/mitodl/odl-video-service/issues/462), it seems like that doesn't actually happen?  

>No, there isn’t a sentry agent or anything like that running on the instances, so unless the app itself raises a sentry error, it just gets logged. In some instances, we have elastalert go through the logs (the majority are sent to an elasitcsearch logging cluster) and send some messages to slack

rhysyngsun:
@mbertrand all set!

Issue #535, opened 2018-03-15
Ensure NotificationSettings have been backpopulated


pdpinch:
- [x] Populate beta test group
  - [x] Populate active users
  - [x] Merge #541 
  - [x] Populate everyone else
  - [ ] add settings link to user drop-down

PR #543, opened 2018-03-20
Ensure new users get the default NotificationSettings

### LINKED ISSUE #541: Ensure new users get NotificationSettings

rhysyngsun:
Fixes #541 

#### What's this PR do?
Adds a call to `ensure_notification_settings` when a user is created so they get the defaults.

Sign up for a new user in MM and navigate to OD, then visit `/api/v0/notification_settings/` and verify you get two entries back with the defaults.


rhysyngsun:
Merge blocked until #535 is ready for rollout to all users

Issue #573, opened 2018-03-28
Delete refresh tokens on 401 error from reddit


noisecapella:
This would fix occasional authentication failures we get from reddit. If we receive the 401 after deleting the refresh token we should just let the exception bubble up

PR #579, opened 2018-03-30
Fix 401 auth errors

noisecapella:
None

#### What's this PR do?
Fixes a recurring 401 error @roberthouse54 has been getting. This deletes the relevant tokens and tries one more time to get a client.

Not sure how to reproduce

### DIFF
--- a/channels/api_test.py
+++ b/channels/api_test.py
@@ -7,6 +7,7 @@ import pytest
 from praw.models import Comment
 from praw.models.comment_forest import CommentForest
 from praw.models.reddit.redditor import Redditor
+from prawcore.exceptions import ResponseException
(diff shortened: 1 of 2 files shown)


channels/api_test.py
```diff
@@ -576,6 +577,73 @@ def test_api_constructor(mocker, settings, verify_ssl):
     assert config.refresh_token == refresh_token
 
 
+def test_api_constructor_error(mocker):
+    """
+    If a non-401 response is received during Api() initialization, we should delete the refresh
```
aliceriot: I think this comment is supposed to read `we should not delete`

noisecapella: haha oops

noisecapella:
Fixes #573

PR #582, opened 2018-04-03
fix missing profile picture in email

### LINKED ISSUE #576: Mail: anonymous profile image is missing for user's without a profile image
img_5587

aliceriot:

closes #576 

#### What's this PR do?

This just does a little bit of work to make the default profile picture URL included in the email into an absolute URL, so that it will actually work when the html is rendered in an email client. Previously we were just shipping a relative URL, which obviously didn't work.

OK! So the following code will render the frontpage email template with your user's current frontpage:

```py
from mail.api import render_email_templates
from channels.utils import ListingParams, get_pagination_and_posts
from channels.api import Api
from channels.serializers import PostSerializer
from profiles.models import Profile

p = Profile.objects.get(name='asdf Pote')

user = p.user

api = Api(user)

lp = ListingParams(None, None, 0, 'hot')

pagination, posts = get_pagination_and_posts(api.front_page(lp), lp)

serialized_posts = PostSerializer(posts,context={'current_user':user}, many=True).data

subject, fallback, html = render_email_templates('frontpage', user, context={'posts': serialized_posts})

print(html)
```

you can either paste that into a django shell or you can save it to a file and use the `%run filename.py` magic in iPython, to print out the rendered HTML. Assuming that works ok, test two things:

1. if your user has no `profile.image_small` you should see a full, absolute URL in the HTML which points to something like `http://localhost:8063/static/images/avatar_default.png`.

2. If your user does have a `profile.image_small`, that URL should be included without being messed with in any way.

I think that's it...

### DIFF
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -1,6 +1,7 @@
 """
 Serializers for channel REST APIs
 """
+from urllib.parse import urljoin
 from datetime import (
     datetime,
     timezone,
@@ -18,6 +19,7 @@ from channels.api import (
 )
 from channels.constants import VALID_CHANNEL_TYPES
 from channels.models import Subscription
+from open_discussions import settings
 
 default_profile_image = "/static/images/avatar_default.png"
 
@@ -171,11 +173,11 @@ class PostSerializer(serializers.Serializer):
         return "[deleted]"
 
     def get_profile_image(self, instance):
-        """Find the Profile for the comment author"""
+        """Find the profile image for the post author"""
         user = self._get_user(instance)
         if user and user.profile.image_small:
             return user.profile.image_small
-        return default_profile_image
+        return urljoin(settings.SITE_BASE_URL, default_profile_image)
 
     def get_text(self, instance):
         """Returns text or null depending on if it's a self post"""
--- a/channels/views/posts_test.py
+++ b/channels/views/posts_test.py
@@ -1,5 +1,6 @@
 """Tests for views for REST APIs for posts"""
 # pylint: disable=unused-argument
+from urllib.parse import urljoin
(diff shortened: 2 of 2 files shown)


mail/templatetags/timeago.py
```diff
@@ -17,3 +21,15 @@ def parse_iso(value):
...
+
+@register.filter
+def absolutize_profile_image_url(url):
+    """
+    Takes a URL, and if it is the default URL makes
+    it an absolute URL
+    """
+
+    if url == default_profile_image:
+        return urljoin(SITE_BASE_URL, url)
```
rhysyngsun: Suggestion: could we instead move this `urljoin` into the `get_profile_image` method of the serializer?

rhysyngsun:
Functionality looks good.

noisecapella:
cc @roberthouse54 This will fix the missing avatar problem on your local instance

mail/templatetags/timeago.py
aliceriot: sure

channels/serializers.py
```diff
@@ -171,11 +173,11 @@ def get_author_name(self, instance):
         return "[deleted]"
 
     def get_profile_image(self, instance):
-        """Find the Profile for the comment author"""
+        """Find the profile image for the post author"""
         user = self._get_user(instance)
         if user and user.profile.image_small:
             return user.profile.image_small
-        return default_profile_image
+        return urljoin(SITE_BASE_URL, default_profile_image)
```
rhysyngsun: Could you actually make this `settings.SITE_BASE_URL`? It avoids any potential issues with the value at import-time.

PR #597, opened 2018-04-09
add comment / post settings

### LINKED ISSUE #474: Mail: Settings UI for automatic follow subscriptions
as a user, I would like to be have the option to disable comment notifications on my posts and comments

aliceriot:

closes #474 

#### What's this PR do?

This adds a setting to the settings page (`/settings`) for managing post / comment replies (basically, it's a checkbox which toggles the value of the comments notification setting between 'never' and 'immediate').

First you'll need to make sure you have a `NotificationSetting` with type `comment` created for your user. Then if you go to the settings page you should see  a checkbox. Checking or unchecking the checkbox should save the value between page loads, and a checked checkbox should correlated with the frequency setting for the `NotificationSetting` being 'immediate' and an unchecked box with 'never'.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "redux-hammock": "^0.2.3",
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.2.0",
+    "rmwc": "^1.5.1",
     "sanctuary": "^0.13.2",
     "sass-lint": "^1.10.2",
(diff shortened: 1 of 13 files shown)


aliceriot:
@rhysyngsun have we back-filled comment notification settings objects on prod yet? I don't think we can merge this until we do

PR #614, opened 2018-04-11
add an error page for 403 responses

### LINKED ISSUE #583: Error message shown instead of forbidden
screenshot from 2018-04-04 10-58-18

See also the 403 responses in the response list below

aliceriot:

closes #583 

#### What's this PR do?

This adds a 403 error page, which we show on the channel, post, and channel moderation pages whenever the user gets a 403 (not-authorized) error back from the API. This can happen when, for instance, the user tried to navigate directly to a channel which they are not part of.

I also moved the 404-page handling logic into the `<Loading />` component, so it lives in one place, instead of in several.

Check that:

- visiting a non-existing channel, post, and channel moderation page shows you a 404 screen
- visiting a channel, post, and channel moderation page for which your user does not have permission (i.e. the channel and post exist but your user isn't in the channel) should show you a 'not authorized' error

Other than that everything should be as it was before.


rhysyngsun:
Functionality is good, looking at code.

Issue #604, opened 2018-04-10
render link URLs in Post view using embed.ly


pdpinch:
As a user making a post, I'd like the URL in my Link post to be parsed and rendered in it's embedded format. I.e., when I post a link to YouTube, I'd like the post to be rendered as a video and when I post a link from Twitter, I'd like the post to be rendered as an embedded Tweet. 

[embed.ly](https://embed.ly) provides a service that parses and renders URLs from multiple social sharing sites.

PR #624, opened 2018-04-18
add embed.ly proxy API

aliceriot:

for #604 

#### What's this PR do?

This adds an API which proxies requests to the [embed.ly embed API](http://docs.embed.ly/docs/oembed).

You should be able to test it out by going to `/api/v0/embedly/${url}`. `url` has to be an encoded URI component. I found that I had to call `encodeURIComponent` twice in order to get it to properly escape all the slashes in a URL.

For now I have defined a function in `lib/api.js` which should allow you to test it, if you open up the discussions app you can find it at `window.getEmbedly`, you can inspect the response in the Chromium network tab. Alternatively, you can try a URL like this one:

http://localhost:8063/api/v0/embedly/https%253A%252F%252Fen.wikipedia.org%252Fwiki%252FArnaldur_Indri%2525C3%2525B0ason/

### DIFF
--- a/embedly/api.py
+++ b/embedly/api.py
@@ -1,0 +1,12 @@
+"""Embedly proxy API"""
+import requests
+
+from django.conf import settings
+
+
+def get_embedly(url):
+    """issue a request to embed.ly's embed API"""
+    return requests.get(settings.EMBEDLY_EMBED_URL, params={
+        'key': settings.EMBEDLY_KEY,
+        'url': url
+    })
--- a/embedly/views.py
+++ b/embedly/views.py
@@ -1,0 +1,20 @@
+"""Embed.ly views"""
+from urllib.parse import unquote
+from rest_framework import status
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+
+from django.conf import settings
+
+from embedly.api import get_embedly
+
+
+@api_view()
+def embedly_view(request, **kwargs):  # pylint: disable=unused-argument
+    """get Embedly API, return the JSON"""
+    if settings.EMBEDLY_KEY:
+        url = unquote(unquote(kwargs["url"]))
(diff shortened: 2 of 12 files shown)


embedly/views.py
```diff
@@ -0,0 +1,23 @@
...
+
+class EmbedlyViews(APIView):
+    """
+    View for getting embedly API
+    """
+
+    def get(self, request, *args, **kwargs):  # pylint: disable=unused-argument
+        """get Embedly API, return the JSON"""
+        if settings.EMBEDLY_KEY:
+            response = get_embedly(unquote(unquote(self.kwargs["url"])))
```
rhysyngsun: Why does this have to be unquoted twice?

aliceriot: it actually doesn't have to be, things seem to work fine if it is only unquoted once. But in order to pass a full URL as a URL parameter I had to call `encodeURIComponent` twice on the frontend, calling `unquote` twice just makes the URL we pass to the `get_embedly` function look like a normal URL again, if that makes sense. I may be missing some easier way to deal with passing a URL as a URL param, haha.

aliceriot:
@rhysyngsun this is now ready for review, I removed the caching code. For testing purposes, there's a `getEmbedly` function in `lib/api.js`, if you assign that to `window.getEmbedly` you can test it out from the JS console in discussions

embedly/api.py
```diff
@@ -0,0 +1,14 @@
+"""Embedly proxy API"""
+import requests
+
+from django.conf import settings
+
+EMBEDLY_URL = "https://api.embed.ly/1/oembed"
```
rhysyngsun: I would suggest putting this into `settings.py` with this as the default so in case it changes on us it doesn't require a code change.

rhysyngsun:
Functionality looks good

PR #626, opened 2018-04-19
Allow anonymous access for channels

### LINKED ISSUE #600: allow anonymous access to channels API

noisecapella:
Fixes #600 

#### What's this PR do?
Adds a feature flag for anonymous access and support in `Api`. Updates permission classes to allow anonymous readonly access to channels.

 - Set `FEATURE_ANONYMOUS_ACCESS=True` in your `.env`
 - In a private session go to these endpoints and verify that you don't get an error: `/api/v0/channels/`, `/api/v0/channels/<some_public_channel_name>/`. `/api/v0/channels/` should have an empty list.


channels/api.py
```diff
@@ -209,11 +211,16 @@ def _get_client(user):
...
 
     Args:
-        user (User): the authenticated user
+        user (User): the authenticated user, or None for anonymous user
 
     Returns:
         praw.Reddit: configured reddit client
     """
+    if user.is_anonymous:
+        return praw.Reddit(
```
rhysyngsun: One thing we'll need to keep an eye out for here is that it doesn't take advantage of the cached tokens so we should keep an eye on reddit load/latency when this release goes out. I don't see a huge need to address this until it becomes a problem though.

noisecapella: :+1:

Issue #487, opened 2018-02-09
Add tests for channels.util.translate_praw_exceptions


rhysyngsun:
This method is getting complicated enough we should probably test it outside of view tests.

noisecapella:
Fixed in #629

PR #629, opened 2018-04-19
Update frontend to allow anonymous access

### LINKED ISSUE #603: don't force redirect to /auth_required on index.html

noisecapella:
WIP

Fixes #603

#### What's this PR do?
Changes behavior to not redirect to the login required page for anonymous users. A 401 response from a REST API will still cause a redirect.

TBD


noisecapella:
@rhysyngsun I just rebased on the other merged PRs

pdpinch:
@rhysyngsun that sounds right to me. 

- Private channel + anonymous user returns a 401, so the front-end has a chance to suggest the user sign in. 
- Private channel + logged-in but not subscribed user returns a 403. I'm not 100% sure what the error message tells the user to do, but it can be the same as the 401. Logging in won't help.

noisecapella:
@rhysyngsun Made the change

rhysyngsun:
BTW, you can close out #487 with this too.

PR #635, opened 2018-04-23
Fixed race condition with NotificationSettings trigger_frequency

### LINKED ISSUE #609: Race condition with settings change to 'never' and pending email
https://sentry.io/mit-office-of-digital-learning/discussions/issues/521711100/

```
InvalidTriggerFrequencyError: Trigger frequency 'never' is invalid for frontpage
  File "notifications/utils.py", line 34, in mark_as_sent_or_canceled
    yield True
  File "notifications/notifiers/email.py", line 71, in send_notification
    data = self._get_notification_data(notification_settings, last_notification)
  File "notifications/notifiers/frontpage.py", line 133, in _get_notification_data
    posts = _posts_since_notification(notification_settings, last_notification)
  File "notifications/notifiers/frontpage.py", line 78, in _posts_since_notification
    params = _get_listing_params(notification_settings.trigger_frequency)
  File "notifications/notifiers/frontpage.py", line 43, in _get_listing_params
    "Trigger frequency '{}' is invalid for frontpage".format(trigger_frequency)

EmailNotification rolled back to pending: 21398
```

rhysyngsun:
Fixes #609 

#### What's this PR do?
Fixes a race condition where we queue an email to be sent but the user changes their settings prior to the actual send

This is a bit tricky to test, but the fix here is very simple so I think passing tests and code review should be sufficient.

### DIFF
--- a/notifications/notifiers/email.py
+++ b/notifications/notifiers/email.py
@@ -64,8 +64,13 @@ class EmailNotifier(BaseNotifier):
         user = email_notification.user
 
         with utils.mark_as_sent_or_canceled(email_notification) as will_send:
+            # check against programmer error
             if user != self.user:
                 raise Exception("Notification user doesn't match settings user")
+
+            # if we're trying to send an email, but the preference is never, we should just cancel it
+            if self.notification_settings.is_triggered_never:
+                raise CancelNotificationError()
 
             if not will_send:
                 return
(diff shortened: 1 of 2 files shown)


notifications/notifiers/email.py
```diff
@@ -70,6 +70,10 @@ def send_notification(self, email_notification):
             if not will_send:
                 return
 
+            # if we're trying to send an email, but the preference is never, we should just cancel it
+            if self.notification_settings.is_triggered_never:
+                raise CancelNotificationError()
```
gsidebo: Seems like the exceptional cases here (this and `user != self.user`) should be grouped together. For that matter, it looks like both of them could be moved above this context processor, which I think would be preferable as it gets the exception/error cases out of the way before the 'meat' of the function

PR #644, opened 2018-04-25
Added handling for praw errors in email notifications

### LINKED ISSUE #639: Cancel email notification for users that have been removed from the channel
https://sentry.io/mit-office-of-digital-learning/discussions/issues/535037643/events/18698378440/

rhysyngsun:
Fixes #639 

#### What's this PR do?
This PR adds a decorator/contextmanager that is used to translate praw errors to notification cancellation errors instead. These errors can happen when the user no longer has permissions to certain objects in Reddit, so sending a notification for them isn't useful.

Most of the translation logic was inspired by `channels.utils.translate_praw_exceptions()`, so this covers a bit more than just the error reported by Sentry so we don't have to revisit this for edge cases we've already hit in other areas.

Only added this to comment notifications because we haven't seen it for the frontpage yet and that situation would be more complicated anyway.

- Stop your celery worker: `docker-compose stop celery`
- Ensure your notification settings for comments are set to "immediate"
- Subscribe to a post (or create a new one) in a private channel
- Comment on the post
- Remove your user from the channel the post is in via the shell: `channels.api.Api.remove_contributor('USERNAME')`
- Start the celery worker: `docker-compose start celery`
- After about a minute it should try to send the email and it should cancel the notification instead of raising an error. You can verify this with:
```python
from discussions.models import EmailNotification
assert EmailNotification.objects.order_by('id').last().state == EmailNotification.STATE_CANCELED
```

This came up when I searched giphy for "cancel" so why not?

Cancel

### DIFF
--- a/notifications/utils.py
+++ b/notifications/utils.py
@@ -3,6 +3,12 @@ from contextlib import contextmanager
 import logging
 
 from django.db import transaction
+from praw.exceptions import APIException, PRAWException
+from prawcore.exceptions import (
+    Forbidden,
+    NotFound,
+    Redirect,
+)
 
 from notifications.notifiers.exceptions import CancelNotificationError
 from notifications.models import EmailNotification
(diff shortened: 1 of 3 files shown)


notifications/utils.py
```diff
@@ -46,3 +52,21 @@ def mark_as_sent_or_canceled(notification):
...
+def praw_error_to_cancelled():
+    """Raises CancelNotificationErrors from certain praw errors, otherwise raises the original error"""
+    try:
+        yield
+    except (Forbidden, NotFound, Redirect) as exc:
+        raise CancelNotificationError() from exc
+    except APIException as exc:
+        if exc.error_type in ('SUBREDDIT_NOTALLOWED', 'SUBREDDIT_NOEXIST', 'DELETED_COMMENT'):
+            raise CancelNotificationError() from exc
+        raise
```
noisecapella: Can you add a test for this case, where the exception doesn't match any of the other cases?

rhysyngsun:
@noisecapella all set

PR #651, opened 2018-04-27
Add management command to index comments and posts

### LINKED ISSUE #641: Create (and re-create) index using existing posts and comments

noisecapella:
Fixes #641 

#### What's this PR do?
Adds a management command to iterate over posts and comments that the user is subscribed to. 

First, delete the elasticsearch docker container and recreate it.

Add a new environment variable `INDEXING_API_USERNAME` with a username which is subscribed to the channels you wish to index (you may want to use `mitodl` which is the default user used to index from micromasters).

Run `./manage.py index_content`. You should see a celery task id being output, then the management command should wait. Run `sudo reddit-restart` in the reddit shell to temporarily interrupt reddit, to assert the autoretry functionality. (Make sure to run `docker-compose build` beforehand, the latest version of celery is required.)

Before it is done, go to `localhost:9101/_alias` and verify that there is an alias ending with `_reindexing` and none ending with `_default`. After the reindex completes go to that URL again. You should not see any alias ending with `_reindexing` but there should be one ending with `_default`.

Go to `localhost:9101/_search`. You should see many documents.


search/indexing_api_test.py
```diff
@@ -73,3 +85,78 @@ def test_increment_document_integer_field(mocked_es):
...
+    post_id = 'post_id'
+    index_post(user.username, post_id)
+
+    get_alias_mock.assert_called_once_with()
+    api_mock.assert_called_once_with(user)
+    client = api_mock.return_value
+    client.get_post.assert_called_once_with(post_id)
+    post = client.get_post.return_value
+
+    assert serialize_mock.call_count == len(aliases)
```
gsidebo: Probably not necessary. The loop at line 155 effectively checks this

search/tasks.py
```diff
@@ -35,3 +74,127 @@ def update_document_with_partial(doc_id, partial_data):
...
+def start_recreate_index(self, api_username):
+    """
+    Wipe and recreate index and mapping, and index all items.
+
+    Args:
+        api_username (str): The reddit username to use when crawling data
+    """
+    from channels.api import Api
+
+    user = User.objects.get(username=api_username)
```
gsidebo: We were planning on having a single admin user for crawling & indexing, right? Would it make more sense to define a setting for the crawler/indexer username and use that in all of these tasks rather than accepting a parameter?

search/tasks.py
```diff
@@ -35,3 +74,127 @@ def update_document_with_partial(doc_id, partial_data):
...
+    raise self.replace(
+        chain(
+            index_channels,
+            finish_recreate_index.si(new_backing_index),
+        )
+    )
+
+
+@app.task
+def finish_recreate_index(backing_index):
```
gsidebo: Our other tasks are basically pass-throughs that turn logic in `indexing_api.py` into asynchronous tasks. I think it might be useful to do the same here. At least in development I imagine we could get some use from having this functionality available outside of celery tasks. There's also the issue of having essential index manipulation logic in 2 different places. If there's a compelling reason to leave this logic inside of tasks, or it's just too annoying, let me know

search/tasks_test.py
```diff
@@ -40,3 +56,180 @@ def test_increment_document_integer_field_task(mocked_api):
     increment_document_integer_field(*indexing_api_args)
     assert mocked_api.increment_document_integer_field.call_count == 1
     assert mocked_api.increment_document_integer_field.call_args[0] == indexing_api_args
+
+
+def test_wrap_retry_exception():
+    """wrap_retry_exception should raise RetryException when other exceptions are raised"""
+    with wrap_retry_exception(KeyError):
+        # Should not raise an exception
+        pass
```
gsidebo: Can you explain what this is doing? Is that docstring accurate?

search/tasks_test.py
```diff
@@ -40,3 +56,180 @@ def test_increment_document_integer_field_task(mocked_api):
...
+    api_username = 'username'
+    post_id = 'post_id'
+    index_post.delay(api_username, post_id)
+
+    index_post_mock.assert_called_once_with(api_username, post_id)
+    wrap_mock.assert_called_once_with(PrawcoreException, PRAWException)
+
+
+def test_index_channel(mocker):
+    """index_channel should all posts of a channel"""
```
gsidebo: should *index* all posts of a channel

search/tasks_test.py
```diff
@@ -40,3 +56,180 @@ def test_increment_document_integer_field_task(mocked_api):
...
+    user = UserFactory.create()
```
gsidebo: Can you use the `fixtures.users.user` fixture here?

search/tasks_test.py
```diff
@@ -40,3 +56,180 @@ def test_increment_document_integer_field_task(mocked_api):
...
+    index_post_mock = mocker.patch('search.tasks.index_post', autospec=True)
+    api_mock = mocker.patch('channels.api.Api', autospec=True)
+
+    expected_exception = ZeroDivisionError
```
gsidebo: Can you explain the need for this?

search/tasks_test.py
```diff
@@ -40,3 +56,180 @@ def test_increment_document_integer_field_task(mocked_api):
...
+    channel_name = 'channel'
+    with pytest.raises(expected_exception):
+        index_channel.delay(user.username, channel_name)
+
+    api_mock.assert_called_once_with(user)
+    get_channel_mock.assert_called_once_with(channel_name)
+
+    wrap_mock.assert_called_once_with(PrawcoreException, PRAWException)
+    assert group_mock.call_count == 1
+    list(group_mock.call_args[0][0])  # iterate through generator
```
gsidebo: Confused about this as well. Needs a more descriptive comment at the very least

search/tasks_test.py
```diff
@@ -40,3 +56,180 @@ def test_increment_document_integer_field_task(mocked_api):
...
+
+    replace_mock = mocker.patch('celery.app.task.Task.replace', return_value=expected_exception)
+    group_mock = mocker.patch('search.tasks.group', autospec=True)
+    get_channel_mock = api_mock.return_value.get_channel
+    post1 = mocker.Mock(id=1)
+    post2 = mocker.Mock(id=2)
+    get_channel_mock.return_value.new.return_value = [
+        post1,
+        post2,
+    ]
```
gsidebo: Nit: can shorten

```python
    posts = [mocker.Mock(id=i) for i in (1,2)]
    get_channel_mock.return_value.new.return_value = posts
```

search/tasks.py
```diff
@@ -35,3 +74,127 @@ def update_document_with_partial(doc_id, partial_data):
...
+    Args:
+        api_username (str): API username
+        channel_name (str): The name of the channel
+    """
+    with wrap_retry_exception(PrawcoreException, PRAWException):
+        from channels.api import Api
+
+        client = Api(User.objects.get(username=api_username))
+        channel = client.get_channel(channel_name)
+        post_ids = [post.id for post in channel.new()]
```
gsidebo: Seems like there should be an added channel API method for this. `get_channel_posts` or something like that

search/tasks_test.py
```diff
@@ -40,3 +56,180 @@ def test_increment_document_integer_field_task(mocked_api):
...
+    chain_mock.assert_called_once_with(
+        group_mock.return_value,
+        finish_recreate_index_mock.si.return_value,
+    )
+    assert replace_mock.call_count == 1
+    assert replace_mock.call_args[0][1] == chain_mock.return_value
+
+
+@pytest.mark.parametrize("default_exists", [True, False])
+def test_finish_recreate_index(mocker, default_exists):
```
gsidebo: This test case is really long, which makes the forked logic even tougher to sort out. You could potentially parametrize like this to get rid of at least some of the ifs:

```python
@pytest.mark.parametrize(
    "default_exists,backing_index_name,expected_alias_remove,expected_index_delete", [
        [True, 'old_index', True, True],
        [False, 'old_index', False, False],
    ]
)
```

search/tasks.py
noisecapella: Makes sense to me

search/tasks_test.py
noisecapella: That's a little confusing, it's meant to show that when an exception is not raised inside the context, it's not raised outside the context either. I'll see if I can modify to explain further

search/tasks_test.py
noisecapella: The Celery tasks are chained together, and to ensure that the parent task waits for all the other child tasks you can `raise self.replace(new_task_here)`. That will tell Celery to stop running code from the parent task and start running code from the child task(s), and it will also keep the same task id, which is important for other code waiting on it.

The Celery docs recommend `raise self.replace(...)` over just `self.replace(...)` because it tells the reader that the code ends there: http://docs.celeryproject.org/en/master/reference/celery.app.task.html#celery.app.task.Task.replace

Should I add a comment explaining that? I can add one at least for `ZeroDivisionError` to say that the exception doesn't matter.

search/tasks_test.py
noisecapella: The `group(...)` which is called is actually operating on a generator, not a list. To assert the functions being called inside the generator we need to force iteration so that the calls are made. Do you think `list(...)` works for that here or should I do something else?

search/tasks.py
noisecapella: There is actually, I used `list_posts`

search/tasks_test.py
noisecapella: I'm not sure that parameterization would help. I moved the logic to test_switch_indices, can you take a second look?

noisecapella:
Ok I think that's all comments

search/tasks_test.py
```diff
@@ -40,3 +54,121 @@ def test_increment_document_integer_field_task(mocked_api):
...
+    matching_exception = RetryException if matching else TabError
+    with pytest.raises(matching_exception):
+        with wrap_retry_exception(KeyError):
+            raise_thing()
+
+
+def test_index_post_with_comments(mocker):
+    """index_post should call the api function of the same name"""
+    index_post_mock = mocker.patch('search.indexing_api.index_post_with_comments')
+    wrap_mock = mocker.patch('search.tasks.wrap_retry_exception')
```
gsidebo: This only shows up in 2 test cases (this one, `test_index_channel`), but I think it's worth it to get rid of some boilerplate by turning this patch and our assertions into a fixture:

```python
@pytest.fixture()
def wrap_retry_mock():
    """Patches the wrap_retry_exception context manager and asserts that it was called by any test that uses it"""
    wrap_mock = mocker.patch('search.tasks.wrap_retry_exception')
    yield
    wrap_mock.assert_called_once_with(PrawcoreException, PRAWException)

def test_index_post_with_comments(mocker, wrap_retry_mock):
    # ...
```

search/tasks_test.py
```diff
@@ -40,3 +54,121 @@ def test_increment_document_integer_field_task(mocked_api):
...
+    wrap_mock.assert_called_once_with(PrawcoreException, PRAWException)
+
+
+def test_index_channel(mocker, settings, user):
+    """index_channel should index all posts of a channel"""
+    settings.INDEXING_API_USERNAME = user.username
+    index_post_mock = mocker.patch('search.tasks.index_post_with_comments', autospec=True)
+    api_mock = mocker.patch('channels.api.Api', autospec=True)
+
+    expected_exception = ZeroDivisionError
```
gsidebo: I read your earlier comment. In this case I think it's best to pull this patching logic out into a separate fixture to make the actual tests clearer. The fixture can patch `celery.app.task.Task.replace`, `search.tasks.group`,`search.tasks.chain`, and any other celery library functions that we might need to patch in the future. The return value for that fixture can also include this exception so the `pytest.raises` assertion would look something like this: `with pytest.raises(celery_mock.expected_exception):`. Anything we can do to isolate this finicky library patching and clarify the actual intent of the tests would be really helpful here IMO.

search/tasks_test.py
```diff
@@ -40,3 +54,121 @@ def test_increment_document_integer_field_task(mocked_api):
...
+    )
+    finish_recreate_index_mock = mocker.patch('search.tasks.finish_recreate_index', autospec=True)
+
+    with pytest.raises(TabError):
+        start_recreate_index.delay()
+
+    create_backing_index_mock.assert_called_once_with()
+    finish_recreate_index_mock.si.assert_called_once_with(backing_index)
+    assert group_mock.call_count == 1
+    list(group_mock.call_args[0][0])  # iterate through generator
```
gsidebo: Suggested comment: 

```
Celery's 'group' function takes a generator as an argument. In order to make assertions about the items
in that generator, 'list' is being called to force iteration through all of those items.
```

search/tasks_test.py
noisecapella: Sounds good

search/indexing_api.py
```diff
@@ -122,3 +143,98 @@ def increment_document_integer_field(doc_id, field_name, incr_amount):
...
+    """
+    from channels.api import Api
+
+    conn = get_conn()
+    api_username = settings.INDEXING_API_USERNAME
+    client = Api(User.objects.get(username=api_username))
+    post = client.get_post(post_id)
+    comments = post.comments
+    # Make sure all morecomments are replaced before serializing
+    comments.replace_more(limit=None)
```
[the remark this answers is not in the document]
noisecapella: I'm going to leave this since I'm not sure how to do that cleanly, if you have a suggestion let me know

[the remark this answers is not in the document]
gsidebo: I was thinking something like: 

```python
def get_all_comments(self, post_id):
  post = self.get_post(post_id)
  comments = post.comments
  comments.replace_more(limit=None)
  return comments
```

Don't know if that does what we would need it to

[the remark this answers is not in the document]
noisecapella: That function would need to return the post so that we could serialize it later in the function. I'm not sure how intuitive it would be to return a post from that function when `list_comments` returns a `CommentTree`

search/indexing_api.py
```diff
@@ -77,12 +96,13 @@ def create_document(doc_id, data):
...
-        body=data,
-        id=doc_id,
-    )
+    for alias in get_active_aliases():
+        return conn.create(
+            index=alias,
+            doc_type=GLOBAL_DOC_TYPE,
+            body=data,
+            id=doc_id,
+        )
```
gsidebo: This returns immediately for the first item. Any items after that will be ignored

search/indexing_api.py
```diff
@@ -94,12 +114,13 @@ def _update_document(doc_id, data, update_key=None):
         data (dict): Full ES document data
     """
     conn = get_conn(verify=True)
-    return conn.update(
-        index=settings.ELASTICSEARCH_INDEX,
-        doc_type=GLOBAL_DOC_TYPE,
-        body={update_key: data},
-        id=doc_id,
-    )
+    for alias in get_active_aliases():
```
gsidebo: Same deal as above

search/indexing_api.py
noisecapella: Yeah that needs to be fixed... do we need the value returned from this function? If not I'm just going to remove that part

noisecapella:
Ok, it's updated now

PR #653, opened 2018-05-01
add embedly frontend

### LINKED ISSUE #604: render link URLs in Post view using embed.ly

aliceriot:

closes #604 

#### What's this PR do?

When viewing a Link-type post this will make a request to the Embedly API and then display the results. Here's a few examples of what that looks like:

nytimes_article
nytimes_article_mob
chihuahuapic
chihuahuapic_mob
youtube_vidv
youtube_vidvmob

You'll need an embedly API key set up to test this, and you'll need to set the `EMBEDLY_KEY` env var with it. Then, if you add a link post, you should see some sort of content get pulled out, as shown in the screenshots.

There are two small other tweaks which should be tested. From the post index (a channel or the frontpage) Link-type posts now link directly to the post detail view, not to the external resource. Also, the domain name of the external link is now put below the content, instead of up by the title, on the detail page.

### DIFF
--- a/static/js/components/Embedly.js
+++ b/static/js/components/Embedly.js
@@ -1,0 +1,55 @@
+// @flow
+import React from "react"
+
+export default class Embedly extends React.Component<*> {
+  renderEmbed() {
+    const { embedly } = this.props
+
+    // the response includes HTML which can be used to load a rich embed
+    // (either a video or the 'rich' type)
+    if (embedly.html) {
+      return (
+        <div
+          className="embed-container"
+          dangerouslySetInnerHTML={{ __html: embedly.html }}
+        />
+      )
+    }
+
+    if (embedly.type === "photo") {
+      return (
+        <div className="photo">
+          <img src={embedly.url} />
+        </div>
+      )
+    }
+
+    return (
+      <a href={embedly.url} target="_blank" className="link">
+        {embedly.thumbnail_url
+          ? <div className="thumbnail">
+            <img src={embedly.thumbnail_url} />
+          </div>
+          : null}
+        <div className="link-summary">
+          <h2>
+            {embedly.title}
+          </h2>
+          <div className="description">
+            {embedly.description}
+          </div>
+        </div>
+      </a>
+    )
+  }
+
+  render() {
+    const { embedly } = this.props
+
+    return (
+      <div className="embedly">
+        {embedly && embedly.type !== "error" ? this.renderEmbed() : null}
+      </div>
+    )
+  }
+}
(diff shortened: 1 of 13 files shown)


pdpinch:
Thanks Alice. I had a couple questions about the UX. 

> From the post index (a channel or the frontpage) Link-type posts now link directly to the post detail view, not to the external resource. 

Is it possible to handle "regular" links differently than embed.ly links? Linking directly to external resource was convenient. 

> Also, the domain name of the external link is now put below the content, instead of up by the title, on the detail page.

Was there a reason to move this?

roberthouse54:
The reason I suggested moving it is that the title of the post is not actually associated with the domain of the external link, and I thought it was a bit misleading.

static/js/components/Embedly.js
```diff
@@ -0,0 +1,55 @@
...
+
+    if (embedly.type === "photo") {
+      return (
+        <div className="photo">
+          <img src={embedly.url} />
+        </div>
+      )
+    }
+
+    return (
```
rhysyngsun: We should handle urls that Embedly considers invalid (e.g. ones it can't resolve). Any easy test case for this is a url to a local OD post.

static/js/containers/App.js
```diff
@@ -55,6 +55,9 @@ class App extends React.Component<*, void> {
       return
     }
 
+    window.actions = actions
```
rhysyngsun: Is this test code?

pdpinch:
Can you revert the "two small tweaks?" They're out of scope and we don't have agreement that they're desirable.

aliceriot:
Sure

aliceriot:
to be clear, you're referring to these changes, right?

- moving the domain name
- linking to the post detail page instead of the external link from the post index

static/js/containers/App.js
aliceriot: :woman_facepalming: yes, haha

pdpinch:
Yes, thanks:

static/js/components/Embedly.js
aliceriot: I added something so that if we get an error response back from embedly we just don't show anything. Curious if @roberthouse54 you have any ideas of what we should do in this case (that a user adds a post with a link but Embedly, for whatever reason, doesn't return any embed information for the link)

aliceriot:
@rhysyngsun this is ready for review again. It looks like the build is failing on the pip install for the python side of things, not sure what's up with that :woman_shrugging:

PR #656, opened 2018-05-02
add zendesk widget

### LINKED ISSUE #549: Add ZenDesk "Help" widget
Same as MicroMasters, in theory. 

If we can, we should pre-populate the user's name and email address. We can leave the "MicroMasters Program" drop down empty.

aliceriot:

closes #549 

#### What's this PR do?

This adds the Zendesk support widget to open discussions. I copied the implementation we have in MM over, and removed the parts which it didn't seem to me like we would need over here. I also added a little thing to hide the MM program selector, and also added the user's email address to the SETTINGS object in order to fully pre-fill the widget form.

Make sure that the widget shows up. It should look roughly the same as in MM. Your user's name and email should be pre-filled in the form, and the form shouldn't have a (visible) drop-down for selecting the program. Also, there shouldn't be any zendesk-related console errors.

### DIFF
--- a/static/js/entry/zendesk.js
+++ b/static/js/entry/zendesk.js
@@ -1,0 +1,184 @@
+/* global SETTINGS:false zE:false _:false */
+__webpack_public_path__ = `${SETTINGS.public_path}` // eslint-disable-line no-undef, camelcase
+import _ from "lodash"
+
+// Start of odl Zendesk Widget script
+/* eslint-disable no-sequences, prefer-const */
+/*<![CDATA[*/
+window.zEmbed ||
+  (function(e, t) {
+    let n,
+      o,
+      d,
+      i,
+      s,
+      a = [],
+      r = document.createElement("iframe")
+    ;(window.zEmbed = function() {
+      a.push(arguments)
+    }), (window.zE = window.zE || window.zEmbed), (r.src =
+      "javascript:false"), (r.title = ""), (r.role =
+      "presentation"), ((r.frameElement || r).style.cssText =
+      "display: none"), (d = document.getElementsByTagName("script")), (d =
+      d[d.length - 1]), d.parentNode.insertBefore(r, d), (i =
+      r.contentWindow), (s = i.document)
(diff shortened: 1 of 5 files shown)


static/js/entry/zendesk.js
```diff
@@ -0,0 +1,172 @@
...
+  const intervalID = setInterval(() => {
+    tries += 1
+    const iframe = document.querySelector(`iframe.zEWidget-${name}`)
+    try {
+      iframeDocument = iframe.contentDocument
+    } catch (err) {
+      // cross-domain exception: can't continue
+    }
+    if (!iframeDocument) {
+      console.error(`Can't access content of Zendesk iframe: ${name}`) // eslint-disable-line no-console
```
rhysyngsun: Did we want to remove this because we concluded it wasn't useful?

aliceriot: The problem before was that we were using this function (`zendeskPollForLoaded`) to check for the existence of a few iframes which the Zendesk embed doesn't make any longer (they were called `ticketSubmissionForm`, `nps`, and `imp`). I changed it so we're just checking for the ones it actually creates.

rhysyngsun: Also that this block is causing `webWidgetLoaded` to sometimes not be called.

static/js/entry/zendesk.js
```diff
@@ -0,0 +1,172 @@
...
+    const iframe = document.querySelector("iframe.zEWidget-webWidget")
+
+    // this is Zendesk's ID for the program selector field
+    // which we want to hide in Discussions
+    const programFieldName = "24690866"
+    const programEl = iframe.contentDocument.querySelector(
+      `input[name="${programFieldName}"]`
+    )
+
+    if (programEl) {
```
rhysyngsun: Hitting the case where this is `null` and we don't hide the field.

aliceriot:
@rhysyngsun ok I think that issue should be address now

Issue #606, opened 2018-04-10
When a comment is made, add it to the search index


pdpinch:

pdpinch:
we'll need to think through the questions in #605 first

pdpinch:
@gsidebo you're working on this right? Please assign it to yourself and move to in progress.

gsidebo:
Setting this to 'blocked' for the time being as we look into https://github.com/mitodl/open-discussions/issues/720. We will likely close this issue in favor of that one

PR #667, opened 2018-05-03
Added Elasticsearch document and added indexing handlers for posts

### LINKED ISSUE #605: When a post is made, add it to the search index
This will require some planning around:
- [ ] permissions (results only include channels learner has access to)
- [ ] rendering (in particular dealing with upvotes and similar post metadata)
- [ ] sorting
- [ ] how to use comments text to assist in searching for posts
- [ ] tags

gsidebo:
Closes #605

#### What's this PR do?
- Adds the elasticsearch document mapping, as well as a function that clears and creates the index in ES
- Wraps various channel API methods to make changes to the index when there are changes made to the information that we index for posts (and only posts)

- Enable the `INDEX_UPDATES` feature (i.e.: add `FEATURE_INDEX_UPDATES=true` to your `.env`)
- Run `search.indexing_api.clear_and_create_index` in a Django shell
- Create posts and alter posts in different ways (changing text, adding comments, upvoting/downvoting)
- Query the index and confirm that those changes are reflected in the indexed document (or that the document was created if it's a new post)

An easy curl command to inspect the documents in the index:
```bash
curl -XGET -s 'http://localhost:9101/discussions_local/_search?pretty=true&filter_path=hits.hits._id,hits.hits._source&sort=created:desc'
```

Probably `channels/api.py`

- This PR does not add a search UI or add any querying capability. That can and should be done in a separate issue.
- This PR does not index changes to comments. That can also be done in a separate PR.
- `search/connection.py` was copied from Micromasters. If we want to make any changes to that connection logic we can, but worth pointing out that it's copied

### DIFF
--- a/channels/api.py
+++ b/channels/api.py
@@ -1,8 +1,10 @@
 """Channels APIs"""
 # pylint: disable=too-many-public-methods, too-many-lines
+import logging
 from datetime import datetime, timedelta
 from urllib.parse import urljoin
 
+from functools import partial
 import pytz
 import requests
 from django.conf import settings
@@ -35,6 +37,9 @@ from channels.constants import (
     VALID_CHANNEL_TYPES,
     VALID_POST_SORT_TYPES,
     VALID_COMMENT_SORT_TYPES,
+    POST_TYPE,
+    COMMENT_TYPE,
+    VoteActions,
 )
 from channels.models import (
     RedditAccessToken,
@@ -44,6 +49,15 @@ from channels.models import (
 
 from open_discussions import features
 from open_discussions.utils import now_in_utc
+from search.task_helpers import (
+    reddit_object_indexer,
+    index_full_post,
+    update_post_text,
+    update_post_removal_status,
+    increment_post_comment_count,
+    decrement_post_comment_count,
+    update_indexed_score,
+)
 
 USER_AGENT = 'MIT-Open: {version}'
 ACCESS_TOKEN_HEADER_NAME = 'X-Access-Token'
@@ -66,6 +80,8 @@ User = get_user_model()
 
 # monkey patch praw's rate limiter to not limit us
 prawcore.rate_limit.RateLimiter.delay = lambda *args: None
+
+log = logging.getLogger()
 
 
 def _get_refresh_token(username):
@@ -273,6 +289,58 @@ def replace_load_comment(original_load_comment):
     return replacement_load_comment
 
 
+def _apply_vote(instance, validated_data, allow_downvote=False, instance_type=None):
+    """
+    Apply a vote provided by the user to a comment or post, if it's different than before.
+
+    Args:
+        instance (Comment or Post): A comment or post
+        validated_data (dict): validated data which contains the new vote from the user
+        allow_downvote (bool): If false, ignore downvotes
+        instance_type (str): A string indicating the reddit object type (comment, post/submission)
+    Returns:
+        bool:
+            True if a change was made, False otherwise
+    """
+    upvote = validated_data.get('upvoted')
+    if allow_downvote:
+        downvote = validated_data.get('downvoted')
+    else:
+        downvote = None
+
+    is_upvoted = instance.likes is True
+    is_downvoted = instance.likes is False
+
+    # Determine vote action to take based on the request and any already-applied votes
+    if upvote and not is_upvoted:
+        vote_action = VoteActions.UPVOTE
+    elif downvote and not is_downvoted:
+        vote_action = VoteActions.DOWNVOTE
+    elif upvote is False and is_upvoted:
+        vote_action = VoteActions.CLEAR_UPVOTE
+    elif downvote is False and is_downvoted:
+        vote_action = VoteActions.CLEAR_DOWNVOTE
+    else:
+        return False
+
+    if vote_action == VoteActions.UPVOTE:
+        instance.upvote()
+    elif vote_action == VoteActions.UPVOTE:
+        instance.downvote()
+    elif vote_action in (VoteActions.CLEAR_UPVOTE, VoteActions.CLEAR_DOWNVOTE):
+        instance.clear_vote()
+
+    try:
+        update_indexed_score(instance, instance_type, vote_action)
+    except Exception:  # pylint: disable=broad-except
+        log.exception('Error occurred while trying to index [%s] object score', instance_type)
+    return True
+
+
+apply_post_vote = partial(_apply_vote, allow_downvote=False, instance_type=POST_TYPE)
+apply_comment_vote = partial(_apply_vote, allow_downvote=True, instance_type=COMMENT_TYPE)
+
+
 class Api:
     """Channel API"""
     def __init__(self, user):
@@ -371,6 +439,7 @@ class Api:
         self.get_channel(name).mod.update(**values)
         return self.get_channel(name)
 
+    @reddit_object_indexer(indexing_func=index_full_post)
     def create_post(self, channel_name, title, text=None, url=None):
         """
         Create a new post in a channel
@@ -469,6 +538,7 @@ class Api:
         """
         return self.reddit.submission(id=post_id)
 
+    @reddit_object_indexer(indexing_func=update_post_text)
     def update_post(self, post_id, text):
         """
         Updates the post
@@ -501,6 +571,7 @@ class Api:
         post = self.get_post(post_id)
         post.mod.sticky(pinned)
 
+    @reddit_object_indexer(indexing_func=update_post_removal_status)
     def remove_post(self, post_id):
         """
         Removes the post, opposite of approve_post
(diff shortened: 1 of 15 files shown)


channels/api.py
```diff
@@ -273,6 +284,46 @@ def replacement_load_comment(*args, **kwargs):
     return replacement_load_comment
 
 
+def _apply_vote(
+        instance, validated_data, allow_downvote=False, instance_type=None
+):  # pylint: disable=unused-argument
```
rhysyngsun: Don't need the lint disable

search/indexing_api.py
```diff
@@ -0,0 +1,95 @@
...
+                        'filter': [
+                            'lowercase',
+                            'asciifolding',  # remove accents if we use folding analyzer
+                        ]
+                    }
+                }
+            }
+        },
+        'mappings': {
+            **({} if skip_mapping else mapping),
```
rhysyngsun: Would suggest either simplifying this to not use `**` or otherwise use `copy.deepcopy` if a seperate copy is what we need here.

search/task_helpers.py
```diff
@@ -0,0 +1,70 @@
...
+    partially_update_document,
+)
+
+
+def reddit_object_indexer(indexing_func=None):
+    def api_indexing_listener_inner(func):
+        @wraps(func)
+        def wrapped_api_func(*args, **kwargs):
+            reddit_obj = func(*args, **kwargs)
+            indexing_func(reddit_obj)
```
rhysyngsun: We may want to `try...except` + `log.exception` this on any errors so an indexing error doesn't result in a 500 error back to the end user.

gsidebo:
@rhysyngsun Noted. I wonder if it's possible/not-too-hacky to trigger a full object reindexing based on the actual object in the data store if some indexing-related celery tasks fail.

noisecapella:
Should we have a backing index like we do for micromasters so we can do a reindex while the other old index is still in use?

gsidebo:
@noisecapella We definitely should, yes. Do you think that should be part of this issue?

noisecapella:
I can take care of it with my issue since mine concerns the tree traversal

pdpinch:
@gsidebo Does this close #606 as well (adding comments to the search index)?

gsidebo:
@pdpinch I left comments out of this to keep the diff a bit more manageable. That issue will be pretty quick (a day or less most likely) once this is merged

gsidebo:
@rhysyngsun @noisecapella Ready for review again

channels/constants.py
```diff
@@ -27,3 +29,14 @@
     COMMENTS_SORT_NEW,
     COMMENTS_SORT_OLD,
 )
+
+POST_TYPE = 'post'
+COMMENT_TYPE = 'comment'
+
+VoteActions = namedtuple('VoteActions', ['UPVOTE', 'DOWNVOTE', 'CLEAR_UPVOTE', 'CLEAR_DOWNVOTE'])
```
rhysyngsun: Any reason not to use an enum for this since they're become nicer to work with in recent Python 3.x and we don't appear to care about the values and they don't cross celery broundaries or anything like that? This would probably be a little simpler:

```python
from enum import Enum, auto
class VoteActions(Enum):
    UPVOTE = auto()
    DOWNVOTE = auto()
    CLEAR_UPVOTE = auto()
    CLEAR_DOWNVOTE = auto()
```

channels/serializers.py
```diff
@@ -269,7 +238,7 @@ def create(self, validated_data):
 
         api.add_post_subscription(post.id)
 
-        changed = _apply_vote(post, validated_data, False)
+        changed = apply_post_vote(post, validated_data)
```
rhysyngsun: I think there's a possible race condition here under multiple workers where the index update for the upvote here gets processed _before_ the full post task index is processed on the `create_post` call above. I would expect since you're doing a partial ES update that call has potential to fail where the document to update doesn't exist. If it's not practical to address that right now, we may want to catch those errors so we don't get a bunch of noise in production.

I have some thoughts on how to handle that race condition, but I'm not sure if that's something we want to chew off right now.

fixtures/common.py
```diff
@@ -38,3 +46,45 @@ def warnings_as_errors():
...
+@pytest.fixture(scope='session', autouse=True)
+def session_indexing_decorator():
+    """
+    Fixture that mocks the reddit object indexer for the test suite by default.
+
+    Decorators require some extra work to patch since they are already applied to functions
+    when a module is loaded. To get around this, we patch the decorator function then
+    reload the relevant modules via importlib.
+    """
+    mock_func = Mock()
```
rhysyngsun: Suggest using `mocker` fixture and call `mocker.Mock()` and `mocker.patch()` instead since those handle `start()`/`stop() shenanigans automatically for you.

search/tasks.py
```diff
@@ -0,0 +1,21 @@
...
+from open_discussions.celery import app
+
+
+@app.task
+def create_document(doc_id, data):
+    """Task that makes a request to create an ES document"""
+    api.create_document(doc_id, data)
+
+
+@app.task
```
rhysyngsun: To address my earlier concern about race conditions between this task and the `create_document` one, I think we should have this task backoff and retry (up to some limit) if a partial update fails because the document doesn't exist yet. That's probably the simplest / most elegant solution.

Same for `increment_document_integer_field` below.

search/task_helpers.py
```diff
@@ -0,0 +1,139 @@
...
+    try:
+        if instance_type != POST_TYPE:
+            return
+        return increment_document_integer_field.delay(
+            gen_post_id(instance.id),
+            field_name='score',
+            incr_amount=vote_increment
+        )
+    except Exception:  # pylint: disable=broad-except
+        log.exception('Error occurred while trying to index [%s] object score', instance_type)
```
rhysyngsun: I think because this task is delayed, the only error's we'd see here are going to be celery queuing errors, correct?

search/task_helpers.py
```diff
@@ -0,0 +1,139 @@
...
+        field_name='num_comments',
+        incr_amount=incr_amount
+    )
+
+
+increment_post_comment_count = partial(_update_post_comment_count, incr_amount=1)
+decrement_post_comment_count = partial(_update_post_comment_count, incr_amount=-1)
+
+
+def update_indexed_score(instance, instance_type, vote_action=None):
```
rhysyngsun: `vote_action` always seems required to me, so I'm not certain it should have a default value

search/task_helpers.py
```diff
@@ -0,0 +1,139 @@
...
+        instance: A PRAW Comment or Submission (a.k.a. post) object
+        instance_type (str): A string indicating the type of object being passed in
+        vote_action (str): A string indicating what kind of vote action was taken
+    """
+    if vote_action == VOTE_ACTIONS.UPVOTE or vote_action == VOTE_ACTIONS.CLEAR_DOWNVOTE:
+        vote_increment = 1
+    elif vote_action == VOTE_ACTIONS.DOWNVOTE or vote_action == VOTE_ACTIONS.CLEAR_UPVOTE:
+        vote_increment = -1
+    else:
+        return
```
rhysyngsun: Should this raise an exception instead of quietly returning? Passing in an unsupported value here should be programmer error and we'd want to catch that.

channels/constants.py
gsidebo: I completely forgot about Enum! Will do

fixtures/common.py
gsidebo: I believe `mocker` is function-scoped, and therefore not usable in this session-scoped fixture. I'll confirm or deny that

rhysyngsun: Ah I think you're right. That being so, carry on.

channels/serializers.py
gsidebo: Yup, that would be possible. I can create an issue to look into this

search/tasks.py
gsidebo: Yeah, any partial update request should be safely retry-able. Celery task retry setting seems like a sensible safeguard against the possibility of partial update tasks running before the document is created

gsidebo:
@rhysyngsun Should be ready to review again

search/tasks.py
```diff
@@ -0,0 +1,37 @@
...
+def create_document(doc_id, data):
+    """Task that makes a request to create an ES document"""
+    api.create_document(doc_id, data)
+
+
+@app.task(**PARTIAL_UPDATE_TASK_SETTINGS)
+def update_document_with_partial(doc_id, partial_data):
```
rhysyngsun: I think we need to feature flag these partial ones so that we don't get a bunch of errors prior to backpopulating.

gsidebo:
@rhysyngsun Feature flag added & comments addressed

gsidebo:
@noisecapella Fixed. Great catch 👍

PR #668, opened 2018-05-04
Don't HTML escape subject lines for frontpage emails

### LINKED ISSUE #611: Don't HTML escape e-mail subjects
A post title of "Courses' level of difficulty" shouldn't be escaped in the email notification to "Courses&#39; level of difficulty"

@rhysyngsun says:
> that's probably django templates escaping that
> just need to add a `|safe` to the end of the variable in the subject template

rhysyngsun:
Fixes #611 

#### What's this PR do?
Ensures the frontpage digest subject lines aren't html escaped

This is pretty simple so the test coverage should be sufficient (the `'` would be escaped to `&apos;` in master branch).


aliceriot:
is it straightforward to do the same thing for comment notification emails too? I was able to create this monstrosity:
title_escaping

PR #685, opened 2018-05-08
add link preview to link post creation

### LINKED ISSUE #694: add embedly preview to link page

aliceriot:

part of #604 
closes #694 

#### What's this PR do?

This adds a preview display to the link post creation form, which shows you what the embedly embed (ha) for your link will look like when it's posted.

Making a text post should work like normal. If you make a link post, when you add a URL to the form, or change the URL, a rich embed thingy like this should pop up:

fjfjasdfasdf

changing the URL should cause the preview to change.

### DIFF
--- a/static/js/containers/CreatePostPage.js
+++ b/static/js/containers/CreatePostPage.js
@@ -26,7 +26,8 @@ type CreatePostPageProps = {
   channel: Channel,
   channels: RestState<Map<string, Channel>>,
   history: Object,
-  processing: boolean
+  processing: boolean,
+  embedly: Object
 }
 
 const CREATE_POST_KEY = "post:new"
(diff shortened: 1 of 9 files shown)


static/js/containers/CreatePostPage.js
```diff
@@ -57,14 +58,19 @@ class CreatePostPage extends React.Component<*, void> {
...
       actions.forms.formUpdate({
         ...CREATE_POST_PAYLOAD,
         value: {
-          [e.target.name]: e.target.value
+          [name]: value
         }
       })
     )
+    if (name === "url") {
+      dispatch(actions.embedly.get(value))
```
rhysyngsun: I think we should consider debouncing this somehow. Granted most users will copy/paste an address, but this can cause a lot of extraneous requests since it'll trigger a request for every keypress.

static/scss/createPost.scss
```diff
@@ -1,5 +1,14 @@
 .new-post-card {
   .card-contents {
+    .embedly-preview {
+      max-width: 770px;
+
+      .preview-header {
```
rhysyngsun: Maybe a little bottom padding for this?

static/js/containers/CreatePostPage.js
rhysyngsun: Also, this should not query embedly if the value is empty. To reproduce, enter a link and then clear the field.

aliceriot: yup, that makes sense >_<

aliceriot:
@rhysyngsun ok comments should be addressed!

PR #704, opened 2018-05-15
allow anonymous users to hit the embedly API

### LINKED ISSUE #703: allow anonymous access to the embedly API
Realized today that anonymous users get a 401 back from it currently.

alicewriteswrongs:

closes #703

#### What's this PR do?

This sets the permission class on the Embedly API to use the `AnonymousAccessReadonlyPermission` which we're using elsewhere.

If you're logged in, things should be normal. If you're not logged in, you should be able to view a link post on a public channel. It shouldn't redirect you to log-in (b/c of a 401) and you should be able to see the rich embed content.

### DIFF
--- a/embedly/views.py
+++ b/embedly/views.py
@@ -1,15 +1,17 @@
 """Embed.ly views"""
 from urllib.parse import unquote
 from rest_framework import status
-from rest_framework.decorators import api_view
+from rest_framework.decorators import api_view, permission_classes
 from rest_framework.response import Response
 
 from django.conf import settings
 
 from embedly.api import get_embedly
+from open_discussions.permissions import AnonymousAccessReadonlyPermission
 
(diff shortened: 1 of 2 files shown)


embedly/views.py
```diff
@@ -1,15 +1,17 @@
...
 from rest_framework.response import Response
 
 from django.conf import settings
 
 from embedly.api import get_embedly
+from open_discussions.permissions import AnonymousAccessReadonlyPermission
 
 
 @api_view()
+@permission_classes([AnonymousAccessReadonlyPermission])
```
rhysyngsun: Could you add a test case for anonymous users to be sure we can catch if this breaks?

aliceriot:
@rhysyngsun feedback addressed

PR #708, opened 2018-05-15
Profile ImageFields

### LINKED ISSUE #693: Profile: Profile image model changes and API
As a user, I want to be able to upload a profile image.

As a micromasters user, I want to be able to use my MicroMasters profile image, or replace it with a new image (I don't need to undo this change) 

The current profile image fields are populated with URLs from MIcroMasters. We need to migrate these fields and their data to an updated model that allows the user to upload new files. Our goal is a similar user experience to MicroMasters. We can take the same approach as MicroMasters, but we should take a careful look at how it was implemented in case there is a better way. 

### acceptance criteria
- [x] existing users don't have to do anything, and their current profile images will continue to work
- [x] new micromasters users will still have their profile images filled when a discussions account is created for them
- [x] existing users can update their profile image
- [x] new users can upload a profile image
- [x] users who haven't uploaded a profile image get some kind of default image
- [x] we can support multiple image sizes
- [x] profile images can be served from a configurable CDN

mbertrand:
Closes #693

#### What's this PR do?
- Adds `image_file`, `image_small_file`, `image_medium_file` ImageFields.  
- Adds a utility function to return the most appropriate image for a profile (for example, `image_small_file.url` if not `None`, else `image_small` if not None, else the default profile image.
- Adds support for enabling S3 media storage 

- Log in to OD with an existing user that has a profile image.  The image should still display.
- From Micromasters, create a new user with a profile image and go to OD.  The profile image should display.
- Log in to OD with a new or existing user that has no profile image.  A bland gray default profile image should display.
- Use the API to update the `image_small_file` field of your profile:
  - Get a `jwt_token`:
    ```python
    from fixtures.users import staff_jwt_token
    user = User.objects.get(username=<your_username>)
    staff_jwt_token(None, user)
    ```
  - Send an API PATCH request to update your profile image:
    ```
    curl --request PATCH -H "Authorization: Bearer  <jwt_token>" -F 'image_small_file=@<image_name.jpg>'  http://od.odl.local:8063/api/v0/profiles/<username>/
    ```
  - Reload OD, you should see the new profile image.


profiles/utils.py
```diff
@@ -0,0 +1,121 @@
...
+    Args:
+        filename(str): The name of the image file
+        suffix(str): 'small', 'medium', or ''
+
+    Returns:
+        str: The filepath for the uploaded image.
+    """
+    name, ext = path.splitext(filename)
+    timestamp = now_in_utc().replace(microsecond=0)
+    path_format = "{prefix}/{name}-{timestamp}{suffix}{ext}"
```
rhysyngsun: Could we namespace this into a directory of `user.username`?

channels/serializers.py
```diff
@@ -175,9 +172,9 @@ def get_author_name(self, instance):
     def get_profile_image(self, instance):
         """Find the profile image for the post author"""
         user = self._get_user(instance)
-        if user and user.profile.image_small:
-            return user.profile.image_small
-        return urljoin(settings.SITE_BASE_URL, default_profile_image)
+        if user and user.profile:
```
rhysyngsun: Could you roll this logic into `image_uri` so we have one single place where the profile image is determined?

profiles/utils_test.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    A name which is too long should get truncated to 100 characters
+    """
+    filename = '{}.jpg'.format('a' * 150)
+    with pytest.raises(ValueError) as ex:
+        _generate_upload_to_uri("x" * 150)(None, filename)  # pylint: disable=protected-access
+    assert str(ex.value).startswith("path is longer than max length even without name")
+
+
+@pytest.mark.django_db()
```
rhysyngsun: Style nit: don't need parens on a decorator with no args

requirements.txt
```diff
@@ -24,6 +25,7 @@ django-anymail[mailgun]==1.2
 django-cors-headers==2.1.0
 django-redis==4.7.0
 django-server-status==0.5.0
+django-storages-redux==1.3.2
```
rhysyngsun: I think this should just be `django-storages` now since this package is no longer getting updates in favor of the official one

profiles/models.py
```diff
@@ -1,7 +1,13 @@
 """Profile models"""
+import logging
 from django.db import models
 from django.conf import settings
 
+from profiles.utils import (
+    profile_image_upload_uri,
+    profile_image_upload_uri_medium,
+    profile_image_upload_uri_small)
```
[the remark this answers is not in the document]
mbertrand: ✅

profiles/models.py
```diff
@@ -14,6 +20,15 @@
     'bio',
 )
 
+PROFILE_IMAGE_PROPS = (
+    'image',
+    'image_small',
+    'image_medium',
+
```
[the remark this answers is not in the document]
mbertrand: ✅

profiles/models.py
```diff
@@ -14,6 +20,15 @@
...
+)
+
+logger = logging.getLogger()
```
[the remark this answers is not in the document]
mbertrand: Removed

profiles/utils.py
mbertrand: ✅

channels/serializers.py
mbertrand: ✅

requirements.txt
mbertrand: The latest version of `django-storages` no longer includes `storages.backends.overwrite.OverwriteStorage` specified in `settings.py` if S3 isn't enabled.  Everything seems to still work fine if I remove this `else` block, and tests pass.  

```python
else:
    # by default use django.core.files.storage.FileSystemStorage with
    # overwrite feature
    DEFAULT_FILE_STORAGE = 'storages.backends.overwrite.OverwriteStorage'
```

profiles/utils_test.py
mbertrand: ✅

mbertrand:
@rhysyngsun Just added the missing env. variables relevant to S3 image upload to `app.json` and `.env.example`

PR #712, opened 2018-05-16
Red dot next to incomplete profiles

### LINKED ISSUE #680: Profile: display red badge when profile is incomplete
As a user, I'd like to see a subtle notification that my profile is incomplete, which will remind me to complete it when I have the time. 

We can display a (red, circular) badge over the profile image to indicate that the profile is not complete. When we add more fields to the profile in the future, we can use the same mechanism to draw user attention.

mbertrand:
The changes in #708 should be merged first so that the check for an existing profile image can be updated.

Closes #680

#### What's this PR do?
Displays a red dot next to the profile image if user is logged in and certain profile fields (name, bio, headline, or image) are blank.

- Add `FEATURE_PROFILE_UI=True` to your `.env` file.
- Log in as a user with empty/null values for any of the following: bio, headline, name, image_small.  A red dot should appear at upper right of your profile image (or default avatar image).
- Using django shell or API request, insert values for all the above fields.  The red dot should no longer appear near your profile image.


noisecapella:
Would the end user have any way to make this go away? Should we put a feature flag on this in the meantime, at least on the frontend portion?

noisecapella:
Functionality looks good, just the one question

mbertrand:
The end user won't be able to make the red dot go away until the UI is in place to edit profiles.  @pdpinch do you want a feature flag to disable the dot (or just block this PR) until that UI is in place?

mbertrand:
@noisecapella I added a feature flag, set `FEATURE_PROFILE_UI=True` to enable the red dot.

PR #715, opened 2018-05-17
embedly styling

roberthouse54:
#### What's this PR do?
Some minor tweaks to embedly styling, margins, and thumbnail image layout.

Make sure that the thumbnail images for posts do not overflow their container, and that the layout does not break.

### DIFF
--- a/static/scss/embedly.scss
+++ b/static/scss/embedly.scss
@@ -19,7 +19,7 @@
   .photo {
     img {
       max-height: 600px;
-      object-fit: scale-down;
+      max-width: 100%;
     }
   }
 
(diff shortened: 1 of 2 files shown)


static/scss/embedly.scss
```diff
@@ -32,33 +33,32 @@
...
-      display: flex;
-      overflow: hidden;
-      flex-direction: column;
-      max-height: 300px;
       width: 100%;
-      justify-content: center;
+
+      img {
+        max-height: 600px;
+        object-fit: scale-down;
```
aliceriot: this might be kind of an edge case, but with a link-type post it kind of does some strange stuff if the image is small on the mobile view:

fasfdasdfasdf

Issue #682, opened 2018-05-08
Profile: UI to upload profile image


pdpinch:
As a new user, I'd like to be able to upload profile images. 

As an existing MicroMasters user, I can upload a new image to replace my image on Open.  

Invision Design at https://invis.io/J8KZT9PR5EN#/301346725_CropProfileImage

Zeplin styling at https://app.zeplin.io/project/5b181d3bf71408ae4d406b94/screen/5b22b71d0fd925e6553dc7f3

pdpinch:
@roberthouse54 are there designs for this? What can we borrow from MicroMasters?

pdpinch:
Waiting on final designs to be posted to Invision

mbertrand:
@pdpinch @roberthouse54 is there an invision link I can look at to see the UI design?

pdpinch:
It starts at https://invis.io/7XK8IN8W348#/301299124_Profile_Open_Drop_Down but it looks like we still don't have the image upload screens there.

pdpinch:
Invision Design at https://invis.io/J8KZT9PR5EN#/301346725_CropProfileImage

Zeplin styling at https://app.zeplin.io/project/5b181d3bf71408ae4d406b94/screen/5b22b71d0fd925e6553dc7f3

PR #729, opened 2018-05-22
Backend modifications for resizing an uploaded image

mbertrand:
- Makes some backend modifications to prepare for #682

#### What's this PR do?
- Modifies the Profile model/serializer/api to resize an uploaded image to different sizes.

- Use the API to update the `image_file` field of your profile:
  - Get a `jwt_token`:
    ```python
    from fixtures.users import staff_jwt_token
    user = User.objects.get(username=<your_username>)
    staff_jwt_token(None, user)
    ```
  - Send an API PATCH request to update your profile image:
    ```
    curl --request PATCH -H "Authorization: Bearer  <jwt_token>" -F 'image_file=@<image_name.jpg>'  http://od.odl.local:8063/api/v0/profiles/<username>/
    ```
 -  The response JSON should include populated URL's for `image_file`, `image_small_file`, and `image_medium_file`.  Open them in a browser to verify that the images load and are correctly sized (max 64x64 pixels for small, 128x128 for medium).
- The `image_small_file` photo should appear as your profile image in OD.

### DIFF
--- a/profiles/models.py
+++ b/profiles/models.py
@@ -9,7 +9,7 @@ from profiles.utils import (
     profile_image_upload_uri,
     profile_image_upload_uri_medium,
     profile_image_upload_uri_small,
-    default_profile_image, image_uri)
+    default_profile_image, image_uri,
     make_thumbnail)
 
 # Max dimension of either height or width for small and medium images
--- a/profiles/models.py
+++ b/profiles/models.py
@@ -80,7 +80,7 @@ class Profile(models.Model):
         """Update thumbnails if necessary"""
         if update_image:
             if self.image_file:
-                small_thumbnail = make_thumbnail(self.image_file.file, IMAGE_SMALL_MAX_DIMENSION)
+                small_thumbnail = make_thumbnail(self.image_file, IMAGE_SMALL_MAX_DIMENSION)
                 medium_thumbnail = make_thumbnail(self.image_file, IMAGE_MEDIUM_MAX_DIMENSION)
 
                 # name doesn't matter here, we use upload_to to produce that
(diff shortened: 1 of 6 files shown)


pdpinch:
How did we decide on the sizes?

mbertrand:
@pdpinch, current sizes are copied from micromasters, will update as necessary once Rob has finalized the design.

profiles/models.py
```diff
@@ -68,5 +75,22 @@ def is_complete(self):
                 return False
         return True
 
+    @transaction.atomic
+    def save(self, *args, update_image=False, **kwargs):  # pylint: disable=arguments-differ
+        """Update thumbnails if necessary"""
+        if update_image:
+            if self.image_file:
+                small_thumbnail = make_thumbnail(self.image_file.file, IMAGE_SMALL_MAX_DIMENSION)
```
noisecapella: Is there a meaningful difference between `self.image_file.file` here and `self.image_file` below?

noisecapella:
Functionality works great

profiles/models.py
mbertrand: First is an `ImageFieldFile` object, the second is a `File` object.  I think the `make_thumbnail` function only works with the latter.
```python
In [3]: profile.image_file
Out[3]: <ImageFieldFile: profile/01CFDJ0PE2S2PQ9QFFQD7MGMR7/0-2018-06-08T183816.jpg>

In [4]: profile.image_file.file
Out[4]: <File: /var/media/profile/01CFDJ0PE2S2PQ9QFFQD7MGMR7/0-2018-06-08T183816.jpg>
```

mbertrand: Actually nvm, seems like either will work, will change

Issue #671, opened 2018-05-07
create account with Touchstone


pdpinch:
As a member of the MIT community, I'd like to be able to create an account using Touchstone.

### acceptance criteria
- [x] use MIT email address for new account
- [x] use MIT name (SAML `displayName`?) to populate profile name

### questions
- [ ] What happens when Touchstone flow fails, or user bails out? 

### related issues (out of scope)
- pre-populate headline with MIT role, 
- set MIT status in profile
- user should be able to reset (add) password after creating account
- user should be able to add another email address to the account, for account recovery

pdpinch:
The related profile issues are blocked until the profile model is created

mbertrand:
- Touchstone (via [django-shibboleth](https://github.com/Brown-University-Library/django-shibboleth-remoteuser/blob/master/shibboleth/backends.py#L59)) will automatically overwrite the email address of the `User` on login if `email` is included in the `SHIBBOLETH_ATTRIBUTE_MAP` Django settings (as they are in OVS).  If excluded, email will not be auto-populated at all on Touchstone login/registration.  Could fork `django-shibboleth` and make these login updates optional based on a new boolean setting, or save user-specified email addresses in `Profile` instead.

- Logging in via Touchstone doesn't necessarily mean you are affiliated with MIT (collaboration accounts open to anyone).

pdpinch:
I updated the description to move the password reset function out of scope. We'll tackle that in a separate issue(s).

pdpinch:
![](https://github.trello.services/images/mini-trello-icon.png) [Log in through EdX or MicroMasters or Touchstone or Email](https://trello.com/c/Wuwcvzzb/42-log-in-through-edx-or-micromasters-or-touchstone-or-email)

mbertrand:
@tmacey do you know if there are any additional attributes returned by Touchstone with successful login, besides `EPPN`, `MAIL`, `DISPLAY_NAME`?

mbertrand:
@blarghmatey  I think we will need to generate our own X509 certificate for each server this will be enabled on (discussions.odl.mit.edu, discussions-rc.odl.mit.edu, discussions-ci.odl.mit.edu?), and then register each with the Touchstone IdP.  

![screen shot 2018-05-25 at 8 54 26 am](https://user-images.githubusercontent.com/187676/40545367-96a8a7bc-5ff9-11e8-85ae-a0343f964f50.png)

mbertrand:
[...]

PR #735, opened 2018-05-24
Support for SAML (Touchstone) login

mbertrand:
- Partially addresses #671

#### What's this PR do?
- Add support for SAML login/registration (used by Touchstone)

- `docker-compose build` (new requirements)
- Follow the steps in the [new section of the README](https://github.com/mitodl/open-discussions/tree/mb/saml#7-testing-integration-with-saml-via-ssocircle)
- Depending on the email you registered with for SSOCircle, a new user with profile should be created or you will be logged in as an existing user with the matching email address.


authentication/pipeline/user.py
```diff
@@ -108,6 +109,44 @@ def require_password_and_profile(
...
 
 
+@partial
+def require_profile_update_user_via_saml(
+        strategy, backend, user=None, is_new=False, *args, **kwargs
+):  # pylint: disable=unused-argument
+    """
+    Sets a new user's password and profile, and updates the user first and last names.
+    Touchstone only returns DisplayName, which is the full name, and that is what the
+    user first_name is initially set to.
```
noisecapella: What happens if the touchstone user's full name is greater than 30 characters (the limit on the field)? We should also check this below and add long names to our tests

Dockerfile
```diff
@@ -14,6 +14,9 @@ RUN apt-get install -y $(grep -vE "^\s*#" apt.txt  | tr "\n" " ")
 # Add repo needed for postgres 9.6 and install it
 RUN apt-get update && apt-get install libpq-dev postgresql-client-9.6 -y
 
+# Add repo needed for python3-saml
```
noisecapella: This shouldn't be necessary since you already added it to apt.txt

noisecapella:
Just went through manual testing, everything looks great

authentication/pipeline/user.py
mbertrand: I modified the name assignment code to handle this and truncate first and last names if they are more than 30 characters; tests included.

mbertrand:
@shaidar @pdpinch before the Touchstone login actually works, there's x509 certificates that have to be created, new environment variables that have to be set, and URL's that have to be registered with Touchstone.  Should I wait on merging this until those are in place?  Nothing should break if it's merged.

mbertrand:
@pdpinch I'm going to go ahead and merge this, it shouldn't affect existing functionality and can be tested with Touchstone whenever we get the x509 certificates and populate the new env. variables.

authentication/pipeline/user.py
pdpinch: I'm not sure 30 characters is a reasonable limit for a full name...

mbertrand: That's 30 characters per first name, 30 per last name: these are the defaults for the `django.contrib.auth.models.User` model (though last_name goes up to 150 in Django 2.0)

PR #742, opened 2018-05-29
Add models to store id information for posts, channels and comments

### LINKED ISSUE #730: create simple models for posts and comments
Due to the 1000 item limit in reddit APIs, we need a way to keep track of posts and comments that we want to index for searching. The idea is that we would make note of the post/comment id whenever one is created and instead of relying on reddit to return a list of all posts in a channel/comments on a post, we could walk this table. 

This may be the starting point of a larger-scale replacement of reddit for keeping track of data for search. 

Before we start on this, we need a Google doc describing the schema and scope. 

### acceptance criteria
- [ ] [Google doc describing the plan](https://docs.google.com/document/d/1V311dVEDBcX3zNjmDVe-mcMxoUQL3ChCVGkp7CxsGzA/edit#heading=h.db6yb9iglzuh)
- [ ] create a new post/comment when user posts/comments
- [ ] walk table when creating index instead of traversing reddit API

### out of scope
- [ ] storing other fields that are indexed
- [ ] updating post/comment when user updates

noisecapella:
Fixes #730 

#### What's this PR do?
Adds models to store id information for posts, channels, and comments. The functions which create new channels, posts and comments and the reindex management command will be updated to also update these tables.

Create a new channel, comment and post. Verify that there is matching information in `Channel` `Comment` and `Post` models.

Run `./manage.py recreate_index`. Verify that all posts, channels and comments have objects associated with them.


channels/api.py
```diff
@@ -618,9 +692,23 @@ def create_comment(self, text, post_id=None, comment_id=None):
...
+            return reply
+
+        reply = self.get_comment(comment_id).reply(text)
+        sync_comment_model(
+            channel_name=reply.subreddit.display_name,
+            post_id=reply.submission.id,
+            comment_id=reply.id,
+            parent_id=comment_id,
+        )
+        return reply
```
gsidebo: Pretty straightforward refactor here:

```python
if post_id is not None:
  reply = self.get_post(post_id).reply(text)
  parent_id=None
else:
  reply = self.get_comment(comment_id).reply(text)
  parent_id = comment_id
sync_comment_model(
  channel_name=reply.subreddit.display_name,
  post_id=reply.submission.id,
  comment_id=reply.id,
  parent_id=parent_id,
)
return reply
```

gsidebo:
Should this not be asynchronous? This shadow data creation only exists to support indexing, which is asynchronous so as not to impact performance or break the app on failures. This also kind of leads to the question of whether we should implement this as a decorator like `@reddit_object_indexer`. This is a very similar use case and it looks like a pretty good fit for what we're trying to do.

noisecapella:
Yeah there's no reason not to make it asynchronous, I'll see what I can refactor

pdpinch:
What's happening with this review?

gsidebo:
@pdpinch Fell off the radar. I'll take another look

channels/api_test.py
```diff
@@ -38,7 +41,15 @@
...
     """Mock reddit get_client"""
-    return mocker.patch('channels.api._get_client', autospec=True)
+    mocked = mocker.patch('channels.api._get_client', autospec=True)
+    mocked.return_value.subreddit.return_value.submit.return_value.id = 'abc'
+    mocked.return_value.submission.return_value.reply.return_value.id = '456'
+    mocked.return_value.submission.return_value.reply.return_value.submission.id = '123'
+    mocked.return_value.submission.return_value.reply.return_value.subreddit.display_name = 'subreddit'
+    mocked.return_value.comment.return_value.reply.return_value.id = '789'
+    mocked.return_value.comment.return_value.reply.return_value.submission.id = '687'
+    mocked.return_value.comment.return_value.reply.return_value.subreddit.display_name = 'other_subreddit'
```
gsidebo: This is a difficult testing pattern to read and maintain. Could we look into using factory boy for something like this? If that's too involved or not the right tool, I don't know if this is _much_ better, but maybe something like this?

```python
Mock = mocker.Mock
mock_client = Mock(
  subreddit=Mock(submit=Mock(id='abc')),
  submission=Mock(reply=Mock(return_value=Mock(
      id='456'
      submission=Mock(id='123', display_name='subreddit')))
  #...
)
mocked = mocker.patch('channels.api._get_client', autospec=True, return_value=mock_client)
```

You could also use `PropertyMock` to patch the `reddit` property instead of patching the `_get_client` function ([doc](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock)). Not a huge benefit but it's one less link in the chain of things you need to mock

noisecapella:
@gsidebo Did you make your larger scope comments?

gsidebo:
I was noticing some code smells while reading through this. 
1. You're basically introducing another set of serializers for comments and posts (albeit simple ones)
1. The channel API methods now have an added concern of calling the sync tasks explicitly. One of the reasons for using a decorator for indexing was that the API methods would only be concerned with their primary function (using PRAW to create a post, etc.) and the decorator would take care of propagating that data to a new store.
1. In the context of our tasks that iterate through reddit objects to serialize and index them, the calls to sync to these new models piggyback on those tasks in a way that is unintuitive and non-obvious, IMO. `index_channel` opens with a call to sync the channel data, then somewhere down the line, the serialized post/comment iterator is wrapped to sync posts/comments before they make it into the request to index those posts/comments. For someone reading this code for the first time, I don't think it would be obvious that the syncing is even being done as part of the indexing task, let alone how it's being done.

For (3), I have some ideas about how to handle that, but a separate tech debt issue might be more appropriate. I think we should have a post/comment generator up front as soon as we start the `index_channel` task (rather than creating it deeper in the execution). We could immediately wrap that generator to run the syncing tasks and the posts/comments are iterated. That way it's much more obvious that indexing and syncing go hand in hand.

For (1) and (2), I think those should be addressed as part of this PR. I think we can rename the decorator to `reddit_object_persist` or something like that, rename `indexing_func` to something like `persistance_func`, and allow it to take multiple functions that would be run in series and receive a reddit object as an arg (like the indexing functions already do). I'm ok with the objects being 'manually' serialized for the sync tasks for now since they're really simple, but down the line we could easily make it so that the object is serialized before it's passed into the indexing/sync functions. Let me know what you think

channels/api.py
```diff
@@ -464,7 +537,12 @@ def create_post(self, channel_name, title, text=None, url=None):
         """
         if len(list(filter(lambda val: val is not None, [text, url]))) != 1:
             raise ValueError('Exactly one of text and url must be provided')
-        return self.get_channel(channel_name).submit(title, selftext=text, url=url)
+        post = self.get_channel(channel_name).submit(title, selftext=text, url=url)
+        tasks.sync_post_model(
```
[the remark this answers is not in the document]
noisecapella: Yes, I'll fix this and move this into a decorator

channels/api_test.py
noisecapella: I'm not sure factory boy can deal with nested patches like this. I'll see about using the Mock you mentioned

noisecapella:
I think this is ready for review again. I changed a bunch of things, let me know if this is what you were thinking

fixtures/common.py
```diff
@@ -69,23 +69,24 @@ def session_indexing_decorator():
     when a module is loaded. To get around this, we patch the decorator function then
     reload the relevant modules via importlib.
     """
-    mock_indexer_func = Mock()
+    mock_indexer_func = Mock(original=[])
```
gsidebo: Should be renamed to `mock_persist_func` or something like that (that also goes for the value returned in the `SimpleNamespace`)

noisecapella:
Comments addressed

channels/api.py
```diff
@@ -603,7 +678,10 @@ def delete_post(self, post_id):
         post.delete()
         return post
 
-    @reddit_object_indexer(indexing_func=index_new_comment)
+    @reddit_object_persist(
+        index_new_comment,
+        task_helpers.sync_comment_model,
```
gsidebo: In my previous comment, I meant that the imports of task helpers *in this module* should be consistent. i.e.: we import the entire search task helper module like we do for the channel task helpers (e.g.: `from search import task_helpers as search_task_helpers`) or we import the channel task helpers by name. Right now we have the search task helpers imported by name and the channel task helpers imported as a whole module. I'm ambivalent about the import rename in the test file

noisecapella: No problem I'll make the same change in this one

PR #745, opened 2018-05-29
fix issue with twitter embeds

### LINKED ISSUE #744: tweet embeds aren't working with embed.ly

aliceriot:

closes #744

#### What's this PR do?

This fixes the bug with embedded tweets not showing up, but ensuring that the twitter JS is on the page and calling its little bootstrap function when necessary.

I think the issue came about because we don't put the twitter HTML on the page at page-load, but afterwards, whenever the embedly request comes back. 

Make sure that if you embed a tweet it will show up in a nice normal tweet embed sort of fashion.

### DIFF
--- a/static/js/containers/PostPage.js
+++ b/static/js/containers/PostPage.js
@@ -64,6 +64,7 @@ import {
   REPORT_CONTENT_NEW_FORM,
   REPORT_CONTENT_PAYLOAD
 } from "../lib/reports"
+import { ensureTwitterEmbedJS, handleTwitterWidgets } from "../lib/embed"
 
 import type { Dispatch } from "redux"
 import type { Match, Location } from "react-router"
@@ -130,6 +131,8 @@ class PostPage extends React.Component<*, void> {
 
   componentWillMount() {
     this.loadData()
+
+    ensureTwitterEmbedJS()
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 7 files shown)


static/js/containers/PostPage.js
```diff
@@ -178,7 +181,14 @@ class PostPage extends React.Component<*, void> {
       ])
 
       if (post.url) {
-        await dispatch(actions.embedly.get(post.url))
+        const embedlyResponse = await dispatch(actions.embedly.get(post.url))
+
+        if (
```
mbertrand: This seems to be identical to code in `CreatePostPage`,  maybe it could be moved to a common function in `lib/embed.js`?

```javascript
export const loadTwitter = (embedlyResponse) => {
  if (embedlyResponse && embedlyResponse.response.provider_name === "Twitter") {
    window.twttr.widgets.load()
  }
}
```

static/js/factories/embedly.js
```diff
@@ -27,6 +27,23 @@ export const makeImage = () => ({
   type:          "photo"
 })
 
+export const makeTweet = () => ({
+  provider_url:    "http://twitter.com",
```
mbertrand: Should this be `https` or does it not really matter?

static/js/containers/PostPage.js
aliceriot: yeah, that's a good idea

static/js/factories/embedly.js
aliceriot: this is just a sample response I captured from embedly, so yeah it should be https but...

in the test it's really just a placeholder so I think it's ok

aliceriot:
@mbertrand comments addressed!

PR #749, opened 2018-05-31
upgrade linting and formatting tooling

aliceriot:

none, just keeping some dependencies up to date

#### What's this PR do?

I updated `babel-eslint`, `eslint`, `eslint-plugin-babel`, `eslint-plugin-flowtype`, `eslint-plugin-mocha`, `eslint-plugin-react`, and `prettier-eslint-cli` to their latest version. There are some related changes:

- prettier now formats a few things differently, in particular ternaries with JSX. it now formats them like this:

```js
{fooBar ? (
  <div>BOO</div>
) : null}
```

instead of

```js
{ fooBar
  ? <div>BOO</div>
  : null}
```

which is kind of nice I think, but did change how a lot of things were formatted.

- I also had to add a setting for the `eslint-plugin-react` package to tell it what react version we're on.

Well, this is just linting and formatting stuff, so nothing should be different! I would click around the app and make sure you can do the normal things.


.eslintrc
```diff
@@ -1,3 +1,9 @@
 {
-  "extends": "eslint-config-mitodl"
+  "extends": "eslint-config-mitodl",
+  "settings": {
+    "react": {
+      "version": "15.6.1",
+      "flowVersion": "0.59.0" // Flow version
```
noisecapella: Is it possible to put these settings in the eslint-config-mitodl repo somehow?

static/js/components/CommentForms.js
```diff
@@ -217,7 +226,7 @@ export const ReplyToCommentForm = connect(mapStateToProps, mapDispatchToProps)(
 
     onSubmit = async (event: Event) => {
       const { formKey, forms, post, onSubmit } = this.props
-      const { post_id, text, comment_id } = R.prop(formKey, forms).value
+      const { post_id, text, comment_id } = R.prop(formKey, forms).value // eslint-disable-line camelcase
```
noisecapella: What was the camelcase lint error? For comment_id and post_id?

noisecapella:
Just a couple questions, nothing blocking this :+1:

.eslintrc
aliceriot: I don't think we'd want to, because then we'd have to be on the same version of react and flow everywhere (and sometimes we run into difficulties upgrading them)

static/js/components/CommentForms.js
aliceriot: yeah, I'm not exactly sure why it wasn't an error before :woman_shrugging:

PR #759, opened 2018-06-01
upgrade react and some associated packages

aliceriot:

none

#### What's this PR do?

this upgrades react and react-dom, and then upgrades a few ancillary packages that had some sort of issues with the new version (basically, enzyme).

There are some code changes, but they aren't too major. The big thing is that `componentWillMount` is now a deprecated method. I was able to change most of our usage over to `componentDidMount` without any apparent effects, but this should obviously be checked closely.

Enzyme also changed substantially from version 2 to version 3. There is a summary here: https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md

A key thing is that the enzyme wrapper is no longer a dynamically updated thing full of pointers to a  mounted react component tree. Instead, it is a snapshot at a given point in time. So, to support our integration-style enzyme tests you need to do something like this:

```js
const wrapper = renderComponent()
wrapper.find('.my-cool-button').simulate('click')
wrapper.update()
assert.ok(wrapper.find('.button-clicked').exists())
```

the `.update` method basically causes a refresh of the `wrapper` object, so that the react tree will reflect any changes that have occurred in the redux store or something like that. It's not too heavy, and I think it is actually nicer this way, because having to manually update the `wrapper` will probably force us to avoid potentially flaky tests. But, anyway, I had a add a liberal sprinkling of `wrapper.update()` around.

Read through the code changes and make sure they're all sensible. Make sure that there are no regressions in the application.


static/js/containers/CreatePostPage_test.js
```diff
@@ -99,63 +100,63 @@ describe("CreatePostPage", () => {
...
+    it(`submits a ${isText ? "text" : "url"} post`, async () => {
       const post = makePost(!isText)
       helper.createPostStub.returns(Promise.resolve(post))
 
-      return renderPage().then(async ([wrapper]) => {
-        const title = "Title"
-        const text = "Text"
-        const url = "http://url.example.com"
-        setTitle(wrapper, title)
+      // return renderPage().then(async (wrapper) => {
```
rhysyngsun: Unused can be removed?

aliceriot: oops!

rhysyngsun:
:+1: looks good other than that one stray comment

PR #782, opened 2018-06-06
Fix comment serialization error, Celery error handling

### LINKED ISSUE #760: Error handling is not working properly for recreate_index
The Celery `group` is not propagating errors up to the parent even though `raise self.replace(...)` is used.

### LINKED ISSUE #762: Comment serialization doesn't work if user exists on reddit but not open-discussions

noisecapella:
Fixes #760 
Fixes #762 

#### What's this PR do?
Fixes a None error with missing reddit authors. Error handling for tasks is also refactored to not rely on Celery propagating the error correctly through groups.

In `serialize_comment` add some code which will fail for a particular comment. Something like `if comment.id == 'abc': raise Exception()` would work. Run `./manage.py recreate_index`. It should fail and show you the comment id on which it failed.

Remove the exception code and run it again. It should succeed.

### DIFF
--- a/open_discussions/utils.py
+++ b/open_discussions/utils.py
@@ -93,3 +93,34 @@ def chunks(iterable, chunk_size=20):
     while len(chunk) > 0:
         yield chunk
         chunk = list(islice(iterable, chunk_size))
+
+
+def merge_strings(list_or_str):
+    """
+    Recursively go through through nested lists of strings and merge into a flattened list.
+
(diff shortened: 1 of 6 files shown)


mbertrand:
@noisecapella Newly created posts and comments get added to the index, but running `recreate_index` just deletes all documents from my index (including on master branch).  Could it be a misconfigured env setting?

PR #789, opened 2018-06-07
Refactored lib/auth*.js files

rhysyngsun:
N/A

#### What's this PR do?
Refactors lib/auth.js files

Quick smoke test should be sufficient


aliceriot:
I think there might be one other import, in `lib/api_test.js`

rhysyngsun:
@aliceriot amended that in the same commit

aliceriot:
ahh that should have been :+1: ! dang autocomplete

Issue #763, opened 2018-06-04
set up SAML environment variables


pdpinch:
from https://github.com/mitodl/open-discussions/issues/671#issuecomment-394346055

```
FEATURE_SAML_AUTH=True
SOCIAL_AUTH_SAML_SP_PUBLIC_CERT=<x509 public certificate string for server>
SOCIAL_AUTH_SAML_SP_PRIVATE_KEY=<x509 private key string for server>
SOCIAL_AUTH_SAML_ORG_DISPLAYNAME=<'MIT ODL' or whatever else you prefer>
SOCIAL_AUTH_SAML_CONTACT_NAME=<'ODL Support' or whatever else you prefer>

SOCIAL_AUTH_SAML_IDP_ENTITY_ID=https://idp.mit.edu/idp/shibboleth
SOCIAL_AUTH_SAML_IDP_URL=https://idp.mit.edu/idp/profile/SAML2/Redirect/SSO
SOCIAL_AUTH_SAML_IDP_ATTRIBUTE_PERM_ID=EPPN
SOCIAL_AUTH_SAML_IDP_ATTRIBUTE_NAME=DISPLAYNAME
SOCIAL_AUTH_SAML_IDP_ATTRIBUTE_EMAIL=MAIL
[...]

PR #797, opened 2018-06-08
Add security config and entityID setting

mbertrand:
- #763 

#### What's this PR do?
- Makes the SAML EntityID configurable via env settings
- Adds an option to set the SAML `wantAssertionsEncrypted` attribute to True or False

- Set the following in your `.env` file:
  - `SOCIAL_AUTH_SAML_SECURITY_ENCRYPTED=True` 
  - `SOCIAL_AUTH_SAML_SP_ENTITY_ID=http://od.odl.local:8063/saml/metadata`
- Go to `/saml/metdata`
  - The XML should include `entityID="http://od.odl.local:8063/saml/metadata"` on the 1st line, as well as a `<md:KeyDescriptor use="encryption">` element.


blarghmatey:
:+1:

PR #798, opened 2018-06-08
Make form pages more ergonomic

rhysyngsun:
Prereq to #650 

#### What's this PR do?
This adds utilities to avoid boilerplate around how we handle form events. You can look in the aforementioned PR to see how it's used. I did not add any usages of these in this PR to keep scope down.

Tests should pass

### DIFF
--- a/static/js/hoc/withForm.js
+++ b/static/js/hoc/withForm.js
@@ -1,0 +1,91 @@
+// @flow
+import React from "react"
+import R from "ramda"
+
+import type { ConfiguredFormProps } from "../lib/forms"
+import type { FormValue } from "../flow/formTypes"
+
+export type WithFormProps = {
+  form: ?FormValue,
+  processing: boolean,
+  onUpdate: Object => void,
+  onSubmit: Object => void,
+  onSubmitResult: Function,
+  validateForm: Function,
+  renderForm: Function
+} & ConfiguredFormProps
+
+export type FormProps = {
+  form: Object,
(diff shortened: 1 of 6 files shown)


static/js/hoc/withForm.js
```diff
@@ -0,0 +1,91 @@
...
+        const { formBeginEdit } = this.props
+        formBeginEdit()
+      }
+
+      componentWillUnmount() {
+        const { formEndEdit } = this.props
+        formEndEdit()
+      }
+
+      onUpdate = (e: Object) => {
```
noisecapella: Can you add a test covering this function?

static/js/lib/redux_props_test.js
```diff
@@ -0,0 +1,32 @@
...
+  beforeEach(() => {
+    sandbox = sinon.sandbox.create()
+  })
+
+  afterEach(() => {
+    sandbox.restore()
+  })
+
+  describe("mergeAndInjectProps", () => {
+    it("should pass stateProps, dispatchProps, and ownProps into the passed method", () => {
```
noisecapella: Can you add some overlapping props verifying the order the spread operator gets applied?

rhysyngsun:
Python tests are failing from an upstream commit in master, since this is just JS code, that shouldn't affect the ability to merge this.

rhysyngsun:
@noisecapella tests are passing after rebase now

PR #739, opened 2018-05-25
Added ES comment document indexing

gsidebo:
#### What are the relevant tickets?
Closes #606

#### What's this PR do?
Adds logic to create/update comment documents in Elasticsearch

#### How should this be manually tested?
- Set `FEATURE_INDEX_UPDATES=True` in your `.env`
- Create/update comments and check to see if those changes are reflected in the ES index
- Set a post to removed, then check to see that all comments associated with that post are also set to removed
- Delete a comment and check to see that the parent post's `num_comments` is decremented

#### Where should the reviewer start?
Probably `search/task_helpers.py`

#### Any background context you want to provide?
There are comments that say as much, but we are going ahead with an approach where we will log information about version conflicts in document updates and let the app continue to function as normal.

PR #802, opened 2018-06-11
Minor serializer test refactor

gsidebo:
No ticket

#### What's this PR do?
Minor refactor for serializer tests that makes use of the reddit object fixtures that were created for a separate feature

Tests should pass

`search/serializers_test.py` and `fixtures/reddit.py`

I initially wrote these as part of #739 since there was a serialization bug that wasn't caught due to lack of tests. Then some tests were written and merged into master while that PR was still open. I decided it was best to drop my tests for that PR and open this separately


gsidebo:
@noisecapella Your tests are more explicit, but I think I prefer this test approach as it makes use of existing fixtures and gets rid of the forked logic. Let me know what you think

PR #810, opened 2018-06-11
Set requestedAuthnContext to False for SAML

mbertrand:
#763

#### What's this PR do?
Adds `"requestedAuthnContext": False` to SAML configuration, which should cause Touchstone to show the 'login via certificate' option.

Nothing should break


blarghmatey:
:+1:

PR #848, opened 2018-06-21
Use gravatar for new profiles without images

### LINKED ISSUE #833: Gravtar integration
As a user, I'd like to save time uploading a profile image by using my existing gravatar image. 

### acceptance criteria:
- [x] By default use the gravatar image as the user’s profile image
- [x] If gravatar not found, use the initials
- [x] allow the user to upload a new image and override the gravatar

mbertrand:
- Closes #833 

#### What's this PR do?
- Checks the user's gravatar URL when a new profile is saved without images; if found the image fields are updated with the gravatar URL's.

- Login/Register with a new user that has a gravatar account associated with the user's email.  The profile image should be the same as the gravatar image.
- Go to the profile page, and upload an image.  The uploaded image should replace the gravatar.
- Login/Register with a new user that does not have a gravatar account associated with the user's email.  The profile "image" should be a circle with the user's initials.

### DIFF
--- a/authentication/pipeline/user_test.py
+++ b/authentication/pipeline/user_test.py
@@ -179,6 +179,7 @@ def test_validate_require_password_and_profile_via_email_exit(mocker, backend_na
 @pytest.mark.django_db
 def test_validate_require_password_and_profile_via_email(mocker):
     """Tests that require_password_and_profile_via_email processes the request"""
+    mocker.patch('profiles.models.requests.get')
     user = UserFactory(profile=None)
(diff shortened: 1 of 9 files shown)


noisecapella:
When I tried with a profile without an avatar I got a null exception:

screenshot from 2018-06-21 16-06-44

mbertrand:
@noisecapella I think that might be caused by the profile not having a name?  Was it manually created?

mbertrand:
I updated the code to not display initials if profile.name is null

PR #852, opened 2018-06-22
Add UI to edit post types

### LINKED ISSUE #831: Channel Settings: Allowed post types
The bottom half of: https://invis.io/XSLE63T6GR2#/303979450_Settings__Channel

Zeplin: https://app.zeplin.io/project/5b181d3bf71408ae4d406b94/dashboard?tag=settings

### acceptance criteria:
- [ ] "Basic" tab
- [ ] Moderators can choose "short text", "link" or both
- [ ] Any moderator can change the channel settings
- [ ] If user chooses only one, remove the other tab in the post form

### out of scope
- [ ] styling

noisecapella:
Fixes #831 

#### What's this PR do?
Adds checkboxes selecting the post types allowed (either text, link, or both). 

Go to the edit channel page. Under the Basic tab uncheck all post types then click 'Save'. You should see an error message. Check one of the boxes and you should be directed to the channel page. If you go back to edit it again you should see that the checkbox you selected is still selected.


noisecapella:
I'm not sure. It's a reddit setting so whatever reddit allows we are allowing. Can you try it as a non-moderator?

aliceriot:
ok, so as a non-moderator it seems to be enforcing the restriction correctly, which is good, haha. I'm not sure if this is out-of-scope for this PR, but when you try to submit a disallowed post type there's no indication in the UI of what's going on, no error message or anything. It looks like the response from the backend is a standard stack trace too, which the redux hammock code doesn't know what to do with (so there's an exception on the frontend from trying to call `JSON.parse` on it).

static/js/lib/channels_test.js
```diff
@@ -39,4 +46,48 @@ describe("Channel utils", () => {
...
+      // The value should be checked if it matches the link value
+      [LINK_TYPE_TEXT, LINK_TYPE_TEXT, true],
+      [LINK_TYPE_TEXT, LINK_TYPE_LINK, false],
+      [LINK_TYPE_LINK, LINK_TYPE_TEXT, false],
+      [LINK_TYPE_LINK, LINK_TYPE_LINK, true],
+      // An empty string means no link value is checked
+      ["", LINK_TYPE_TEXT, false],
+      ["", LINK_TYPE_LINK, false]
+    ].forEach(([linkType, value, expected]) => {
+      assert.equal(isLinkTypeChecked(linkType, value), expected)
```
aliceriot: I think you need to wrap the assert it an `it()` call for it to be run

static/js/lib/validation.js
```diff
@@ -89,6 +89,14 @@ export const validateChannelEditForm = validate([
   )
 ])
 
+export const validateChannelBasicEditForm = validate([
+  validation(
+    R.isEmpty,
+    R.lensPath(["value", "link_type"]),
+    "One of the post type options must be selected"
```
aliceriot: Maybe `"At least one"` ?

aliceriot: also it's super trivial, but it would probably be good to at least have a simple regression unit test for this

noisecapella:
Right, I forgot about the part of the PR to remove the tab on the create post form. I'll work on that now

noisecapella:
Done!

aliceriot:
cool, taking a look now!

noisecapella:
I think I've handled the additional edge case you found. There's an additional `FORM_UPDATE` but it should only be there when you switch channels or load for the first time

aliceriot:
yep, works great! reading code now.

static/js/lib/channels.js
```diff
@@ -34,3 +47,53 @@ export const userCanPost = (channel: Channel) =>
...
+  }
+}
+
+export const isLinkTypeChecked = (linkType: string, value: string) => {
+  if (linkType === LINK_TYPE_ANY) {
+    return true
+  } else {
+    return linkType === value
+  }
+}
```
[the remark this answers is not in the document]
aliceriot: ah, I see that it's being used in both the UI form code for setting the link types and in the code for figuring out what tabs to show

static/js/containers/CreatePostPage.js
```diff
@@ -51,6 +61,32 @@ class CreatePostPage extends React.Component<*, void> {
...
+
+    if (
+      postForm &&
+      channel &&
+      (postForm.value.postType === null ||
+        !isLinkTypeChecked(channel.link_type, postForm.value.postType))
+    ) {
+      const postType = isTextTabSelected(null, channel)
+        ? LINK_TYPE_TEXT
+        : LINK_TYPE_LINK
```
aliceriot: reading through this is a bit difficult to follow, since `isTextTabSelected` is called in a bunch of places for different functions. it seems that what we're doing here is check if the form postType value is either null or a value which is not allowed for this channel, if it's not allowed for the channel we want to set it to the right one.

I think it may make sense to define another function, other than `isLinkTypeChecked`, which is called `isLinkTypeAllowed` - it may check a very similar, or even the same, thing, but it can improve readability by making it more clear what we're doing. It might also be more readable to just check

```js
(postForm.value.postType === null ||
  (channel.link_type !== LINK_TYPE_ANY && (channel.link_type !== postForm.value.postType)))
```
just mentioning this because when reading this it took me a while to read through the different functions, parse out what they were doing with null arguments, etc.

also, in this case, I *think*, if I'm reading things right, we can set the `postType` variable here to be

```js
const postType = channel.link_type === LINK_TYPE_ANY
  ? LINK_TYPE_TEXT
  : channel.link_type
```

if the channel link type is `LINK_TYPE_ANY`, then we're in this block because `postForm.value.postType` is `null`, so we can just pick `LINK_TYPE_TEXT` as a default value. On the other hand, if the `channel.link_type` is not `LINK_TYPE_ANY`, then the value of `channel.link_type` is going to be the only valid choice anyhow.

noisecapella:
I added some extra tests and did the refactoring you suggested. Does it make sense to you?

static/js/containers/CreatePostPage_test.js
```diff
@@ -336,5 +346,57 @@ describe("CreatePostPage", () => {
...
+      it(`${
+        shouldDispatch ? "dispatches" : "doesn't dispatch"
+      } FORM_UPDATE if the post types when it goes from ${fromLinkType} to ${toLinkType}`, async () => {
+        currentChannel.link_type = fromLinkType
+        const wrapper = await renderPage()
+        const startingActions = [...helper.actionsLog]
+        channels[7].link_type = toLinkType
+        const select = wrapper.find("select")
+        select.simulate("change", { target: { value: channels[7].name } })
+        // Not sure how to wait for something which might not to happen
```
aliceriot: can we stub out `dispatch` or something and assert it doesn't get called?

aliceriot: if not, or if it's impractical, I think this approach is fine

aliceriot:
yeah, I think it's a lot more clear now. I just had one tiny question, other than that things look good

noisecapella:
I replace the tests with a test calling the function directly

PR #862, opened 2018-07-02
Add preventDefault to toolbar click handler

noisecapella:
None

#### What's this PR do?
Adds a `preventDefault()` call to the click handler for the hamburger menu click handler

Click the menu. You should not see a `#` appear in the URL


aliceriot:
Not sure why, but I'm still seeing the `#` appear in the URL when I click. could it be happening somewhere else?

aliceriot:
nvm, ignore that! I was looking at the wrong tab :woman_facepalming:

PR #863, opened 2018-07-02
Upgrade javascript dependencies

noisecapella:
None

#### What's this PR do?
Upgrades javascript packages and node version

Nothing should break


static/js/util/withTracker.js
```diff
@@ -3,18 +3,19 @@
 
 // From https://github.com/ReactTraining/react-router/issues/4278#issuecomment-299692502
 import React from "react"
-import ga from "react-ga"
+import ReactGA from "react-ga"
 
 const withTracker = (WrappedComponent: Class<React.Component<*, *>>) => {
   const debug = SETTINGS.reactGaDebug === "true"
+  const test = !!SETTINGS.isTest
```
aliceriot: is this just to make sure it's in test mode in the tests? if so I think it might be a little cleaner to check if a node-only global is defined, for instance `process`, rather than adding a new `isText` boolean to our `SETTINGS` global.

aliceriot: and maybe additionally check the `process.env` is a value we expect (think it's `'development'` normally)

noisecapella:
It looks like `process` is actually provided by webpack by default: https://webpack.js.org/configuration/node/#node-process. There's a `browser` field which is set to true for node and undefined for the browser environment, from the docs: https://www.npmjs.com/package/process#browser-implementation

aliceriot:
Oh I didn't realize that! Sounds like what we need, haha

PR #869, opened 2018-07-03
post detail page changes

### LINKED ISSUE #834: Detail page: layout of title / author / submitted time / vote / follow /  share
detailpage

from https://projects.invisionapp.com/d/main#/console/14224393/303520620/preview

aliceriot:
closes #834 

#### What's this PR do?

This updates the detail page display of the post for the most recent designs. The design on the linked issue isn't the most up-to-date, look on invision instead.

Basically, some stuff is moved around, and some of the post actions are now moved into a little dropdown menu (similar to on the post list view). There is also a little 'share' menu, which lets you easily copy a link to the post or share it to your fb or whatever.

Check that the changes implement the design, and that the new stuff works alright. Nothing should be weird...


pdpinch:
Do you have the link to Invision?

aliceriot:
they're all in the 'detail' section, here: https://invis.io/B2MF7PWGMY7

aliceriot:
this one in particular shows the final design for the buttons along the bottom: https://projects.invisionapp.com/share/B2MF7PWGMY7#/screens/305904717

static/js/containers/ProfileImage.js
```diff
@@ -36,6 +36,7 @@ class ProfileImage extends React.Component<*> {
     startPhotoEdit: (p: File) => void,
     updatePhotoEdit: (b: Blob) => void,
     useSmall?: boolean,
+    useMicro?: boolean,
```
rhysyngsun: Rather than having multiple boolean fields for each type (which must/should all agree or we could end up in undefined behavior, e.g. `<ProfileImage useMicro useSmall />` is valid with this implementation), I'd suggest passing a single `size` prop that is typed as a union and determine whether you're using micro, small, medium off of that.

static/js/components/ExpandedPostDisplay.js
```diff
@@ -153,17 +194,19 @@ export default class ExpandedPostDisplay extends React.Component<*, void> {
...
-          />
           <div className="post-title">{formatPostTitle(post)}</div>
           <div className="authored-by">
-            by <span className="author-name">{post.author_name}</span>,{" "}
-            {formattedDate}
+            <div className="left">
+              <ProfileImage
+                profile={{
+                  name:        post.author_name,
+                  image_small: post.profile_image
```
rhysyngsun: Here, because `useMicro` is specified `makeProfileImageUrl` ends up returning the default profile image because the `useSmall` value passed to it is `undefined` and there is no medium profile image passed here.

static/js/components/ExpandedPostDisplay.js
```diff
@@ -153,17 +194,19 @@ export default class ExpandedPostDisplay extends React.Component<*, void> {
...
       <div className="post-summary expanded">
         <div className="summary">
-          <ProfileImage
-            profile={{
-              name:        post.author_name,
-              image_small: post.profile_image
-            }}
-            useSmall={true}
-          />
           <div className="post-title">{formatPostTitle(post)}</div>
```
rhysyngsun: I think the font size of this should increase to match the mockup. I'm uncertain as to what exact size it should be.

static/js/components/ExpandedPostDisplay.js
```diff
@@ -88,60 +98,91 @@ export default class ExpandedPostDisplay extends React.Component<*, void> {
...
-          </div>
-        ) : null}
-        {!userIsAnonymous() ? (
-          <div
-            className="comment-action-button report-post"
-            onClick={showPostReportDialog}
-          >
-            <a href="#">report</a>
+        <div className="left">
+          <PostVotingButtons
```
rhysyngsun: Are you addressing the icon change for the upvote button in a separate PR?

static/js/components/SharePopup.js
```diff
@@ -0,0 +1,56 @@
...
+    const { url } = this.props
+
+    return (
+      <div className="share-popup">
+        <div className="triangle" />
+        <div className="share-title">Share a link to this post:</div>
+        <input ref={ref => (this.input = ref)} readOnly value={url} />
+        <div className="bottom-row">
+          <div className="share-buttons">
+            <FacebookShareButton url={url}>
```
rhysyngsun: Nit: these buttons could probably use a little horizontal padding between each other

static/js/components/SharePopup.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  render() {
+    const { url } = this.props
+
+    return (
+      <div className="share-popup">
+        <div className="triangle" />
+        <div className="share-title">Share a link to this post:</div>
+        <input ref={ref => (this.input = ref)} readOnly value={url} />
```
rhysyngsun: This input hangs off the end of the popup:

2018-07-03-164628_1920x1080_scrot

static/js/components/SharePopup.js
```diff
@@ -0,0 +1,56 @@
...
+
+export class SharePopupHelper extends React.Component<*> {
+  input: ?HTMLInputElement
+
+  handleClickOutside = () => {
+    const { closePopup } = this.props
+    closePopup()
+  }
+
+  selectInputText = () => {
```
rhysyngsun: This needs some `preventDefault()` magic so the page doesn't scroll when you click it.

static/js/components/SharePopup.js
```diff
@@ -0,0 +1,56 @@
...
+import onClickOutside from "react-onclickoutside"
+import {
+  FacebookShareButton,
+  LinkedinShareButton,
+  TwitterShareButton
+} from "react-share"
+import { FacebookIcon, TwitterIcon, LinkedinIcon } from "react-share"
+
+export class SharePopupHelper extends React.Component<*> {
+  input: ?HTMLInputElement
```
rhysyngsun: Should use `React.createRef()` instead: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

static/js/components/ExpandedPostDisplay.js
```diff
@@ -88,60 +98,91 @@ export default class ExpandedPostDisplay extends React.Component<*, void> {
...
+            </div>
+          ) : null}
+          <div className="post-action share-action">
+            <div onClick={showPostShareMenu}>
+              <i className="material-icons reply">reply</i>
+              Share
+            </div>
+            {postShareMenuOpen ? (
+              <SharePopup
+                url={window.location.href}
```
rhysyngsun: I don't think we want to do `window.location.href` here because it's not always the canoncial link to the post. A counterexample can be seen if you click on the "permalink" button for a comment and then click the "share" button, you get the url to the comment, not the post, which I don't think is what we want.

static/js/containers/ProfileImage.js
aliceriot: yeah, that makes sense to me

static/js/components/ExpandedPostDisplay.js
aliceriot: The last I heard the design wasn't yet finalized, so I have held off on it so far

static/js/containers/ProfileImage.js
aliceriot: unfortunately it seems that because of the `connect` call the props for the `ProfileImage` aren't actually checked anywhere when the component is mounted...seeing if there's an easy way to work around that

rhysyngsun: Would `flow-typed` defs for `react-redux` help flow infer that? https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-redux_v5.x.x/flow_v0.63.0-/react-redux_v5.x.x.js

aliceriot: yeah I added them, unfortunately that broke flow typings for a bunch of other things so I had to fiddle with it for a while. I just got it all typechecking now (haven't pushed the commits yet though)

static/js/components/ExpandedPostDisplay.js
aliceriot: yeah, it's 28px on the Zeplin thing so I'll change it to that

static/js/components/SharePopup.js
aliceriot: Firefox seems to set a different default `box-sizing` on `<input>` tags than chrome, it should be fixed now

static/js/components/SharePopup.js
aliceriot: nice, I didn't know about the new API

static/js/components/ExpandedPostDisplay.js
aliceriot: makes sense, I'll write a library function for getting a post permalink

aliceriot:
@rhysyngsun ok I think all your comments should be addressed now

aliceriot:
nvm, I have one more thing to do (we don't want to show the social sharing buttons on private channels)

aliceriot:
ok, I think things should be ready for review now

static/js/containers/ProfileImage.js
```diff
@@ -68,27 +72,29 @@ class ProfileImage extends React.Component<*> {
...
 
   render() {
-    const { profile, useSmall, editable, onClick } = this.props
+    const { profile, imageSize, editable, onClick } = this.props
 
-    const imageUrl = makeProfileImageUrl(profile, useSmall)
-    const imageSizeClass = useSmall ? "small" : "medium"
+    const imageUrl = makeProfileImageUrl(
+      profile,
+      imageSize === "micro" || imageSize === "small"
```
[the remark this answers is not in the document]
aliceriot: I was on the fence when I was writing it, on the one hand it would prevent typos, but Flow *should* do that automatically. One thing that is annoying about these union types is that it doesn't seem to allow you to declare constants (as JS variables) and then declare a type using those variables. I wanted to do something like this:

```js
const PROFILE_IMAGE_MICRO = "micro"
const PROFILE_IMAGE_SMALL = "small"
const PROFILE_IMAGE_MEDIUM = "medium"

type ImageSize =
  | PROFILE_IMAGE_MICRO
  | PROFILE_IMAGE_SMALL
  | PROFILE_IMAGE_MEDIUM
```

But when I tried it gave me an error about needing to use the `typeof` operator - but when I do that the type for `ImageSize` ends up being `string | string | string`! With the current setup for the `ImageSize` type it throws a type error if I pass, for instance, "Micro" as the `imageSize` prop, but if the type is `string | string | string` it will obviously accept any string.

I'm not sure if this is a limitation for Flow or I just don't know the right syntax. Anyway, so we could declare constants for the three strings, which would make it easier to change them in one place later if we wanted, but it doesn't seem like they can be used in the flow type :/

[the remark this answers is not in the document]
aliceriot: this provides a little context for that: https://github.com/facebook/flow/issues/2639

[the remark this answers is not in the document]
aliceriot: I'll go ahead and declare constants to use when setting the size prop though

static/js/components/SharePopup_test.js
```diff
@@ -0,0 +1,85 @@
...
+    ].forEach(([button, icon]) => {
+      const buttonEl = wrapper.find(button)
+      assert.ok(buttonEl.exists())
+      assert.ok(buttonEl.find(icon).exists())
+    })
+  })
+
+  it("should hide buttons, if hideSocialButtons === true", () => {
+    const wrapper = renderSharePopupHelper({ hideSocialButtons: true })
+    ;[FacebookShareButton, LinkedinShareButton, TwitterShareButton].forEach(
```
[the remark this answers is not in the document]
aliceriot: sure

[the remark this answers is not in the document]
aliceriot: so I gave that a try and, just changing all the component names to strings, my tests are failing. I think maybe the names of these components aren't set to the same thing as the same of the export (that can happen with things created with HOCs, and looking at the source the author wrote a system to create all the different icons with a HOC). I'm going to leave it as-is, since it looks like all the icons have the same name (`CreatedButton`).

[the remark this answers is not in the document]
rhysyngsun: ah ok, that's a bit odd that such a low-level component is an hoc

PR #876, opened 2018-07-05
Replace 'channel' with 'c' in URLs, redirect old URLs to new ones

### LINKED ISSUE #867: replace `channel` in page URLs with `c`
In order to shorten URLs, and allow us more flexibility about UI labelling, we'd like to obscure the "channel" in page URLs. 

No need to change API URLs.

mbertrand:
Fixes #867

#### What's this PR do?
Replaces `channel/` with `c/` in URLs.

- Navigate to various channel, post, comment URL's.  They should all start with `c/` instead of `channel/`.   Try replacing `c/` with `channel/`, you should be redirected to the `c` URL.
- Click 'View reported posts & comments' for a channel you have moderation access to.  The link should include `moderation/c/` instead of `moderation/channel` and should work properly.
- Channel management links should work with a `c` instead of `channel`, for example `http://od.odl.local:8063/manage/c/edit/<your_channel>/basic`


aliceriot:
there doesn't seem to be a redirect for going to `/moderation/channel/$CHANNEL_NAME`/, I'm not sure if that's important or not, since it's an internal-only page

aliceriot:
other than that the functionality looks good!

mbertrand:
I didn't include a redirect for the moderation or admin URL's since they should be invisible to search engines.  But I can add them in if you or @pdpinch think it would be good to have.

static/js/components/ChannelBreadcrumbs_test.js
```diff
@@ -17,7 +17,7 @@ describe("ChannelBreadcrumbs", () => {
     const [first, second] = wrapper.find(Link)
     assert.equal(first.props.to, "/")
     assert.equal(first.props.children, "Home")
-    assert.equal(second.props.to, `/channel/${channel.name}`)
+    assert.equal(second.props.to, `/c/${channel.name}`)
```
aliceriot: I think for these tests, while we're here, it would be good to replace the template literals with calls to the `channelURL` function, so if we make changes in the future we don't have to fiddle with it everywhere

static/js/components/ChannelBreadcrumbs_test.js
```diff
@@ -27,9 +27,9 @@ describe("ChannelBreadcrumbs", () => {
     const [first, second, third] = wrapper.find(Link)
     assert.equal(first.props.to, "/")
     assert.equal(first.props.children, "Home")
-    assert.equal(second.props.to, `/channel/${channel.name}`)
+    assert.equal(second.props.to, `/c/${channel.name}`)
     assert.equal(second.props.children, channel.title)
-    assert.equal(third.props.to, `/moderation/channel/${channel.name}`)
+    assert.equal(third.props.to, `/moderation/c/${channel.name}`)
```
aliceriot: same thing here for the `channelModerationURL` functoin

static/js/components/ChannelBreadcrumbs_test.js
aliceriot: (and in the other test files where similar changes were made too)

Issue #864, opened 2018-07-03
Channel settings: access control for moderators and contributors


pdpinch:
As a channel owner and moderator, I need to be able to add and remove moderators and contributors (where appropriate) and also manage the banned user list. 

See https://invis.io/JUMDQ9SSWAZ#/303979447_Settings__Channel_Roles and https://zpl.io/2ZJLM4q

- Public channels:
One tab : Moderators
Anyone with an MIT OPEN account is a contributor
- Private channels + Restricted channels
Two tabs:  Moderators + Contributors
Only contributors can post
- Special case :  Channels managed via MM
Maybe show list of Moderators + Contributors  but  no "add" / "remove"
Add a note saying that these are managed via MM

### acceptance criteria
- [ ] Add members tab to channel settings
- [ ] Restricted channels have moderators, contributors (and banned list)
- [ ] don't allow changes to MicroMasters private channels
- [ ] add user by email address 
- [ ] error message if no user matches email address
- [ ] use existing REST APIs for listing, adding moderator and contributors

### questions
- How should we prevent changes to MicroMasters channels?
- Do we need pagination? 

### out of scope
- banned user list
- REST API for banning users

pdpinch:
I edited the description to make it clearer which tabs go on which channel types, and that we won't be doing banned users in this issue: 

- Public channels:
One tab : Moderators
Anyone with an MIT OPEN account is a contributor
- Private channels + Protected channels
Two tabs:  Moderators + Contributors
Only contributors can post
- Special case :  Channels managed via MM
Maybe show list of Moderators + Contributors  but  no "add" / "remove"
Add a note saying that these are managed via MM

PR #881, opened 2018-07-06
Add membership field to Channel and REST API serializer

noisecapella:
Part of #864 

#### What's this PR do?
Adds a `membership_is_managed` field which defaults to true for channels created via REST API.

 - Create a channel in micromasters and verify that there is a `Channel` object in discussions for that channel and `membership_is_managed` is `True`
 - View `/api/v0/channels/` and verify that the field shows up in the results


channels/serializers.py
```diff
@@ -58,20 +75,34 @@ def get_user_is_moderator(self, channel):
         """
         return bool(channel.user_is_moderator)
 
+    def get_membership_is_managed(self, channel):
+        """
+        Get membership_is_managed from the associated Channel model
+        """
+        try:
+            channel_obj = Channel.objects.get(name=channel.display_name)
```
mbertrand: I've got an older channel where name != display_name, so this line raises the DoesNotExist exception and returns `False` for `membership_is_managed` even though the actual value is `True`; is this the intended behavior?

noisecapella: As far as I know `name` should always be the same as `display_name`. When you have a moment can you show me the channel in person?

noisecapella:
Just removed the `except` clause

noisecapella:
I think all channels accessible to the end user should be accessible to the API user. A user could produce this error by typing in the URL to access a channel which exists on the reddit instance but was not created by that particular discussions server, but I feel like having the error would be preferable to just returning false on that value. Though maybe we should add this check to our permissions at some point to avoid extra 500 errors

PR #887, opened 2018-07-09
Standardize creating users and social auth

### LINKED ISSUE #841: Create user API should use an existing account if it matches on email
When MM attempts to create a new user, we want to ensure it uses an existing account and creates a PSA record for MM. Needs some testing of how to do this, but naively this would use the PSA pipeline to create the user.

### LINKED ISSUE #871: Auth: Ensure only one account per email when registering
A user can register two times with the same email address. They can register, click the link to activate the profile, then go back and register again, then fill in the profile again. If you then log out you are unable to log in with that account again.

In theory this shouldn't be happening, but there might be a particular ordering of events that is allowing this, which we need to determine.

Discovered testing #824

rhysyngsun:

Two for the price of one since they're both highly related to the refactor done here:

Fixes #871 
Fixes #841

#### What's this PR do?

- Ensures duplicate accounts don't happen (either via register or the user api)
- User API now creates `UserSocialAuth` objects (hardcoded to the MicroMasters provider)

- Test duplicate account creation
  - Go to `/register` and start registration twice for the same email (so your receive two confirmation emails)
  - Confirm clicking on the each link only allows creation of 1 account
- Test that creating a new MM user creates a `UserSocialAuth` record for them

### DIFF
--- a/authentication/pipeline/user.py
+++ b/authentication/pipeline/user.py
@@ -11,10 +11,8 @@ from authentication.exceptions import (
     RequireRegistrationException,
 )
 from authentication.utils import SocialAuthState
-from channels import api as channel_api
-from notifications import api as notifications_api
 from open_discussions.settings import SOCIAL_AUTH_SAML_IDP_ATTRIBUTE_NAME
-from profiles.models import Profile
+from profiles import api as profile_api
 from profiles.utils import update_full_name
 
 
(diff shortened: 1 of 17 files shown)


authentication/pipeline/user.py
```diff
@@ -125,12 +118,9 @@ def require_profile_update_user_via_saml(
         # No name information passed, skipping
         pass
 
-    profile, _ = Profile.objects.update_or_create(
-        user=user,
-        defaults={
-            'name': user.get_full_name()
-        },
-    )
+    profile = user.profile
```
mbertrand: Registering a new user via SAML breaks at this point (there is no existing profile).

```
web_1      |   File "./authentication/pipeline/user.py", line 121, in require_profile_update_user_via_saml
web_1      |     profile = user.profile
web_1      |   File "/usr/local/lib/python3.6/site-packages/nplusone/core/signals.py", line 23, in wrapped
web_1      |     ret = func(*args, **kwargs)
web_1      |   File "/usr/local/lib/python3.6/site-packages/django/db/models/fields/related_descriptors.py", line 407, in __get__
web_1      |     self.related.get_accessor_name()
web_1      | django.db.models.fields.related_descriptors.RelatedObjectDoesNotExist: User has no profile.
```

rhysyngsun:
@mbertrand all set!

PR #897, opened 2018-07-11
Added password change UI

### LINKED ISSUE #827: Account Settings: Change Password & Notification Settings
See https://invis.io/WZLBNGQ5CKY

### acceptance criteria
- [ ] Add tab interface to (account) settings page
- [ ] Put current notification settings under "Email Notifications" tab (keep existing styling). Note that this form is also accessed by users that are not logged in and we should preserve that behavior 
- [ ] Add "Account" tab that displays all social auths . 
- [ ] Click to change password
- [ ] Users who do not have a email social auth should not be allowed to change their password
- [ ] Change password form (see https://projects.invisionapp.com/share/WZLBNGQ5CKY#/screens/303520669)

### out of scope
- "remove" account 
- change e-mail address
- styling update

gsidebo:
Closes #827 

#### What's this PR do?
Adds the password change UI

With an email-authenticated user, go to the settings page, click the "Account" tab, click "Change Password", then use the form to change your password

Probably `static/js/containers/AccountSettingsPage.js` and `static/js/containers/PasswordChangePage.js`

The designs for the account settings page are incomplete. The output for Micromasters-authenticated users is just a placeholder for now.

#### Screenshots

ss 2018-07-11 at 10 42 48
ss 2018-07-11 at 10 42 59
ss 2018-07-11 at 10 47 45

### DIFF
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -1,19 +1,24 @@
 """Authentication views"""
 from django.conf import settings
 from django.core import mail as django_mail
-from django.contrib.auth import get_user_model
+from django.contrib.auth import get_user_model, update_session_auth_hash
 from django.shortcuts import render, redirect
 from social_core.backends.email import EmailAuth
+from social_django.models import UserSocialAuth
 from social_django.utils import load_backend
 from rest_framework import status
 from rest_framework.views import APIView
 from rest_framework.response import Response
+from rest_framework.decorators import api_view, permission_classes
+from rest_framework.permissions import IsAuthenticated
 from rest_framework_jwt.settings import api_settings
 from anymail.message import AnymailMessage
 from djoser.views import (
     PasswordResetView as DjoserPasswordResetView,
-    PasswordResetConfirmView as DjoserPasswordResetConfirmView
+    PasswordResetConfirmView as DjoserPasswordResetConfirmView,
+    SetPasswordView as DjoserSetPasswordView
 )
+from djoser.utils import ActionViewMixin
 from djoser.email import PasswordResetEmail as DjoserPasswordResetEmail
 
 from open_discussions import features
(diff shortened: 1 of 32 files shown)


gsidebo:
Marking this WIP for the following reasons:

1. After the password change form successfully submits, a message is shown on the same page, but subsequent visits to the form page still show the success message instead of showing the form again (if there was no browser refresh). We don't seem to have a clear pattern for this type of UX and the options for a direct solution don't seem very straightforward. We might want to change the approach
1. If you partially fill out the password change form, then use the Cancel link button, chrome asks if you want to save the password even though you didn't submit the form. I have some ideas for tackling that but wanted to get a PR up before coming up with a possibly-finicky solution to that minor bug
1. Still have tests + linting to do

pdpinch:
When you get a chance, can you add some screenshots to the description?

gsidebo:
Yup, was planning on doing that this AM. Coming right up...

gsidebo:
@pdpinch Screenshots added. The first one is the most important to review. A few decisions we need to make:

1. If we want to add the logos as part of this PR
1. If the answer to (1) is yes, what logo to use for Micromasters
1. What exactly to show for the Micromasters auth line. I had this showing the user email earlier, and I can add that back in. I was wondering if that would be confusing since the user didn't actually use that email to log into OD, and if they authenticate via email as well, we would be showing the email twice. That seems like it wouldn't be too jarring but thought it was worth flagging

gsidebo:
@pdpinch One other thing to consider: long email addresses on the account settings page. A long email would bump the "Change Password" link to the next line. One way to handle this would be to have the end of the email address "fade" at the end before bumping into the link. This could be accomplished with some HTML/CSS tweaks. Might be worth asking Rob/Lucas for their opinions

pdpinch:
We can do the logos later. 

For the long e-mail address, are you just talking about making the div wider?

gsidebo:
We won't want an extra-wide div for narrow browser widths (mobile in particular). I was suggesting that everything could stay on one line, but the "Change Password" link would basically hide the email address underneath with a fading effect if the email was long enough. Something like [this](https://codepen.io/martinwolf/pen/DkKgJ) (just imagine the "Change Password" link to the right of that faded portion)

authentication/views.py
```diff
@@ -137,41 +169,51 @@ def send(self, to, *args, **kwargs):
...
-            return Response({}, status=status.HTTP_200_OK)
+        Overrides CustomDjoserAPIView._action. If the settings indicate that the user
+        should not be logged out after a successful password change, the session is
+        updated. Without this explicit refresh, the user's session would be invalid
+        and they would be logged out.
+        """
+        response = super()._action(serializer)
+        if (
+                response.status_code in (status.HTTP_200_OK, status.HTTP_204_NO_CONTENT) and
+                settings.DJOSER['LOGOUT_ON_PASSWORD_CHANGE'] is False
```
mbertrand: Would this setting ever be `True`?  If not, maybe remove the check for it here?

mbertrand: And remove the setting altogether?

gsidebo: I included this just to guard against a future case where we change this setting, but looking at it now I agree with you. I'll remove this logic. I'll keep the setting just to be explicit about the behavior we want

gsidebo:
My bad. Fixed

PR #898, opened 2018-07-11
Make contributors API moderator-only and add moderator-only serializer for contributors

noisecapella:
Part of #864 

#### What's this PR do?
 - Make contributors list view accessible only to moderators, and adds full name and email to the information shown. At the moment the UI does not use the contributors API at all so this should not cause any issues.
 - Remove contributor detail GET view since there is no use case for it.
 - Add logic to detect if a user is a moderator of the channel in order to determine whether to show the email and name of moderators. For non-moderators only the username is shown.
 - Remove moderator detail GET view since there is no use case for it.
- Add code to `Api.is_moderator` to verify that the given username is not falsey. Reddit interprets a falsey username as a request for all moderator information.
 - Add explicit check for anonymous users in `is_moderator` to fix reddit issue.

Note that none of these changes should cause any user facing behavior changes.

Log in as a contributor and view `/api/v0/channels/<name>/moderators/`. You should see a list of dicts with only the username.

Log in as a moderator and view the same URL. You should now see full names and email addresses.


channels/views/moderators_test.py
```diff
@@ -22,6 +22,50 @@ def test_list_moderators(client, private_channel_and_contributor, staff_user):
     assert resp.json() == [{'moderator_name': staff_user.username}]
 
 
+def test_list_moderators_staff(  # pylint: disable=too-many-arguments
+        client, private_channel, staff_user, staff_api, reddit_factories, staff_jwt_header
+):
+    """
+    List moderators in a channel
```
mbertrand: Minor nitpick: Change to something like `List moderators in a channel as a staff user`?

channels/views/moderators_test.py
```diff
@@ -12,7 +12,7 @@
 
 def test_list_moderators(client, private_channel_and_contributor, staff_user):
     """
-    List moderators in a channel
+    List moderators in a channel as a logged in user
```
mbertrand: Minor nitpick: change to something like `List moderators in a channel as a logged in contributor`?

channels/views/moderators_test.py
```diff
@@ -22,6 +22,50 @@ def test_list_moderators(client, private_channel_and_contributor, staff_user):
...
+    assert resp.json() == [{
+        'moderator_name': mod_user.username,
+        'full_name': mod_user.profile.name,
+        'email': mod_user.email,
+    }]
+
+
+def test_list_moderators_moderator(client, private_channel, staff_user, staff_api, reddit_factories):
+    """
+    List moderators in a channel
```
mbertrand: `List moderators in a channel as a logged in moderator of that channel`

PR #901, opened 2018-07-11
Don't fetch from moderators list to check whether user is mod

### LINKED ISSUE #900: Remove calls to moderators API
In #857 `user_is_contributor` and `user_is_moderator` were added to the channels API, which originally come from reddit. We currently fetch the moderator list every time the user fetches the channels page. We should instead just check `user_is_moderator` and avoid the unnecessary API request.

noisecapella:
Fixes #900 

#### What's this PR do?
Removes requests that fetch the moderators list. To check if the user is a moderator `user_is_moderator` is used instead.

Log in as a regular user and verify that you don't see the edit channel button on the channel page. Log in as a mod and verify that you do see this button. Also verify that the moderation tools in the post page look correct.

I didn't delete the reducer or API functions because we will need this soon for the moderator list page.

### DIFF
--- a/static/js/containers/ChannelModerationPage.js
+++ b/static/js/containers/ChannelModerationPage.js
@@ -37,7 +37,6 @@ class ChannelModerationPage extends React.Component<*, *> {
 
     try {
       await dispatch(actions.channels.get(channelName))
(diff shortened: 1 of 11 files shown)


aliceriot:
functionality looks good!

PR #903, opened 2018-07-11
update submit post form

### LINKED ISSUE #870: Post :  Improve form for posting short post and links
See https://invis.io/4EMF2CMATZN#/306442580_Draft_Freeform_Start

### acceptance criteria
- [x] dropdown list of channels defaults to "current" channel
- [x] dropdown list of channels limited to channels user is allowed to post to
- [x] "write something" button only appears for channels where text posts are allowed
- [x] "share a link" button only appears for channels where link posts are allowed
- [x] "Cancel" button closes the form

### out of scope
- WYSIWYG editor

aliceriot:

closes #870 

#### What's this PR do?

This updates the submit post form to match the latest designs (see linked from the issue). There are a few changes which necessitated re-writing some of the logic.

the user interaction now looks like 'adding' a text or url component to the form, rather than select a text post or link post tab. this means that we don't default to having one of those two things selected anymore. before we always had to have one of the two selected, which meant some complicated logic (especially when dealing with changing between subreddits, which might have different posting poermissions). with the new design, though, we default to having neither of the options selected, so things got a little easier.

make sure that things work smoothly and there are not issues! especially test edge cases around starting to fill out the form, changing the subreddit selection, and so on, with a variety of different permissions settings. hopefully I prevented any weird things but I'm sure there is still a dark corner or two.

### DIFF
--- a/static/js/components/CreatePostForm_test.js
+++ b/static/js/components/CreatePostForm_test.js
@@ -1,19 +1,29 @@
-// @flow
 import React from "react"
 import { shallow } from "enzyme"
 import { assert } from "chai"
 import sinon from "sinon"
 
 import CreatePostForm from "./CreatePostForm"
-import Embedly from "./Embedly"
+import Embedly, { EmbedlyLoader } from "./Embedly"
 
 import { LINK_TYPE_LINK, LINK_TYPE_TEXT } from "../lib/channels"
 import * as channels from "../lib/channels"
 import { makeChannel } from "../factories/channels"
 import { makeArticle } from "../factories/embedly"
+import { newPostForm } from "../lib/posts"
 
 describe("CreatePostForm", () => {
   let sandbox, isTextTabSelectedStub, isLinkTypeAllowedStub
+
+  const renderPostForm = (props = {}) =>
+    shallow(
+      <CreatePostForm
+        validation={{}}
+        channels={[]}
+        postForm={newPostForm()}
+        {...props}
+      />
+    )
 
   beforeEach(() => {
     sandbox = sinon.createSandbox()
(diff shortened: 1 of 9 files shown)


noisecapella:
If I fill out only the title and I click submit without clicking 'Write something' or 'share a link', I don't see any validation error message

noisecapella:
If I type in a link in the link box sometimes it gets replaced by the embedly box, even if I'm still typing. Is the embedly box meant to replace the link text box?

aliceriot:
yeah, if you look at the designs in invision the embedly preview replaces the text box. if it's changing too fast while you're typing we can up the debounce, my thought is that most likely most users will be pasting things in though

noisecapella:
Another thing I'm noticing is that if I paste in a link which doesn't have an embedly preview there is a flicker where the embedly box appears and then disappears. Not sure how much control you have over that though

static/js/components/CreatePostForm_test.js
```diff
@@ -1,11 +1,10 @@
-// @flow
```
noisecapella: Just curious, what did flow complain about here?

aliceriot:
yeah, that's when the request is in-flight, so we have no way of knowing that the link will fail. there aren't designs currently for what to do if embedly doesn't return anything or return an error, which might be a larger question we should answer, haha

static/js/containers/CreatePostPage.js
```diff
@@ -60,34 +57,31 @@ class CreatePostPage extends React.Component<*, void> {
       })
     )
     ensureTwitterEmbedJS()
-    this.updateTabSelection()
-  }
 
-  componentDidUpdate() {
-    this.updateTabSelection()
+    window.actions = actions
```
noisecapella: Was this leftover from debugging?

aliceriot:
@noisecapella I address the `post_type` validation issue, and then also added tests for the other validation messages while I was at it

noisecapella:
The validation works now, just wondering about the text: `One of text or post tabs must be selected`. Should `tab` be something else here?

static/js/containers/CreatePostPage.js
aliceriot: haha, indeed it was

static/js/components/CreatePostForm_test.js
aliceriot: in a lot of places in the codebase we declare the props for our components like this:

```js
class Foo extends React.Component<*, void> {
  props: FooProps

 ...
}
```

it turns out that, when you type a component like that, it type-checks the props for methods *inside* the component but doesn't type check the props being passed in. So obviously this is not at all what we want. Instead, what we want to do is this:

```js
class Foo extends React.Component<FooProps, void> {
  ...
}
```

I noticed this when I was working on the detail page changes, and I've been changing it as I go through things since then. But basically, once I made that it change it started complaining about props that we missing in the test file and so on, so I decided to just turn off flow for the test file

aliceriot:
Oh yeah, that should probably be changed, haha

aliceriot:
@noisecapella sorry, I forgot to mention, there are a few design things to still work out and work on for this, I'll ping you when it's ready for review again

aliceriot:
support for 'title only' posts

aliceriot:
not exactly a design change, more a new requirement which means some backend changes and so on

pdpinch:
We could do "title only" as a follow-up.

I was under the impression that you were concerned about the scope of that.

aliceriot:
I think it will be a fairly substantial change overall, I have just been fiddling with it a little bit and haven't yet gotten it working. If we're not concerned with supporting title-only posts right now then this is ready to merge.

pdpinch:
:shipit: 

We can do the "title only" posts as a follow up, esp. since it's a change in the existing functionality.

Issue #891, opened 2018-07-09
Filter out channel creator from moderators list


noisecapella:
Using `settings.INDEXING_API_USERNAME` is probably best here, I think that user should be creating all the channels for that server

PR #906, opened 2018-07-11
Add readonly moderator and contributor tabs

noisecapella:
Part of #864 

#### What's this PR do?
Adds a members tab with moderator and contributor tabs inside it. These list the current moderators and contributors respectively. There is no UI to edit or delete moderators and contributors yet, that will be next PR.

As a moderator, click the edit channel link. Click on the members tab then each of the two tabs inside it. You should see a list of moderators and contributors.

Note that you will see a blank row for the system user which created the channel. At some point we should remove it but it's out of scope of this issue: #891

### DIFF
--- a/static/js/components/admin/EditChannelContributorsForm.js
+++ b/static/js/components/admin/EditChannelContributorsForm.js
@@ -1,0 +1,49 @@
+// @flow
+import React from "react"
+import { Link } from "react-router-dom"
+
+import Card from "../Card"
+import MembersNavbar from "./MembersNavbar"
+
+import { profileURL } from "../../lib/url"
+
+import type { ChannelContributorsForm } from "../../flow/discussionTypes"
+
+export default class EditChannelContributorsForm extends React.Component<
+  *,
+  void
+> {
+  props: {
+    form: ChannelContributorsForm
+  }
+
+  render() {
+    const { form } = this.props
+    return (
+      <form className="form contributors-form">
+        <Card>
+          <MembersNavbar channelName={form.channel.name} />
+          <div className="contributors">
+            {form.contributors.map((contributor, index) => (
+              <div key={index} className="row">
+                {contributor.full_name ? (
+                  <Link
+                    className="name"
+                    to={profileURL(contributor.contributor_name)}
+                  >
+                    {contributor.full_name}
+                  </Link>
+                ) : (
+                  <span className="name">{"<missing>"}</span>
+                )}
+                <span className="email">
+                  {contributor.email || "<missing>"}
+                </span>
+              </div>
+            ))}
+          </div>
+        </Card>
+      </form>
+    )
+  }
+}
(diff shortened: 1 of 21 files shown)


static/js/components/admin/EditChannelContributorsForm.js
```diff
@@ -0,0 +1,49 @@
...
+  void
+> {
+  props: {
+    form: ChannelContributorsForm
+  }
+
+  render() {
+    const { form } = this.props
+    return (
+      <form className="form contributors-form">
```
gsidebo: This follows from the Slack convo, but I'm not sure that all of this should be part of a form or that this data should be part of a form object. I think we should just use forms for form-like behavior. The only user inputs that I imagine will be part of this eventually are the inputs for a new contributor/moderator. If we just have delete buttons that will sit on each row for each user, those can just be buttons with onClick events. Is there some more involved user input that I'm not aware of?

static/js/components/admin/EditChannelContributorsForm.js
```diff
@@ -0,0 +1,49 @@
...
+                  {contributor.email || "<missing>"}
+                </span>
+              </div>
+            ))}
+          </div>
+        </Card>
+      </form>
+    )
+  }
+}
```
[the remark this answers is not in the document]
noisecapella: Sure, I created https://github.com/mitodl/open-discussions/issues/908 to address it later

static/js/components/admin/EditChannelContributorsForm.js
noisecapella: Are you saying I should remove the `<form />` or that we should merge the `...Form` into the `...Page` object?

gsidebo: I would say remove the form and pass the contributor/moderator data into each component as standalone arrays

noisecapella:
The form elements and types are removed and I'm passing in the data directly

static/js/components/admin/EditChannelContributorsForm.js
```diff
@@ -0,0 +1,46 @@
...
+            <div key={index} className="row">
+              {contributor.full_name ? (
+                <Link
+                  className="name"
+                  to={profileURL(contributor.contributor_name)}
+                >
+                  {contributor.full_name}
+                </Link>
+              ) : (
+                <span className="name">{"<missing>"}</span>
```
gsidebo: `<missing>` appears 8 times in this PR. Seems like a good idea to create a constant for it

static/js/containers/admin/EditChannelContributorsPage_test.js
```diff
@@ -0,0 +1,65 @@
...
+        SET_CHANNEL_DATA
+      ]
+    )
+    return wrapper.update()
+  }
+
+  it("should set the document title", async () => {
+    await renderPage()
+    assert.equal(document.title, formatTitle("Edit Channel"))
+  })
```
gsidebo: In this file and the moderators test file, I feel like you should at least check that the `Edit*Form` component is being rendered and that we're passing it a list of contributors/moderators

noisecapella:
Comments addressed

PR #914, opened 2018-07-13
Remove duplicates when adding a new moderator or contributor

noisecapella:
Part of #864 

#### What's this PR do?
Add some logic to remove duplicate moderators and contributors

No front end UI yet

### DIFF
--- a/static/js/reducers/channel_contributors.js
+++ b/static/js/reducers/channel_contributors.js
@@ -25,7 +25,11 @@ const addContributor = (
   data: Map<string, ChannelContributors>
 ): Map<string, ChannelContributors> => {
   const update = new Map(data)
-  const contributors = update.get(channelName) || []
+  let contributors = update.get(channelName) || []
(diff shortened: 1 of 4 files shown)


static/js/reducers/channel_contributors.js
```diff
@@ -25,7 +25,11 @@ const addContributor = (
   data: Map<string, ChannelContributors>
 ): Map<string, ChannelContributors> => {
   const update = new Map(data)
-  const contributors = update.get(channelName) || []
+  let contributors = update.get(channelName) || []
+  contributors = contributors.filter(
+    _contributor =>
+      _contributor.contributor_name !== contributor.contributor_name
+  )
```
gsidebo: Would it not be better to scan the list for a matching contributor (using `R.find` or something like that) and only concat if one isn't found?

noisecapella: Sure

PR #915, opened 2018-07-13
Fixed password reset UI and refactored redirect/load logic

### LINKED ISSUE #912: Fix password reset and refactor anonymous access logic
Password reset is broken in a couple ways:

1. It's not accessible by logged-out users
2. The JS email form validator (`validateEmailForm`) was changed to invalidate the form if an 'mit.edu' address is entered and show "MIT users please login with Touchstone below". There is no Touchstone link on the password reset page

Our logic to determine whether a user can access certain pages on the front end is duplicated and unclear, so a minor refactor to that logic can be done as part of this

gsidebo:
Fixes #912

#### What's this PR do?
Refactors front-end redirect/load logic (including tests) and adds the touchstone link to the password reset page to agree with the MIT address form validation error

- Use the password reset page
- Try to get to the settings page and any channel page with a non-logged-in user with the `ANONYMOUS_ACCESS` feature on and off. Make sure it redirects as expected

ss 2018-07-13 at 17 46 48

### DIFF
--- a/static/js/containers/App.js
+++ b/static/js/containers/App.js
@@ -38,6 +38,7 @@ import {
 } from "../actions/ui"
 import { setChannelData } from "../actions/channel"
 import { AUTH_REQUIRED_URL, SETTINGS_URL } from "../lib/url"
+import { isAnonAccessiblePath, needsAuthedSite } from "../lib/auth"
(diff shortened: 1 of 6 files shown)


pdpinch:
So is this still necessary?

PR #916, opened 2018-07-13
Implement adding and removing moderators and contributors

### LINKED ISSUE #864: Channel settings: access control for moderators and contributors

noisecapella:
Fixes #864 

There is a question of pagination, not sure if that's a requirement for closing the issue

#### What's this PR do?
Adds a UI  to add and remove contributors and moderators

Go to the moderators tab and try adding and removing contributors via email address. If the add fails you should get a simple validation error. If the input is blank you should see a validation error about that. If you add a user correctly (even if they already exist) it should clear the field and the list should be updated.


rhysyngsun:
I'd suggest we deal with pagination when it becomes a performance issue (e.g. we have 100s) of moderators (so probably not really anytime soon?), @pdpinch @noisecapella

static/js/containers/admin/EditChannelModeratorsPage.js
```diff
@@ -35,7 +42,37 @@ export class EditChannelModeratorsPage extends React.Component<Props> {
 const mapDispatchToProps = (dispatch: Dispatch<*>, ownProps) => ({
   loadMembers: () =>
     dispatch(actions.channelModerators.get(getChannelName(ownProps))),
-  loadChannel: () => dispatch(actions.channels.get(getChannelName(ownProps)))
+  loadChannel:  () => dispatch(actions.channels.get(getChannelName(ownProps))),
+  formValidate: (payload: FormValue<AddMemberForm>) =>
```
rhysyngsun: This is duplicating a lot of what `withForm` and friends give you for free, is there a reason to avoid using that here?

static/js/hoc/withMemberForm.js
```diff
@@ -31,21 +43,73 @@ const withMemberForm = (WrappedComponent: *) => {
...
-      const { channel, loadChannel, loadMembers, members } = this.props
+      const {
+        channel,
+        loadChannel,
+        loadMembers,
+        members,
+        beginFormEdit
+      } = this.props
       if (!channel) {
         await loadChannel()
```
rhysyngsun: Is there a reason we load the channel and then members serially here? Now that we're initializing the form after both of these it is probably worth parallelizing those loads with a `Promise.all`

static/js/components/admin/EditChannelMembersForm.js
```diff
@@ -3,32 +3,113 @@ import React from "react"
...
+  usernameGetter: (member: Member) => string,
+  addMember: Function,
+  removeMember: Function,
+  updateEmail: Function,
+  channel: Channel,
+  validation: AddMemberValidation,
+  form: AddMemberForm,
+  processing: boolean,
+  beginFormEdit: Function,
+  noun: string
```
rhysyngsun: Could you give this a more descriptive variable name? It doesn't tell me a lot about what it is and since it's being passed in via state through `withMemberForm`, it can be unclear at the page container level what a state property `noun` is supposed to be for.

static/js/components/admin/EditChannelMembersForm.js
```diff
@@ -3,32 +3,113 @@ import React from "react"
...
+                  </td>
+                ) : (
+                  <td>
+                    <span className="name">{MISSING_TEXT}</span>
+                  </td>
+                )}
+                <td>
+                  <span className="email">{member.email || MISSING_TEXT}</span>
+                </td>
+                {!editable ? null : (
```
rhysyngsun: Nit: reverse this to a `editable ? <dom> : null` so it's easier to read.

static/js/containers/admin/EditChannelContributorsPage.js
```diff
@@ -49,7 +86,9 @@ const mapStateToProps = (state, ownProps) => {
     members,
     channelName,
     processing,
-    usernameGetter: (member: any): string => member.contributor_name
+    noun:           "contributor",
+    usernameGetter: (member: any): string => member.contributor_name,
```
rhysyngsun: Nit: could just be `R.prop('contributor_name')`.

static/js/components/admin/EditChannelMembersForm.js
```diff
@@ -3,32 +3,113 @@ import React from "react"
...
+          </React.Fragment>
+        )}
+        <table>
+          <tbody>
+            {members.map((member, index) => (
+              <tr
+                key={index}
+                className={`row ${editable ? "three-cols" : "two-cols"}`}
+              >
+                {member.full_name ? (
```
rhysyngsun: Take it or leave it: could move the conditional into the `<td>`

static/js/components/admin/EditChannelMembersForm.js
```diff
@@ -3,32 +3,113 @@ import React from "react"
...
+                onChange={this.onUpdateEmail}
+                name="email"
+                placeholder={`The email of the ${noun} you want to add`}
+                value={form.email}
+              />{" "}
+              <button onClick={addMember} type="submit" disabled={processing}>
+                Submit
+              </button>
+            </div>
+            {validation ? validationMessage(validation.email) : null}
```
rhysyngsun: I don't think you should have to do this check here

noisecapella:
@rhysyngsun regarding pagination we will probably have many more contributors than moderators, though I don't think it's that important for right now since only moderators can access that list

static/js/hoc/withMemberForm.js
noisecapella: No reason, I'll change it. By the way there are other places where we do this kind of sequential requests, I made an issue to address those: https://github.com/mitodl/open-discussions/issues/966

static/js/containers/admin/EditChannelModeratorsPage.js
noisecapella: What do you mean? Can you go into more detail?

rhysyngsun: Combining `withForm` (https://github.com/mitodl/open-discussions/blob/master/static/js/hoc/withForm.js) and `configureForm` (https://github.com/mitodl/open-discussions/blob/master/static/js/lib/forms.js#L36) allows you to get rid of a lot of the boilerplate code like this around forms.

Take a look at this page as an example: https://github.com/mitodl/open-discussions/blob/master/static/js/containers/auth/LoginPage.js

noisecapella:
@rhysyngsun I finished the refactoring. What do you think?

noisecapella:
>The table columns don't have any widths specified so they shift around as you add moderators/subscribers

What should the widths be?

pdpinch:
I was expecting I could read or infer the widths from Zeplin, but I don't think it's Lucas's intention that each column be a different, fractional width. I'll ask him.

pdpinch:
I left Lucas a comment in Zeplin. I'm not sure when/if he'll see it. It looks to me like -- using the grid -- the name and email columns are 3 columns wide and the remove button is 2 columns wide.

noisecapella:
I fixed the margin issues, I'm unsure what to do for styling the error message, the submit button text or the cell widths

pdpinch:
George, you can merge this with those 3 design issues. We'll do a fix up -- if it's even necessary.

noisecapella:
@rhysyngsun Anything else?

static/js/components/admin/EditChannelMembersForm.js
```diff
@@ -1,34 +1,43 @@
...
-              <Link className="name" to={profileURL(usernameGetter(member))}>
-                {member.full_name}
-              </Link>
-            ) : (
-              <span className="name">{MISSING_TEXT}</span>
-            )}
-            <span className="email">{member.email || MISSING_TEXT}</span>
-          </div>
-        ))}
+        <React.Fragment>
```
rhysyngsun: Don't need this fragment anymore.

static/js/components/admin/MembersList.js
```diff
@@ -0,0 +1,69 @@
...
+
+type Props = {
+  channel: Channel,
+  editable: boolean,
+  members: Array<Member>,
+  usernameGetter: (member: Member) => string,
+  removeMember: (channel: Channel, username: string) => Promise<*>
+}
+
+export default class MembersList extends React.Component<Props> {
```
rhysyngsun: Nit: could refactor this into a stateless component since react can optimize those a bit

rhysyngsun:
Functionally this looks good.

noisecapella:
Comment addressed

PR #918, opened 2018-07-16
fix flow typing issues

aliceriot:

none

#### What's this PR do?

This fixes our flow typing for a number of components. Before, we had a lot of components typed like this:

```js
class FooBar extends React.Component<*> {
  props: FooBarProps
}
```

This isn't right! If you use Flow to get the type of this class it is `React.Component<*>` when it should be `React.Component<FooBarProps>`. This setup will type-check the usage of props *inside* the component but will *not* check that the right props are passed in to the component. This means that we're effectively not using like 50% of what Flow can give us.

We should instead be typing things like this:

```js
class FooBar extends React.Component<FooBarProps> {
}
```

this actually type-checks that the right props are passed in to components when we use them elsewhere in the codebase.

I went through and made this change almost everywhere, there were a few places that it ended up being too difficult to get it to type-check so I left it alone. 

Make sure that I didn't break anything!


static/js/components/PostList.js
```diff
@@ -5,50 +5,56 @@ import CompactPostDisplay from "./CompactPostDisplay"
...
-    posts,
-    showChannelLinks,
-    showPinUI,
-    isModerator,
-    toggleUpvote,
-    togglePinPost,
-    reportPost,
-    removePost
-  } = props
+export default class PostList extends React.Component<Props> {
```
noisecapella: Do functions as react components not work for flow typing purposes?

static/js/reducers/settings.js
```diff
@@ -2,10 +2,10 @@
 import * as api from "../lib/api"
 import { GET, PATCH } from "redux-hammock/constants"
 
-export const FREQUENCY_IMMEDIATE = "immediate"
-export const FREQUENCY_DAILY = "daily"
-export const FREQUENCY_WEEKLY = "weekly"
-export const FREQUENCY_NEVER = "never"
+export const FREQUENCY_IMMEDIATE: "immediate" = "immediate"
```
noisecapella: This is to tell flow that it can use these as constants across module boundaries? Should we do this for any of the other constants?

static/js/components/PostList.js
aliceriot: Oh yeah, they do just fine - I refactored this component into a class in order to be able to set `defaultProps`

static/js/reducers/settings.js
aliceriot: yeah, basically what it does is make `FREQUENCY_DAILY` actually have type `"daily"` instead of type `string`. then we can use the types of the constants in a union type to have stricter checking on string constant options.

I think we should use this anywhere that we have a strictly limited set of constant strings that a variable or property could hold (for instance some of the properties in API responses might qualify). One place I used it recently was the `imageSize` prop for the profile image, which is now type-checked by flow to be `"micro" | "small" | "medium"`. I'll take a quick look for other places we could take advantage of this rn.

aliceriot: I just made a few more changes along those lines for some of the LinkType business, there may be a few more places where we can take advantage of this stuff more, but I think we can just try to make those changes going forward as we notice them

PR #940, opened 2018-07-17
Check on server that channels are not managed before letting users moderate them

noisecapella:
Part of #864 

#### What's this PR do?
Adds a check that users who are moderators can't moderate a channel which is managed by micromasters. Also some permissions refactoring

N/A


noisecapella:
Sounds good, comment addressed

PR #941, opened 2018-07-17
Refactor moderator and contributor forms

### LINKED ISSUE #908: Refactor channel editing forms to decrease duplication

noisecapella:
Fixes #908

#### What's this PR do?
Merges forms into a form which displays the same tabular member data for both, with certain props where we need to differentiate.

View the moderator and contributor pages. Nothing should break.


gsidebo:
A few thoughts: 
- I don't think we should implement the contributor/moderator components as simple pass-throughs to a more general member component. For one, that results in a component hierarchy like this:
    
    ```
    <EditChannelContributorsPage>
        <EditChannelMembersPage />
    </EditChannelContributorsPage>
    ```
    That doesn't make a lot of sense. Maybe it's better to implement this as an HOC (`withMemberForm` or something like that)? Following the current implementation you have here, the actual contributor/moderator components would probably just be empty functions. That's a little wonky, but it still gives us a single component instead of one meaningful component nested inside of a basically meaningless one. 
- Passing a string key for the reducer name makes the code less grep-friendly, and it's also a little odd to read `state[reducerName]` and `actions[reducerName]`. Not a big deal, and I don't have any great ideas off the top of my head for changing that implementation, but thought I'd flag it
- I think the `<Card>` and the members navbar should be moved out of the form component and into the container
- I think instead of the username key, the form component should just take a getter function that pulls the username from the given object (probably just `R.prop`). That's one less component that needs to know the implementation details of properly fetching the username from a member object

noisecapella:
How does the refactor look?

gsidebo:
@noisecapella This HOC is basically the same as the previous approach but it's a pass-through class instead of a pass-through function. It's kind of like a factory to dynamically generate a component. I think this is in contrast to the HOC pattern, which has the aim of defining a component via composition. These components are pretty simple and very similar right now, so the 'factory' does the job, but (1) it still yields a strange component hierarchy, and (2) it will get very hairy as the needs of each component drift. Have a look at [this file](https://github.com/mitodl/open-discussions/blob/master/static/js/containers/PasswordChangePage.js). That page, and several others, make use of the `withForm` HOC that @rhysyngsun wrote, and I think the general pattern would be useful here. Might look something like the example below. Please forgive the length - I was trying to conceptualize this myself and it was easiest just to create a rough code sketch

```javascript
// ChannelModeratorPage.js

export class ChannelModeratorPage extends React.Component<Props> { 
  render () {
    const { renderBody } = this.props
    return (
      <React.Fragment>
        <MetaTags>
          <title>{formatTitle("Edit Channel")}</title>
        </MetaTags>
        {renderBody()}
      </React.Fragment>
    )
  }
}

const mapDispatchToProps = (dispatch, ownProps) => ({
  loadMembers: () => {
    dispatch(actions.channelModerators.get(getChannelName(ownProps)))
  }
})

const mapStateToProps = state => {
  const channelName = getChannelName(ownProps)
  const members = state.channelModerators.data.get(channelName)
  const usernameGetter = R.prop('moderator_name')
  return { channelName, members }
}

export default R.compose(
  connect(mapStateToProps, mapDispatchToProps),
  withMemberForm,
  withSingleColumn("edit-channel")
)(ChannelModeratorPage)

// withMemberForm.js

const withMemberForm = ( WrappedComponent ) => {
  class withMemberForm extends React.Component {
    componentDidMount() {
      this.loadData()
    }
  
    componentDidUpdate(prevProps) {
      if (shouldLoadData(prevProps, this.props)) {
        this.loadData()
      }
    }
  
    loadData = async () => {
      const { dispatch, channel, channelName, members, loadMembers } = this.props
      if (!channel) {
        await dispatch(actions.channels.get(channelName))
      }
  
      if (!members) {
        await loadMembers()
      }
    }
    
    renderBody() {
      if (!channel || !members) { return null }
      return <div>
        <EditChannelNavbar channelName={channel.name} />
        <Card>
          <EditChannelMembersForm />
      </div>
    }

    render() {
      return <WrappedComponent {...this.props} renderBody={renderBody} />
    }
  }
}

```

Using this approach, we would have defined top level containers for both the moderator and contributor pages, they would both define their own dispatched actions and handlers, and it would be much easier to handle cases where the UI/logic needs for each component start to drift

static/js/components/admin/EditChannelMembersForm.js
```diff
@@ -0,0 +1,41 @@
...
+  members: Array<Member>,
+  usernameGetter: (member: Member) => string
+}
+
+export default class EditChannelMembersForm extends React.Component<Props> {
+  render() {
+    const { channelName, members, usernameGetter } = this.props
+    return (
+      <React.Fragment>
+        <MembersNavbar channelName={channelName} />
```
gsidebo: Still think this navbar should be moved out of the form component

static/js/containers/admin/EditChannelContributorsPage.js
```diff
@@ -1,89 +1,63 @@
...
       </React.Fragment>
     )
   }
 }
 
+const mapDispatchToProps = (dispatch: Dispatch<*>, ownProps) => ({
+  loadMembers: () =>
+    dispatch(actions.channelContributors.get(getChannelName(ownProps))),
+  loadChannel:    () => dispatch(actions.channels.get(getChannelName(ownProps))),
+  usernameGetter: (member: any): string => member.contributor_name
```
gsidebo: This doesn't dispatch any actions, so I would argue it doesn't belong here. Probably makes more sense in `mapStateToProps`

static/js/hoc/withMemberForm.js
```diff
@@ -0,0 +1,79 @@
...
+
+      if (!channel || !members) {
+        return null
+      }
+
+      return (
+        <React.Fragment>
+          <MetaTags>
+            <title>{formatTitle("Edit Channel")}</title>
+          </MetaTags>
```
gsidebo: The wrapped components already render this

noisecapella:
Comments addressed

Issue #926, opened 2018-07-17
Post form: save link to thumbnail image


pdpinch:
When a user posts a link, we should store a link to its thumbnail image for later use, like for building list views and for displaying search results

### acceptance criteria:
- [x] add a thumbnail field to the post model
- [x] When the results of the embed.ly API are returned, store a link to the thumbnail
- [x] when adding the post to the search index, include the thumbnail link
- [x] #778 When displaying the post in a list view, use the [embed.ly display API](https://embed.ly/docs/explore/display/resize) to resize the thumbnail appropriately

### out of scope
- image caching
- CDN

We considered some options for caching the image and/or the resized image, but at this time it seems simplest to just use the embedly display API, which uses a CDN to deliver resized images. If we start approaching the month limit of 25,000 display API requests, we may reconsider this.

mbertrand:
@pdpinch The url for an embedly resized image includes the API key, which should not be exposed on the front end:  `https://i.embed.ly/1/display/resize?key=<api_key>&url=<image_url>&width=<pixels>&height=<pixels>`

We could have a Django view that makes the request to embedly on the backend and passes on the response to the frontend.  This will hide the API key but add some performance overhead (client -> backend -> embedly -> backend -> client).  Or the embedly response could be saved to our own CDN (Cloudfront) and that URL could then be used instead, but this issue mentions that the CDN is out of scope for now.

pdpinch:
Embed.ly doesn't have any other authentication mechanism, like JWT? 

I could ask if there's some approach we would like.

pdpinch:
What do you mean by saving the "embedly response" ?

> Or the embedly response could be saved to our own CDN (Cloudfront) and that URL could then be used instead, but this issue mentions that the CDN is out of scope for now.

mbertrand:
embed.ly's authorization options are either [a key in the url or oauth](https://docs.embed.ly/docs/authentication).  Currently the embedly requests are [routed through the backend](https://github.com/mitodl/open-discussions/blob/master/embedly/api.py#L9), hiding the key from the frontend.   Might be able to use referrer whitelisting to safely expose the key on the frontend.

By saving the embedly response, I mean saving the resized image generated by embedly to Cloudfront and using that URL instead.

pdpinch:
FWIW, embed.ly has sample code that includes the key in front-end javascript.... 

If you turn on referer whitelisting, what does it do if there is no referer passed? (something the paranoid internet users do)

[...]

PR #944, opened 2018-07-18
Save embedly thumbnail URL's

mbertrand:
Partially addresses #926 

#### What's this PR do?
- Adds a `LinkMeta` model with attributes `url` and `thumbnail`
- Adds a nullable ForeignKey from `Post` to `LinkMeta` (but this isn't really used for anything yet).
- Sets the `Post.link_meta` foreign key value for link posts in `channels.api.sync_post_model`
- Calls `LinkMeta.objects.get_or_create` every time the `embedly.api.views.embedly_view` is invoked (ie when creating or viewing a link post)
- Calls `LinkMeta.objects.get_or_create` on `PostSerializer.create` and adds `thumbnail` to the `channels.serializers.BasePostSerializer` class.
- Adds the thumbnail (`post_link_thumbnail`) to the `search.serializers.ESPostSerializer` class so that it is included in the index.

- Create some link posts and text posts.  There should not be any obvious changes on the front end.  
- Check the REST API response for link posts and text posts (`/api/v0/posts/<id>).  There should be a `thumbnail` attribute for both, that should always be null for text posts, but equal to whatever thumbnail was returned by embedly for link posts.  (They will also be null for old link posts created before this PR, until the post detail page is viewed or a new link post is made with the same URL).
- Do an elasticsearch query on your new link posts and make sure that `post_link_thumbnail` is populated with the correct values.


channels/models.py
```diff
@@ -98,12 +99,21 @@ def __str__(self):
         return f"{self.name}"
 
 
+class LinkMeta(TimestampedModel):
+    """
+    The thumbnail embedly provides for a particular URL
+    """
+    url = URLField(unique=True, max_length=2048)
+    thumbnail = URLField(blank=True, null=True, max_length=2048)
```
noisecapella: Can you add a `__str__` method?

channels/utils.py
```diff
@@ -244,3 +245,25 @@ def get_reddit_slug(permalink):
         str: the reddit slug for a submission
     """
     return list(filter(None, permalink.split('/')))[-1]
+
+
+def get_or_create_link_meta(url):
```
noisecapella: Can you add a `@transaction.atomic` decorator here?

mbertrand: done

channels/models.py
mbertrand: done

mbertrand:
@noisecapella addressed comments and rebased with master to fix a travis failure

PR #953, opened 2018-07-20
Fix adding contributors and moderators by email

noisecapella:
Part of #864 

#### What's this PR do?
The REST API to add contributors and moderators by email wasn't working. This adds a fix for those cases and some tests which exercise the REST API view rather than just the serializer.

N/A, no UI yet

### DIFF
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -605,6 +605,14 @@ class ContributorSerializer(serializers.Serializer):
             raise ValidationError("contributor name is not a valid user")
         return {'contributor_name': value}
 
+    def validate_email(self, value):
+        """Validates the contributor email"""
+        if not isinstance(value, str):
+            raise ValidationError("email must be a string")
+        if not User.objects.filter(email__iexact=value).exists():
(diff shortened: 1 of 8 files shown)


channels/serializers.py
```diff
@@ -659,6 +667,14 @@ def validate_moderator_name(self, value):
             raise ValidationError("moderator name is not a valid user")
         return {'moderator_name': value}
 
+    def validate_email(self, value):
```
mbertrand: Maybe move the bulk of this code into a separate function called by both `Contributor.validate_email` and `Moderator.validate_email`, to avoid repetition?

```
    def validate_email(self, value):
        """Validates the contributor email"""
        return is_valid_email(value)
```

```
def is_valid_email(value):
    if not isinstance(value, str):
        raise ValidationError("email must be a string")
    if not User.objects.filter(email__iexact=value).exists():
        raise ValidationError("email does not exist")
    return {'email': value}
```

noisecapella:
Comment addressed

PR #955, opened 2018-07-20
Add membership notice and alert tab visibility based on channel type

noisecapella:
Part of #864 

#### What's this PR do?
Hides the contributors tab if the channel is public

For a public channel go to the members tab in the edit channel page. You should only see a tab for moderators.


noisecapella:
@mbertrand Thanks, I think #961 should fix that

PR #956, opened 2018-07-20
Display link post thumbnails in list view

### LINKED ISSUE #778: List View: display images for link posts
### acceptance criteria:
- [ ] store the url of thumbnail images returned by embedly service
- [ ] resize images as necessary
- [ ] consider how thumbnail images will work in search results 
- [ ] if there is no thumbnail, display nothing (e.g. if the site is password-protected)

mbertrand:
Closes #778 

#### What's this PR do?
Displays a thumbnail at the top right of each link post card.  The thumbnail is cropped and sized via the Embedly API (103 x 201 pixels as in [Zeplin design](https://app.zeplin.io/project/5b181d3bf71408ae4d406b94)).  CSS is used to crop further to 103x94 pixels in mobile/phone mode.

Create a mix of text and list posts in a channel and then view the channel's posts list.  Each link post should display a thumbnail, 103x201 pixels.  If you reduce the width of your browser window enough, the thumbnail should shrink to 103x94 pixels.  Clicking on the thumbnail should open the link in a new tab.

https://app.zeplin.io/project/5b181d3bf71408ae4d406b94

Wide browser:
screen shot 2018-07-23 at 2 08 38 pm

Narrow browser/mobile:
screen shot 2018-07-23 at 2 08 50 pm


pdpinch:
Couple questions about mobile view:

1. Was there a Zeplin design for the mobile view that specified the 94x94 dimensions?

2. Why do we do the mobile resize and crop in CSS instead of actually giving the mobile user a smaller file size?

pdpinch:
FYI @mbertrand we decided that when there is not thumbnail, we'd like to display nothing.

I'm not sure what is the easiest way to do that -- perhaps a 1 pixel transparent PNG?

mbertrand:
@pdpinch a transparent 1px pixel makes sense

pdpinch:
@mbertrand looks like we don't have a mobile design ready yet, so just make sure you're following the size in https://zpl.io/agnKAMO -- we'll update it later if necessary

mbertrand:
@pdpinch I was basing the 94 pixel mobile width on [this design](https://app.zeplin.io/project/5b181d3bf71408ae4d406b94/screen/5b22b6eee093e5262bdf4f58)

aliceriot:
I'm thinking we might want to go to the smaller display size at wider screen widths - I ran into a few things like this with long titles:

weird_width_mixup

longtitle

(these are both on the channel page, so the card doesn't fill the whole width of the screen b/c of the sidebar). what do you think?

static/js/lib/url_test.js
```diff
@@ -207,4 +208,18 @@ describe("url helper functions", () => {
...
+      const height = 100
+      const width = 200
+      const url = "http://www.fake.url/fake.html"
+      const embedlyUrl = embedlyThumbnail(key, url, height, width)
+      assert.ok(embedlyUrl.includes(`height=${height}`))
+      assert.ok(embedlyUrl.includes(`width=${width}`))
+      assert.ok(embedlyUrl.includes(`key=${key}`))
+      assert.ok(embedlyUrl.includes(`url=${encodeURIComponent(url)}`))
+    })
+  })
```
aliceriot: might be good to just have a super basic test for the `blankThumbnailUrl` function too

static/scss/post-list.scss
```diff
@@ -13,3 +13,55 @@
...
+    min-width: 94px;
+    width: 94px;
+    height: 103px;
+  }
+
+  @include breakpoint(phone) {
+    min-width: 94px;
+    width: 94px;
+    height: 103px;
+  }
```
aliceriot: it looks like the stuff in `mobile` and `phone` is the same - the `mobile` breakpoint should make the `phone` breakpoint redundant here (the `mobile` one is `max-width: 800px` and the `phone` is `max-width: 580px`, so `mobile` will cover all the screen widths that `phone` will)

static/scss/post-list.scss
```diff
@@ -13,3 +13,55 @@
...
 }
+
+.post-toprow {
+  min-height: 103px;
+  height: auto;
+  margin-bottom: 10px;
+  overflow: hidden;
+}
+
+.post-toprow-column2 {
```
aliceriot: I might suggest nesting `.post-toprow-column2` inside of `.post-toprow` so you can just call the columns something like `col-1` and `col-2` (just for readability)

static/js/components/CompactPostDisplay.js
```diff
@@ -67,28 +68,47 @@ export class CompactPostDisplay extends React.Component<Props> {
...
+          {post.url ? (
+            <div className="post-toprow-column2">
+              <div className="post-row-top-right">
+                <a href={post.url} target="_blank" rel="noopener noreferrer">
+                  <i className="material-icons open_in_new post-row-top-right overlay-icon">
+                    open_in_new
+                  </i>
+                  {post.thumbnail ? (
+                    <img
+                      ref={`img_${post.id}`}
```
aliceriot: what's the ref here for?

mbertrand: That's left over from when I was trying to track the card width via state, I'll get rid of it.

static/scss/post-list.scss
mbertrand: got rid of the `phone` breakpoint

static/scss/post-list.scss
mbertrand: done

static/js/lib/url_test.js
mbertrand: added a test

mbertrand:
@aliceriot I added `word-break: break-word;` to the `.post-toprow` css to deal with the long titles:

screen shot 2018-07-24 at 3 49 21 pm

aliceriot:
I think that helps, but I think we may still want to adjust things - this is what I'm seeing:

beeeeeepsdfasdf

at intermediate screen-widths where the drawer defaults to open things can get pretty squished. Here's an another example, with the placeholder image:

noimage

I think we need to show the smaller version of the image at that screen width (for reference it's about 1050px wide) and probably also try to figure out a way to hide the second column if we don't have a preview image, as the current solution with `blank.png` leaves a lot of whitespace

mbertrand:
@pdpinch @aliceriot maybe the smaller 94-pixel-wide thumbnail should just be used in all cases?

Ferdi:
@mbertrand  when you can't get an image from embedly,  we should not provide any image.

mbertrand:
@ferdi when embedly can't retrieve the specified image, the options are either a broken image tag or a blank image.

mbertrand:
@aliceriot I've updated the html/css to reduce the default width of `column2` for link posts to the same as the link icon (24 px) unless the link post has a thumbnail.  However, if the post does have a thumbnail, and for whatever reason that thumbnail can't be loaded (404 or whatever), embedly returns the `blank.png` image (the href url doesn't change though, that's included as the `errorurl` parameter) and it will still have the thumbnail width of 201/94.  

Not sure yet how best to handle the narrow card + long title issue (94 pixel thumbnails all the time, 94 pixels when browser width <= 1050 pixels, or some other solution).

mbertrand:
@aliceriot I think the thumbnails should behave better now (dynamically resize while staying centered on the middle of the image).

aliceriot:
yeah, I think that addresses it - I think we may want to change when we hide / show the drawer too but we can fiddle with that later.

PR #970, opened 2018-07-24
Removed front-end email suggestion

### LINKED ISSUE #932: Login/Sign up: remove email suggestion
As a user, the email suggestion is obtrusive and annoying and doesn't actually guarantee that we get a working e-mail address. We would be better off without it.

gsidebo:
Closes #932

#### What's this PR do?
What the title says

Use any email form (`/login`, `/signup`), and type in something that would normally prompt a suggestion (e.g.: `someone@m1t.edu`). There should be no suggestion and the form should have the same behavior otherwise


aliceriot:
I'm still seeing the HTML5 native email validation like this (in chrome at least):

validation_thing

I'm assuming we still want that? I don't know exactly what the functionality looked like before

static/js/components/auth/AuthEmailForm.js
```diff
@@ -21,7 +21,7 @@ const AuthEmailForm = ({
   processing,
   formError
 }: LoginFormProps) => {
-  const suggestion = validEmail(form.email) ? suggestEmail(form.email) : null
+  const suggestion = !isEmptyText(form.email) ? suggestEmail(form.email) : null
```
aliceriot: Do we still need the ternary here? Just wondering b/c it looks like `suggestEmail` returns `null` if it doesn't have a valid suggestion to offer

gsidebo:
@aliceriot @pdpinch I updated this to remove email suggestion and leave basic email validation in place

PR #973, opened 2018-07-25
Updated user api to create social auth if provider_username is present

### LINKED ISSUE #967: Update MM social auth backend to store the MM username as the UID
Update discussions `python-social-auth` implementation for MM to store the MM username as the `uid`. This means that MM will have to change to asserting who the MM user is, rather than effectively forcing the identity as a particular discussions user by setting the JWT token username to the discussions username. There might be some nuance here regarding how we currently craft our JWT token and authenticate against the DRF API with it.

rhysyngsun:
Fixes #967 

#### What's this PR do?
This PR takes a few steps to handling `UserSocialAuth` where the `uid` is the MicroMasters username:

- Updates the user/profile API to accept an optional `uid` (optional for backwards compat). If it's passed, the API creates/updates the username.
- Updates the `BaseJwtAuth` backend to more accurately process the JWT. It does so by passing the JWT payload as `response`, so that the core PSA implementation will populate `uid` based on the looking up `ID_KEY` in the response. 
- Updates the JWT token handling to look up the username if a `provider` is provided. MM will later be updated to pass this *only for user-facing JWTs*. Backend JWTs will not have this so we also preserve the old behavior both for that, and to ensure that we can time delivery of the MM changes separately.

- You could make a test call to the API, but since it's just another optional field this isn't strictly necessary
- You should still be able to authenticate via JWT to the discussions API as a user
- You'll need to test with the MM branch `nl/social_auth_tokens`. If you create a `UserSocialAuth` record for your user with `provider='micromasters', uid='$MM_USERNAME'` you should be able to update MM `OPEN_DISCUSSIONS_REDIRECT_URL` to point to `/complete/micromasters` and see that properly give you a session for that user. **NOTE:** the MM username needs to be `User.username`, this is *not* the same as `SETTINGS.user.username`.


gsidebo:
Don't know if this was an error on my end, but I followed your instructions to test and the resulting `extra_data` property of the `UserSocialAuth` record looks like this after I linked from Micromasters: `{'auth_time': 1532639760, 'email': '', 'username': 'gsidebo'}`. The `email` is supposed to be populated, right? If I just created my initial `UserSocialAuth` record incorrectly, or if that value is supposed to be blank, let me know. Everything seems to work otherwise

authentication/api.py
```diff
@@ -52,12 +52,14 @@ def create_user(username, email, profile_data=None, user_extra=None):
     return user
 
 
-def create_micromasters_social_auth(user):
+def create_or_update_micromasters_social_auth(user, uid, details):
     """
-    Creates a MicroMasters social auth for a user
+    Creates a or updates MicroMasters social auth for a user
```
gsidebo: ~~Creates a or updates~~ --> Creates or updates

authentication/api.py
```diff
@@ -68,15 +70,15 @@ def create_micromasters_social_auth(user):
...
-            'username': str(user.username),
-        }, {})
-        social.set_extra_data(extra_data)
-        return social
+        social = storage.user.create_social_auth(user, uid, MicroMastersAuth.name)
     except IntegrityError:
         # if the user already has a social auth for MM, we don't want to fail
         # so just return the existing one
-        return storage.user.get_social_auth_for_user(user, provider=MicroMastersAuth.name)
+        social = storage.user.get_social_auth_for_user(user, provider=MicroMastersAuth.name).filter(uid=uid).first()
```
gsidebo: You're not returning here anymore, so the comment above needs to be updated

rhysyngsun:
@gsidebo addressed feedback

gsidebo:
@rhysyngsun Sorry if it was unclear above. I'm still seeing an empty `email` value in my `extra_data` field for the social auth object created when I log in via Micromasters. I produced this by logging out of OD, deleting the existing Micromasters `UserSocialAuth` object, then using the discussions link in the MM dashboard.

```python
{'_state': <django.db.models.base.ModelState at 0x7f540a4562e8>,
  'extra_data': {'auth_time': 1532965804, 'email': '', 'username': 'gsidebo'},
  'id': 7,
  'provider': 'micromasters',
  'uid': 'gsidebo',
  'user_id': 37}
```

Don't know if it's relevant or if it's even expected behavior, but it appears that a new OD user was created instead of associating with my existing user that shares an email address with the MM user and was authenticated via email

rhysyngsun:
@gsidebo until https://github.com/mitodl/open-discussions-client/pull/31 is merged and MM updated you won't see email get populated since it needs to come in using the `uid` passed in the code introduced in this PR.

gsidebo:
@rhysyngsun Ah, got it

PR #975, opened 2018-07-26
Profile image improvements

mbertrand:
(Required)

#### What's this PR do?
(Required)

(Required)

(Optional)

(Optional)

(Optional)

(Optional)


pdpinch:
nit: can your commit messages be more descriptive? You can update them now or when you squash and merge.

mbertrand:
@pdpinch this isn't ready for review yet and those commits will likely disappear before I put this in ready status, and definitely when I squash and merge.  I like to push my branches to github early and often just in case my laptop dies.

profiles/views.py
```diff
@@ -26,3 +33,13 @@ class ProfileViewSet(mixins.RetrieveModelMixin, mixins.UpdateModelMixin, viewset
     serializer_class = ProfileSerializer
     queryset = Profile.objects.all()
     lookup_field = 'user__username'
+
+
+@cache_page(60*60*24)
+def initialized_avatar(request, username, size, color, bgcolor):  # pylint:disable=unused-argument
```
gsidebo: `initialized` can be a very confusing term here (it was for me at least!). Can we change this name to `name_initials_avatar_view` or something like that?

profiles/utils.py
```diff
@@ -12,21 +15,90 @@
...
+        pointer-events="auto" font-family="{font-family}"
+        font-weight="400" fill="#{color}" style="{text-style}">{text}</text>
+</svg>
+""".strip()
+SVG_TEMPLATE = re.sub(r'(\s+|\n)', ' ', SVG_TEMPLATE)
+
+DEFAULT_PROFILE_IMAGE = urljoin(settings.STATIC_URL, "images/avatar_default.png")
+
+
+def get_gravatar_image(user, image_field=None):
```
gsidebo: `generate_gravatar_image_url` is probably a better name

profiles/utils.py
```diff
@@ -173,3 +245,67 @@ def update_full_name(user, name):
...
+
+    initials = '=)'
+
+    text = text.strip()
+    if text:
+        split_text = text.split(' ')
+        if len(split_text) > 1:
+            initials = split_text[0][0] + split_text[-1][0]
+        else:
+            initials = split_text[0][0]
```
gsidebo: I think the function to derive initials from the text should be split out into a separate method. `initials = '=)'` should stay in this method and be called `default_text` or something

profiles/views.py
```diff
@@ -26,3 +33,13 @@ class ProfileViewSet(mixins.RetrieveModelMixin, mixins.UpdateModelMixin, viewset
...
+    """ View for initial avatar """
+    user = User.objects.filter(username=username).first()
```
gsidebo: We're primarily concerned with the user's profile name, right? In that case I think we should change this query to fetch the profile name only:

```python
profile_name = Profile.objects.filter(user__username=username).values_list('name', flat=True).first()
```

profiles/views_test.py
```diff
@@ -210,3 +218,39 @@ def test_patch_profile_by_user(client, logged_in_profile):
...
+        }
+    )
+    resp = client.get(url)
+    assert resp.status_code == 200
+    assert resp.__getitem__('Content-Type') == 'image/png'
+
+
+def test_initialized_avatar_fake_user(mocker, client):
+    """
+    Test that a PNG avatar image is returned for a fake user
```
gsidebo: `Test that a PNG avatar image` --> `Test that a default avatar image`

static/js/containers/ProfileImage.js
```diff
@@ -81,29 +80,20 @@ class ProfileImage extends React.Component<ProfileImageProps> {
...
 
-    const imageUrl = makeProfileImageUrl(
-      profile,
-      imageSize === "micro" || imageSize === "small"
-    )
+    const imageUrl = profile
+      ? imageSize === "micro" || imageSize === "small"
+        ? profile.profile_image_small
+        : profile.profile_image_medium
+      : defaultProfileImageUrl
```
gsidebo: Logic here is harder to follow than necessary, IMO. This double ternary and the ternary below for the profile name would be more readable like this:

```javascript
let imageUrl, profileName
if (profile) {
  imageUrl = imageSize === "micro" || imageSize === "small" ? profile.profile_image_small : profile.profile_image_medium
  profileName = profile.name
} else {
  // ...
}
```

profiles/utils.py
```diff
@@ -173,3 +245,67 @@ def update_full_name(user, name):
...
+        style_dict(dict): A dictionary of style attributes/values
+
+    Returns:
+        str: An HTML style string
+    """
+
+    return '; '.join(['{}: {}'.format(k, v) for k, v in style_dict.items()])
+
+
+def get_svg_avatar(text, size, color, bgcolor):
```
gsidebo: `get_svg_avatar` --> `generate_svg_avatar`

profiles/utils_test.py
```diff
@@ -76,21 +78,40 @@ def test_too_long_prefix(user):
...
+    profile_image = image_uri(profile.user, 'image_small')
+    assert profile_image.startswith(
+        profile.image_small_file.url if image else url if url else 'https://www.gravatar.com/avatar/'
     )
+    if not image and not url:
+        params_d = parse_qs(urlparse(profile_image).query)['d'][0]
+        if fullname:
+            assert params_d.endswith('profile/{}/64/fff/579cf9.png'.format(profile.user.username))
+        else:
+            assert params_d.endswith(DEFAULT_PROFILE_IMAGE)
```
gsidebo: The forked logic in this test makes this pretty difficult to reason about (I know some of it existed before this PR). Can you split this into as many test cases as it takes to get rid of all the booleans? 1 case each for the following scenarios (maybe more): (1) profile has an image URL, (2) profile has an image, (3) profile has neither image URL nor image. (3) could be parametrized so that you pass in the user's username, profile name, and expected `d` querystring parameter value

profiles/api_test.py
```diff
@@ -33,24 +27,8 @@ def test_ensure_profile(mocker, profile_data, no_profile):
...
         else:
-            base_img_url = 'https://www.gravatar.com/avatar/7ec7606c46a14a7ef514d1f1f9038823.jpg'
-            assert profile.image == base_img_url
-            assert profile.image_small == '{}?s={}'.format(base_img_url, IMAGE_SMALL_MAX_DIMENSION)
-            assert profile.image_medium == '{}?s={}'.format(base_img_url, IMAGE_MEDIUM_MAX_DIMENSION)
+            assert not profile.image
     else:
         assert 'gravatar.com' not in profile.image
         assert 'gravatar.com' not in profile.image_small
         assert 'gravatar.com' not in profile.image_medium
```
gsidebo: Do you know why this testing logic is still necessary (or ever was)? From what I can tell, this is just checking that some fields generated by our `ProfileFactory` don't have the string 'gravatar.com' in them.

profiles/urls.py
```diff
@@ -3,12 +3,14 @@
...
 
 router = DefaultRouter()
 router.register(r'users', UserViewSet, base_name='user_api')
 router.register(r'profiles', ProfileViewSet, base_name='profile_api')
 
 urlpatterns = [
-    url(r'^api/v0/', include(router.urls))
+    url(r'^api/v0/', include(router.urls)),
+    url(r'^profile/(?P<username>[A-Za-z0-9_]+)/(?P<size>\d+)/(?P<color>[A-Za-z0-9]+)/(?P<bgcolor>[A-Za-z0-9]+).png',
+        initialized_avatar, name="profile_avatar")
```
gsidebo: Can you change the `name` parameter to whatever you end up calling the view method, and change the underscores to slashes?

mbertrand:
@gsidebo addressed all the comments, should be ready for review again.

channels/serializers.py
```diff
@@ -149,6 +149,20 @@ def _get_user(self, instance):
...
+        """ Return a user profile if it exists, else None
+        Args:
+            instance (praw.models.Submission):
+                The post to look up the user for
+
+        Returns:
+            profiles.models.Profile: the user profile if it exists
+        """
+        user = self._get_user(instance)
+        if user and hasattr(user, 'profile'):
```
gsidebo: I don't think it's possible for a user not to have a `profile` property at all. If there's no associated profile in the db, this is just `None`, right?

profiles/urls.py
```diff
@@ -3,12 +3,15 @@
...
 router = DefaultRouter()
 router.register(r'users', UserViewSet, base_name='user_api')
 router.register(r'profiles', ProfileViewSet, base_name='profile_api')
 
 urlpatterns = [
-    url(r'^api/v0/', include(router.urls))
+    url(r'^api/v0/', include(router.urls)),
+    # The URL that gravatar will redirect to if no gravatar exists for the user (no query parameters allowed).
+    url(r'^profile/(?P<username>[A-Za-z0-9_]+)/(?P<size>\d+)/(?P<color>[A-Za-z0-9]+)/(?P<bgcolor>[A-Za-z0-9]+).png',
+        name_initials_avatar_view, name="profile_avatar")
```
gsidebo: can you change `profile_avatar` to `name-initials-avatar`?

profiles/utils.py
```diff
@@ -173,3 +245,81 @@ def update_full_name(user, name):
...
+        if len(split_text) > 1:
+            return (split_text[0][0] + split_text[-1][0]).upper()
+        else:
+            return split_text[0][0].upper()
+    return None
+
+
+def generate_svg_avatar(text, size, color, bgcolor):
+    """
+    Generate an SVG avatar based om input text.  Adopted from https://github.com/CraveFood/avinit
```
gsidebo: Nit: `om` -> `on`

profiles/utils.py
```diff
@@ -173,3 +245,81 @@ def update_full_name(user, name):
...
+    if text:
+        split_text = text.split(' ')
+        if len(split_text) > 1:
+            return (split_text[0][0] + split_text[-1][0]).upper()
+        else:
+            return split_text[0][0].upper()
+    return None
+
+
+def generate_svg_avatar(text, size, color, bgcolor):
```
gsidebo: Nit: can you rename the `text` param to `name` so we're clearer about what the function is expecting?

static/js/containers/ProfileImage.js
```diff
@@ -80,30 +79,24 @@ class ProfileImage extends React.Component<ProfileImageProps> {
...
-            ) : (
-              <img
-                src={imageUrl}
-                alt={`Profile image for ${profile ? profile.name : "anonymous"}`}
-                className={`profile-image ${imageSize}`}
-              />
-            )}
+          <img
+            src={imageUrl}
+            alt={`Profile image for ${profile ? profile.name : "anonymous"}`}
```
gsidebo: You already have a conditional above that checks for a nil profile object. You can add another variable called `profileName` and determine the value in that `if (profile)` conditional.

static/js/containers/ProfileImage_test.js
```diff
@@ -131,11 +129,14 @@ describe("ProfileImage", () => {
...
     [PROFILE_IMAGE_MEDIUM, false]
   ].forEach(([imageSize, exp]) => {
-    it(`should call makeProfileImageUrl with ${String(
-      exp
-    )} if imageSize is ${imageSize}`, () => {
-      renderProfileImage({ imageSize })
-      assert.ok(makeProfileImageUrlStub.calledWith(thatProfile, exp))
+    it(`image url should ${
+      exp ? "" : " not"
+    } be small if imageSize is ${imageSize}`, () => {
```
gsidebo: Optional change, you can take or leave: The `shouldIf` helper func can be used here. 

`image url ${shouldIf(exp)} be small if imageSize is ${imageSize}`

profiles/utils.py
```diff
@@ -12,21 +15,90 @@
...
+    if user.profile.name:
+        d_param = urljoin(
+            settings.SITE_BASE_URL,
+            '/profile/{}/{}/fff/579cf9.png'.format(
+                user.username,
+                max_dimension
+            )
+        )
+    else:
+        d_param = settings.SITE_BASE_URL + DEFAULT_PROFILE_IMAGE
```
gsidebo: Just wondering: could/should `urljoin` be used here instead of concatenating via `+`?

channels/serializers.py
gsidebo: OK. In that case, I think it would definitely be considered exceptional that a user does not have an associated profile. Given that, it's a bit more pythonic to implement this as a try/catch and return `None` when `RelatedObjectDoesNotExist` is raised

PR #981, opened 2018-07-27
prosemirror editor

### LINKED ISSUE #725: WYSIWYG editor for basic posts
As a user, I'd like to be able to author basic posts with a WYSIWYG editor. 

As a developer, I'd like to store post texts as markdown, as we do currently. 

We'll try using [prosemirror](https://prosemirror.net/) 

### acceptance criteria:
- [x] bold
- [x] italic 
- [x] links
- [x] list 
- [x] quote (likely needs add'l styling for rendering)

### related issues
- WYSIWYG editor for comments

### LINKED ISSUE #965: Disable horizontal resizing of textareas on new form posts
See attached screenshot, I don't think this is desirable

2018-07-23-160322_1920x1080_scrot

aliceriot:

closes #725 
closes #965 

#### What's this PR do?

This uses Prosemirror to implement a basic wysiwyg editor which we can use for posts and comments. The editor is basically a react wrapper (and menu) around the normal prosemirror editor. It is set-up to consume and produce Markdown, so it should drop-in to our existing use-cases without any problems. 

To start, we're just implementing a few basic editing actions (see the issue for the complete list) and if we want to get more fancy we'll come back later.

I think just play around with it and try to find bugs? Make sure that it's nice to use and all that.


static/js/components/Editor.js
```diff
@@ -0,0 +1,262 @@
...
+      </div>
+    )
+  }
+
+  renderLinkMenu = () => {
+    return <div className="link-menu">LINK</div>
+  }
+
+  formHelpers = () => {
+    const { getForm, actionCreators } = configureForm(this.uuid, () => ({
```
rhysyngsun: This is called in a few places and the values shouldn't be changing, but there's some overhead involved here with generating these new functions. Thoughts on maybe moving this into the constructor or at least `componentDidMount` to only incur this overhead once per component lifecycle?

rhysyngsun: Actually, on second thought, thoughts on splitting this and the `uuid` thing out into a separate HOC that wraps `withForm`? That way you can push some of this form-specific code out of your implementation here.

aliceriot: hm, I hadn't thought of that but it could make this a little more compact, I'll give it a go and see how I do, haha

aliceriot: I think after looking into this a little bit I'm going to just go with putting the call to `configureForm` in the constructor, just to keep things simple

gsidebo:
Here's a list of some UX things I noticed. You addressed some of these directly in retro but I'll just list them all for posterity:

**Issues with Prosemirror using their [example UI](https://prosemirror.net/), i.e.: not specific to your PR (as far as I can tell)**
1. Odd bulleting behavior - If you start a bulleted list, type something, hit enter for a new bullet, then click the bullet icon....
    - The active bullet indents a level, and repeating the cycle of typing, hitting enter, and clicking the bullet icon indents them even further. Is that desired behavior? Definitely wasn't what I was expecting.
    - Moving the cursor back up to the parent bullet hides the bullet icon and the decrease indent icon. You also can't use delete/backspace at the beginning of the line to delete the bullet. Seems odd, but tolerable I guess.
   - Hitting delete/backspace on an indented bullet with no text results in indented text. Hitting enter after that results in a bullet on the new line. Found that strange.
1. Applying both bold and italic still shows up as just italic in the WYSIWYG window for me. Don't know if this is browser+OS specific or what
ss 2018-07-30 at 17 21 37
1. When you click to create a link on some highlighted text, the text being linked is no longer highlighted, and there is no UI element showing you the text content to which the link will be applied. I feel like we should do at least one of those things.
1. You can apply links to text selections that only contain whitespace characters (normal blank space, line breaks, etc). I feel like we should prevent that.
1. Can't edit existing links. You can only remove the link and apply a new one

**Issues specific to this PR (again, as far as I can tell)**
1. Like you mentioned, exiting the Add Link UI
1. No UI to remove a link (in the Prosemirror example the link icon is highlighted if the selection contains an active link, and clicking it removes the link)
1. If you select text with any character that has a link applied and follow the UI to add a link, the result is that all text within the selection becomes unlinked.
1. Some UI elements that exist in the prosemirror example that don't exist in this PR: unindent button, heading/code formatting options, visual indicator for active bold/italic style or linked text

aliceriot:
I just added a 'close' button and also a thing for removing links in the current selection

gsidebo:
Link UI looks much better now, and yeah the bold+italic thing is looking good on this branch 👍

gsidebo:
A few UI/UX notes before I get into code particulars:

1. I said this before, but I'll mention here: If it's not possible to keep the text highlighted when applying a link, it would be helpful to display a copy of the text to which the link would be applied in the "Add Link" popup. Google Docs does this:
        ss 2018-08-03 at 14 12 31
1. Also mentioned this before, but I think it could be improved: I only see a _very_ slight difference between italicized text and bold + italicized text (OSX Chrome). Not sure if there's anything we can do about it. Maybe just increasing the font-weight?
        ss 2018-08-03 at 11 42 43
        ss 2018-08-03 at 11 42 54
1. Another "Add Link" quirk: if you (1) create a link, (2) highlight text that contains some of that link, (3) click the link icon to remove, then (4) click link icon again, you can apply a link to the previously-highlighted text even though that text selection was un-highlighted like 4 actions ago. I would argue that the text selection should be scrapped when you remove a link, and the cursor should be placed at the beginning of the original selection.
1. Weird backspace behavior with bullets/quotes - if you (1) open a quote block or start a bulleted list, (2) type something, (3) hit enter, (4) hit enter again (which removes the quote/bullet), then (5) hit backspace, that reapplies the quote/bullet. I would expect it to behave like a normal backspace.
1. Bullet icon changes
  1. If your cursor is on the first bullet of a particular indent level (let's call it a 'parent' bullet), clicking the bullet icon does nothing. It would be nice to change that button to disabled and style it as such
  1. If your cursor is on a non-'parent' bullet, that indents the bullet one level. It would be nice to change the bullet icon to a [format_indent_increase](https://material.io/tools/icons/?icon=format_indent_increase&style=baseline) icon in those cases
1. You can create a bullet inside of a quote block, but you can't create a quote block inside of a bulleted list. It would be best to disable the quote block when the cursor is in a bulleted list
1. We could use a clearer 'active' style for the bold & italic buttons. Here's a comparison of our current UI with the prosemirror example UI. In both cases, bold is applied and italic is not. The prosemirror example is much clearer about what is applied.
        ss 2018-08-03 at 11 50 02
        ss 2018-08-03 at 11 50 25

static/js/components/Editor.js
```diff
@@ -0,0 +1,310 @@
...
+
+    this.setState({ showLinkMenu: true })
+    dispatch(formBeginEdit())
+  }
+
+  closeLinkMenu = () => {
+    this.setState({ showLinkMenu: false })
+  }
+
+  updateLinkMenuForm = (e: Object) => {
```
gsidebo: Is `e` not of the type `Event`?

static/js/components/Editor.js
```diff
@@ -0,0 +1,310 @@
...
+      return true
+    }
+
+    return (
+      this.view.state.selection.empty ||
+      this.view.state.selection
+        .content()
+        .content.content.map(node => node.textContent)
+        .join("")
+        .trim() === ""
```
gsidebo: This is a beefy bit of logic. Can you explain this with a comment?

static/js/components/Editor.js
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  menuClickHandlerCreator = (command: Function) => (e: Object) => {
+    e.preventDefault()
+    command(this.view.state, this.view.dispatch, this.view)
+    this.view.focus()
+  }
+
+  renderMenuBar = () => {
+    const selectEmpty = this.selectionIsEmpty()
```
gsidebo: Nit: can you change to something like `isSelectionEmpty`?

static/js/components/Editor.js
```diff
@@ -0,0 +1,310 @@
...
+
+    const linkIcon =
```
gsidebo: Nit: can you change to something like `linkIconName`?

aliceriot:
1. I'll add a thing to show the text in the link menu, but I'm not going to implement editing it since it's fairly complicated to do so (I looked into it before). If we want that I think that it would be better to file an issue and circle back after the initial implementation merges.

2. this is the bold vs bold and italic thing on my machine:

browserthing

so I think that might be a browser stylesheet thing? not sure, on my browser the user agent stylesheet sets `<strong>` to have `font-weight: bold;` - can you check in your inspector to see what it's setting for you? 

3. I think it's fairly easy to add the link removal behavior you described, I'll give it a go.

4. tbh I haven't dug into this level of behavior since it's implemented by prosemirror itself, if you think the backspace behavior is a dealbreaker we can try to implement our own key handlers but I think there's a fairly big investment there in terms of time and code.

5. I can work on the indenting stuff, I had left that off since it wasn't mentioned in the MVP list but it probably makes sense to just do.

6. yeah that makes sense, I'll add a thing to disable it

7. yeah, one issue I ran into when working on this is that none of the behavior of these buttons or their different states is described in the mocks, so I made it all up mostly. I'll add a little grey background like that.

the link removal behavior you described might be doable, I haven't tried. I think it might involve some complicated searching around to find the borders of the link, but not having tried that I'm not sure how complicated it is, so I'll take a look and see.

static/js/components/Editor.js
```diff
@@ -0,0 +1,310 @@
...
+    onChange(defaultMarkdownSerializer.serialize(this.view.state.doc))
+  }
+
+  menuClickHandlerCreator = (command: Function) => (e: Object) => {
+    e.preventDefault()
+    command(this.view.state, this.view.dispatch, this.view)
+    this.view.focus()
+  }
+
+  renderMenuBar = () => {
```
gsidebo: Personally I like keeping the rendering methods next to each other. I'd prefer it if this function was defined right above the main `render` method but I'll leave that to you

static/js/components/Editor.js
```diff
@@ -0,0 +1,310 @@
...
+  markIsActive = (mark: Object) => {
+    // on the first render call this.view isn't there yet
+    if (!this.view) {
+      return false
+    }
+
+    const { from, $from, to, empty } = this.view.state.selection
+    return empty
+      ? !!mark.isInSet(this.view.state.storedMarks || $from.marks())
+      : this.view.state.doc.rangeHasMark(from, to, mark)
```
gsidebo: Can you include a brief comment(s) about this logic? Not clear what `empty` means and what exactly is being done as a result of that value

static/js/components/Editor.js
```diff
@@ -0,0 +1,310 @@
...
+      </div>
+    )
+  }
+}
+
+export default connect(({ forms }) => ({ forms }))(Editor)
+
+// a little helper to make it easier to drop the <Editor />
+// in to existing form-based components
+export const updateFormShim = R.curry((name, onUpdate, serialized) => {
```
gsidebo: Nit: can you call this something like `updateEditorFormShim`

static/js/components/Editor_test.js
```diff
@@ -0,0 +1,351 @@
...
+        if (markIsActive) {
+          assert(toggleMarkStub.calledWith(schema.marks.link))
+          assert(
+            doubleStub.calledWith(
+              wrapper.instance().view.state,
+              wrapper.instance().view.dispatch
+            )
+          )
+        } else {
+          assert(openLinkMenuStub.called)
```
gsidebo: I was under the impression that we prefer using sinon to assert various things about stubbed methods (e.g.: `sinon.assert.called(openLinkMenuStub)` since they're more descriptive when they fail

static/js/components/Editor_test.js
```diff
@@ -0,0 +1,351 @@
...
+      const wrapper = renderEditor()
+      assert.equal(
+        wrapper.find(".button-group").length,
+        menuBarManifest.length + 1
+      )
+      const allButtons = wrapper.find(".menu-button")
+      assert.equal(allButtons.length, R.flatten(menuBarManifest).length + 1)
+      assert.deepEqual(allButtons.map(button => button.text()), [
+        ...R.flatten(menuBarManifest).map(item => item.icon.props.children),
+        "link"
```
gsidebo: Can you create variables for these lists so it's a little more readable & clearer about what you're comparing?

static/js/components/Editor_test.js
```diff
@@ -0,0 +1,351 @@
...
+          const markIsActiveStub = sandbox
+            .stub(wrapper.instance(), "markIsActive")
+            .returns(linkActive)
+          wrapper.instance().forceUpdate()
+          assert.equal(
+            wrapper
+              .find(".menu-button")
+              .last()
+              .text(),
+            linkActive ? "link_off" : "link"
```
gsidebo: You can parameterize the expected link class

static/js/components/Editor_test.js
```diff
@@ -0,0 +1,351 @@
...
+      assert.isFalse(wrapper.state().showLinkMenu)
+    })
+
+    //
+    ;[true, false].forEach(empty => {
+      it(`should disable the link button if the selection is ${
+        empty ? "" : "not "
+      }empty`, () => {
+        const wrapper = renderEditor()
+        // console.log(wrapper.instance().view);
```
gsidebo: comment

static/js/components/Editor_test.js
```diff
@@ -0,0 +1,351 @@
...
+          wrapper.instance().view
+        )
+      )
+      assert(preventDefault.called)
+      assert(focusStub.called)
+    })
+  })
+
+  describe("menu bar", () => {
+    it("should have the things we tell it do", () => {
```
gsidebo: "should have the right buttons based on the manifest"?

static/js/components/Editor_test.js
```diff
@@ -0,0 +1,351 @@
...
+    assert.isDefined(wrapper.instance().view)
+  })
+
+  it("should use the initialValue prop, if passed", () => {})
+
+  //
+  ;[true, false].forEach(autoFocus => {
+    it(`${
+      autoFocus ? "should" : "shoudn't"
+    } call .focus on view if autoFocus = ${String(autoFocus)}`, () => {
```
gsidebo: I don't know if I like this or if anyone else likes this, but I created this silly helper function '[shouldIf](https://github.com/mitodl/open-discussions/blob/master/static/js/lib/test_utils.js#L24)' since I was doing this all the time. This could be written as: 

```javascript
`${shouldIf(autoFocus)} call .focus on view if autoFocus = ${String(autoFocus)}`
```

You can take or leave

static/scss/close-button.scss
```diff
@@ -0,0 +1,19 @@
+.close-button {
```
gsidebo: Could this just be added to `static/scss/button.scss` instead of creating a new file?

static/scss/add-link-menu.scss
```diff
@@ -0,0 +1,19 @@
+.add-link-menu {
```
gsidebo: Shouldn't we add this to `editor.scss` instead of creating a new file?

static/js/components/Editor.js
```diff
@@ -0,0 +1,310 @@
...
+  constructor(props: Props) {
+    super(props)
+    this.node = React.createRef()
+    this.uuid = makeUUID(10)
+
+    const { getForm, actionCreators } = configureForm(this.uuid, newLinkForm)
+    this.formHelpers = {
+      getForm,
+      ...actionCreators
+    }
```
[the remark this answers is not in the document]
aliceriot: yeah, the `uuid` is so that we can have more than one editor component on the page at the same time without them trying to edit the same form. unfortunately that means I need to calculate the `formHelpers` and so on dynamically at runtime, so I can't use the `mapDispatchToProps` pattern here. in general I think it's best not to even have a constructor, but in cases like this component there are things that have to be done there (like setting up refs and stuff like that).

static/js/components/Editor.js
aliceriot: I had typed this as `Object` because it gives me an error about not finding the `name` and `value` props in `target` if I set it to have type `Event`

aliceriot: ah, but it works if I do `SyntheticInputEvent`! huzzah

static/js/components/Editor_test.js
aliceriot: sure, I can use that

static/scss/add-link-menu.scss
aliceriot: I've been preferring to keep CSS specific to a particular component in separate files recently, I think it prevents CSS files from getting too big (which is a big hindrance to refactoring).

aliceriot:
@gsidebo ok I think your comments should be addressed now, lmk what you think!

static/scss/add-link-menu.scss
gsidebo: In this case, the menu is only ever relevant to the editor component. I agree that bloated CSS files are a problem broadly speaking, but I would argue that (a) the menu that is being styled is only ever relevant to the editor, (b) it's a different form of bloat to have a ton of separate files for somewhat limited definitions, and (c) we may end up finding that it's not a good pattern to couple our approach to separating CSS files with the concept of React components. I don't feel too strongly about it, so if you prefer to leave as is, no worries

static/js/components/Editor_test.js
```diff
@@ -0,0 +1,379 @@
...
+import * as pmMarkdown from "prosemirror-markdown"
+import { Provider } from "react-redux"
+
+import AddLinkMenu from "./AddLinkMenu"
+import ConnectedEditor, { Editor, newLinkForm, menuBarManifest } from "./Editor"
+
+import IntegrationTestHelper from "../util/integration_test_helper"
+import { shouldIf } from "../lib/test_utils"
+import * as utilFuncs from "../lib/util"
+import * as pmLibFuncs from "../lib/prosemirror"
```
gsidebo: Nit: I think it would be more readable just to import these with the names `util` and `prosemirrorLib` (or `pmLib`) respectively

aliceriot:
yeah, with the backspace stuff that's provided by prosemirror's starter setup, I think in order to address that we'd need to implement our own setup for handling all the keypress events. as I said above, I think that's a pretty big lift in terms of code and effort, so given that this PR is already 1300 lines I'd say that we should not worry about it unless it's a dealbreaker. IMO it's a little odd but not the biggest deal, although maybe it would be good to hear @pdpinch's thoughts on whether we should worry about it for the MVP or not.

static/js/components/Editor.js
```diff
@@ -0,0 +1,353 @@
...
+  editorState: Object,
+  showLinkMenu: boolean
+}
+
+export class Editor extends React.Component<Props, State> {
+  node: { current: null | React$ElementRef<typeof HTMLDivElement> }
+  view: Object
+  uuid: string
+  formHelpers: Object
+  dispatchTransaction: Function
```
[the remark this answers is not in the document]
aliceriot: yeah it's a bit weird, if you look you'll see that `dispatchTransaction` is a class method (rather than a class field, which we normally do). to get `this` to be bound properly I then have to bind `dispatchTransaction` in the constructor, and Flow was complaining about that property not having a type annotation on the class, so I had to add this annotation to get it to be quiet.

I had to do all that silliness in the first place so that I could intercept the `dispatchTransaction` method with a `sinon` stub. I wanted to do this to check that it was being properly passed into the `EditorView` object in a test. The issue, though, was that class field type methods (which we normally use, declaring them with an arrow function) are not defined on the class' prototype in the same way that a class method is, so there is not a way to stub them out before instantiating the react component.

With enzyme it's possible to stub out a class-field type method after instantiating the component, using `wrapper.instance().methodName` to grab a reference to the method, but I couldn't do that in this particular case because we pass `this.dispatchTransaction` to the `EditorView` instance which lives at `this.view` in the `componentDidMount` callback. This meant that, if I mounted the component, the `EditorView` object was already holding on to a reference to the un-stubbed `dispatchTransaction` method, and stubbing out the method on `.instance()` didn't have any effect. So kind of a chicken-and-egg situation, and the solution was to move to a class method, so it can be stubbed out on `Editor.prototype` before instantiating the class, so that when the `EditorView` is created it's holding on to a reference to the stubbed method, instead of the real one.

static/js/components/Editor.js
```diff
@@ -0,0 +1,353 @@
...
+    const linkIconName =
+      this.view && markIsActive(schema.marks.link, this.view.state)
+        ? "link_off"
+        : "link"
+
+    return (
+      <div className="menu-bar">
+        {menuBarManifest.map((buttonGroup, idx) => (
+          <div className="button-group" key={idx}>
+            {buttonGroup.map(({ command, title, icon, active }) => {
```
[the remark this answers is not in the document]
aliceriot: sure. I just used the immediately invoked function for fun, haha, I can change it to something else.

aliceriot:
@gsidebo ok those comments should be addressed

Issue #927, opened 2018-07-17
Channel Settings: banner image and avatar image


pdpinch:
As a channel owner, I'd like to brand my channel using a custom banner for the page header on my channel and a custom avatar. The avatar appears in the header and is also used in the navigation menu. 

See https://invis.io/SVMZO4WWCT9#/301262024_Channel_ClimateX for wireframes of navigation menu, channel header. You can get final designs from Lucas later this week. 

See https://projects.invisionapp.com/d/main#/console/14224393/303979448/preview for wireframes of channel settings. You can get final designs from Lucas later in the sprint. 

### acceptance criteria:
- [x] add channel banner and channel avatar to channel model
- [x] add forms to the channel appearance setting to upload the banner image and the avatar image

### related issues
- #1061 update the navigation to use the channel avatars
- #989 update the channel listing template to use the banner and avatar
- channels with no banner will have a white background
- channels with no avatar will have an avatar based off their initials

### questions
- [ ] what color(s) should be used for channels with no custom avatar?

pdpinch:
Moving this to In Progress

noisecapella:
Do we know what the dimensions are for the channel avatar and banner?

pdpinch:
I broke out the channel template into another issue (#989) and updated the InVision link in the description.

pdpinch:
I updated the description to move the changes to the left nav into a separate issue. 

We can close this after #983 merges.

PR #983, opened 2018-07-27
Implement uploading channel avatar and banner

noisecapella:
Part of #927 

#### What's this PR do?
Provides the UI to edit channel avatars and banners

Edit the channel avatar and banner. Note that if you do not click 'Save' the changes should be discarded. (There isn't clear indication of this at the moment.)

### DIFF
--- a/static/js/components/ImageUploaderForm.js
+++ b/static/js/components/ImageUploaderForm.js
@@ -9,6 +9,8 @@ import { validationMessage } from "../lib/validation"
 import type { FormProps } from "../flow/formTypes"
 import type { FormErrors, FormValue } from "../flow/formTypes"
 
+const uploaderBodyHeight = (): number => R.min(500, window.innerHeight * 0.44)
+
 type ImageUploaderProps<Form> = {
   description: string,
   dialogOpen: boolean,
@@ -57,10 +59,6 @@ export default class ImageUploaderForm<Form> extends React.Component<
     formValidate({ image: message })
   }
 
-  uploaderBodyHeight = (): number => {
-    return R.min(500, window.innerHeight * 0.44)
-  }
-
   dialogContents = () => {
     const {
       height,
@@ -78,7 +76,7 @@ export default class ImageUploaderForm<Form> extends React.Component<
             updatePhotoEdit={this.updatePhotoEdit}
             photo={image}
             guides={true}
-            uploaderBodyHeight={this.uploaderBodyHeight}
+            uploaderBodyHeight={uploaderBodyHeight}
             height={height}
             width={width}
           />
@@ -93,7 +91,7 @@ export default class ImageUploaderForm<Form> extends React.Component<
   spinner = () => (
     <div
       className="photo-upload-dialog photo-active-item photo-dropzone dashed-border"
-      style={{ height: this.uploaderBodyHeight() }}
+      style={{ height: uploaderBodyHeight() }}
     >
       <div className="sk-fading-circle">
         <div className="sk-circle1 sk-circle" />
@@ -119,7 +117,7 @@ export default class ImageUploaderForm<Form> extends React.Component<
       <Dropzone
         onDrop={onDrop(this.startPhotoEdit)}
         className="photo-upload-dialog photo-active-item photo-dropzone dashed-border"
-        style={{ height: this.uploaderBodyHeight() }}
+        style={{ height: uploaderBodyHeight() }}
         accept="image/*"
         onDropRejected={() => this.setPhotoError("Please select a valid photo")}
       >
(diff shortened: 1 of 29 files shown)


static/js/util/integration_test_helper.js
```diff
@@ -40,7 +40,7 @@ export default class IntegrationTestHelper {
         const stubName = `${methodName}Stub`
         this[stubName] = this.sandbox
           .stub(api, methodName)
-          .throws(new Error(`${stubName} not implemented`))
+          .throws(() => new Error(`${stubName} not implemented`))
```
noisecapella: This makes the stack trace for the error more accurate

static/js/containers/ImageUploader.js
```diff
@@ -99,6 +109,7 @@ const mapStateToProps = (state, ownProps) => {
     dialogOpen,
     processing,
     onUpdate,
+    name,
```
rhysyngsun: Shouldn't `name` be passed into the component anyway? If you need it in `mergeAndInjectProps` you can destructure it form the third arg passed there.

static/js/containers/admin/EditChannelAppearancePage.js
```diff
@@ -142,6 +174,8 @@ const mapStateToProps = (state, ownProps) => {
   }
 }
 
+export { EditChannelAppearancePage }
```
rhysyngsun: Is there a reason to export like this instead of exporting inline with the class def?

static/js/containers/ChannelBanner.js
```diff
@@ -0,0 +1,92 @@
...
+    {
+      showDialog: () => showDialog(makeDialogKey(ownProps.name))
+    },
+    dispatch
+  )
+
+export { ChannelBanner }
+
+export default connect(
+  mapStateToProps,
```
rhysyngsun: Can pass `null` here.

static/js/containers/ChannelBanner.js
```diff
@@ -0,0 +1,92 @@
...
+const mapStateToProps = () => ({})
+const mapDispatchToProps = (dispatch: Dispatch<*>, ownProps) =>
+  bindActionCreators(
+    {
+      showDialog: () => showDialog(makeDialogKey(ownProps.name))
+    },
+    dispatch
+  )
+
+export { ChannelBanner }
```
rhysyngsun: Same question as below.

static/js/containers/ChannelBanner.js
```diff
@@ -0,0 +1,92 @@
...
+                showButton={false}
+                onUpdate={onUpdate}
+                isAdd={imageUrl === defaultChannelBannerUrl}
+                description="Channel Banner"
+                width={1150}
+                height={125}
+              />
+              <a
+                onClick={e => {
+                  e.preventDefault()
```
rhysyngsun: There's a utility to handle this - `preventDefaultAndInvoke(showDialog)` would do the trick.

static/js/containers/ChannelAvatar.js
```diff
@@ -0,0 +1,101 @@
...
+const mapDispatchToProps = (dispatch: Dispatch<*>, ownProps) =>
+  bindActionCreators(
+    {
+      showDialog: () => showDialog(makeDialogKey(ownProps.name))
+    },
+    dispatch
+  )
+
+export default connect(
+  mapStateToProps,
```
rhysyngsun: Can be null.

static/js/containers/ChannelAvatar.js
```diff
@@ -0,0 +1,101 @@
...
+              width={512}
+              height={512}
+              showButton={false}
+            />
+          ) : null}
+        </div>
+        {editable ? (
+          <a
+            className="upload-avatar"
+            onClick={e => {
```
rhysyngsun: `preventDefaultAndInvoke` (see my comment below as I'm weird and am reviewing this backwards).

noisecapella:
I'll play around with the banner selection and see if I can get something more useful

noisecapella:
I addressed your code comments, still working on the UI question

static/js/components/ImageUploaderForm.js
```diff
@@ -122,6 +57,80 @@ export default class ImageUploaderForm<Form> extends React.Component<
     formValidate({ image: message })
   }
 
+  uploaderBodyHeight = (): number => {
```
rhysyngsun: This could just be a `const` now instead of a class method since it's not using `this.props` anymore.

rhysyngsun:
Looks good, aside from the backend issue I reported as #1067

PR #984, opened 2018-07-30
fix 'rich' embed display

aliceriot:

none

#### What's this PR do?

this fixes the 'rich' type embed display a little bit. basically, this just changes things so that, if the 'rich' type is an iframe, we do the same responsive iframe thingy we do for videos currently. sometimes, notably with wikipedia articles, 'rich' type embeds are *not* iframes, so we don't always want to do this.

Check that `<iframe>`-using 'rich' embeds render in a nice, full-width responsive container that looks like this:

nice_google_iframe

(google docs are a good thing to test with)

Also test that the `.iframe-container` class (and associated styling) is *not* applied for rich embeds which are not iframes. Wikipedia articles are the main example of this that I know of, but there may be others. A wikipedia article should still look something like this:

wiki_elephant

### DIFF
--- a/static/js/components/Embedly.js
+++ b/static/js/components/Embedly.js
@@ -4,6 +4,7 @@ import React from "react"
 import ContentLoader from "react-content-loader"
 
 import { urlHostname } from "../lib/url"
+import { hasIframe } from "../lib/embed"
 
 export const EmbedlyLoader = (props: Object = {}) => (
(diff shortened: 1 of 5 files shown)


mbertrand:
@aliceriot do you have an example of an iframe url with rich embeds to test this with?

aliceriot:
any public google doc should do the trick

PR #996, opened 2018-07-30
Add avatar and banner fields to serializer and models

noisecapella:
Part of #927 

#### What's this PR do?
Add fields and REST API support for the avatar and banner images for channels

No UI to test with


channels/views/channels.py
```diff
@@ -52,6 +57,9 @@ def get_serializer_context(self):
         """Context for the request and view"""
         return {
             'channel_api': self.request.channel_api,
+            'channels': {
```
mbertrand: Can you explain what the benefit is of adding `channels` to the context?

channels/serializers.py
```diff
@@ -80,9 +82,45 @@ def get_membership_is_managed(self, channel):
...
+            raise ValidationError("Expected avatar to be a file")
+        return {"avatar": value}
+
+    def validate_banner(self, value):
+        """Empty validation function, but this is required for WriteableSerializerMethodField"""
+        if not hasattr(value, 'name'):
+            raise ValidationError("Expected banner to be a file")
+        return {"banner": value}
+
+    def _get_channel(self, name):
```
mbertrand: Related to other comment, what is the benefit of having `channels` in the context and trying to get the channel from there first instead of just calling `Channel.objects.get(name=name)`?

channels/views/channels.py
noisecapella: It's to cache `Channel` objects so they aren't fetched repeatedly. Otherwise we would need to fetch it once per field for `avatar`, `banner` and `membership_is_managed`, times once per `Channel` being shown in the list

channels/serializers.py
noisecapella: The `channels` is there to cache `Channel` objects so we can fetch them all at once and reuse them rather than fetching once per field or channel.

The `try` `catch` here is because of an edge case where we are creating a new channel. The `Channel` object isn't created until some point within the `create` method. Since the context from `get_serializer_context` is created before `create` is invoked there won't be any channels in the object.

PR #1008, opened 2018-07-31
fix location of profile dot

### LINKED ISSUE #988: Incomplete profile dot is misplaced
The red dot that appears for incomplete profiles used to be positioned at the upper right edge of the profile image, now it's appearing below the image.

screen shot 2018-07-30 at 12 47 43 pm

### LINKED ISSUE #1003: Click area for the user menu dropdown needs to be expanded
The clickable area on the user avatar that reveals the user menu is not big enough. As an example, clicking the small arrow in the corner of the avatar does not reveal the menu.

aliceriot:

closes #988 
closes #1003 

#### What's this PR do?

moves the profile dot to where it should be and also increases the size of the click area (just to encompass the little dropdown basically).

check that it appears in the bottom left of the user menu, and check that clicking on the little arrow opens the user menu too.

### DIFF
--- a/static/js/components/UserMenu.js
+++ b/static/js/components/UserMenu.js
@@ -31,17 +31,18 @@ export default class UserMenu extends React.Component<Props> {
 
     return (
       <div className="user-menu">
-        <ProfileImage
-          editable={false}
-          userName={SETTINGS.username}
-          profile={profile}
-          onClick={toggleShowUserMenu}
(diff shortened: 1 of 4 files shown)


noisecapella:
I'm now seeing the dot in the lower left, is that correct? 
screenshot from 2018-07-31 14-58-07

aliceriot:
the designs are somewhat AWOL, Peter said lower left is ok for now

Issue #1012, opened 2018-07-31
Snackbar UI bugs


gsidebo:
There are several UI bugs with our snackbar component:

**1. Various UI elements are rendered above the snackbar message when they shouldn't be (avatar, reply/edit links, upvote/downvote icons, etc)**
        ![ss 2018-07-31 at 16.45.35 .png](https://images.zenhubusercontent.com/5730c094bf93871ccd9abb95/017835d7-4702-4ceb-b3b1-884a4259a3c0)
        ![ss 2018-07-31 at 16.42.12 .png](https://images.zenhubusercontent.com/5730c094bf93871ccd9abb95/09f506e7-c880-4941-80cf-bcd948eeeaa9)

**2. If the snackbar text is sufficiently long, it expands beyond its container, and some of the contents are hidden under the drawer.**
        ![ss 2018-07-31 at 16.55.25 .png](https://images.zenhubusercontent.com/5730c094bf93871ccd9abb95/c879b3f0-9250-406e-b185-47e59833c3a9)
        ![ss 2018-07-31 at 16.53.20 .png](https://images.zenhubusercontent.com/5730c094bf93871ccd9abb95/7d016ae2-23cf-47e9-bd8b-4cec5ddb21d2)

**3. When the window width is narrow enough to bring up a drawer that covers the page content with a semi-transparent overlay, opening the drawer while a snackbar message is being shown causes the snackbar to remain where it is until the page is refreshed**

PR #1013, opened 2018-07-31
Set snackbar message when posts/comments are followed/unfollowed

### LINKED ISSUE #998: Display snackbar when user follows a post
Proposed messages:

When the user follows a post:

`Followed post <post title>. You will receive email for every comment.`

When the user unfollows a post:

`Unfollowed post <post title>. You will no longer receive email about this post.`

### LINKED ISSUE #999: Display snackbar when user follows a comment
Proposed messages:

When the user follows a comment:

`Followed comments. You will receive email for every reply to this comment.`

When the user unfollows a post:

`Unfollowed comment. You will no longer receive email about replies to this comment.`

### related issues
- #998 display snackbar when user follows a post

gsidebo:
Closes #998
Closes #999 

#### What's this PR do?
What the title says

Follow/unfollow posts and comments & observe what the snackbar does

I noticed a few snackbar bugs while I was working on this. I created an issue for them here: #1012. Since the snackbar text is a bit longer than the others we have in the app, it's more likely that you'll see the bug where snackbar text is hidden underneath an open drawer

ss 2018-07-31 at 17 19 57

ss 2018-07-31 at 17 20 12


static/js/util/api_actions.js
```diff
@@ -46,13 +47,30 @@ export const removeComment = R.curry(
...
+    const newSubscribedValue = !post.subscribed
+    await dispatch(
+      actions.posts.patch(post.id, { subscribed: newSubscribedValue })
+    )
+    const message = newSubscribedValue
+      ? `Followed post ${post.title}. You will receive email for every comment.`
+      : `Unfollowed post ${
+        post.title
+      }. You will no longer receive email about this post.`
+    return dispatch(setSnackbarMessage({ message: message }))
```
aliceriot: the object here can just be `{ message }`

static/js/util/api_actions.js
```diff
@@ -46,13 +47,30 @@ export const removeComment = R.curry(
...
-      actions.comments.patch(comment.id, { subscribed: !comment.subscribed })
+  async (dispatch: Dispatch<*>, comment: CommentInTree) => {
+    const newSubscribedValue = !comment.subscribed
+    await dispatch(
+      actions.comments.patch(comment.id, { subscribed: newSubscribedValue })
     )
+    const message = newSubscribedValue
+      ? "Followed comment. You will receive email for every reply to this comment."
+      : "Unfollowed comment. You will no longer receive email about replies to this comment."
+    return dispatch(setSnackbarMessage({ message: message }))
```
aliceriot: same comment about object, also I'm curious why you're returning the call to `dispatch(setSnackbarMessage(...))` here?

gsidebo: No other reason than to keep consistency with the functions that I modified. They both returned the results of a `dispatch()` call before I modified them. If that's not important I'm happy to change it

aliceriot: well, `setSnackbarMessage` is a synchronous action, so I think `dispatch(setSnackbarMessage)` will just evaluate to `undefined`, and since this is an `async` it always returned a promise anyway. I was just wondering if there was something I was missing that you were doing with the return value. I think you could either change it or leave it, it won't make too much difference

aliceriot: actually I just checked at it returns the action! not `undefined` as I thought it did.

PR #1015, opened 2018-08-01
Touched up account settings UI and added SAML auth type

### LINKED ISSUE #922: Add basic UI for the 'saml' provider type in account settings page
Related to #897 

The account settings page currently displays something for the 'email' and 'micromasters' provider types. We have an additional provider type ('saml') that we should show something for. We may be able to use the Touchstone logo that appears in the Touchstone login buttons.

gsidebo:
Closes #922

#### What's this PR do?
Displays the 'saml' social auth provider type in the account settings page and makes some basic stylistic changes to that page based on various bits of feedback

Authenticate via Touchstone, or just hardcode a list of provider types that includes `'saml'` and return it in the `authentication.views.get_social_auth_types` view function

From a UI perspective, we decided we no longer care about the email associated with each particular auth type

ss 2018-08-01 at 14 37 20

### DIFF
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -112,17 +112,11 @@ def get_social_auth_types(request):
         UserSocialAuth
         .objects
         .filter(user=request.user)
-        .values('provider', 'user__email')
+        .values('provider')
+        .distinct()
     )
-    serialized_social_auths = [
-        {
-            'provider': social_auth['provider'],
-            'email': social_auth['user__email']
-        }
-        for social_auth in social_auths
-    ]
     return Response(
-        data=serialized_social_auths,
+        data=social_auths,
         status=status.HTTP_200_OK
     )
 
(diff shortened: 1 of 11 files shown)


authentication/views.py
```diff
@@ -112,17 +112,11 @@ def get_social_auth_types(request):
...
-    serialized_social_auths = [
-        {
-            'provider': social_auth['provider'],
-            'email': social_auth['user__email']
-        }
-        for social_auth in social_auths
-    ]
     return Response(
-        data=serialized_social_auths,
+        data=social_auths,
```
rhysyngsun: Suggestion: keep this as a list of objects in case we need to return additional properties in the future (probably a fair chance at that).

gsidebo: Yeah, makes sense. I'll just leave it as a list of objects with `provider` as the only key

gsidebo:
@rhysyngsun All comments adressed

PR #1018, opened 2018-08-02
Notify user via snackbar when URL is copied

### LINKED ISSUE #1000: Display snackbar when copying a share URL
When the user copies a URL to the clipboard for sharing

Proposed message: `Copied to clipboard`

gsidebo:
Closes #1000 

#### What's this PR do?
Notifies a user via snackbar when URL is copied from the Share popup

Copy a URL in the Share popup, which can be seen on the post detail page

ss 2018-08-02 at 14 45 17

### DIFF
--- a/static/js/components/SharePopup.js
+++ b/static/js/components/SharePopup.js
@@ -1,5 +1,7 @@
 // @flow
+import { connect } from "react-redux"
 import React from "react"
+import R from "ramda"
 import onClickOutside from "react-onclickoutside"
 import {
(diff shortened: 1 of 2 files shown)


gsidebo:
@aliceriot Since the `SharePopup` component only really needs to dispatch the `setSnackbarMessage`, to me it made sense to connect it to the store and provide that action via `mapDispatchToProps`. Maybe that isn't the right approach. Let me know what you think

PR #1019, opened 2018-08-02
Enable channel-specific google analytics tracking

### LINKED ISSUE #997: Google Analytics for each channel
As a channel owner, I'd like to be capture web traffic with Google Analytics for later review. 

Similar to what we did for MicroMasters, we can set configure a GA account for each channel. Page events for the channel page and any detail pages should be sent to the appropriate GA account. 

### acceptance criteria
- [x] add a GA UA id to the channel model
- [x] admin only UI for now
- [x] include the channel GA UA in the channel list page and on all post detail pages within the channel
- [x] confirm that page view events are recorded 
- [x] allow for multiple channels to use the same GA UA for aggregating channel analytics

mbertrand:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #997 

#### What's this PR do?
- adds a GA tracking id to the `Channel` model (`ga_tracking_id`)
- adds a simple admin page to allow manual entry of GA tracking id's for channels
- modifies the Django `index` view to include a dict of channel:tracking_id keys/values in `js_settings`
- modifies the `withTracker` React component to:
  -  initialize a GA tracker for each channel that has a tracker id
  - record a pageview with the appropriate tracker id if the current page path includes a tracked channel name.

- Turn off your ad blockers
- Set `GA_TRACKING_ID` to some value in your .env file
- Go to `admin/channels/channel/` and enter a value (real or dummy) for `ga_tracking_id` for 1+ channels
- With the network developer tab open, go to the home page.  You should see one request made to `https://www.google-analytics.com/collect` with request parameter `tid=<GA_TRACKING_ID>`
- Go to a channel page (or post page) for a channel with no `ga_tracking_id`.  Again there should be only one request made to `https://www.google-analytics.com/collect`, with request parameter `tid=<GA_TRACKING_ID>`
- Go to a channel page, post detail page, and comments page for a channel that you did enter a `ga_tracking_id` value for.  In addition to the usual GA pageview request as above, there should be one with request parameter `tid=<Channel.ga_tracking_id>` on each of these pages.
- If you happened to use real tracking id's, go to that GA account's "real-time" view and check that the data was received.


open_discussions/views.py
```diff
@@ -62,6 +62,9 @@ def index(request, **kwargs):  # pylint: disable=unused-argument
         user_email = user.email
     js_settings = {
         "gaTrackingID": settings.GA_TRACKING_ID,
+        "gaChannelTrackers": dict(
```
rhysyngsun: I have two concerns about this:
1) this list could grow large
2) this exposes the names of our channels to people who aren't members.

I'd suggest that we should consider returning this in the channel api instead, since that will be limited to channels the user can see.

mbertrand:
@rhysyngsun any tips on how to call the the channel API inside `withTracker`? As far as I can tell it doesn't have access to state, dispatch, etc.

rhysyngsun:
@mbertrand is there any way to dynamically configure react-ga as we sections of the site instead?

mbertrand:
@rhysyngsun ideally all the necessary tracking ids would be included at once during the GA initialization.  But it's also possible to do something like `ReactGA.ga('create', 'UA-000000-02','auto' , {name: 'trackerName' });` later: https://github.com/react-ga/react-ga/issues/217 

There is also a problem with using the channels API - it seems to only retrieve channels the user is subscribed to, and doesn't include public channels they are not subscribed to.  So if users arrive on a public channel with a custom tracking id via an external link, no pageviews will be recorded to that tracking id unless they happen to be subscribed to it.

@pdpinch how many channels do you estimate we may likely have with custom tracking ids? 10? 100? 1000+?

mbertrand:
@rhysyngsun I'll try out something like a `ChannelTracker` component that would be included in every channel-related container (ChannelPage, PostPage, etc) if the channel happens to have a tracking id, and see if I can get it to create the tracker and record pageviews properly that way.

mbertrand:
@rhysyngsun ready for review

open_discussions/views.py
```diff
@@ -29,7 +29,6 @@ def index(request, **kwargs):  # pylint: disable=unused-argument
     user = None
     auth = request.COOKIES.get(api_settings.JWT_AUTH_COOKIE)
     site_key = settings.OPEN_DISCUSSIONS_DEFAULT_SITE_KEY
-
     if request.user.is_authenticated:
```
rhysyngsun: This is an unrelated change, could you remove it? I also personally much prefer this spacing being here because it makes the code more readable to me and I had just added it back in a few commits ago.

static/js/components/ChannelTracker.js
```diff
@@ -0,0 +1,27 @@
...
+  channel: Channel,
+  location: Location
+}
+
+export class ChannelTracker extends React.Component<ChannelTrackerProps> {
+  componentDidMount() {
+    const { channel, location } = this.props
+    if (channel.ga_tracking_id) {
+      const trackerName = channel.ga_tracking_id.replace(/-/g, "_")
+      ReactGA.ga("create", channel.ga_tracking_id, "auto", {
```
rhysyngsun: Is there any issue repeatedly creating this or should you check to see if GA already has it first?

static/js/components/ChannelTracker.js
```diff
@@ -0,0 +1,27 @@
...
+import ReactGA from "react-ga"
+import type { Location } from "react-router"
+import type { Channel } from "../flow/discussionTypes"
+
+type ChannelTrackerProps = {
+  channel: Channel,
+  location: Location
+}
+
+export class ChannelTracker extends React.Component<ChannelTrackerProps> {
```
rhysyngsun: I think this would be better done as an HOC so it doesn't have to be included in the page's render().

static/js/hoc/withChannelTracker.js
```diff
@@ -0,0 +1,37 @@
...
+
+    loadGA() {
+      const { channel, location } = this.props
+      if (channel && channel.ga_tracking_id && !this.isTracked) {
+        const trackerName = channel.ga_tracking_id.replace(/-/g, "_")
+        ReactGA.ga("create", channel.ga_tracking_id, "auto", {
+          name: trackerName
+        })
+        ReactGA.ga(`${trackerName}.send`, "pageview", location.pathname)
+        this.isTracked = true
```
rhysyngsun: Hmmm, gotta be careful with this. I *think* this breaks the tracking if someone switches from channel A to channel B because the component never reinitializes. We ran into similar issue with `ChannelPage` and solved it with this: https://github.com/mitodl/open-discussions/blob/master/static/js/containers/ChannelPage.js#L73

static/js/hoc/withChannelTracker.js
```diff
@@ -0,0 +1,37 @@
...
+    componentDidUpdate() {
+      this.loadGA()
+    }
+
+    render() {
+      return <WrappedComponent {...this.props} />
+    }
+  }
+
+  return WithChannelTracker
```
rhysyngsun: Could you add `WithChannelTracker.WrappedComponent = ...`?

static/js/components/ChannelTracker.js
mbertrand: No method AFAIK for checking if GA already has it, but this doesn't seem to be an issue. [MM does the same thing](https://github.com/mitodl/micromasters/blob/3c6a4c73dc38e9e590e956d40aef2755cc5b1f92/ui/templates/analytics.html#L18) and in manual testing I haven't noticed any side-effects like duplicated calls.  I'd prefer to initialize all trackers at once along with the main default tracker but that would require having all subscribed + public channels available in `/util/withTracker.js`.

static/js/hoc/withChannelTracker.js
rhysyngsun: Confirmed that this definitely happens.

static/js/hoc/withChannelTracker.js
mbertrand: Not sure what the rest of that line is supposed to be?  The other HOC's are similar (`return WithSingleColumn`, `return WithPostModeration`, etc).

rhysyngsun: Here's an example of what I'm talking about: https://github.com/mitodl/open-discussions/blob/master/static/js/hoc/withPostModeration.js#L193

PR #1024, opened 2018-08-03
Improvements to moderator/contributor UI

### LINKED ISSUE #938: Handle case where user wants to remove themself as a moderator
I'm not sure what the design should be otherwise but at least we should redirect the user back to the channel page.

See https://invis.io/9EN908ZQPAX#/311617721_Channel_Settings_Members

### LINKED ISSUE #971: Provide confirmation dialogs on contributor and moderator removal
Suggest that we add confirmation dialogs to these actions since the remove buttons are visually proximate to each other and especially on mobile devices you'd want to be sure you performing the right action since it's semi-destructive.

### LINKED ISSUE #1004: Handle moderators that can’t be removed b/c they predate the current user
Moderators that predate you being added as a moderator cannot be removed by you via the reddit api. So we need to either:

- [ ] Hide the remove button if there is data in the reddit moderator data that can inform us of who you can/can't do this to
- [ ] Message an error when the remove fails for this reason

noisecapella:
Fixes #1004 
Fixes #971 
Fixes #938 

#### What's this PR do?
 - Adds "Can't remove" text if the moderator is not removable due to being older than logged in user
 - Adds confirmation dialog for removing a user
 - Changes in text when the user you're removing is yourself

Go to the moderators page. You should likely see some moderator which says "Can't remove" instead of a "Remove" link. If not, add a new moderator, log in as that moderator and view the list again.

You should also see `You are a moderator of this channel` in the list, and `Leave` instead of `Remove` in the text. Click the link and you should see a dialog "Remove yourself?"

Find a user you can remove and remove them via the dialog to confirm that behavior still works. You should also be able to dismiss the dialog without anyone getting removed.

All of this should be exactly the same for contributors

### DIFF
--- a/static/js/actions/ui.js
+++ b/static/js/actions/ui.js
@@ -4,6 +4,7 @@ import { createAction } from "redux-actions"
 export const DIALOG_REMOVE_POST = "DIALOG_REMOVE_POST"
 export const DIALOG_REMOVE_COMMENT = "DIALOG_REMOVE_COMMENT"
 export const DIALOG_PROFILE_IMAGE = "DIALOG_PROFILE_IMAGE"
+export const DIALOG_REMOVE_MEMBER = "DIALOG_REMOVE_MEMBER"
 
 export const SET_SHOW_DRAWER_DESKTOP = "SET_SHOW_DRAWER_DESKTOP"
 export const setShowDrawerDesktop = createAction(SET_SHOW_DRAWER_DESKTOP)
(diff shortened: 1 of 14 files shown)


noisecapella:
Can you try the remove link again on the moderators which aren't removable? I thought I had rebased it against the latest master before. There could be another issue which we need to address

mbertrand:
@noisecapella I cleared out my data last night, but with new users/moderators and channels plus the latest code all the mods listed as removable can be removed.  

However I noticed something strange : I removed a user from the mods list in a Chrome window, while also logged in as the removed user in a FF window, and this is what the moderators tab looked like afterward for the removed user:

screen shot 2018-08-08 at 11 56 41 am

However, the link to that page no longer appeared on the channel page for the removed user.  Probably non-mods shouldn't be allowed to load that page at all?

mbertrand:
PS Also if I click `Leave`, then manually reload the page, the same thing happens as above.

noisecapella:
Both of those cases should be handled now

PR #1026, opened 2018-08-03
Added a setting to change the default for feature flags

rhysyngsun:
N/A

#### What's this PR do?
Adds a setting (default to `False` as before) for setting the default for all feature flags.

Comment out a feature flag setting in `.env` and verify it is still enabled (default set in docker-compose.yml)


docker-compose.yml
```diff
@@ -46,6 +46,7 @@ services:
       OPEN_DISCUSSIONS_REDDIT_VALIDATE_SSL: 'False'
       OPEN_DISCUSSIONS_REDDIT_CLIENT_ID: od_client_id
       OPEN_DISCUSSIONS_REDDIT_SECRET: od_client_secret
+      OPEN_DISCUSSIONS_FEATURE_DEFAULT: 'True'
```
mbertrand: Missing an `s` in `FEATURE`, should be `OPEN_DISCUSSIONS_FEATURES_DEFAULT`.
Also value should be `${OPEN_DISCUSSIONS_FEATURES_DEFAULT:-False}` instead of `'True'`?

rhysyngsun: @mbertrand we want this to default to `True` for development environments, which are the only ones that use docker-compose.yml. If individual devs was to opt out of that, they can override it in `.env`. I'll fix the name though.

rhysyngsun:
@mbertrand feedback addressed

PR #1027, opened 2018-08-06
Added redirect to MM on login

### LINKED ISSUE #673: recommend users with MicroMasters account login with MicroMasters
As a user with an email address associated with a MicroMasters account, I should be directed to login with MicroMasters instead of creating a new account.

rhysyngsun:
Closes #673 

#### What's this PR do?
Adds a login step that prompts a user to login with MM if that's their only authentication method

- Set `SOCIAL_AUTH_MICROMASTERS_LOGIN_URL=http://mm.dev.local:8079/discussions/`
- Ensure your user only has a "micromasters" social auth 
- Go to `/login` and enter the email for this user. You should see the page and be directed to login in MM.


gsidebo:
A few issues trying to test this. When I try to log into OD via MM for the first time with one of my users (via the "View All" link in the dashboard) I get a 401 in `discussions.api.create_discussion_user`: `HTTPError: 401 Client Error: Unauthorized for url: http://docker.for.mac.localhost:8063/api/v0/users/`. The error message is just `Error creating discussion user for gsidebo3`, so I inspected the content of the `create` API call, which is: `{"detail":"Invalid signature."}`. I could dive in and see what's wrong, but I have a feeling we'll save quite a lot of time if I just wait to hear back about what you think that is.

I also want to understand what you mean by "try to log in". Does this just mean that I should try to use the email-based login form? I ask because last time I looked at the "micromasters" provider functionality, the email for the MM account was not being captured in the `UserSocialAuth` record, so I wouldn't expect an input email to match a micromasters-authenticated user.

rhysyngsun:
@gsidebo the exception is almost certainly a JWT secret mismatch between the two applications

I've updated the description. MM users are actually now getting social auth objects.

gsidebo:
`OPEN_DISCUSSIONS_JWT_SECRET` is the same for both. I'll poke around for a few minutes. If I can't figure it out I'll just manufacture a `UserSocialAuth` record for the purpose of this PR

gsidebo:
@rhysyngsun Worth mentioning: I commented out the `permission_classes` for `profiles.views.UserViewSet` in OD, and still I get the same 401 when Micromasters makes the OD client `create` request.

gsidebo:
Another note: if you have an MIT address for your Micromasters email, you will not be able to reach this redirect page due to the Touchstone validation error. I'm guessing we're OK with that (if you have an MIT address, you *should* be able to log in via Touchstone) but worth mentioning

pdpinch:
This is what I was expecting, but I do want to test it carefully on RC to make sure we're making login easier, not harder. 

> Another note: if you have an MIT address for your Micromasters email, you will not be able to reach this redirect page due to the Touchstone validation error. I'm guessing we're OK with that (if you have an MIT address, you should be able to log in via Touchstone) but worth mentioning

PR #1031, opened 2018-08-07
Move container level form code out of ProfileImage

noisecapella:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

Part of #927 

#### What's this PR do?
Refactors `ProfileImage` further to provide an `onUpdate` prop which will separate the image upload form from other forms and treat it as something more like an input instead. This will help in #927 to allow the image to be treated as a field for the rest of the form

Upload an image. Nothing should look different than before, everything should work.


noisecapella:
Made some comments

PR #1032, opened 2018-08-07
Changed unrecognized email UX to a validation message

### LINKED ISSUE #931: Login: display an error message when email isn't recognized
When a user is logging in and we don't recognize their email address, we can simply display a validation error on the login form, instead of going to another form.

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #931 

#### What's this PR do?
Changes the unrecognized email UX so a validation message is shown instead of redirecting the user to the signup page where they would click a button to continue signing up.

Enter an unrecognized email address in the login page

Back-end errors for the auth pipeline are essentially considered non-field errors. Getting the unrecognized email address to show up directly beneath the email input (as it does for normal front-end validation errors) would require some extra work. If we want to put in that extra work, that could be done in a different issue.

ss 2018-08-06 at 12 28 00

### DIFF
--- a/static/js/containers/auth/LoginPage.js
+++ b/static/js/containers/auth/LoginPage.js
@@ -29,31 +29,47 @@ import ExternalLogins from "../../components/ExternalLogins"
 
 type LoginPageProps = {
   history: Object,
-  formError: ?string
+  formError: ?string,
+  clearEndpointState: Function
 } & WithFormProps<EmailForm>
 
-export const LoginPage = ({ renderForm, formError }: LoginPageProps) => (
-  <div className="content auth-page login-page">
-    <div className="main-content">
-      <Card className="login-card">
-        <h3>Log In</h3>
-        <MetaTags>
-          <title>{formatTitle("Log In")}</title>
-        </MetaTags>
-        {renderForm({ formError })}
-        <ExternalLogins />
-        <div className="alternate-auth-link">
-          Not a member? <Link to={REGISTER_URL}>Sign up &gt;</Link>
+export class LoginPage extends React.Component<LoginPageProps> {
+  componentWillUnmount() {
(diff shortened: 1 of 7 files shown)


static/js/containers/auth/LoginPage.js
```diff
@@ -29,32 +29,48 @@ import ExternalLogins from "../../components/ExternalLogins"
...
+        <div className="main-content">
+          <Card className="login-card">
+            <h3>Log In</h3>
+            <MetaTags>
+              <title>{formatTitle("Log In")}</title>
+            </MetaTags>
+            {renderForm({ formError })}
+            <ExternalLogins />
+            <div className="alternate-auth-link">
+              Not a member? <Link to={REGISTER_URL}>Sign up &gt;</Link>
```
pdpinch: Since you're touching this line anyway, can you remove the &gt; so it's the same as https://docs.google.com/document/d/1_sD0-ng7PsYmvmmOjoTYcguygLVctx2RZqUbJsDPvR8/edit#heading=h.45j4wcfi4l54

static/js/containers/auth/LoginPage.js
```diff
@@ -29,32 +29,48 @@ import ExternalLogins from "../../components/ExternalLogins"
...
+  }
+
+  render() {
+    const { renderForm, formError } = this.props
+
+    return (
+      <div className="content auth-page login-page">
+        <div className="main-content">
+          <Card className="login-card">
+            <h3>Log In</h3>
```
pdpinch: I think we're also preferring "Login" here.

gsidebo:
@mbertrand Great catch. Fixed & added test

PR #1033, opened 2018-08-07
Added redirect for new JWT tokens to /complete/provider

### LINKED ISSUE #1029: Index pages should redirect to provider auth complete urls
If we detect a JWT with a `"provider"` key, we should redirect to `/complete/{provider}/` from the index page. Practically, this is an abstract way of supporting the MicroMasters provider JWTs without being specific to that provider yet.

rhysyngsun:
#### Pre-Flight checklist

- [x] ~Migrations~
  - [x] ~Migration is backwards-compatible with current production code~
- [x] Testing
  - [x] Code is tested
- [x] ~Settings~
  - [x] ~New settings are documented and present in `app.json`~
  - [x] ~New settings have reasonable development defaults, if applicable~

Fixes #1029 

#### What's this PR do?
This adds a redirect to /complete/micromasters/ if you have a JWT token that has the `"provider"` key on it. This causes that JWT to be redeemed for a django session so we can lookup the user correctly in the index view

- First, make sure you're on master for discussions and micromasters
- Make sure in MM, this env var does not have `/complete/micromasters/` in it: `OPEN_DISCUSSIONS_REDIRECT_URL=http://od.dev.local:8063/`
- Login to micromasters and navigate to discussions. You should see a 500 error regarding not being able to find the `User` record.
- Checkout this branch and reload the page, you should see a redirect to /complete/micromasters, then get redirected back to your original page.

### DIFF
--- a/open_discussions/views.py
+++ b/open_discussions/views.py
@@ -3,13 +3,14 @@ open_discussions views
 """
 import json
 from datetime import timedelta
+from urllib.parse import urlencode
 
 import jwt
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ImproperlyConfigured
 from django.http import Http404, HttpResponse, HttpResponsePermanentRedirect
-from django.shortcuts import render
+from django.shortcuts import render, redirect
 from django.urls import reverse
 from rest_framework_jwt.settings import api_settings
 from social_django.utils import load_strategy, load_backend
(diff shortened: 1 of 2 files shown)


mbertrand:
@rhysyngsun Getting some strange issues when testing this (new OD user being created instead of getting linked to existing user), I will try again tomorrow morning with fresh new user accounts and clear out my existing data.

PR #1070, opened 2018-08-14
Add upload_to to banner and avatar

### LINKED ISSUE #1067: Channel avatar and banner fields on backend don't specify upload_to
These fields should specify `upload_to`, otherwise every image uploaded will have the same name as the last one and will be subject to browser caching and (more critically) CDN caching. This will have the effect of making it look like the image changes did not stick to end users.

noisecapella:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Fixes #1067 

#### What's this PR do?
Adds upload_to parameter so that filenames are distinguished from each other

Upload an avatar


profiles/utils.py
```diff
@@ -132,53 +133,68 @@ def generate_filepath(filename, username, suffix=''):
...
         function:
             A function to use with upload_to to specify an upload directory and filename
     """
 
     def _upload_to(instance, filename):
         """Function passed to upload_to on an ImageField"""
-        return generate_filepath(filename, instance.user.username, suffix)
+        return generate_filepath(filename, directory_getter(instance), suffix)
 
     return _upload_to
```
rhysyngsun: I feel like this is getting too complex now, we're not passing in a `getter` for `directory_name` because we're return a function which we invoke immediately.

I think we should instead remove this `_generate_upload_to_uri` function altogether and instead directly call `generate_filepath` in the other 5 functions

profiles/utils_test.py
```diff
@@ -71,7 +71,9 @@ def test_too_long_prefix(user):
     """
     filename = '{}.jpg'.format('a' * 150)
     with pytest.raises(ValueError) as ex:
-        _generate_upload_to_uri("x" * 150)(user.profile, filename)  # pylint: disable=protected-access
+        _generate_upload_to_uri("x" * 150, lambda _profile: _profile.user.username)(
```
rhysyngsun: This should probably be testing `generate_filepath` instead.

profiles/utils.py
```diff
@@ -101,25 +101,26 @@ def image_uri(profile, image_field=IMAGE_SMALL):
...
+        directory_name (str): A directory name
         suffix(str): 'small', 'medium', or ''
 
     Returns:
         str: The filepath for the uploaded image.
     """
     name, ext = path.splitext(filename)
     timestamp = now_in_utc().replace(microsecond=0)
-    path_format = "{prefix}/{username}/{name}-{timestamp}{suffix}{ext}"
+    path_format = "{prefix}/{directory_name}/{name}-{timestamp}{suffix}{ext}"
```
rhysyngsun: `prefix` is statically `"profile"`

PR #1075, opened 2018-08-15
Use profile for IMAGE_PATH_PREFIX

noisecapella:
None

#### What's this PR do?
Use 'profile' for the image path prefix to avoid tweaking S3 settings

N/A, we should only see the result in S3

### DIFF
--- a/profiles/utils.py
+++ b/profiles/utils.py
@@ -33,6 +33,7 @@ IMAGE_FIELDS = {
 
 # This is the Django ImageField max path size
 IMAGE_PATH_MAX_LENGTH = 100
+IMAGE_PATH_PREFIX = 'profile'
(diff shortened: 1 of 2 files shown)


mbertrand:
@gschneel I wonder if the cloudfront distribution's behavior is hardcoded to look for the `/profile` path, if so adjusting that could fix the issue.  However, avatar and banner uploads still work with this PR so 👍 

screen shot 2018-08-15 at 11 24 33 am

mbertrand:
PS I think some AWS settings might still need to be configured for production? 
```
AWS_STORAGE_BUCKET_NAME
AWS_SECRET_ACCESS_KEY
AWS_ACCESS_KEY_ID
OPEN_DISCUSSIONS_USE_S3=True
```

PR #1083, opened 2018-08-16
Use embedly image api to resize thumbnails in Embedly component

### LINKED ISSUE #1022: Use embedly display api for displaying image in detail page
@Ferdi commented on [Thu Aug 02 2018](https://github.com/mitodl/micromasters/issues/4086)

Currently, in the detail page, we show the image directly from the source. Sometimes these images can be big and slow down the page. Consider using embedly display api to resize images.

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #1022 

#### What's this PR do?
Instead of displaying the embedly thumbnail url directly, makes another call to the embedly image api to resize it if necessary (max 500 or 600 pixels, based on embedly CSS styles `.photo img` and `.link .thumbnail img`).

Create some new link posts for links with images of various sizes; try direct links to images as well as links to URL's containing images.  Embedly should appear to work the same as before.  Inspect the URL of the displayed image; it should be in the form of `https://i.embed.ly/1/display/resize/?key=<key>&url=<source_url>&height=<500|600>&grow=false&errorurl=http://od.odl.local:8063/static/images/blank.png`

### DIFF
--- a/static/js/components/Embedly_test.js
+++ b/static/js/components/Embedly_test.js
@@ -9,6 +9,7 @@ import Embedly from "./Embedly"
 import { urlHostname } from "../lib/url"
 import { makeArticle, makeYoutubeVideo, makeImage } from "../factories/embedly"
 import * as embedFuncs from "../lib/embed"
+import * as urlFuncs from "../lib/url"
 
 const renderEmbedly = embedlyResponse =>
   mount(<Embedly embedly={embedlyResponse} />)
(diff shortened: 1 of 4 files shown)


static/js/components/Embedly_test.js
```diff
@@ -35,7 +35,10 @@ describe("Embedly", () => {
     const wrapper = renderEmbedly(image)
     assert.ok(wrapper.find(".photo"))
     const img = wrapper.find("img")
-    assert.equal(img.props().src, image.url)
+    assert.ok(img.props().src.includes(encodeURIComponent(image.url)))
+    assert.ok(
+      img.props().src.startsWith("https://i.embed.ly/1/display/resize/")
+    )
```
aliceriot: I might suggest, for this test, stubbing out the `embedlyImage` function and just checking that `img.props().src` is equal to whatever the return value of the stubbed function is (otherwise this test here is kind of indirectly also testing the implementation of the `embedlyImage` function)

static/js/lib/url.js
```diff
@@ -52,6 +52,11 @@ export const embedlyThumbnail = (
     url
   )}&height=${height}&width=${width}&grow=true&errorurl=${blankThumbnailUrl()}`
 
+export const embedlyImage = (key: string, url: string, height: number) =>
```
aliceriot: what do you think about calling this `embedlyResizeURL`? just something to make it a little more clear exactly what it's doing

mbertrand: Renamed to `embedlyResizeImage`

static/js/components/Embedly_test.js
mbertrand: Stubbed out

PR #1086, opened 2018-08-16
Add avatar_small and avatar_medium

### LINKED ISSUE #1077: Add avatar_small and avatar_medium fields for channel avatars
Currently the channel avatars, introduced in #928, don't have a medium or small variation. We should figure out for sure what dimensions these will have and create those thumbnails, similar to how we do it for the profiles.

Do we need to do something similar for the banners?

noisecapella:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Fixes #1077 

#### What's this PR do?
Adds `avatar_small` and `avatar_medium` fields to reduce the bandwidth of users fetching images

View and upload the channel avatar.

The migration will wipe any existing avatars. Since they aren't yet being displayed this seemed easier to me


pdpinch:
Hey George, I wasn't anticipating you working on this so quickly. I had a couple of thoughts:

1. As far as I know, we only have 2 sizes for the channel avatars. However, we haven't worked out the details of all the mobile views yet. 

2. For post thumbnails we're using a 3rd party API for handling creation and caching of different image sizes (embed.ly's display API). Would the approach in this PR work with, or work against, an API-based approach for resizing?

noisecapella:
I'm not sure what the embedly approach is for resizing, can you go into more detail?

mbertrand:
@gschneel here are links to a couple of the embedly api calls being made for link post image resizing/cropping:
https://docs.embed.ly/docs/resize : used to resize a link post image on the detail page to a max height (`height=X&grow=false`)
https://docs.embed.ly/docs/crop: used to used to resize a link post image on a list page to a specific height & width (`height=X&width=Y&grow=true`)

However, profile images are created in different sizes just like in this PR, not using embedly API for that.

noisecapella:
Ok, I don't think this approach would prevent that, since the original image is still being exposed via REST API

channels/migrations/0010_avatar_small_medium.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def nullify_avatars(apps, schema_editor):
+    """
+    Set all avatars to None. Since we don't show avatars yet
+    this avoids the work of populating avatar_small and avatar_medium
+    """
+    Channel = apps.get_model('channels', 'Channel')
+    # At the point the migration runs
+    Channel.objects.all().update(avatar_medium=None, avatar_small=None, avatar=None)
```
mbertrand: What do you think of changing this to the following, just in case someone has uploaded a channel avatar (even though it doesn't display anywhere yet):
```python
for channel in Channel.objects.filter(avatar__isnull=False):
    channel.save(update_image=True)
```

noisecapella: Since `Channel` is a [historical model](https://docs.djangoproject.com/en/2.1/topics/migrations/#historical-models) the `save` method won't be the same as the one we defined. However I can import the `make_thumbnail` method and use it here

noisecapella:
> I noticed an issue with avatars and banners 

Which server are you testing it against?

channels/migrations/0010_avatar_small_medium.py
mbertrand: Never mind, it looks like there are no avatars on production yet anyway, so this is fine.

mbertrand:
It actually happens on my local instance regardless of whether or not I have cloudfront/s3 enabled. Just realized it happens on an initial upload too, not just a re-upload - unless I click 'Save' on the channel appearance form after clicking 'Save' on the upload image dialog. This could be user error on my part, maybe?  I assumed that when I clicked 'Save' on the image upload dialog, it would be saved to the Channel object then - that's how the profile image upload works.

PR #1093, opened 2018-08-16
Make Session auth take priority over JWT tokens

### LINKED ISSUE #1100: Make session auth take priority in DRF
This will stop JWT tokens from stomping on existing django sessions, which is causing APIs to behave ambiguously.

This will allow MM to continue generating JWTs but they will effectively be ignored if you already have a session.

It should preserve existing JWT auth from MM

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1100

#### What's this PR do?

- Swaps the order of the authentication classes
- This caused a change to status code handling for unauthenticated users: status codes that used to be 401 changed to 403 as the first authentication class error handling is used if no auth is provided and a 401 does not make sense for session auth (See https://github.com/encode/django-rest-framework/issues/5968#issuecomment-386519574 and http://www.django-rest-framework.org/api-guide/authentication/#unauthorized-and-forbidden-responses)
- Added an `error_type` field to error responses that serializes the value of `exc.__class__.__name__`
- Updated frontend checks against status codes of 401 and 403 to use this new field (`PostPage.js` and `fetch_auth.js` primarily).

- Verify the app works as usual as a logged in user
  - Particularly, verify that you get the `NotAuthenticated` page if you try to view a private channel you're not a contributor on
- Verify as an unauthenticated user the app works as usual
  - Particularly, verify if you try to visit a channel/post that is private you are prompted to login

### DIFF
--- a/channels/views/channels_test.py
+++ b/channels/views/channels_test.py
@@ -9,19 +9,22 @@ from rest_framework import status
 from channels.constants import LINK_TYPE_ANY
 from channels.factories import STRATEGY_BUILD
 from channels.models import Channel
-from open_discussions.factories import UserFactory
+from open_discussions.constants import (
+    NOT_AUTHENTICATED_ERROR_TYPE,
+    PERMISSION_DENIED_ERROR_TYPE,
+)
 from open_discussions.features import ANONYMOUS_ACCESS
 
 pytestmark = pytest.mark.betamax
 
 
-def test_list_channels(client, jwt_header, private_channel_and_contributor, request):
+def test_list_channels(user_client, private_channel_and_contributor, request):
     """
     List channels the user is subscribed to
     """
     channel, _ = private_channel_and_contributor
     url = reverse('channel-list')
-    resp = client.get(url, **jwt_header)
+    resp = user_client.get(url)
     assert resp.status_code == status.HTTP_200_OK
     assert resp.json() == [
         {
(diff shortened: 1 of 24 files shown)


channels/views/channels_test.py
```diff
@@ -56,10 +55,11 @@ def test_list_channels_anonymous(client, settings, allow_anonymous):
         # Until we decide otherwise this should always be an empty list.
         assert resp.json() == []
     else:
-        assert resp.status_code == status.HTTP_401_UNAUTHORIZED
+        assert resp.status_code in (status.HTTP_401_UNAUTHORIZED, status.HTTP_403_FORBIDDEN)
+        assert resp.data['error_type'] == 'NotAuthenticated'
```
gsidebo: Can you declare constants for the HTTP status code tuple, `'NotAuthenticated'`, and `'PermissionDenied'`? At a glance it seems like it should be easy enough to make those changes via find/replace, but if it's not and it would take a while to do, don't worry about it

static/js/util/rest.js
```diff
@@ -35,6 +35,20 @@ const anySpecificError = code =>
   )
 
 export const any404Error = anySpecificError(404)
-export const any403Error = anySpecificError(403)
 export const anyErrorExcept404 = anyErrorExcept([404])
 export const anyErrorExcept404or410 = anyErrorExcept([404, 410])
+
+// the below code is for our new error_type logic and 401/403 status codes are unreliable indicators
```
gsidebo: Suggested edit: `401/403 status codes are unreliable for detecting authentication errors. Our auth endpoints return an error type in the response, and these values can be matched to that error type to properly detect authentication errors.`

rhysyngsun:
@gsidebo code feedback addressed

PR #1104, opened 2018-08-21
Implement infinite scroll

### LINKED ISSUE #1045: Infinite scroll in channel page, home page
Once we move the footer into the navigation bar (#986), we can implement an infinite scroll and remove the next/prev navigation.

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1045 

#### What's this PR do?
Replaces the pagination links on the home and channel pages with an infinite scroll

TBD: do we care about pagination in the URL?

Scroll down the home and channel pages, making sure there are enough posts to paginate. The loader that appears should look appropriate.

If you change the sort it should clear and reload the posts. Also verify that refreshing the browser with the `sort=...` in the query parameters will show the appropriately sorted posts.


aliceriot:
I think we discussed in planning that it probably made sense to keep pagination in the URL, cc @rhysyngsun

noisecapella:
At retro @pdpinch said leaving out the pagination information in the URL is fine at least for now.

Also @pdpinch I just wanted to confirm, do we care that the information on the right will be pushed down out of reach for mobile users?

static/js/containers/HomePage.js
```diff
@@ -63,6 +63,26 @@ class HomePage extends React.Component<Props> {
     dispatch(setPostData(response.posts))
   }
 
+  loadMore = async () => {
```
gsidebo: It's very clear that the home page and channel page share a lot of common functionality for displaying posts, including infinite scroll. The `loadMore` logic, as well as the `<PostList>` and `<InfiniteScroll>` rendering is basically identical. I know that the channel page and home page are very likely to be the only two components that need to show posts, so if we keep this repetition I think that would be OK, but if you see a really quick, straightforward way to move this logic into an HOC or something, I would say go for it. Worth remembering that we're repeating testing logic on top of the application logic

static/js/containers/ChannelPage_test.js
```diff
@@ -91,168 +159,248 @@ describe("ChannelPage", () => {
...
+      }
+    })
+    assert.isTrue(wrapper.props().postsForChannelProcessing)
+  })
+
+  describe("pagination", () => {
+    [true, false].forEach(hasPagination => {
+      it(`has both PostList and InfiniteScroll if there is ${
+        hasPagination ? "" : "no "
+      }pagination`, async () => {
```
gsidebo: This can be parametrized to avoid the boolean flags.

```javascript
[
  [{sort: "new"}, 1, "not null"]
  [null, 0, "null"],
].forEach([paginationParam, expectedInfScroll, paramDesc] => {
      it(`has both PostList and InfiniteScroll if pagination param is ${paramDesc}`, async () => {
        //...
        assert.equal(inner.find("InfiniteScroll").length, expectedInfScroll)
      })
}
```

static/js/containers/HomePage_test.js
```diff
@@ -23,59 +28,304 @@ describe("HomePage", () => {
...
+    [true, true, true],
+    [true, false, false],
+    [false, true, true],
+    [false, false, true]
+  ].forEach(([processing, hasPostIds, expectedLoaded]) => {
+    it(`sets loaded to ${String(expectedLoaded)} when the API request is ${
+      processing ? "" : "not "
+    }in progress and there ${
+      hasPostIds ? "are" : "aren't"
+    } post ids`, async () => {
```
gsidebo: The list of post ids can be defined outside of this function and passed in as a parameter. Also, the description could be easier to read

```javascript
;[
    [true, mockPostIds, true],
    [true, [], false],
    [false, mockPostIds, true],
    [false, [], true]
].forEach(([processing, postIds, expectedLoaded]) => {
  it(`sets loaded to ${String(expectedLoaded)} when there are ${String(postIds.length)} and API request processing=${String(processing)}
}
```

static/js/containers/HomePage_test.js
```diff
@@ -23,59 +28,304 @@ describe("HomePage", () => {
...
+      })
+      assert.equal(wrapper.props().loaded, expectedLoaded)
+    })
+  })
+
+  describe("pagination", () => {
+    [true, false].forEach(hasPagination => {
+      it(`has both PostList and InfiniteScroll if there is ${
+        hasPagination ? "" : "no "
+      }pagination`, async () => {
```
gsidebo: Same comment as above

static/js/containers/HomePage_test.js
```diff
@@ -23,59 +28,304 @@ describe("HomePage", () => {
...
+          }
+        })
+        assert.equal(inner.find("InfiniteScroll").length, hasPagination ? 1 : 0)
+        assert.equal(inner.find("PostList").length, 1)
+      })
+    })
+    ;[true, false].forEach(hasAfterUrl => {
+      it(`sets hasMore=${String(hasAfterUrl)} if it ${
+        hasAfterUrl ? "has" : "doesn't have"
+      } an after url`, async () => {
```
gsidebo: Same comment as above

static/js/lib/posts_test.js
```diff
@@ -95,4 +97,76 @@ describe("Post utils", () => {
       )
     })
   })
+
+  describe("mapPostListResponse", () => {
+    it("clears existing post ids if there is no existing pagination", () => {
+      const posts = makeChannelPostList(),
+        oldPosts = makeChannelPostList()
```
gsidebo: Please define `posts`, `oldPosts`, and the lists of `oldPosts` and `posts` ids in a `beforeEach` function

static/js/containers/HomePage.js
noisecapella: That's a good point, I'll look into it

static/js/containers/HomePage_test.js
noisecapella: I'm not sure that's possible since the posts are defined in `beforeEach` and the test name is evaluated before that

static/js/containers/HomePage_test.js
noisecapella: What is this referring to?

noisecapella: Nevermind I think I understand

static/js/containers/HomePage_test.js
gsidebo: That list (or a separate list) could be initialized outside of `beforeEach`. You could also include a small description as a parameter if that makes it more readable

static/js/hoc/withInfiniteScroll.js
```diff
@@ -0,0 +1,129 @@
...
+  toggleUpvote: (post: Post) => Promise<*>,
+  dispatch: Dispatch<any>,
+  showChannelLinks: boolean,
+  showReportPost: boolean,
+  showRemovePost: boolean,
+  showTogglePinPost: boolean,
+  showPinUI: boolean
+}
+
+const withInfiniteScroll = (WrappedComponent: Class<React.Component<*, *>>) => {
```
gsidebo: I would call this `withPostList` or `withPostDisplay` or something. The purpose of this HOC is to handle the logic of rendering a list of posts. The fact that the posts are navigable via infinite scroll is an implementation detail

static/js/containers/HomePage_test.js
```diff
@@ -1,83 +1,191 @@
...
-    }
+  it("sets props used by withPostList", async () => {
+    const { wrapper } = await render()
+
+    assert.isFalse(wrapper.props().showPinUI)
+    assert.isTrue(wrapper.props().showReportPost)
+    assert.isFalse(wrapper.props().showRemovePost)
+    assert.isFalse(wrapper.props().showTogglePinPost)
+    assert.isTrue(wrapper.props().showChannelLinks)
+    assert.isFalse(wrapper.props().isModerator)
```
gsidebo: You can set the result of `wrapper.props()` to a variable and access it instead of the repeated calls

static/js/hoc/withPostList_test.js
```diff
@@ -0,0 +1,218 @@
...
+              }
+            }
+          )
+          const props = inner.find("InfiniteScroll").props()
+          await props.loadMore()
+          assert.equal(loadPosts.callCount, 1)
+          sinon.assert.calledWith(loadPosts, {
+            after: pagination.after,
+            count: pagination.after_count,
+            ...(hasParams ? { sort: "some_sort" } : {})
```
gsidebo: Can you parametrize the `location.search` value and the expected expected `loadPosts` params?

static/js/hoc/withPostList_test.js
```diff
@@ -0,0 +1,218 @@
...
+                [showProp]: showProp
+              }
+            )
+            assert.equal(
+              inner.find("PostList").props()[prop],
+              showProp ? prop : null
+            )
+          })
+        })
+      })
```
gsidebo: This is kind of hard to read. I think as a general rule we should avoid nested loops for declaring test cases. I think it's better just to flatten it out. Something like this:

```javascript
    ;[
        ["reportPost", "showReportPost", true, true],
        ["reportPost", "showReportPost", false, false],
        ["removePost", "showRemovePost", true, true],
        ["removePost", "showRemovePost", false, false]
      ].forEach(([prop, showProp, showPropValue, shouldBePassed]) => {
        it(`${shouldIf(showProp)} pass down ${prop} prop when ${showProp}=${String(
          showPropValue
        )}`, async () => {
          const { inner } = await render(
            {},
            {
              [prop]:     prop,
              [showProp]: showPropValue
            }
          )
          assert.equal(
            notNil(inner.find("PostList").prop(prop)),
            shouldBePassed
          )
        })
      })
```

noisecapella:
Any other comments?

static/js/hoc/withPostList_test.js
```diff
@@ -0,0 +1,221 @@
...
+        it(`${
+          showProp ? "passes" : "doesn't pass"
+        } through ${prop} when ${String(showProp)}=${String(
+          showPropValue
+        )}`, async () => {
+          const { inner } = await render(
+            {},
+            {
+              [prop]:     prop,
+              [showProp]: showProp
```
gsidebo: I don't think this test is correctly written. `showProp` is a string, and the value of the prop should be a boolean. Looks like you need to replace `showProp` with `showPropValue` for the value here and at line 174. Also, you can use `inner.find("PostList").prop(prop)` instead of `inner.find("PostList").props()[prop]`

noisecapella:
Any other comments?

Issue #1051, opened 2018-08-13
Detail page: Anonymous view


pdpinch:
Anonymous users should be able to see the detail page, but when they interact with it they should get a pop-up encouraging them to log in or create an account. 

Design at https://invis.io/9EN908ZQPAX#/314585748_Detail_Link_Not_Logged

### acceptance criteria
- [x] protect upvote, similar to channel listing page
- [ ] protect comment form
- [x] protect follow button
- [ ] protect comment reply button 
- [ ] anything else?

pdpinch:
Blocked waiting for updated designs

pdpinch:
We're still working out some of the fine details about the design, but in general it will be similar to the "pop-up" for sharing a URL. I expect it will have a button to log in / sign up and a close button.

mbertrand:
The following are not displayed at all on a post detail page for anonymous users: comment form, comment reply form, follow button.  The only thing mentioned in the checklist above that is displayed is the upvote (& downvote) button.  Should that change?

aliceriot:
Yeah, I'm working on the reply stuff and refactoring a lot of stuff on the comment component, might be best to hold off on that to avoid merge conflicts or duplicate work

mbertrand:
Thanks Alice, I will start on creating the login popup and attach it to the vote and follow buttons for now, leave the comment changes for later.

aliceriot:
we also have this existing popup / tooltip thing, which I'm guessing we'll want to get rid of:

![login-popup](https://user-images.githubusercontent.com/6207644/44281399-a32e2480-a225-11e8-8ba4-8128b656de23.png)

pdpinch:
Yes, we would like to replace that -- on the detail page and on the listing page.

PR #1106, opened 2018-08-22
LoginPopup for Follow button

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Partially addresses #1051

#### What's this PR do?
- Adds the `LoginPopup` component to the follow button
- Creates a new class for `FollowButton` instead of creating it via a function within ExpandedPostDisplay.
- Adds `className` to the LoginPopup so its styling can be customized (necessary for the Follow button, to avoid it overflowing past the right edge of the browser window).

- When logged in, the Follow/Unfollow button should work as before on the post detail page.
- When not logged in, the Follow button should still display and clicking it should display the login popup.


aliceriot:
functionality looks great! reading code now

static/js/components/ExpandedPostDisplay_test.js
```diff
@@ -197,23 +197,6 @@ describe("ExpandedPostDisplay", () => {
...
-        const wrapper = renderPostDisplay()
-        const button = wrapper.find(
-          subscribed ? ".subscribed" : ".unsubscribed"
-        )
-        assert.include(button.text(), buttonText)
-        button.simulate("click")
-        assert.ok(toggleFollowPostStub.called)
-      })
-    }
-  )
```
aliceriot: might be good to add a test here just confirming that the right props are passed down to the `FollowButton` component

static/js/components/FollowButton_test.js
```diff
@@ -0,0 +1,85 @@
...
+import { shouldIf } from "../lib/test_utils"
+import * as utilFuncs from "../lib/util"
+import Router from "../Router"
+import IntegrationTestHelper from "../util/integration_test_helper"
+
+describe("FollowButton", () => {
+  let helper, toggleFollowPostStub, post
+
+  beforeEach(() => {
+    //sandbox = sinon.createSandbox()
```
aliceriot: stray comment?

static/js/components/LoginPopup.js
```diff
@@ -1,14 +1,15 @@
 // @flow
-import { Link } from "react-router-dom"
 import React from "react"
 import onClickOutside from "react-onclickoutside"
 
 import Button from "./Button"
+import { Link } from "react-router-dom"
```
aliceriot: small thing, but we usually keep outside dependencies together at the top

static/scss/share-popup.scss
```diff
@@ -46,6 +46,14 @@
   }
 }
```
aliceriot: this file is styling both the share and login popups I think, what do you think about renaming it to just `popup.scss` while we're in here? I figure that's a little more clear

mbertrand: 👍

static/js/components/ExpandedPostDisplay_test.js
mbertrand: ✅

PR #1110, opened 2018-08-22
add material grid

### LINKED ISSUE #1007: add material layout grid
As a designer, I'd like to be able to specify certain interface elements in terms of how many columns they use. Columns can shrink and grow as necessary for a responsive layout. 

As an engineer, I'd like to be able to apply a simple CSS style that will location of an element as specified by the designer. 

We will try applying [mdc-layout-grid](https://github.com/material-components/material-components-web/tree/master/packages/mdc-layout-grid) to see if it can give us a simple 12 column grid for desktop layout.

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #1007

#### What's this PR do?

This PR redoes our layout to use the material grid for column widths, instead of our own custom stuff. I added two react components, `<Grid />` and `<Cell />`, which are basically just divs with the right classes set on them. These can be used to make a material grid wherever we want. I went ahead and replaced the existing layout in our `singleColumn` and `withChannelSidebar` components with this setup.

Basically everything should look the same throughout the app, and things should behave at all screenwidths and so on. Code should make sense.

### DIFF
--- a/static/js/components/Grid_test.js
+++ b/static/js/components/Grid_test.js
@@ -1,0 +1,35 @@
+// @flow
+import React from "react"
+import { shallow } from "enzyme"
+import { assert } from "chai"
+
+import { Grid, Cell } from "./Grid"
+
+describe("Grid components", () => {
+  it("should render a grid with children", () => {
+    const wrapper = shallow(
+      <Grid>
+        <div className="child" />
+      </Grid>
+    )
+    assert.equal(wrapper.props().className, "mdc-layout-grid")
+    assert.ok(wrapper.find(".mdc-layout-grid__inner").exists())
+    assert.ok(wrapper.find(".child").exists())
+  })
+
+  //
+  ;[1, 12].forEach(width => {
+    it(`should create a cell with width ${width}`, () => {
+      const wrapper = shallow(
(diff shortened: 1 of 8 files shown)


static/js/components/Grid_test.js
```diff
@@ -0,0 +1,35 @@
...
+        <div className="child" />
+      </Grid>
+    )
+    assert.equal(wrapper.props().className, "mdc-layout-grid")
+    assert.ok(wrapper.find(".mdc-layout-grid__inner").exists())
+    assert.ok(wrapper.find(".child").exists())
+  })
+
+  //
+  ;[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12].forEach(width => {
```
mbertrand: Minor nitpick: could probably just test values `0` and `12` for `width` here, test passes no matter what `width` is: 0-12, or 99, or 'foo'.  It's just Flow that will complain about the bad values.

aliceriot: yeah, I had added this mainly because the test file is checked with flow, but I can change it to just check 1 and 12

PR #1114, opened 2018-08-23
Added allow_top arg to channel creation and updates

### LINKED ISSUE #948: Home: anonymous user
As an anonymous user, I'd like to see posts on the home page from all the public channels. I'd also like to be able to create an account. 

~~If our reddit supports an r/all endpoint, we can use that.~~ We'll try using the existing reddit default channel functionality so anonymous users can see public channels. 

The "Become a Member" button should be behind the same feature flag that controls creating email accounts.

See mock-ups at https://invis.io/9EN908ZQPAX#/311617690_Home_Not_Logged
Desktop CSS at https://zpl.io/2EAZJpY
Mobile at https://zpl.io/ad7AKeL

### acceptance criteria
- [ ] For anonymous users, display posts from all public channels

### out of scope
-  "discover" button: this is displayed only to logged in users
-  "Learn, Share, Connect" intro text
- "Become a Member" button (behind feature flag) 
-  Without the feature flag, continue to direct anonymous users to log in with MM (i.e. /auth_required/ )

### LINKED ISSUE #950: subscribe new users to all public channels
When a user creates their account, we would like to subscribe them to all public channels so they will see something on their home page. 

Instead of subscribing new users, we're considering using the default channel functionality of reddit.

### acceptance criteria
- [x] when a new account is created, subscribe the user to all the default channels

### out of scope
- #1064 backfilling existing accounts
- In the future, we might want a way to manage a list of channels that new users are subscribed to -- but it will be some time before we have enough public channels to justify this

rhysyngsun:
**Note:** this PR generates **a lot** of changes to fixtures. Save your time and sanity and ignore them. I split the code between two commits with the cassette changes in the second. Github is cutting off the files list in this PR diff, so look at this commit to see the actual code changes: https://github.com/mitodl/open-discussions/pull/1114/commits/69d1041013c059f01b87eb7131e7a48af27ba268

#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #948 
Closes #950

#### What's this PR do?
- Adds `allow_top` with a default of `True` to channel creation
- Adds `allow_top` as an allowed channel update property (but doesn't add it to the API)
- Adds a command to manage this setting for channels


channels/management/commands/set_channel_allow_top.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument('--name', nargs='?', action='append')
+        parser.add_argument('--allow-top', action='store', default=True)
+
+    def handle(self, *args, **options):
+        """Sets allow_top on channels"""
+        channels = Channel.objects.all()
+
+        print(options)
```
noisecapella: Stray print

channels/management/commands/set_channel_allow_top.py
```diff
@@ -0,0 +1,40 @@
...
+
+        for channel in channels:
+            try:
+                api.update_channel(channel.name, allow_top=allow_top)
+                self.stdout.write("Channel '{}': set allow_top={}".format(
+                    channel.name,
+                    allow_top,
+                ))
+            except Exception as exc:  # pylint: disable=broad-except
+                self.stderr.write(str(exc))
```
noisecapella: Should we let the exception raise instead? If one up the updates fails the user should be able to rerun the command to try again

channels/management/commands/set_channel_allow_top.py
```diff
@@ -0,0 +1,40 @@
...
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from django.contrib.auth.models import User
+from channels.api import Api
+from channels.models import Channel
+
+
+class Command(BaseCommand):
+    """Sets allow_top on channels"""
+    help = 'Sets allow_top on channels'
```
noisecapella: Can you change to `Sets allow_top=true on channels`?

channels/management/commands/set_channel_allow_top.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def handle(self, *args, **options):
+        """Sets allow_top on channels"""
+        channels = Channel.objects.all()
+
+        print(options)
+        allow_top = options['allow_top']
+
+        if options['name']:
+            channels.filter(name=options['name'])
```
noisecapella: I think this needs to be `channels = channels.filter(...)`

channels/management/commands/set_channel_allow_top.py
rhysyngsun: So, I ran into 404s on my local setup (not quite sure why), but I think I would prefer to have the command partially succeed, because otherwise the command could abort part way through and the user would be unable to complete the command until the errors were fixed.

I'll update this so it also prints the channel name though, because it's not super helpful as-is.

noisecapella: That's fine, can we also make it exit with a non-zero code if there are any errors?

noisecapella:
Any comment on this? https://github.com/mitodl/open-discussions/pull/1114#discussion_r213354562

rhysyngsun:
@noisecapella all set

PR #1119, opened 2018-08-23
Fix regression regarding loading behavior for post and channel moderation page

### LINKED ISSUE #1115: Post and channel moderation pages are broken
Looks like a regression from #1111

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1115 

#### What's this PR do?
Fixes post and channel moderation page initialization to work properly on fresh reloads

Go to a post page and do a full browser refresh. Go to the channel reports moderation page and refresh the browser again. It should work fine both times.

I also went through the other pages which were using the old `withLoader` hoc component and they appear to be working for me.

You should also be able to copy over the tests in this PR into master and reproduce the errors thrown.

### DIFF
--- a/static/js/containers/ChannelModerationPage.js
+++ b/static/js/containers/ChannelModerationPage.js
@@ -49,7 +49,7 @@ type Props = {
   ignoreCommentReports: Function
 } & LoadingProps
 
-class ChannelModerationPage extends React.Component<Props> {
+export class ChannelModerationPage extends React.Component<Props> {
   componentDidMount() {
     this.loadData()
   }
@@ -63,7 +63,7 @@ class ChannelModerationPage extends React.Component<Props> {
     } catch (_) {} // eslint-disable-line no-empty
   }
 
-  renderReport = report => {
+  renderReport = (report: PostReportRecord | CommentReportRecord) => {
     const {
       isModerator,
       removePost,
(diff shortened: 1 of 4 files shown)


aliceriot:
I don't think this is the right approach overall, for a few reasons:

- the pattern we had before for the `withLoading` HOC was I think a really good pattern for container components which need to load certain data on page load and then display only after all the data is there. This is the general requirements for the post page, the channel edit page, etc. Using the `withLoading` HOC as it was written before prevented us from having to do cumbersome stuff like constantly checking that data was present before rendering. Instead, we just had to define a relatively simple contract in the `connectStateToProps` function - when everything is loaded, return `loaded === true`, else, return `false`. Then the whole component can be written as if all data is present all the time, without a lot of null checks. I think this is a superior pattern, because it puts the information about what data the component requires in one place and simplifies the code.
- the data requirements for a page like the infinite scroll page are different. We want to initialize the page, fetch some initial data, and then as the user interacts with it we want to fetch more data. So this requirement is different than, say, the `PostPage`, where we just fetch stuff on mount and are then done (aside from adding comments or editing things, which don't trigger a loading icon). since the data requirements for these two types of pages are different, it doesn't make sense to me to refactor the way we handle loading on the `PostPage` and so on just because we need a new pattern on the channel list and frontpage. We need a different pattern for these two kinds of components which takes their different requirements into account.

I think we should come up with a good pattern for handling the loading on the new infinite scroll pages, and we should retain the existing `withLoading` HOC pattern for all the pages which fetch all their data on mount. I might suggest actually wrapping the `PostList` component with the `withLoading` HOC, and removing the HOC from the channel page and home page containers - then each new page of posts can be shown as loading independent of the others which are already loaded. The channel page component could manage fetching the data, calculating the `loaded`, `errored`, etc props. Just an idea, but I don't think the refactor that's been done is the right way forward.

PR #1120, opened 2018-08-23
Fix sidebar scrolling

### LINKED ISSUE #1113: Scrolling doesn't work on channel sidebar

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Fixes #1113 

#### What's this PR do?
Tweaks CSS to enable sidebar scrolling

Try out different browser window sizes and make sure the sidebar scrolling looks good.

If there is enough room for everything the channel links should be on the top and the footer should be on the bottom. There should be no scroll bar visible and scrolling should not have any effect on the sidebar.

If there isn't enough room the channels should be right above the footer and you should see a scrollbar. If you scroll up or down it should act as expected

### DIFF
--- a/static/scss/navigation.scss
+++ b/static/scss/navigation.scss
@@ -188,6 +188,7 @@ $link-padding: 11px 20px 11px 19px;
   display: flex;
   flex-direction: column;
   height: 100%;
+  overflow: scroll;
 
   .navigation {
     flex: 1 0 auto;


aliceriot:
ah, sorry I didn't get a chance to look at this last night. I don't think this is the right approach overall, for a few reasons:

- the pattern we had before for the `withLoading` HOC was I think a really good pattern for container components which need to load certain data on page load and then display only after all the data is there. This is the general requirements for the post page, the channel edit page, etc. Using the `withLoading` HOC as it was written before prevented us from having to do cumbersome stuff like constantly checking that data was present before rendering. Instead, we just had to define a relatively simple contract in the `connectStateToProps` function - when everything is loaded, return `loaded === true`, else, return `false`. Then the whole component can be written as if all data is present all the time, without a lot of null checks. I think this is a superior pattern, because it puts the information about what data the component requires in one place and simplifies the code.
- the data requirements for a page like the infinite scroll page are different. We want to initialize the page, fetch some initial data, and then as the user interacts with it we want to fetch more data. So this requirement is different than, say, the `PostPage`, where we just fetch stuff on mount and are then done (aside from adding comments or editing things, which don't trigger a loading icon). since the data requirements for these two types of pages are different, it doesn't make sense to me to refactor the way we handle loading on the `PostPage` and so on just because we need a new pattern on the channel list and frontpage. We need a different pattern for these two kinds of components which takes their different requirements into account.

I think we should come up with a good pattern for handling the loading on the new infinite scroll pages, and we should retain the existing `withLoading` HOC pattern for all the pages which fetch their data on mount. I might suggest actually wrapping the `PostList` component with the `withLoading` HOC, and removing the HOC from the channel page and home page containers - then each new page of posts can be shown as loading independent of the others which are already loaded. The channel page component could manage fetching the data, calculating the `loaded`, `errored`, etc props. Just an idea, but I don't think the refactor that's been done is the right way forward.

aliceriot:
ah, sorry comment was on the wrong PR! deleting and moving :woman_facepalming:

PR #1123, opened 2018-08-24
remove problematic CSS to fix scrolling

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

address this additional scrolling issue:

scrolling-fix

#### What's this PR do?

this fixes an issue with the `flex` property, which seems to only show up in Chrome. I changed it to use `justify-content` to space things out instead.


aliceriot:
ok that seems to be the simplest way to address it, lmk what you think

PR #1131, opened 2018-08-27
LoginPopup for comment reply buttons and post reply form

### LINKED ISSUE #1051: Detail page: Anonymous view

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #1051

#### What's this PR do?
- Adds the `LoginPopup` to comment Reply buttons and the post reply text area.

- When logged in, the comment reply buttons and post reply form on a post detail page should work the same as before.
- When not logged in, the comment reply buttons and post reply form should still be rendered.  Clicking on a reply button or the text area of the post reply form should cause the `LoginPopup` to appear (just under the clicked Reply button, or over the post reply text area).  Clicking again outside the popup or text area should make the popup disappear.

### DIFF
--- a/static/js/components/CommentForms.js
+++ b/static/js/components/CommentForms.js
@@ -7,14 +7,13 @@ import { connect } from "react-redux"
 import { actions } from "../actions"
 import { setPostData } from "../actions/post"
 import { setBannerMessage } from "../actions/ui"
-import { isEmptyText, userIsAnonymous } from "../lib/util"
+import { isEmptyText, commentLoginText, userIsAnonymous } from "../lib/util"
 import Editor, { editorUpdateFormShim } from "./Editor"
+import LoginPopup from "./LoginPopup"
 
 import type { CommentForm, CommentInTree, Post } from "../flow/discussionTypes"
 import type { FormsState } from "../flow/formTypes"
 import type { Dispatch } from "redux"
-import LoginPopup from "./LoginPopup"
-import { commentLoginText } from "./CommentTree"
 
 type CommentFormProps = {
   dispatch: Dispatch<*>,
--- a/static/js/components/CommentForms.js
+++ b/static/js/components/CommentForms.js
@@ -82,7 +82,7 @@ const commentForm = (
             !disabled &&
             !isEmptyText(text)
           ) {
-            userIsAnonymous() && onTogglePopup ? onTogglePopup() : onSubmit(e)
+            onSubmit(e)
           }
         }}
       >
@@ -465,7 +465,7 @@ export const ReplyToPostForm: Class<React$Component<*, *>> = connect(
             onUpdate,
             cancelReply,
             false,
-            replying,
+            replying || userIsAnonymous(),
             false,
             false,
             this.onTogglePopup
--- a/static/js/components/CommentForms.js
+++ b/static/js/components/CommentForms.js
@@ -72,7 +72,14 @@ const commentForm = (
   return (
     <div className="reply-form">
       <form
-        onSubmit={onSubmit}
+        onSubmit={
+          userIsAnonymous() && onTogglePopup
+            ? R.compose(
+              onTogglePopup,
+              e => e.preventDefault()
+            )
+            : onSubmit
+        }
         className="form"
         onKeyDown={e => {
           if (
@@ -110,7 +117,7 @@ const commentForm = (
         <button
           type="submit"
           className={`blue-button ${disabled ? "disabled" : ""}`}
-          disabled={disabled || isEmptyText(text)}
+          disabled={(disabled || isEmptyText(text)) && !userIsAnonymous()}
         >
           Submit
         </button>
--- a/static/js/components/CommentForms.js
+++ b/static/js/components/CommentForms.js
@@ -458,6 +458,7 @@ export const ReplyToPostForm: Class<React$Component<*, *>> = connect(
             message={commentLoginText}
             visible={popupVisible}
             closePopup={this.onTogglePopup}
+            className="downshift"
           />
           {commentForm(
             this.onSubmit,
--- a/static/js/components/CommentForms.js
+++ b/static/js/components/CommentForms.js
@@ -7,12 +7,14 @@ import { connect } from "react-redux"
 import { actions } from "../actions"
 import { setPostData } from "../actions/post"
 import { setBannerMessage } from "../actions/ui"
-import { isEmptyText } from "../lib/util"
+import { isEmptyText, userIsAnonymous } from "../lib/util"
 import Editor, { editorUpdateFormShim } from "./Editor"
 
 import type { CommentForm, CommentInTree, Post } from "../flow/discussionTypes"
 import type { FormsState } from "../flow/formTypes"
 import type { Dispatch } from "redux"
+import LoginPopup from "./LoginPopup"
+import { commentLoginText } from "./CommentTree"
 
 type CommentFormProps = {
   dispatch: Dispatch<*>,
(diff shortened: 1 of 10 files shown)


aliceriot:
I think it might be good to put up the popup if the user clicks here, on the 'Submit' button, too:

reply-button-popup

(for the top-level post-reply form)

aliceriot:
oh, that's another option, I'm not sure - I think either option would be good

mbertrand:
I did both (button should appear disabled until text is entered; clicking on it as an anonymous user should show the popup)

aliceriot:
functionality looks good! reading code now

static/js/components/CommentForms.js
```diff
@@ -65,59 +66,77 @@ const commentForm = (
...
+) => {
+  return (
+    <div className="reply-form">
+      <form
+        onSubmit={
+          userIsAnonymous() && onTogglePopup
+            ? R.compose(
+              onTogglePopup,
+              e => e.preventDefault()
+            )
```
aliceriot: we have a little helper called `preventDefaultAndInvoke`, can you use that here instead of the `R.compose` setup?

static/js/components/CommentForms.js
```diff
@@ -65,59 +66,77 @@ const commentForm = (
...
-    <form
-      onSubmit={onSubmit}
-      className="form"
-      onKeyDown={e => {
-        if (e.key === "Enter" && e.ctrlKey && !disabled && !isEmptyText(text)) {
-          onSubmit(e)
+  wysiwyg: boolean = false,
+  onTogglePopup?: Function
+) => {
+  return (
```
aliceriot: can we keep this as a one-line arrow function (like `() => (<div />)` rather than

```js
() => {
  return (
    <div />
  )
}
```
?

I think the one-line version is just a little cleaner, and it will mean fewer spurious changes showing up in the diff

static/js/components/CommentForms.js
```diff
@@ -65,59 +66,77 @@ const commentForm = (
...
-)
+          Submit
+        </button>
+        {isComment ? (
+          <a
+            href="#"
+            onClick={R.compose(
+              cancelReply,
+              e => e.preventDefault()
+            )}
```
aliceriot: I think we could update this to use `preventDefaultAndInvoke` as well, while we're in here

static/js/components/CommentForms_test.js
```diff
@@ -33,7 +36,9 @@ describe("CommentForms", () => {
   const renderPostForm = (props = {}) =>
     mount(
       <Provider store={helper.store}>
-        <ReplyToPostForm post={post} {...props} />
+        <Router store={helper.store} history={helper.browserHistory}>
+          <ReplyToPostForm post={post} {...props} />
+        </Router>
```
aliceriot: why is the `<Router>` necessary here?

static/js/components/CommentTree.js
```diff
@@ -64,7 +69,14 @@ export const commentDropdownKey = (c: CommentInTree) =>
...
   `COMMENT_SHARE_MENU_${c.id}`
 
-export default class CommentTree extends React.Component<Props> {
+export default class CommentTree extends React.Component<Props, State> {
+  constructor() {
+    super()
+    this.state = {
+      popupVisible: false
+    }
+  }
```
aliceriot: is the state here used for anything?

static/js/components/ReplyButton.js
```diff
@@ -0,0 +1,61 @@
...
+      <div>
+        <div
+          className="comment-action-button reply-button"
+          onClick={e => {
+            if (beginEditing && !userIsAnonymous()) {
+              beginEditing(formKey, initialValue, e)
+            } else {
+              this.onTogglePopup()
+            }
+          }}
```
aliceriot: I think here we might want to do something like

```js
onClick={
  beginEditing && !userIsAnonymous()
    ? beginEditing(formKey, initialvalue)
    : this.onTogglePopup
}
```

then we can add an `e.preventDefault` in `this.onTogglePopup` to make sure we handle that.

static/js/components/ReplyButton_test.js
```diff
@@ -0,0 +1,77 @@
...
+    comment = makeComment(makePost())
+  })
+
+  afterEach(() => {
+    helper.cleanup()
+  })
+
+  const renderButton = () =>
+    mount(
+      <Router store={helper.store} history={helper.browserHistory}>
```
aliceriot: wondering about the `Router` here too

aliceriot: oh, because of the `<Link>` in the `LoginPopupHelper`? you may be able to avoid this by using `shallow` instead of `mount` here

static/scss/popup.scss
```diff
@@ -55,6 +55,10 @@
   }
 }
 
+.downshift {
+  margin-top: 50px !important;
```
aliceriot: any way we can avoid the `!important` here?

static/js/components/CommentForms_test.js
mbertrand: Without it, this error is thrown due to `<Link>` being in the `LoginPopup`:
```
Invariant Violation: You should not use <Link> outside a <Router>
```

static/js/components/CommentTree.js
mbertrand: Not anymore, forgot to remove it.

static/scss/popup.scss
mbertrand: Doesn't seem to be effective without it

aliceriot: yeah, I was wondering if there is a way you could refactor it so that it wouldn't be necessary. Sometimes `!important` is necessary to get a style to stick because there is already a more-specific selector styling that element - in that case, we should specify a more specific selector to apply this style, not override the style cascase with `!important`. The issue here, I think, is that `.login-popup` already has a `margin-top` applied to it - if this selector is changed to `.login-popup.downshift` instead of just `.downshift` is should override `.login-popup` automatically. I think if you move it into the `.login-popup` block by doing:

```scss
.login-popup {
  &.downshift {
  }
}
```

you should be able to remove `!important`. This is probably good practice anyway, since currently the `.downshift` selector is not focused to the `login-popup` at all.

static/js/components/CommentForms_test.js
aliceriot: gotcha

static/js/components/CommentForms.js
```diff
@@ -65,11 +71,17 @@ const commentForm = (
...
+  wysiwyg: boolean = false,
+  onTogglePopup?: Function
 ) => (
   <div className="reply-form">
     <form
-      onSubmit={onSubmit}
+      onSubmit={
+        userIsAnonymous() && onTogglePopup
+          ? // $FlowFixMe - the above ensures onTogglePopup is defined
+          preventDefaultAndInvoke(() => onTogglePopup())
```
aliceriot: can this just be `preventDefaultAndInvoke(onTogglePopup)`?

static/js/components/CommentForms.js
```diff
@@ -91,25 +103,28 @@ const commentForm = (
             className="input"
             placeholder="Write a reply here..."
             value={text || ""}
-            onChange={onUpdate}
+            onChange={disabled ? null : onUpdate}
+            onClick={
+              userIsAnonymous() && onTogglePopup
+                ? // $FlowFixMe: the above
+                preventDefaultAndInvoke(() => onTogglePopup())
```
aliceriot: same question here

static/js/components/CommentForms.js
```diff
@@ -91,25 +103,28 @@ const commentForm = (
...
         Submit
       </button>
       {isComment ? (
         <a
           href="#"
-          onClick={R.compose(
-            cancelReply,
-            e => e.preventDefault()
-          )}
+          onClick={preventDefaultAndInvoke(() => cancelReply())}
```
aliceriot: same here :)

static/js/components/CommentForms.js
```diff
@@ -416,6 +433,13 @@ export const ReplyToPostForm: Class<React$Component<*, *>> = connect(
       this.ensureInitialState()
     }
 
+    onTogglePopup = async () => {
```
aliceriot: it doesn't look like this needs to be an `async` function

static/js/components/CommentTree.js
```diff
@@ -125,21 +124,13 @@ export default class CommentTree extends React.Component<Props> {
...
-              }}
-            >
-            reply
-            </div>
-          )}
+        {atMaxDepth || moderationUI || comment.deleted ? null : (
+          <ReplyButton
+            beginEditing={beginEditing}
+            formKey={replyToCommentKey(comment)}
+            initialValue={getCommentReplyInitialValue(comment)}
```
aliceriot: what do you think about just passing down a `beginEditing` func to `ReplyButton`? It seems like the `ReplyButton` shouldn't have to know about `formKey` and `initialValue` and all that

static/scss/popup.scss
mbertrand: Tried this, doesn't work.

static/js/components/ReplyButton_test.js
```diff
@@ -0,0 +1,74 @@
...
+  })
+
+  const renderButton = () =>
+    shallow(
+      <ReplyButton
+        formKey={replyToCommentKey(comment)}
+        initialValue={getCommentReplyInitialValue(comment)}
+        beginEditing={beginEditingStub}
+      />
+    )
```
aliceriot: I recently wrote a little function, called `configureShallowRenderer`, which might be helpful here. you call it with a component and a default props object, and it returns a little test renderer - you can use it if you want :)

static/scss/popup.scss
aliceriot: that's odd, I tried it locally and it works on my machine. maybe it's a browser specific thing? I'm using Chromium.

static/js/components/ReplyButton_test.js
mbertrand: OK

static/js/components/CommentTree.js
mbertrand: sure

static/scss/popup.scss
mbertrand: I'm using Chrome too, no luck here.  Would also require changing ```<div className={`popup login-popup ${className || ""}`}>``` in `ReplyButton` to use a period instead of a space between `login-popup` and the specified `className` if any, not sure I like that idea.

aliceriot: No, you shouldn't need to do that. The `className` prop is set as an HTML attribute on the DOM node that ReactDOM eventually produces, you don't need to put any periods in here.

I think if you add the `&.downshift` to the existing CSS block for `.login-popup` and make sure that the `&.downshift` block is located after the pre-existing declaration for `margin-top` it should work. Here's the diff of what I did which worked for me:

```diff
diff --git a/static/scss/popup.scss b/static/scss/popup.scss
index 2b46f5ae..dc5e1515 100644
--- a/static/scss/popup.scss
+++ b/static/scss/popup.scss
@@ -55,10 +55,6 @@
   }
 }
 
-.downshift {
-  margin-top: 50px !important;
-}
-
 .share-popup {
   .triangle {
     right: 50%;
@@ -71,6 +67,10 @@
   margin-top: 10px;
   min-width: 255px;
 
+  &.downshift {
+    margin-top: 50px;
+  }
+
   .close-popup {
     position: absolute;
     top: 0px;
```

mbertrand: OK, I interpreted the `.login-popup.downshift` too literally.  Just made the change.

PR #1139, opened 2018-08-29
Configured login flow to show greeting for external auth providers

### LINKED ISSUE #1058: Login: Update UI for recognized email to include micromasters & touchstone login
PR #1039 changed the login flow to present the user with a greeting, profile image, and email address when they enter a recognized email address into the login form. We want to do something similar if the email matches a Micromasters- or Touchstone-authenticated user. We will show them a the same greeting/image/email, and under that we will show a button/link that redirects to Micromasters or a Touchstone login link button respectively. 

See [MIT OPEN onboarding google doc for guidance](https://docs.google.com/document/d/1_sD0-ng7PsYmvmmOjoTYcguygLVctx2RZqUbJsDPvR8)

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #1058

#### What's this PR do?
Changes the login flow so that emails that match micromasters- or touchstone-authenticated users are shown a personalized greeting after the email is submitted

On the `/login` page...
- Enter an email for a user that has _only_ authenticated via Micromasters
- Enter an email for a user that has authenticated via Touchstone at any time (if you're running locally, you may have to fudge this - ask me for details if you need to do that)

In both cases, the name, email, and profile image should be shown to the user after submission, just like we do for users that have an email authentication

ss 2018-08-29 at 10 32 26
ss 2018-08-29 at 10 32 18

### DIFF
--- a/authentication/pipeline/user.py
+++ b/authentication/pipeline/user.py
@@ -3,6 +3,7 @@ import ulid
 from social_core.backends.email import EmailAuth
 from social_core.backends.saml import SAMLAuth
 from social_core.pipeline.partial import partial
+from django.conf import settings
 
 from authentication.backends.micromasters import MicroMastersAuth
 from authentication.exceptions import (
@@ -13,6 +14,7 @@ from authentication.exceptions import (
     RequireRegistrationException,
 )
 from authentication.utils import SocialAuthState
+from open_discussions import features
 from open_discussions.settings import SOCIAL_AUTH_SAML_IDP_ATTRIBUTE_NAME
 from profiles import api as profile_api
 from profiles.utils import update_full_name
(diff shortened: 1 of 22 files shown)


static/js/lib/validation.js
```diff
@@ -48,11 +48,6 @@ export const validEmail = R.compose(
   R.trim
 )
 
-export const validNotMIT = R.compose(
```
rhysyngsun: So it looks like we're losing this now, which I think was meant to require MIT users to always login via touchstone, but now if they don't already have an account, they don't have to. It doesn't look like registration was guarded by this, so this "require MIT users to login via touchstone" thing is basically entirely going away here.

authentication/pipeline/user.py
```diff
@@ -190,10 +222,11 @@ def require_micromasters_provider(
...
-            user_storage.get_social_auth_for_user(user, provider=MicroMastersAuth.name).exists() and
-            user_storage.get_social_auth_for_user(user).count() == 1
-    ):
-        raise RequireProviderException(backend, MicroMastersAuth.name)
+    social_auths = user_storage.get_social_auth_for_user(user)
+    if len(social_auths) != 1:
+        return {}
 
+    social_auth = social_auths[0]
+    if social_auth.provider == MicroMastersAuth.name:
```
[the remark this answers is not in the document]
gsidebo: You're right that this isn't related to the issue for this PR. I was tinkering with this step early on and this was sort of a leftover from that. I'm not sure that it reintroduces the race condition though. `len(social_auths)` runs the query, and any subsequent access to the `social_auths` variable uses Django's cache instead of running another query. `social_auths` could be wrapped in a `list()` call to make that more explicit. In short, there's only one query being run here, so no chance of a race condition that I can spot. More generally, I think this is more efficient. I would definitely expect one query that fetches no more than 3 objects to be faster than 2 separate queries.

static/js/lib/validation.js
gsidebo: Ah, good catch. I forgot about the register page. I'll add this back in and add a separate validation function for the register page since it will be different from the login page now

authentication/pipeline/user.py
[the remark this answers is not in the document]
rhysyngsun: You make a good point. We can leave it for now since you've already written it, but I think it's generally important to partition unrelated changes into separate commits/PRs.

rhysyngsun:
@gsidebo JS tests are failing

authentication/pipeline/user.py
```diff
@@ -170,13 +171,44 @@ def validate_password(
...
+    """
+    If the user is attempting to log in via email and has authenticated via Touchstone/SAML, require
+    them to log in that way.
+
+    Args:
+        strategy (social_django.strategy.DjangoStrategy): the strategy used to authenticate
+        backend (social_core.backends.base.BaseAuth): the backend being used to authenticate
+        flow (str): the type of flow (login or register)
+    """
+    if backend.name != EmailAuth.name or flow != SocialAuthState.FLOW_LOGIN:
```
rhysyngsun: Just realized this should probably be returning here if the `SAML_AUTH` feature is turned off, because there's no way for the user to fix the login at that point.

gsidebo: Good point. Will do

gsidebo: Should be all set

authentication/pipeline/user_test.py
```diff
@@ -290,3 +288,36 @@ def test_require_micromasters_provider(
...
+])
+def test_require_touchstone_login(
+        mocker, backend_name, flow, social_auth, feature_flag_val, raises
+):  # pylint: disable=too-many-arguments
+    """
+    Tests that verify test_require_touchstone_login raises a RequireProviderException
+    if (a) the user is logging in via email, (b) has a SAML-type authentication, and
+    (c) the SAML auth feature flag in on
+    """
+    settings.FEATURES[features.SAML_AUTH] = feature_flag_val
```
rhysyngsun: Could you adjust this to use the `settings` fixture instead of the import so you get its rollback behavior and don't have this leak into other tests?

PR #1158, opened 2018-09-05
Added link tags with rel=canonical to improve SEO and analytics

### LINKED ISSUE #1011: SEO: Add rel=canonical tags to post detail page template
As a search engine crawler, I don't like to see the same content at multiple URLs. It spams my index leading to lower ranking in search results and possibly even penalizing the ranking. 

One way to address this is to add a `<link rel="canonical" href="">` tag to each page with the value of the href being what we consider the canonical URL for the page. For post detail pages, this should include the page slug, since it helps describe the page. 

Exposing canonical URLs should also fix an issue that we're seeing in Google Analytics where traffic to `/c/channel/reddit_id` and `/c/channel/reddit_id/` are getting counted separately. 

Ref: https://support.google.com/webmasters/answer/139066

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #1011 

#### What's this PR do?
Adds a `<link>` tag with `rel="canonical"` to improve SEO and simplify our analytics ([reference](https://support.google.com/webmasters/answer/139066))

Open up various pages on the site and check that there's a `<link>` tag with `rel="canonical"` and a consistent `href` attribute (URLs with and without a trailing slash should have the same value)

ss 2018-09-05 at 10 12 22

### DIFF
--- a/static/js/containers/ProfileEditPage.js
+++ b/static/js/containers/ProfileEditPage.js
@@ -9,6 +9,7 @@ import ProfileForm from "../components/ProfileForm"
 import Card from "../components/Card"
 import withLoading from "../components/Loading"
 import withSingleColumn from "../hoc/withSingleColumn"
+import CanonicalLink from "../components/CanonicalLink"
 
 import { actions } from "../actions"
 import { formatTitle } from "../lib/title"
@@ -120,6 +121,7 @@ class ProfileEditPage extends React.Component<Props> {
       <div className="profile-page">
         <MetaTags>
           <title>{formatTitle("Edit your profile")}</title>
+          <CanonicalLink relativeUrl="profile/edit" />
         </MetaTags>
         <div className="main-content">
           <Card>
--- a/static/js/lib/util.js
+++ b/static/js/lib/util.js
@@ -111,3 +111,8 @@ export const makeUUID = (len: number) =>
     .map(int => int.toString(16))
     .join("")
     .slice(0, len)
+
+export const removeTrailingSlash = (str: string) =>
+  str.length > 0 && str[str.length - 1] === "/"
+    ? str.substr(0, str.length - 1)
+    : str
--- a/static/js/components/CanonicalLink.js
+++ b/static/js/components/CanonicalLink.js
@@ -1,0 +1,31 @@
+// @flow
+/* global SETTINGS: false */
+import React from "react"
+
+import { removeTrailingSlash } from "../lib/util"
+
+import type { Match } from "react-router"
+
+type Props = {
+  relativeUrl?: ?string,
+  match?: ?Match
+}
+
+const CanonicalLink = ({ relativeUrl, match }: Props) => {
+  let partialUrl
+  if (relativeUrl) {
+    partialUrl = relativeUrl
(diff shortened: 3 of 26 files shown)


pdpinch:
Wow, this is a poster child for react making things complicated. 30 lines of JavaScript to add one line of HTML? 

Were you able to confirm whether this had the desired effect in Google Analytics? I ask because I'm wondering about the utility of all the user settings and login pages. If they don't show up in GA, then we should probably skip them.

gsidebo:
@pdpinch I wouldn't say this is an example of react making things complicated. The component is doing a few things

1. It accepts a string or a match object, and chooses which one of those values takes priority. It accepts both because many pages will just have the actual URL as the canonical URL, but some others specify a custom URL to use as the canonical (e.g.: `password_reset/confirm/:uid/:token`)
1. It appends the partial URL to the absolute site URL and remove any trailing slashes (URLs with and without trailing slashes create 2 different entries in GA)
1. It renders the `<link>` element

The only actual work the component is doing is (1) choosing the partial URL based on the props provided, (2) building the canonical URL without a trailing slash, and (3) rendering. The size of the file is appropriate IMO

As for the actual GA results, I'll look into it in the code, but it would still be helpful to see what it looks like for a CI/prod deployment. I took a better-safe-than-sorry approach, but it's very likely that some of these pages don't need this link tag

static/js/lib/util.js
```diff
@@ -111,3 +111,11 @@ export const makeUUID = (len: number) =>
...
     .join("")
     .slice(0, len)
+
+export const removeTrailingSlash = (str: string) =>
+  R.compose(
+    R.equals("/"),
+    R.last
+  )(str)
+    ? R.dropLast(1, str)
+    : str
```
aliceriot: could this just be `str.replace(/\/$/, "")`?

static/js/containers/ProfileEditPage.js
```diff
@@ -120,6 +121,7 @@ class ProfileEditPage extends React.Component<Props> {
       <div className="profile-page">
         <MetaTags>
           <title>{formatTitle("Edit your profile")}</title>
+          <CanonicalLink relativeUrl="profile/edit" />
```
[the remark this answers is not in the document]
gsidebo: Yeah that's the intent. I don't think we get any value from have 10 different entries for 10 different users visiting the profile edit page.

static/js/lib/util.js
gsidebo: Yeah, I tried to get too cute with this. I generally try to avoid regexes in really simple use cases like this, but I did a quick performance test and this is definitely worse than the regex equivalent. I also tested a dirt-simple solution using `substr` and that performed twice as well as the regex, so I went with that.

static/js/containers/ProfileEditPage.js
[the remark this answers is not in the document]
aliceriot: true enough, haha.

static/js/lib/util.js
aliceriot: sure, my concern is more code readability than performance for a trivial function like this, but I think the new version is fine

PR #1159, opened 2018-09-05
Recaptcha for new signups

### LINKED ISSUE #1065: Sign up: add recaptcha
As an admin, I'd like to avoid having a bunch of fake accounts created by bots.

We can add the Google recaptcha tool to the signup page.

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #1065

#### What's this PR do?
Adds a recaptcha field to the signup form

- Sign up for a Google recaptcha key
- In your `.env` file, set:
 ```
  RECAPTCHA_SITE_KEY=<your_site_key>
  RECAPTCHA_SECRET_KEY=<your_secret_key>
  ``` 
- Go to the signup page.  A recpatcha form should be visible.
- If you try to submit the form without checking it, you should get a validation message.
- When you check it, it should make you pass a test such as 'Select all images with cars"
- Once it is checked, you should be able to successfully submit the form.
- Go to the login page, it should not have a recaptcha field and should work as before.
- If you remove `RECAPTCHA_SITE_KEY` from your `.env` file, the recaptcha should not appear on the signup page and the form should work as before.
- If you use an invalid `RECAPTCHA_SECRET_KEY` value, you should receive a validation error message after submitting the form ('Error validating your submission.')

screen shot 2018-09-05 at 11 21 45 am

screen shot 2018-09-05 at 1 41 07 pm


noisecapella:
I think as is a bot user could easily work around the recaptcha by setting `SETTINGS.recaptchaKey = null`. I think there needs to be some backend change to verify the recaptcha token which was received in onChange: https://developers.google.com/recaptcha/docs/verify

mbertrand:
@noisecapella I added backend verification of the recaptcha token in `authentication.views.RegisterEmailView`

authentication/views.py
```diff
@@ -84,6 +85,20 @@ def get_serializer_cls(self):
         """Return the serializer cls"""
         return RegisterEmailSerializer
 
+    def post(self, request):
+        """ Verify recaptcha response before proceeding """
+        if settings.RECAPTCHA_SITE_KEY:
+            r = requests.post(
+                "https://www.google.com/recaptcha/api/siteverify?secret={}&response={}".format(
+                    settings.RECAPTCHA_SECRET_KEY,
```
noisecapella: Can you escape these two parameters, since `request.data` comes from the outside world?

authentication/views.py
```diff
@@ -84,6 +85,20 @@ def get_serializer_cls(self):
...
+                    request.data["recaptcha"]
```
noisecapella: Can we also make these named parameters just for clarity?

authentication/views.py
mbertrand: done

authentication/views.py
mbertrand: done

authentication/views.py
```diff
@@ -84,6 +87,20 @@ def get_serializer_cls(self):
         """Return the serializer cls"""
         return RegisterEmailSerializer
 
+    def post(self, request):
+        """ Verify recaptcha response before proceeding """
+        if settings.RECAPTCHA_SITE_KEY:
+            r = requests.post(
+                "https://www.google.com/recaptcha/api/siteverify?secret={key}&response={captcha}".format(
+                    key=escape(settings.RECAPTCHA_SECRET_KEY),
```
noisecapella: I think you need to use [urllib.parse.quote](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote) here instead of `html.escape`

mbertrand: replaced `escape` with `quote`

PR #1165, opened 2018-09-07
Don't show an error page if comment posting fails

### LINKED ISSUE #1069: Don't display error page if creating a comment fails.
If posting a comment on a post detail page fails for whatever reason, the page will reload as an error page, onto which the snackbar/banner error message will appear.  Ideally the page would continue to display as is.

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

- Fixes #1069

#### What's this PR do?
- Adds a `clearCommentError` dispatch action, similar to `clearPostError`, to prevent an error page being shown after an error occurs creating a comment.

- Go to a post detail page.
- Make comment posting certain to fail by turning off the reddit vagrant box; or adding `raise Exception()` to the beginning of the `channels.api.create_comment` method.
- Reply to the post or an existing comment with a new comment and click Submit.
- An error banner should appear at the top of the page and the post page should still display correctly.
- The same should happen if you edit one of your existing comments and click Submit.

### DIFF
--- a/static/js/components/CommentForms.js
+++ b/static/js/components/CommentForms.js
@@ -195,64 +195,65 @@ const mapDispatchToProps = (dispatch: any, ownProps: CommentFormProps) => {
     },
     cancelReply:  cancelReply(dispatch, formKey),
     beginEditing: beginEditing(dispatch, formKey),
-    onSubmit:     R.curry((postID, text, commentID, post, e) => {
+    onSubmit:     R.curry(async (postID, text, commentID, post, e) => {
       e.preventDefault()
-      return dispatch(actions.comments.post(postID, text, commentID))
-        .then(() => {
+      try {
+        await dispatch(actions.comments.post(postID, text, commentID))
+        dispatch(
+          setPostData({
+            ...post,
+            num_comments: post.num_comments + 1
+          })
+        )
+        cancelReply(dispatch, formKey)()
+      } catch (err) {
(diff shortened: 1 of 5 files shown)


static/js/components/CommentForms.js
```diff
@@ -227,9 +229,30 @@ const mapDispatchToProps = (dispatch: any, ownProps: CommentFormProps) => {
...
+            // Unknown errors
+            dispatch(
+              setBannerMessage(
+                `Something went wrong editing your comment. Please try again or contact us at ${
+                  SETTINGS.support_email
+                }`
+              )
+            )
+          }
+        }),
```
aliceriot: two questions about this function:

- I think it might be a little easier to read if it were written with the `async / await` sugar, up to you
- if the comment creation fails, should we also do `cancelReply` in that case, as well as in the success case? as it you see the banner message that your comment failed, but the form stays open. might prompt the user to spam submitting the comment, even if it won't ever work. what do you think?

mbertrand: I rewrote the patch and post functions to use async/await.  I decided to use cancelReply if the error is `410` (comment deleted) only, because another error might be a temporary issue and the user might want to try again, or at least copy their reply text before it disappears and try again later.

aliceriot: cool, just tried that out and it works well

PR #1169, opened 2018-09-10
Added banner message for PSA error messages

### LINKED ISSUE #1057: Handle account already in use errors
This happens when the user already has a OD session but tries to login with an account (MM or touchstone) that is linked to another user:

https://sentry.io/mit-office-of-digital-learning/discussions/issues/635378179/
https://sentry.io/mit-office-of-digital-learning/discussions/issues/621321627/

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1057

#### What's this PR do?
This fixes this issue by displaying a banner with the PSA error message in it. To do so, it also removes the `django.contrib.messages` app, since that causes PSA to write the messages to a cookie, which is more complex to handle and we're not actually using messages in any way elsewhere anyway.

- Login to discussions as one user via email (User A)
- Login to discussions as another user via MM (User B)
- You should remained logged in as User A, but see a banner message indicating the MM account was already linked.

### DIFF
--- a/authentication/middleware.py
+++ b/authentication/middleware.py
@@ -1,0 +1,34 @@
+"""Authentication middleware"""
+from django.shortcuts import redirect
+from django.utils.http import urlquote
+
+from social_core.exceptions import SocialAuthBaseException
+from social_django.middleware import SocialAuthExceptionMiddleware
+
+
+class SocialAuthExceptionRedirectMiddleware(SocialAuthExceptionMiddleware):
+    """
+    This middleware subclasses SocialAuthExceptionMiddleware and overrides
+    process_exception to provide an implementation that does not use
+    django.contrib.messages and instead only issues a redirect
+    """
+    def process_exception(self, request, exception):
+        """
+        Note: this is a subset of the SocialAuthExceptionMiddleware implementation
+        """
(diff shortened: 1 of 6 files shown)


mbertrand:
I remain logged in as the email user but don't see a banner message.

rhysyngsun:
@mbertrand sorry, there's one extra config step you need to do that i forgot about, could you temporarily change `DEBUG` to `False` in docker-compose.override.yml` (it doesn't work to override in `.env` because of how those compose files are setup right now) and retest?

mbertrand:
Still no luck, it seems as though there's no socialauth exception being thrown at all for me because the new middleware isn't being hit.

In incognito browser:
- Login to OD with email # 1 (existing user)
- Login to MM via EdX with email #2 (tried with both existing user and new user)
- Go to https://mm.odl.local:8079/dashboard
- Click 'View All' for Discussions
- Redirected to `http://od.odl.local:8063/`, still logged in as email # 1, no banner message.

mbertrand:
```python
In [1]: from django.conf import settings

In [2]: settings.DEBUG
Out[2]: False
```

PR #1172, opened 2018-09-10
Added empty post loading animation when posts are being loaded

### LINKED ISSUE #1050: When loading the channel page, display empty posts with loading animation...
... instead of animated ⋯

Similar to how the detail page behaves when loading from embed.ly.

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #1050

#### What's this PR do?
Adds a loading animation that shows some dummy empty post elements while posts are being loaded

Visit the home page or any channel page and observe the loading behavior. If you want to render the loading animation permanently, you can change `static/js/components/PostLoading.js` line 63 to: `if (true) {`

`static/js/components/PostLoading.js`

- ~~Some of the `PostLoading.js` code was copied over from `Loading.js`. I figured this was small, isolated, and unlikely to be repeated~~ No longer relevant
- I was thinking about possibly rendering this animation HTML by default and hiding/revealing it instead of rendering it fresh every time a list of posts needs to load. I figured any performance penalty would be small enough that it wasn't worth spending a lot of time on for this PR, but I'd be interested in other opinions

ss 2018-09-10 at 12 20 38

### DIFF
--- a/static/js/components/Loading.js
+++ b/static/js/components/Loading.js
@@ -1,7 +1,11 @@
 // @flow
 import React from "react"
+import R from "ramda"
+import ContentLoader from "react-content-loader"
 
 import { NotFound, NotAuthorized } from "../components/ErrorPages"
+import Card from "./Card"
+import { contentLoaderSpeed } from "../lib/constants"
 
 type LoadingProps = {
   loaded: boolean,
@@ -14,6 +18,8 @@ type SpinnerProps = {
   className?: string
 }
 
+const emptyPostsToRender = 5
+
 export const Loading = (props: SpinnerProps) => (
   <div className={`loading ${props.className ? props.className : ""}`}>
     <div className="sk-three-bounce">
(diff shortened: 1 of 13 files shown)


static/js/containers/auth/LoginPage.js
```diff
@@ -55,6 +57,22 @@ export class LoginPage extends React.Component<LoginPageProps> {
     return (
       <div className="content auth-page login-page">
         <div className="main-content">
+          <div className="content-loader">
```
mbertrand: Why is this on the login page?

gsidebo:
@mbertrand Comments addressed, and I decided to go ahead an abstract the common bits of the 2 different loading components. Ending up being pretty straightforward

static/js/components/Loading.js
```diff
@@ -14,6 +18,8 @@ type SpinnerProps = {
   className?: string
 }
 
+const emptyPostsToRender = 5
```
mbertrand: maybe this should go in the new `lib/constants.js` file too?

gsidebo: It's only relevant to the post loading component in this file, so I thought it was better to put it here. I put it in a constant instead of inlining it just to make it more obvious where the count was coming from, and easier to find and change

PR #1180, opened 2018-09-11
Validate a new link post URL before calling embedly

### LINKED ISSUE #1156: When entering a link for a link post, don't fetch the embed if its not a valid URL
Currently, when the user starts typing in a URL for a link post, we attempt to fetch the URL whenever there's a pause in the user's typing. If they type slowly, we'll end up checking on `h` and then `ht` and then `htt` etc. 

We really shouldn't both trying to fetch the URL unless and until what there is enough to make a valid URL.

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #1156

#### What's this PR do?
Installs the [validator.js](https://www.npmjs.com/package/validator) module and uses its `isURL` function to validate a new link post's URL before calling embedly.

Create a new link post and enter different values for the URL.  Embedly should not be called until a valid URL is entered, such as `http://reddit.com` or `reddit.com`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
     "spinkit": "^1.2.5",
     "style-loader": "^0.21.0",
     "url-loader": "^1.0.1",
+    "validator": "^10.7.1",
     "webpack": "^3.12.0",
     "webpack-bundle-tracker": "^0.3.0",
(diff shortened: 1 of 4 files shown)


aliceriot:
works great! reading code

PR #1184, opened 2018-09-12
Better handling of non-existent channels

### LINKED ISSUE #741: don't throw 500 when channel doesn't exist
I'm not really sure how this happened the first time, but if you make an API request for a channel that doesn't exist, we throw a 500 instead of just passing the 404 back. 

We should display a meaningful error message in the case of a 404, instead of "Error loading page"

NotFound: received 404 HTTP response 

https://sentry.io/mit-office-of-digital-learning/discussions/issues/562748222/

```
NotFound: received 404 HTTP response
(14 additional frame(s) were not displayed)
...
  File "rest_framework/views.py", line 491, in dispatch
    response = handler(request, *args, **kwargs)
  File "rest_framework/generics.py", line 241, in get
    return self.list(request, *args, **kwargs)
  File "rest_framework/mixins.py", line 40, in list
    queryset = self.filter_queryset(self.get_queryset())
  File "channels/views/moderators.py", line 29, in get_queryset
    return api.list_moderators(channel_name)
  File "channels/api.py", line 896, in list_moderators
    return self.get_channel(channel_name).moderator(redditor=moderator_name)

NotFound: received 404 HTTP response
```

### LINKED ISSUE #894: Add check for matching Channel object in permission classes
This would make sure that only channels which the discussions server knows about are accessible

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #894 
Closes #741 

#### What's this PR do?
- In permission classes, raise an `Http404` if the view contains a channel name and it does not exist
- In `PostPage.js`, check the `channels` state for 404 errors.

Try various URL's related to channels, using a channel name that doesn't exist, for example:
- `http://od.odl.local:8063/c/fake`
- `http://od.odl.local:8063/c/fake/<real_post_id>/mypost/`
- `http://od.odl.local:8063/c/fake/<real_post_id>/mypost/comment/<real_comment_id>/`

You should get a 404 page for each one.  There should not be any 500 errors in the console.


mbertrand:
Initial approach (commit # 1) was to return `False` in those permission classes that queried for `Channel` objects by channel name if it didn't exist, and raise a 404 in the relevant views.  However, that resulted in some views returning a 403 and others returning a 404 depending on which permission classes they used, and the preference is to consistently return 404.  

Reverted the change of making all permission classes inherit from the new `ChannelPermission` and instead added `ChannelPermission` to the views that need it.

mbertrand:
@rhysyngsun I'm going to look this over a bit more before I put it back to 'Needs Review', I suspect I'm implementing this in a way that's more complicated than it needs to be.

open_discussions/permissions.py
```diff
@@ -1,14 +1,32 @@
...
+
+    Args:
+        view (rest_framework.views.APIView): django DRF view
+
+    Returns:
+        bool: True if Channel exists (or there is no channel name)
+    """
+    channel_name = view.kwargs.get('channel_name', None)
+    if channel_name:
+        get_object_or_404(Channel, name=channel_name)
```
rhysyngsun: Can you change this to an `exists()`, so the overhead is a little lower since we'll be executing this for everything?

mbertrand: Change has been made

PR #1190, opened 2018-09-13
Add subscriber when a moderator adds another moderator

### LINKED ISSUE #1168: when adding a moderator, subscribe them as well
As a moderator, when I add another users as a moderator, I expect them to be able to see the channel in the side nav. However, if the new moderator isn't subscribed, they won't see it in the side nav, despite being a moderator of the channel.

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1168 

#### What's this PR do?
The UI for adding a moderator now also makes a call to add the user as a subscriber also. 

The permissions for adding a subscriber are altered slightly so moderators can add other moderators for the channels they moderate as a subscriber. *Note*: this means that via the REST API any moderator can add any user as a subscriber to their channel, even if they are not also adding the user as a moderator.

Add a moderator and view the network tab. You should see a successful addition of the moderator user as a subscriber.

Log in as the moderator user and verify that they have the channel listed on the left side.

### DIFF
--- a/static/js/reducers/channel_subscribers_test.js
+++ b/static/js/reducers/channel_subscribers_test.js
@@ -1,0 +1,54 @@
+// @flow
+import { actions } from "../actions"
+import IntegrationTestHelper from "../util/integration_test_helper"
+import { makeSubscriber } from "../factories/channels"
+
+describe("channelSubscribers reducers", () => {
+  let helper, dispatchThen, newSubscriber
+
+  beforeEach(() => {
+    helper = new IntegrationTestHelper()
+    newSubscriber = makeSubscriber("new_mod")
+    dispatchThen = helper.store.createDispatchThen(
+      state => state.channelSubscribers
+    )
+    helper.addChannelSubscriberStub.returns(Promise.resolve(newSubscriber))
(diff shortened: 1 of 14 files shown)


static/js/reducers/channel_subscribers_test.js
```diff
@@ -0,0 +1,54 @@
...
+  let helper, dispatchThen, newSubscriber
+
+  beforeEach(() => {
+    helper = new IntegrationTestHelper()
+    newSubscriber = makeSubscriber("new_mod")
+    dispatchThen = helper.store.createDispatchThen(
+      state => state.channelSubscribers
+    )
+    helper.addChannelSubscriberStub.returns(Promise.resolve(newSubscriber))
+    helper.deleteChannelSubscriberStub.returns(Promise.resolve())
```
gsidebo: Is this needed?

noisecapella: Ah, I'll remove that

noisecapella: Actually it is needed for the test for deleting a subscription

PR #1192, opened 2018-09-14
Refactored tests and added pytest-lazy-fixture

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

No ticket

#### What's this PR do?
Makes some improvements to some test cases, including:
- Adds [pytest-lazy-fixture](https://github.com/tvorog/pytest-lazy-fixture) library and rewrites some tests to make use of parametrized fixtures
- Refactors some test cases to use proper parametrization

Tests should pass


gsidebo:
@rhysyngsun Did you want to take a look at `jwt_token` changes before this merges?

rhysyngsun:
@gsidebo I think this should be all set because all of the usages I was concerned about aren't around anymore

PR #1206, opened 2018-09-18
Simplify exception handling for emails

### LINKED ISSUE #618: Better logging for failed emails
Locally I get this message due to misconfiguration: `mail.exceptions.SendEmailsException: [(<anymail.message.AnymailMessage object at 0x7f6effc34ef0>, AnymailRequestsAPIError())]`

If I let it raise it gives me enough information to know that it's a misconfigured domain name. Is there a way we can put the error message + stacktrace in the `failed_message_errors` list so that it gets output by the loggers?

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #618 

#### What's this PR do?
Logs an exception whenever an error occurs sending an email

- Enter an invalid value for `MAILGUN_KEY` in your `.env` file
- Reply to some posts & comments
- Eventually, celery errors should appear for each email that the app tries to send.


mail/api.py
```diff
@@ -22,6 +22,7 @@
     pass  # handle failed emails
```
rhysyngsun: Could you update this exception in the usage example?

mail/api.py
```diff
@@ -159,22 +161,13 @@ def messages_for_recipients(recipients_and_contexts, template_name):
...
     for msg in messages:
         try:
             msg.send()
-        except Exception as exc:  # pylint: disable=broad-except
-            failed_message_errors.append(
-                (msg, exc)
-            )
-
-    if failed_message_errors:
-        raise SendEmailsException(failed_message_errors)
```
rhysyngsun: I don't believe we're using this anymore, could you remove the exception definition itself too while you're in here?

mail/api_test.py
```diff
@@ -77,8 +77,10 @@ def test_messages_for_recipients():
         assert msg.subject == "Welcome {}".format(user.profile.name)
 
 
-def test_send_message(mailoutbox):
+def test_send_message(mocker, mailoutbox):
     """Tests that send_messages works as expected"""
+    sendmail = mocker.patch('mail.api.AnymailMessage.send', side_effect=ConnectionError)
```
rhysyngsun: Could you do this as a separate failure test case? this is supposed to be a affirmative test case

mail/api.py
mbertrand: Already did (`mail/exceptions.py` removed).

PR #1213, opened 2018-09-19
Changed HTTP response error handling to behave like form validation

### LINKED ISSUE #1129: Login: aborted wrong password login error message sticks
If you try to login via email and enter an incorrect password, and then click the "Login" button in the header, you still see the same error message on the first login page.

### LINKED ISSUE #1130: Bad UX when using back/forward buttons on login pages
Steps:
1. Enter a recognized email into the login form
2. Click submit/next
3. Enter a bad password
4. Click the browser's back button, then forward
5. Try entering a password again

Result: The request data only includes the password value, and the endpoint expects more properties (flow, partial_token, etc). The resulting error is not handled by the UI, nor should it be. If we don't have the email/flow values in the state, the password page should not be usable at all

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1129 
Fixes #1130 

#### What's this PR do?
- Changes our form component (HOC) to accept some optional functions that detect HTTP response errors and 'coerce' them to form validation errors
  - Example case: a user enters an unrecognized email in a form, and the HTTP response indicates that the email doesn't exist. You can now define a function that inspects that response and returns an object mapping that error message to the 'email' field of the form
- Updates several components to define these functions
- Removes some behavior that treated HTTP endpoint state as UI state (more on that in the background context section)

Try both of these cases and check to see if the given error shows up like a validation message
- Enter an unrecognized email on the login page (/login), password reset page (/password_reset) and the add moderator/contributor pages (/manage/c/edit/<channel_name>/members/moderators/)
- Enter a valid but incorrect password for a user with email auth

Also try that second case (incorrect password), then click back then forward in your browser. You should no longer see the incorrect password message. You should also be able to enter an email on the first login page with no issues.

`static/js/hoc/withForm.js` to understand the function props that were changed/added

The core of the problem being addressed in this PR is that we have been treating HTTP responses as UI state (#1129 has a short thread about this). That presented a few problems for us. If the last request to some endpoint resulted in an error, that really doesn't tell us what we should be showing in the UI. If a form submission results in an error, navigate to a different page, then come back to the same form, the endpoint state will still be the same, but we don't want to continue showing that error any more than we want to show a validation message for a required field that wasn't filled out. Overall, in the context of a form submission, I felt it made the most sense to treat HTTP response errors exactly the same as validation errors.

ss 2018-09-19 at 12 27 45

### DIFF
--- a/static/js/components/auth/AuthEmailForm.js
+++ b/static/js/components/auth/AuthEmailForm.js
@@ -1,7 +1,6 @@
 // @flow
 /* global SETTINGS:false */
 import React from "react"
-import R from "ramda"
 import ReCAPTCHA from "react-google-recaptcha"
 
 import { validationMessage } from "../../lib/validation"
@@ -9,19 +8,14 @@ import { validationMessage } from "../../lib/validation"
 import type { EmailForm } from "../../flow/authTypes"
 import type { FormProps } from "../../flow/formTypes"
 
-type LoginFormProps = {
-  formError: ?string
-} & FormProps<EmailForm>
-
 const AuthEmailForm = ({
   form,
   validation,
   onSubmit,
   onUpdate,
   onRecaptcha,
-  processing,
-  formError
-}: LoginFormProps) => {
+  processing
+}: FormProps<EmailForm>) => {
   return (
     <form onSubmit={onSubmit} className="form">
       <div className="emailfield row">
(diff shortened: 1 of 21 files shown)


gsidebo:
@mbertrand I just looked at the files you mentioned and I can't say I'm clear on any UX differences there might be between those pages and these forms I changed, nor the effort involved to update them if the usage is similar. Without knowing either of those I think as long as those pages are working as expected we should leave that functionality alone

PR #1217, opened 2018-09-19
fix comment share url

### LINKED ISSUE #1216: Share url for comments does not include the hostname, just the relative url
This should have the full url

2018-09-19-135147_1920x1080_scrot

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #1216 
#### What's this PR do?

this fixes the comment share url, so that it is an absolute URL and not a relative one.

check that the URL in the comment share popup is an absolute one.

### DIFF
--- a/static/js/lib/url.js
+++ b/static/js/lib/url.js
@@ -57,11 +57,11 @@ export const embedlyResizeImage = (key: string, url: string, height: number) =>
     url
   )}&height=${height}&grow=false&errorurl=${blankThumbnailUrl()}`
 
+export const absolutizeURL = (url: string) =>
+  new URL(url, window.location.origin).toString()
+
 export const postPermalink = (post: Post): string =>
-  new URL(
-    postDetailURL(post.channel_name, post.id),
-    window.location.origin
-  ).toString()
+  absolutizeURL(postDetailURL(post.channel_name, post.id, post.slug))
 
 // pull the channel name out of location.pathname
 // see here for why this hackish approach was necessary:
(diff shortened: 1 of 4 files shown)


static/js/lib/url.js
```diff
@@ -57,11 +57,11 @@ export const embedlyResizeImage = (key: string, url: string, height: number) =>
...
 
+export const absolutizeURL = (url: string) =>
+  new URL(url, window.location.origin).toString()
+
 export const postPermalink = (post: Post): string =>
-  new URL(
-    postDetailURL(post.channel_name, post.id),
-    window.location.origin
-  ).toString()
+  absolutizeURL(postDetailURL(post.channel_name, post.id))
```
mbertrand: I noticed that the post slug isn't being included but probably should be (for example, `/c/mypost/u` vs `/c/mypost/u/myslug`).  Not the same issue the PR is addressing, but wondering if it should be fixed here anyway since the line of code is being modified?

```python
absolutizeURL(postDetailURL(post.channel_name, post.id, post.slug))
```

pdpinch:
Yeah, if it’s not too much trouble we should include the slug in the shared URL. It helps with SEO. 


> I noticed that the post slug isn't being included but probably should be (for example, /c/mypost/u vs /c/mypost/u/myslug). Not the same issue the PR is addressing, but wondering if it should be fixed here anyway since the line of code is being modified?

static/js/lib/url.js
aliceriot: yeah, I think that's a good idea, I'll go ahead and add it

aliceriot:
@mbertrand I added the slug

PR #1232, opened 2018-09-20
add channel header to channel page

### LINKED ISSUE #989: Channel header with channel banner and channel avatar
As a channel owner, I want to see my branding reflected in the channel page. 

See https://invis.io/9EN908ZQPAX#/311617692_Channel_ClimateX and [Zeplin mockup](https://app.zeplin.io/project/5b181d3bf71408ae4d406b94/screen/5b60e6947d6f9ad509224cee)

### acceptance criteria
- [x] Display channel banner when there is one
  - [x] the banner should always be 105 pixels tall, at all screen widths
  - [ ] when the screen is too narrow to show the full banner width, left-align it
- [x] Transparent background when there is no channel banner
- [x] Display channel avatar when one was uploaded
- [x] Display channel initials avatar when there isn't an avatar uploaded
- [x] Display channel headline, if there is one (see #1094)

### related issues
- #927 Channel Settings: banner image and avatar image
- #990 Add channel tabs to channel template

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #989 

#### What's this PR do?

this adds the channel avatar and the banner to the channel page. it also shows the channel title and the headline up there (the headline only if it's been set).

make sure that the things described above show up. it should look ok and so on. everything should work decently well on mobile as well.

here's what it looks like with the gradient:

channel-header

here's with an image:

channel-headewimager

one possible concern is that certain images will make the headline and channel title text pretty hard to read (as the one above does, for instance).

### DIFF
--- a/static/js/components/admin/EditChannelAppearanceForm.js
+++ b/static/js/components/admin/EditChannelAppearanceForm.js
@@ -50,7 +50,7 @@ export default class EditChannelAppearanceForm extends React.Component<Props> {
                 form.avatar.edit &&
                 URL.createObjectURL(form.avatar.edit)
               }
-              imageSize="medium"
+              imageSize="large"
             />
             <div className="title-container">
               <input
--- a/static/js/components/ChannelSidebar.js
+++ b/static/js/components/ChannelSidebar.js
@@ -10,7 +10,7 @@ import { editChannelBasicURL } from "../lib/url"
 import type { Channel } from "../flow/discussionTypes"
 
 type ChannelSidebarProps = {
-  channel: Channel,
+  channel: ?Channel,
   isModerator: boolean
 }
 
--- a/static/js/components/Grid.js
+++ b/static/js/components/Grid.js
@@ -46,10 +46,14 @@ type CellWidth =
   | typeof twelve
 
 type CellProps = {
-  children: React$Element<*>,
-  width: CellWidth
+  children: any,
+  width: CellWidth,
+  className?: string
 }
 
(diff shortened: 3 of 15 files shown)


pdpinch:
We're thinking about moving the channel edit button from the left nav into the channel header (only visible to channel moderators). Does it make sense to add that to this PR, or a separate one?

aliceriot:
sure, we could do that here

aliceriot:
@pdpinch is there a design for what that should look like?

pdpinch:
@aliceriot never mind about moving the channel edit button. We want to think about that some more.

static/js/components/admin/EditChannelAppearanceForm.js
```diff
@@ -50,7 +50,7 @@ export default class EditChannelAppearanceForm extends React.Component<Props> {
                 form.avatar.edit &&
                 URL.createObjectURL(form.avatar.edit)
               }
-              imageSize="medium"
+              imageSize="large"
```
mbertrand: Maybe change to:
```python
imageSize=CHANNEL_AVATAR_LARGE
```

`CHANNEL_AVATAR_LARGE` was added in `ChannelAvatar.js`, doesn't seem to be referenced anywhere else but this might be a good place to make use of it.

mbertrand: Actually never mind, glanced over that part of the code too quickly, it's fine.

PR #1245, opened 2018-09-21
grey out submit button on create post page

### LINKED ISSUE #1237: Post form: grey out post button when choosing post type
As implemented, the "Post" button competes with the post type buttons ("write something" and "share a link"). It should be greyed out until the user choses a post type. 

currently:

compete

as designed: 

draft freeform start

aliceriot:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

closes #1237 

#### What's this PR do?

This PR just makes it so that, when you're creating a post, the 'submit' button will be greyed out until you enter a title and pick a post type. it doesn't actually disable the button (by passing `disabled={true}`) but instead sets a `.disabled` class on it. we do want the user to be able to click, because unless they've filled anything out they'll see validation errors anyhow.

make sure that the behavior described above (and in the issue) is implemented.

### DIFF
--- a/static/js/components/CreatePostForm.js
+++ b/static/js/components/CreatePostForm.js
@@ -152,6 +152,7 @@ export default class CreatePostForm extends React.Component<Props> {
     }
 
     const { postType, title } = postForm
+    const readyToSubmit = postType !== null && title !== ""
 
     return (
       <div className="new-post-form">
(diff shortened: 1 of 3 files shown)


aliceriot:
ah yes I forgot, we allow the user to submit just a title because we wanted to allow title-only posts like that (just like you can make on reddit). so we probably don't want to grey out the button in that case. @pdpinch do you have any thoughts on this?

it seems like the only case where we want the button to be grey is if the user has not yet entered a title.

PR #1248, opened 2018-09-21
Show post type buttons after switching channels if empty

### LINKED ISSUE #1215: Switching channels on compose page causes post type buttons to hide permanently
#### Steps to reproduce

- Have a mix of channels that support all post types and some that support only one
- Switch between channels that support all types (the type selection buttons should be visible) and ones that only support 1, and then back again

Expected behavior(?): if you didn't type anything in, the UI should go back to offering you the buttons.
Actual behavior: the UI sticks to the post type from the more restricted channel

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes #1215 

#### What's this PR do?
Alters the `CreatePostPage.componentDidUpdate` logic so that if the new post form is switched to a channel that accepts both text and link posts, and both the text and url form values are null, those input fields will be hidden if visible and the post type buttons will be displayed instead.

- Create a few channels, one that accepts any post types, one that accepts only link posts, one that accepts only text posts.
- From the home page, click 'Compose' to create a new post. Select the links-only channel.  The URL field should appear. Don't put anything in it, and switch to the 'any' channel.  The URL field should disappear and you should see the post type buttons.
- Select the link post type.  Don't put anything in the URL field.  Switch to the links-only channel.  The URL field should still be visible.
- Enter a value in the URL field.  Either before or after embedly shows a screenshot, change back to the 'any' channel.  The URL field or embedly screenshot should still display.
- Change to the text-only channel.  The URL field or embedly screenshot should be replaced with a textarea.

`CreatePostPage.componentDidUpdate`

### DIFF
--- a/static/js/containers/CreatePostPage.js
+++ b/static/js/containers/CreatePostPage.js
@@ -20,7 +20,7 @@ import { newPostForm } from "../lib/posts"
 import { postDetailURL } from "../lib/url"
 import { getChannelName } from "../lib/util"
 import { formatTitle } from "../lib/title"
-import { validatePostCreateForm } from "../lib/validation"
+import { emptyOrNil, validatePostCreateForm } from "../lib/validation"
 import { ensureTwitterEmbedJS, handleTwitterWidgets } from "../lib/embed"
 import { anyErrorExcept404 } from "../util/rest"
 
@@ -87,20 +87,24 @@ class CreatePostPage extends React.Component<CreatePostPageProps> {
     // we may need to change out the postType under certain conditions
     // if the user switches channels and the new channel isn't LINK_TYPE_ANY,
     // and the current postType is null or a postType which is incompatible with the
-    // new channel, we want to set the postType to be the link_type for the channel
+    // new channel, we want to set the postType to be the link_type for the channel.
+    // If the new channel is LINK_TYPE_ANY and both text and url values are empty,
+    // we want to set the postType to null to display the post type options.
     if (
       postForm &&
       channel &&
       ((prevProps.channel && prevProps.channel.name !== channel.name) ||
         !prevProps.channel) &&
-      channel.link_type !== LINK_TYPE_ANY &&
+      (channel.link_type !== LINK_TYPE_ANY ||
+        emptyOrNil(postForm.value.text + postForm.value.url)) &&
       channel.link_type !== postForm.value.postType
     ) {
       dispatch(
         actions.forms.formUpdate({
           ...CREATE_POST_PAYLOAD,
           value: {
-            postType:  channel.link_type,
+            postType:
+              channel.link_type === LINK_TYPE_ANY ? null : channel.link_type,
             url:       "",
             text:      "",
             thumbnail: null
--- a/static/js/containers/CreatePostPage.js
+++ b/static/js/containers/CreatePostPage.js
@@ -83,8 +83,11 @@ class CreatePostPage extends React.Component<CreatePostPageProps> {
 
   shouldTriggerUpdate() {
     const { postForm } = this.props
-    return (emptyOrNil(postForm.value.text + postForm.value.url) &&
-          !emptyOrNil(postForm.value.postType))
+    return (
+      postForm &&
+      emptyOrNil(postForm.value.text + postForm.value.url) &&
+      !emptyOrNil(postForm.value.postType)
+    )
   }
 
   componentDidUpdate(prevProps: CreatePostPageProps) {
--- a/static/js/containers/CreatePostPage.js
+++ b/static/js/containers/CreatePostPage.js
@@ -85,7 +85,7 @@ class CreatePostPage extends React.Component<CreatePostPageProps> {
     const { postForm } = this.props
     return (
       postForm &&
-      emptyOrNil(postForm.value.text + postForm.value.url) &&
+      R.all(emptyOrNil, [postForm.value.text, postForm.value.url]) &&
       !emptyOrNil(postForm.value.postType)
     )
   }
--- a/static/js/containers/CreatePostPage.js
+++ b/static/js/containers/CreatePostPage.js
@@ -96,7 +96,8 @@ class CreatePostPage extends React.Component<CreatePostPageProps> {
       ((prevProps.channel && prevProps.channel.name !== channel.name) ||
         !prevProps.channel) &&
       (channel.link_type !== LINK_TYPE_ANY ||
-        emptyOrNil(postForm.value.text + postForm.value.url)) &&
+        (emptyOrNil(postForm.value.text + postForm.value.url) &&
+          !emptyOrNil(postForm.value.postType))) &&
       channel.link_type !== postForm.value.postType
     ) {
       dispatch(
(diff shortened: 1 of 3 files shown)


static/js/containers/CreatePostPage_test.js
```diff
@@ -414,66 +414,83 @@ describe("CreatePostPage", () => {
...
+      ([fromLinkType, toLinkType, formValue, hasUrlOrText, shouldDispatch]) => {
+        it(`${shouldIf(
+          shouldDispatch
+        )} dispatch FORM_UPDATE if the postType is ${String(
+          formValue
+        )} when it goes from ${String(
+          fromLinkType
+        )} to ${toLinkType} and user has ${
+          hasUrlOrText ? "" : "not "
+        }entered a value`, () => {
```
gsidebo: This test case is huge now. If it's clean enough to do so, please create a separate test case for switching from one post type to another

static/js/containers/CreatePostPage.js
```diff
@@ -87,20 +87,25 @@ class CreatePostPage extends React.Component<CreatePostPageProps> {
...
+    // we want to set the postType to null to display the post type options.
     if (
       postForm &&
       channel &&
       ((prevProps.channel && prevProps.channel.name !== channel.name) ||
         !prevProps.channel) &&
-      channel.link_type !== LINK_TYPE_ANY &&
+      (channel.link_type !== LINK_TYPE_ANY ||
+        (emptyOrNil(postForm.value.text + postForm.value.url) &&
+          !emptyOrNil(postForm.value.postType))) &&
```
gsidebo: Shouldn't the added logic just be a check to see if the `postType` value changed? I'm pretty sure you can just compare `prevProps.postForm.value.postType` and `this.props.postForm.value.postType` for that. If that value changes, we don't need to know anything else to determine that the form needs an update, right? 

Also, I would create a named helper function for this new logic. The existing conditional is really verbose and it would help to have a named function to read here instead of yet another comparison.

mbertrand: If you start creating a post in a link-only channel, the link field will display by default.  If you then switch to a channel that allows both links and text posts, whether or not the form should update depends on whether anything was entered in that link field yet.  If not, the link field should be hidden and the 'Write something' & 'Share a link' buttons should be shown instead.  Otherwise, the URL entered by the user should continue to display.

mbertrand: I put the additional logic into a separate function

static/js/containers/CreatePostPage_test.js
mbertrand: Test has been split up.

static/js/containers/CreatePostPage.js
```diff
@@ -81,26 +81,38 @@ class CreatePostPage extends React.Component<CreatePostPageProps> {
...
+    // new channel, we want to set the postType to be the link_type for the channel.
+    // If the new channel is LINK_TYPE_ANY and both text and url values are empty,
+    // we want to set the postType to null to display the post type options.
     if (
       postForm &&
       channel &&
       ((prevProps.channel && prevProps.channel.name !== channel.name) ||
         !prevProps.channel) &&
-      channel.link_type !== LINK_TYPE_ANY &&
+      (channel.link_type !== LINK_TYPE_ANY || this.shouldTriggerUpdate()) &&
```
gsidebo: This boolean and the comment above it were confusing me enough that I tinkered with it in a branch. This is what I came up with:

```javascript
    // If a newly-selected channel cannot be rendered with the same form state as the
    // previously-selected channel, we need to reset the form state.
    if (
      postForm &&
      channel &&
      prevProps.channel &&
      // The selected channel's link type is different from the previously-selected channel
      channel.link_type !== prevProps.channel.link_type &&
      (
        // The selected channel accepts any type of post, and no input has been entered
        (channel.link_type === LINK_TYPE_ANY && R.all(emptyOrNil)(postForm.value.url, postForm.value.text)) ||
        // The selected channel only takes specific posts, and the post type of the
        // current form doesn't match the channel's post type
        (channel.link_type !== LINK_TYPE_ANY && channel.link_type !== postForm.value.postType)
      )
    ) {
```

It's still a big statement, but to me it's much more clear about what should trigger a form reset. As far as I could tell, the channel name comparison and the check for an empty post type are unnecessary. The 2 last conditionals (the ones that make up the 'or' expression) could be moved to helper functions, but I think it's OK leaving them inline with a comment. `R.all(emptyOrNil)` could be a new helper function if you feel like doing that. Let me know what you think about all that

Also, if you decide to go with something like this and the tests all explode in horrible ways, we could move this to a separate tech debt issue

static/js/containers/CreatePostPage_test.js
```diff
@@ -512,5 +507,76 @@ describe("CreatePostPage", () => {
...
+      ]) => {
+        it(`${shouldIf(
+          shouldDispatch
+        )} dispatch FORM_UPDATE if the postType is ${String(
+          formValue
+        )} when it goes from ${String(
+          fromChannelType
+        )} to ${toChannelType} and user has ${
+          hasUrlOrText ? "" : "not "
+        }entered a value`, () => {
```
gsidebo: Oof... I think I misinterpreted this test before when I suggested we should split them up. This is basically a carbon copy of the test case above with some added , and I now I understand why it needs all of this logic. So going against my ill-conceived suggestion before, this should definitely be just 1 parametrized test case.

I think the description and parameters could be clearer though. `formValue` should be `postType` or `formPostType`, `hasUrlOrText` could be something like `hasInput`, and the description could be changed to something like this:

```javascript
it(`${shouldIf(
          shouldDispatch
        )} reset form if channel type changes from ${String(
          fromChannelType
        )} to ${toChannelType}, postType=${String(
          formPostType
        )}, user input is ${
          hasInput ? "not " : ""
        }empty`
```

static/js/containers/CreatePostPage.js
mbertrand: Using the above code, if you:
- Choose a text-only channel and enter some text
- Change to a channel that accepts both text and links

The text field will disappear along with your text.  This should not happen unless it is blank.  With the current code in the PR, it behaves as expected.

gsidebo: Ah, the problem with with the `R.all` call. Should be:

`R.all(emptyOrNil, [postForm.value.url, postForm.value.text])`

mbertrand: - Click 'Compose' from the home page
- Without selecting a channel, click 'Write something' to show the textarea but don't write anything.
- Now select a links-only channel.  The textarea does not disappear as it should.

Also a few unit tests break related to this scenario (from null postType to link/text).  I agree the current conditional logic is complex and not that easy to follow, but it seems to handle all the possible permutations correctly and passes all the tests, so I'd prefer to keep it as is.

gsidebo: Sounds good. I'll create a tech debt issue. I appreciate your patience!

mbertrand: No problem, thanks for taking the time to try finding a simpler solution.

static/js/containers/CreatePostPage_test.js
mbertrand: Test has been updated with new parameters/description

PR #1269, opened 2018-09-26
fix display of auth cards at all screen widths

### LINKED ISSUE #1267: Viewport a little too wide for iPhone SE
I'm not sure what the best solution is to this, but it looks like the viewport is too wide when using an iPhone SE. There's a little bit of horizontal scroll and some buttons go off screen. 

For example:

bs_realios_mobile_iphone se-11 0 2

aliceriot:

closes #1267 

#### What's this PR do?

fixes the display for the auth cards at all page widths

Look at various auth pages and make sure that the display bug shown in the attached issue doesn't crop up anywhere. make sure you're testing in the chrome device simulator or a mobile device emulator.

### DIFF
--- a/static/scss/auth.scss
+++ b/static/scss/auth.scss
@@ -1,27 +1,25 @@
+.content .auth-page .main-content {
+  max-width: 470px;
+}
+
 @include breakpoint(desktop) {
   .content .auth-page .main-content {
     margin-top: 70px !important;
-    max-width: 470px;
-  }
-
-  .auth-page {
-    form,
-    .card-contents {
-      width: 350px;
-    }
   }
 }
 
 .auth-page {
   @include breakpoint(mobile) {
     width: 95%;
-    margin: auto;
 
-    form,
-    .actions,
+    .main-content,
     .card-contents {
-      width: 100%;
+      margin: auto;
     }
+  }
+
+  form {
+    width: 100%;
   }
 
   .form-header {
@@ -33,8 +31,22 @@
     color: $font-grey;
   }
 
-  .link-button {
+  .link-button,
+  button[type="submit"] {
     margin-right: 0;
+  }
+
+  button {
+    width: 100%;
+
+    &[type="submit"] {
+      width: 35%;
+    }
+
+    &.submit-password-reset {
+      width: 50%;
+      padding: 9px 19px;
+    }
   }
 
   .touchstone-text-logo {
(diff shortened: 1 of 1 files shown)


static/scss/auth.scss
```diff
@@ -15,12 +15,21 @@
...
+      margin: auto;
+      max-width: 470px;
+    }
 
     form,
-    .actions,
-    .card-contents {
+    .actions {
       width: 100%;
+      max-width: 350px;
```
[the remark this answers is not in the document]
aliceriot: I think we don't need it on `.actions` (I looked at it's declared already below), but this happens if it's not set on `form`:

weird_withd

vs with it:

weird_withdfixed

but I think I can change it to have one `max-width` value for `form` at all display widths

static/scss/auth.scss
```diff
@@ -15,12 +15,21 @@
...
+    }
+
+    .card-contents {
+      // this sets the width to 100% minus the padding
+      width: calc(100% - 36px);
```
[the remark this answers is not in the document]
aliceriot: yeah, I think I inadvertently left that in from something I was trying initially

static/scss/auth.scss
```diff
@@ -15,15 +15,18 @@
...
-    .card-contents {
-      width: 100%;
+    .main-content {
+      margin: auto;
+      max-width: 470px;
     }
   }
 
+  form {
+    max-width: 350px;
```
gsidebo: I guess this is a nitpick, but I think the width should be fixed on `.card-contents`, not `form`. The 350px width is defined for `.card-contents` when the window is >800px wide. I think it should just be set to that width for any window width above mobile. That way you wouldn't have the `.card-contents` div suddenly growing to ~450px wide when the screen width drops below 800px. I don't think there's any real reason for that

ss 2018-09-27 at 17 27 24

aliceriot: are you seeing the card contents div growing super wide like that? I can't repro that on my machine. The reason I set the the max-width on `form` was the mobile display, if I set the `max-width` on `.card-contents` instead it breaks the layout with flexbox:

max-width-on-card

there may be another way around it, but setting the max-width on the form and actions was the simplest way I found.

aliceriot: so I can refactor it to set `max-width: 350px;` on `.card-contents` at all widths, and as long as I keep `margin: auto;` on at mobile the `width: 100%;` on `.form` and `.actions` will preserve the layout - I can do that if you think it's more logical.

Ferdi: @aliceriot can you right-align the "next" button with the "email" box ?

aliceriot: yeah, that's pretty simple

aliceriot:
k should be fixed now

PR #1279, opened 2018-09-28
Replace underscores with dashes in post slugs

### LINKED ISSUE #1278: SEO: use dashes in slugs instead of underscores
As a user, I find it easier to read and share URLs with dashes in them than underscores. 

the post.url returned from reddit includes a slug, but it uses underscores as the word separator instead of dashes. However, the content of the slug isn't used to look up the resource anyway. So we can simply replaces the `_` with `-` in the URLs before returning them to the front-end. 

### to be complete, the slugs need to change on:
- [ ] post card links to detail pages
- [ ] comment card date/time links
- [ ] emails
- [ ] share URLs

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #1278 

#### What's this PR do?
- Modifies `channels.utils.get_reddit_slug ` to replace underscores in the slug with dashes 

Click on links posts that have spaces in their titles, and links to comments for those posts.  The post slug in each URL should have dashes but no underscores.

### DIFF
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -278,8 +278,8 @@ class BasePostSerializer(RedditObjectSerializer):
         return None
 
     def get_slug(self, instance):
-        """Returns the post slug from reddit and replaces '_' with '-'"""
-        return get_reddit_slug(instance.permalink).replace('_', '-')
+        """Returns the post slug"""
+        return get_reddit_slug(instance.permalink)
 
     def get_author_name(self, instance):
         """get the authors name"""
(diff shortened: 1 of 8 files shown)


mbertrand:
@rhysyngsun The actual value of the post slug is currently ignored, it can be anything and the page will still load correctly.  What do you think of changing the `CanonicalLink` on the `PostPage` to:
```javascript
<CanonicalLink relativeUrl={postDetailURL(channel.name, post.id, post.slug, commentID)} />
```
and modifying the `postDetailURL` function to take another optional parameter `commentID`?  That way the CanonicalLink should always be correct regardless of what slug in the URL is used to get to the page.

rhysyngsun:
@mbertrand that's fine, except use `commentPermalink` instead.

PR #1315, opened 2018-10-05
Channel-specific analytics should trigger on direct URL load

### LINKED ISSUE #1306: Channel-specific GA tracking not triggered when going direct to a URL
If a channel has its own Google Analytics tracking id, hits will be sent to GA only if that channel or it's posts is reached via links in the app.  Going directly to one of those URL's (for instance from an email) does not trigger a google analytics hit.

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes #1306

#### What's this PR do?
Triggers a channel-specific pageview to Google Analytics when a channel/post URL is loaded without navigating to it from within the app.

- Set `GA_TRACKING_ID` to any value in your `.env` file.
- In django admin, enter a fake google analytics tracking id for a channel (different from above).
- Manually enter the URL for that channel, and a URL for any post in that channel, into a browser tab, with network console open.  For each URL, two `collect` calls to Google Analytics should be made, one of which should be for the above tracking id, the other for the main tracking id.
- Navigate to Home, then navigate back to the channel, and then a post in that channel.  Two calls should be made for each of that channel's URLs.

### DIFF
--- a/static/js/hoc/withChannelTracker_test.js
+++ b/static/js/hoc/withChannelTracker_test.js
@@ -56,11 +56,12 @@ describe("withTracker", () => {
   })
 
   //
-  ;[true, false].forEach(updateChannel => {
+  ;[[true, 4], [false, 2]].forEach(([missingPrevChannel, gaCalls]) => {
     it(`${shouldIf(
-      updateChannel
-    )} call loadGA() on componentDidUpdate`, async () => {
+      missingPrevChannel
+    )} call google analytics on componentDidUpdate`, async () => {
       channel.ga_tracking_id = "UA-FAKE-01"
+      const prevChannel = missingPrevChannel ? null : channel
       window.location = "http://fake/c/path"
       const { wrapper } = await render(
         {},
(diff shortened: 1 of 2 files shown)


static/js/hoc/withChannelTracker_test.js
```diff
@@ -42,11 +48,30 @@ describe("withTracker", () => {
...
+      const { wrapper } = await render(
+        {},
+        { location: window.location, channel: channel }
+      )
+      const prevProps = {
+        location: window.location,
+        channel:  updateChannel ? null : channel
+      }
+      wrapper.instance().componentDidUpdate(prevProps)
+      assert.equal(gaGaStub.callCount, updateChannel ? 4 : 2)
```
gsidebo: You can parametrize the `prevProps.channel` value at line 71 as well as this expected call count. Also, it's a little odd that the test case description says that `loadGA()` shouldn't be called in one case, but the assertion says that it should be called twice

gsidebo:
Comment above should be a requested change (I misclicked). Otherwise everything works as expected

PR #1336, opened 2018-10-09
Added new authentication class to ignore expired JWTs

### LINKED ISSUE #1331: Invalid JWT token cookies end up in an infinite redirect loop on the login page
We still request the channel list for this page and an expired JWT token is causing an authentication error, which then triggers an infinite redirect loop here. Solutions in preferred order are:

- Update the auth middleware to ignore the cookie if the internal JWT is expired
- Send a header to purge the cookie if it's expired

rhysyngsun:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Fixes #1331 

#### What's this PR do?
Replaces the authentication class for JWTs with one that ignores JWTs that have expired, thereby avoiding an auth error.

(Required)

(Optional)

(Optional)

(Optional)

(Optional)


open_discussions/auth_utils.py
```diff
@@ -0,0 +1,38 @@
...
+        str: the decoded username
+    """
+
+    signer = signing.TimestampSigner()
+    try:
+        return signer.unsign(
+            token,
+            max_age=settings.OPEN_DISCUSSIONS_UNSUBSCRIBE_TOKEN_MAX_AGE_SECONDS
+        )
+    except:  # pylint: disable=bare-except
```
noisecapella: Is it possible to be more specific here?

open_discussions/authentication.py
```diff
@@ -14,39 +15,25 @@
...
+        try:
+            # try to decode the value just to see if it's expired
+            from rest_framework_jwt.settings import api_settings
+            jwt_decode_handler = api_settings.JWT_DECODE_HANDLER
+            jwt_decode_handler(value)
+        except jwt.ExpiredSignature:
+            # if it is expired, treat it as if the user never passed a token
+            return None
+        except:  # pylint: disable=bare-except
+            pass
```
noisecapella: What exceptions is this silencing? Should this raise instead if they're unknown?

open_discussions/settings.py
```diff
@@ -697,7 +697,7 @@ def get_all_config_keys():
     ),
     'DEFAULT_AUTHENTICATION_CLASSES': (
         'rest_framework.authentication.SessionAuthentication',
-        'rest_framework_jwt.authentication.JSONWebTokenAuthentication',
+        # 'open_discussions.authentication.IgnoreExpiredJwtAuthentication',
```
noisecapella: Was this meant to be uncommented?

rhysyngsun: Nope

open_discussions/authentication.py
rhysyngsun: It's silencing all exceptions because the token gets parsed later in the request lifecycle and we want that code to do the error handling of whatever other exceptions there are. We're only concerned with the more specific exception handled above for purposes of ignoring the token.

rhysyngsun:
@noisecapella all set

open_discussions/authentication.py
noisecapella: Ok, can you add that as a comment?

rhysyngsun:
@rhysyngsun fix is not reproducible

rhysyngsun:
Added a useful debug statement

PR #1340, opened 2018-10-10
Hide useless asteroid warnings when running tests

### LINKED ISSUE #1349: Hide astroid warnings from pylint
Pylint outputs thousands of warnings which make it difficult to find the relevant test output details.

mbertrand:
Closes #1349

#### What's this PR do?
Prevents asteroid warnings (deprecations, etc) from showing up after running tests locally.

Run `docker-compose run web tox`.  There should not be a huge stream of asteroid warnings afterward.

### DIFF
--- a/authentication/pipeline/user.py
+++ b/authentication/pipeline/user.py
@@ -20,7 +20,7 @@ from profiles import api as profile_api
 from profiles.utils import update_full_name
 
 
-def validate_email_auth_request(strategy, backend, user=None, *args, **kwargs):  # pylint: disable=unused-argument
+def validate_email_auth_request(strategy, backend, user=None, **kwargs):  # pylint: disable=unused-argument
     """
     Validates an auth request for email
 
@@ -41,7 +41,7 @@ def validate_email_auth_request(strategy, backend, user=None, *args, **kwargs): 
     return {}
 
 
-def get_username(strategy, backend, user=None, *args, **kwargs):  # pylint: disable=unused-argument
+def get_username(strategy, backend, user=None, **kwargs):  # pylint: disable=unused-argument
     """
     Gets the username for a user
 
@@ -62,7 +62,7 @@ def get_username(strategy, backend, user=None, *args, **kwargs):  # pylint: disa
 
 @partial
 def require_password_and_profile_via_email(
-        strategy, backend, user=None, flow=None, current_partial=None, *args, **kwargs
+        strategy, backend, user=None, flow=None, current_partial=None, **kwargs
 ):  # pylint: disable=unused-argument
     """
     Sets a new user's password and profile
@@ -102,7 +102,7 @@ def require_password_and_profile_via_email(
 
 @partial
 def require_profile_update_user_via_saml(
-        strategy, backend, user=None, is_new=False, *args, **kwargs
+        strategy, backend, user=None, is_new=False, **kwargs
 ):  # pylint: disable=unused-argument
     """
     Sets a new user's password and profile, and updates the user first and last names.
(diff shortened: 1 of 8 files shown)


noisecapella:
It seems like this prevents pylint from running... I think we need to revert the change in the other repo that @amir-qayyum-khan made

noisecapella:
I think the trailing comma is what's preventing pylint from issuing any errors. But if I remove the trailing comma and leave 'astroid' I see all the usual warnings and errors like I do on master

mbertrand:
Sorry about that, I thought pylint was responsible for complaining about pep8 violations so assumed everything was still working when I added some extra lines in a file and got an error about it.

Took a different approach and updated astroid, pylint, etc to latest versions.  This got rid of the astroid errors but also introduced some new lint violations.  I added frequent ones to the disable list in `pylintrc`, then added `disable` lines where appropriate for the rest.

pylintrc
```diff
@@ -17,6 +17,7 @@ ignored-classes=
 ignored-modules=
 	six,
 	six.moves,
+	betamax,
 
 [MESSAGES CONTROL]
-disable = no-member, old-style-class, no-init, too-few-public-methods, abstract-method, invalid-name, too-many-ancestors, line-too-long, no-self-use, len-as-condition, no-else-return
+disable = no-member, old-style-class, no-init, too-few-public-methods, abstract-method, invalid-name, too-many-ancestors, line-too-long, no-self-use, len-as-condition, no-else-return, cyclic-import, duplicate-code, inconsistent-return-statements, keyword-arg-before-vararg, consider-using-set-comprehension, inconsistent-return-statements, raising-format-tuple
```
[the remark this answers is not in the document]
mbertrand: Removed keyword-arg-before-vararg, consider-using-set-comprehension, raising-format-tuple from the ignore list

authentication/pipeline/user.py
```diff
@@ -20,7 +20,7 @@
 from profiles.utils import update_full_name
 
 
-def validate_email_auth_request(strategy, backend, user=None, *args, **kwargs):  # pylint: disable=unused-argument
+def validate_email_auth_request(strategy, backend, user=None, **kwargs):  # pylint: disable=unused-argument
```
noisecapella: We should still keep these around just in case there is an extra positional argument and we don't want to error on that. But we can move `*args` before `user=None` to fix the warning

channels/serializers.py
```diff
@@ -900,4 +900,4 @@ def get_reasons(self, instance):
         Returns:
             list of str: list of reasons a post/comment has been reported for
         """
-        return sorted(list(set([report[0] for report in instance.user_reports + instance.mod_reports])))
+        return sorted(list({report[0] for report in instance.user_reports + instance.mod_reports}))
```
noisecapella: You can also remove the `list` at the beginning since `sorted` will accept a set as input

authentication/pipeline/user.py
mbertrand: Moving `*args` before `user` causes a test to fail (`auth.pipeline.user_test.test_get_username`).  I think it might be best to just leave these functions in their original state and add an ignore for `keyword-arg-before-vararg` to the top, so as to avoid any nasty surprises with social-auth.

channels/serializers.py
mbertrand: 👍

authentication/pipeline/user.py
mbertrand: I reverted the code for `auth.pipeline.user` to be the same as master and added `# pylint: disable=keyword-arg-before-vararg` to the top of the file, I think this is safer for now.

Issue #1325, opened 2018-10-08
Email: front page template round 2


pdpinch:
Current version (iPhone 6s):

![img_d106984fb3ba-1](https://user-images.githubusercontent.com/430126/46623363-67b02800-cafb-11e8-9389-6ed78c65d0fb.jpeg)

New version, on Zeplin: https://zpl.io/2pBW1dE (updated 10/10)

### acceptance criteria:
- [x] stop centering the text at the narrowest window width
- [x] change title to "Top Stories for You"
- [x] Tighten up the vertical spacing between elements, to make it clearer where the card boundaries are (in other words, follow the Zeplin spacing)
- [x] restore the channel name
- ~~restore a count of comments~~
- [x] change subject line to "[`sitename`] Top stories for You"

### questions:
- Can we display thumbnails from link posts, or is there more infrastructure we need for that?
  - A: maybe 
- How much weight does the "comment" styling add?
  - A: doesn't matter, we're cutting the comment count

pdpinch:
I made a few changes to the description: we're dropping the comment count. And I simplified the email subject.

rhysyngsun:
This should be closed.

PR #1346, opened 2018-10-11
Third pass of email templates

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#1325 

#### What's this PR do?
Updates email templates and fixes a bug where `{{ site_name }}` is not being set in the email context. Also updated quite a few tests to use the default response data stuff @gsidebo wrote to cut down on the amount of code.

Go to `/__emaildebugger__/` and test render the frontpage email and verify it matches the mockup (sans thumbnail images which we're still skipping)

#### Screenshots
screenshot_2018-10-12 email debugger

### DIFF
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -5,6 +5,7 @@ from datetime import (
     datetime,
     timezone,
 )
+from urllib.parse import urlparse
 
 from django.contrib.auth import get_user_model
 from praw.models import Comment, MoreComments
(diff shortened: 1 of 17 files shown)


Ferdi:
@rhysyngsun 
- Lucas is saying the font type looks different in your screenshot ?
- Can you align the "Read more" button left ?
- Can you align `--`  left ?

rhysyngsun:
@Ferdi the above screenshot was rendered in the browser-based email debugger I wrote, the font definition in the template we're using doesn't appear to work there, but it's definitely working in email clients so I think we can just trust that will render correctly in the actual emails.

Reason for using this debugger is there's not a reliable way to preview an email outside of CI/RC/prod (images break due to email client's image proxy), so local testing requires a combination of the browser-based testing and client-based testing to get enough confidence in the result prior to merging.

I've updated the code and screenshot to reflect the other feedback.

PR #1350, opened 2018-10-12
Use black code formatter

rhysyngsun:
N/A

#### What's this PR do?
Adds the black code formatter

Everything should still work because the code has just been reformatted


tox.ini
```diff
@@ -10,6 +10,7 @@ deps =
     -r{toxinidir}/test_requirements.txt
 commands =
     {toxinidir}/scripts/test/check_pip.sh
+    black --check
```
noisecapella: If the user isn't specifying an arg to `tox` this won't do anything

rhysyngsun:
@noisecapella I changed the tox line to `black --check .` and also updated the README with instructions on how to format the code

PR #1373, opened 2018-10-15
Index user profiles in Elasticsearch

### LINKED ISSUE #1317: Search: index user profiles

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #1317

#### What's this PR do?
- Adds a profile index
- Adds a profile mapping and refactors common mapping fields
- Index new profiles
- Update the profile document when a profile is changed.
- Update all relevant post/comment docs when a profile name or avatar is changed.
- Corrects a mismatch between channel name and title when saving `Channel` objects.

- Run `python manage.py recreate_index`.  When done, there should be a `discussions_local_profile_default` index with documents for all existing profiles.
- Edit your profile, changing bio and headline.  The changes should be reflected in the index doc for your profile.
- Change the name and avatar of your profile.  The `author_name` and `author_avatar_small` fields of all your posts and comments should also be changed in ES.
- Register a new user by email, and another new user via Micromasters.  When complete, the ES doc for each profile should exist and have the correct info.


search/indexing_api.py
```diff
@@ -33,34 +33,49 @@
 SCRIPTING_LANG = 'painless'
 UPDATE_CONFLICT_SETTING = 'proceed'
 
-BASE_CONTENT_TYPE = {
+BASE_OBJECT_TYPE = {
     'object_type': {'type': 'keyword'},
     'author_id': {'type': 'keyword'},
     'author_name': {'type': 'keyword'},
+    'author_avatar_small': {'type': 'keyword'},
```
mbertrand: @noisecapella I've added `author_avatar_small` to post and comment docs (derived from `profile_image` contained in their respective base serializers); profile docs will also include this and `author_avatar_medium` as well - does that sound good to you?

search/tasks.py
```diff
@@ -63,19 +63,34 @@ def update_document_with_partial(doc_id, partial_data, object_type):
...
     """
     Task that makes a request to update a field value for all ES documents that match some query.
     """
-    return api.update_field_values_by_query(
-        query, field_name, field_value, object_types
-    )
+    return api.update_field_values_by_query(query, field_dict, object_types)
+
+
+@app.task(autoretry_for=(RetryException,), retry_backoff=True, rate_limit="600/m")
```
noisecapella: Since we're not contacting reddit when indexing profiles, I don't think we need this decorator or any of the rate limit or exponential backoff logic.

search/indexing_api.py
```diff
@@ -249,7 +274,7 @@ def sync_comments(serialized):
 
     for item in serialized:
         sync_comment_model(
-            channel_name=item["channel_title"],
+            channel_name=item["channel_name"],
```
noisecapella: Good catch!

noisecapella:
Functionality works great

authentication/pipeline/user.py
```diff
@@ -90,6 +91,7 @@ def require_password_and_profile_via_email(
     if "name" in data:
         profile.name = data["name"]
         profile.save()
+    index_new_profile(profile)
```
noisecapella: Instead of putting calling `index_new_profile` here and below, can we add a signal so it gets called every time `profile.save()` happens?

search/tasks.py
mbertrand: ✅

search/indexing_api.py
mbertrand: I think some erroneously created channels will need to be manually deleted from admin.

profiles/views_test.py
```diff
@@ -8,11 +8,18 @@
 from social_django.models import UserSocialAuth
 
 from authentication.backends.micromasters import MicroMastersAuth
+from open_discussions.features import INDEX_UPDATES
```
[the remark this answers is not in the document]
mbertrand: For tests to pass, `INDEX_UPDATES` needs to be `False` or the index functions need to be mocked.  I went for the first option, but on second thought mocking the functions seems like a better choice.

authentication/pipeline/user.py
mbertrand: The `Profile` model already has a custom `save` function, how about I put it there?

mbertrand:
@noisecapella are there any other changes that would make sense to do in this PR to unblock what you're working on?

search/indexing_api.py
noisecapella: Ah yes, I bet that's where some of the extra `Channel` objects are coming from

PR #1395, opened 2018-10-18
Fixed signup page UI issues

### LINKED ISSUE #1347: fixup sign-up page layout
The signup page is looking pretty ragged (desktop and mobile)

### acceptance criteria
- [x] if possible, size the captcha box to be the same width as the input field and the touchstone button
- [x] change the "Next" button text to "Sign Up" 
- [x] change the `<hr>` width to be the same as the  input field and the touchstone button

rhysyngsun:
#### Pre-Flight checklist

  - [x] Desktop screenshots 
- [x] Testing

Fixes #1347 

#### What's this PR do?
Fixes various signup page UI issues:
- Fixed maximum width of for to 304px to match largest width of ReCAPTCHA
- Dynamically scale ReCAPTCHA on resizes and smaller devices
- Renamed the button to "Sign up" and adjusted its width to a pixel value instead of percentage so it doesn't get scrunched on mobile widths

Test the UI at various browser sizes and make sure things look good.

screen shot 2018-10-18 at 16 46 43
screen shot 2018-10-18 at 16 56 36

### DIFF
--- a/static/js/components/auth/AuthEmailForm.js
+++ b/static/js/components/auth/AuthEmailForm.js
@@ -2,49 +2,116 @@
 /* global SETTINGS:false */
 import React from "react"
 import ReCAPTCHA from "react-google-recaptcha"
+import styled from "styled-components"
+import debounce from "lodash/debounce"
 
 import { validationMessage } from "../../lib/validation"
 
 import type { EmailForm } from "../../flow/authTypes"
 import type { FormProps } from "../../flow/formTypes"
 
-const AuthEmailForm = ({
-  form,
-  validation,
-  onSubmit,
-  onUpdate,
-  onRecaptcha,
-  processing
-}: FormProps<EmailForm>) => {
-  return (
-    <form onSubmit={onSubmit} className="form">
-      <div className="emailfield row">
-        <input
-          type="email"
-          name="email"
-          value={form.email}
-          onChange={onUpdate}
-          placeholder="Email"
-        />
-        {validationMessage(validation.email)}
-      </div>
-      {SETTINGS.recaptchaKey && onRecaptcha ? (
-        <div className="recaptcha row">
-          <ReCAPTCHA sitekey={SETTINGS.recaptchaKey} onChange={onRecaptcha} />
-          {validationMessage(validation.recaptcha)}
+type Props = FormProps<EmailForm>
+type State = {
+  recaptchaScale: number
+}
+
+// These needs to stay in sync with ReCAPTCHA's runtime width
+const RECAPTCHA_NATURAL_WIDTH = 304
+const RECAPTCHA_NATURAL_HEIGHT = 78
+
+const StyledReCAPTCHA = styled(ReCAPTCHA)`
+  transform: scale(${props => props.scale.toFixed(3)});
+  transform-origin: 0 0;
+  height: ${props => (props.scale * RECAPTCHA_NATURAL_HEIGHT).toFixed(0)}px;
+`
+
+class AuthEmailForm extends React.Component<Props, State> {
+  recaptcha: null
+
+  constructor(props: Props) {
+    super(props)
+    this.recaptcha = null
+    // only rescale up to 4x a second to salvage some performance
+    this.scaleRecaptcha = debounce(this.scaleRecaptcha.bind(this), 250)
+    this.state = {
+      recaptchaScale: 1.0
+    }
+  }
+
+  componentDidMount() {
+    window.addEventListener("resize", this.scaleRecaptcha)
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener("resize", this.scaleRecaptcha)
+  }
+
+  scaleRecaptcha = () => {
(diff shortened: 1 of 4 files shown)


aliceriot:
functionality looks great! reading code now

static/js/components/auth/AuthEmailForm.js
```diff
@@ -2,49 +2,114 @@
...
+
+  componentDidMount() {
+    window.addEventListener("resize", this.scaleRecaptcha)
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener("resize", this.scaleRecaptcha)
+  }
+
+  scaleRecaptcha = () => {
```
aliceriot: since we're binding this in the constructor, should it just be a standard class method instead of an arrow-function class property?

static/js/components/auth/AuthEmailForm.js
```diff
@@ -2,49 +2,114 @@
...
+    if (this.recaptcha) {
+      const {
+        captcha: { clientWidth }
+      } = this.recaptcha
+      // compute this as a fractional scale of the scale that ReCAPTCHAwants to render at if our container is smaller
```
aliceriot: typo: `ReCAPTCHAwants`

static/js/components/auth/AuthEmailForm.js
```diff
@@ -2,49 +2,114 @@
...
-      {SETTINGS.recaptchaKey && onRecaptcha ? (
-        <div className="recaptcha row">
-          <ReCAPTCHA sitekey={SETTINGS.recaptchaKey} onChange={onRecaptcha} />
-          {validationMessage(validation.recaptcha)}
+type Props = FormProps<EmailForm>
+type State = {
+  recaptchaScale: number
+}
+
+// These needs to stay in sync with ReCAPTCHA's runtime width
```
aliceriot: I'm not sure what this comment means - these constants are the width and height the captcha component renders at normally, right? if so, maybe calling them something like `RECAPTCHA_NATURAL_WIDTH` or `RECAPTCHA_INHERENT_WIDTH` or something along those lines would capture what the value represents more clearly.

pdpinch:
Nathan showed this to me on his laptop, and the captcha looked good at multiple sizes. His config did not have the Touchstone button and the not-an-hr horizontal rule, so I couldn't comment on that.

static/js/components/auth/AuthEmailForm.js
```diff
@@ -2,49 +2,114 @@
...
+            value={form.email}
+            onChange={onUpdate}
+            placeholder="Email"
+          />
+          {validationMessage(validation.email)}
+        </div>
+        {SETTINGS.recaptchaKey && onRecaptcha ? (
+          <div className="recaptcha row">
+            <StyledReCAPTCHA
+              ref={this.setRecaptcha}
```
aliceriot: should we use the newer `ref` API here, with `React.createRef` and so on? I think then the `scaleCaptcha` method might have to be called in `componentDidMount` or something, not sure if the specific use-case here prevents us from using that API

rhysyngsun: So the reason I didn't do that is because I needed some hook to call when the component first instantiates and I have a realistic idea of the size of the container, when the ref is created seemed like the best in practice unless you have a better suggestion.

aliceriot: hmm, yeah that probably makes sense. I think there are a few guarantees that refs will be populated at certain points in the component life cycle, but I think if we let the DOM elements be rendered without immediately grabbing the ref we could potentially run into a situation where the element flashes on the screen at a wider width for a few frames, which we probably want to avoid. So yeah, I think it's fine to go with this.

rhysyngsun:
@aliceriot feedback addressed

static/js/components/auth/AuthEmailForm.js
rhysyngsun: @aliceriot I'm going to back that out, since I'm getting this `flow` error now:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ static/js/components/auth/AuthEmailForm.js:35:10
Cannot assign debounce(...) to this.scaleRecaptcha because property
scaleRecaptcha is not writable.
     32│     super(props)
     33│     this.recaptcha = null
     34│     // only rescale up to 4x a second to salvage some performance
     35│     this.scaleRecaptcha = debounce(this.scaleRecaptcha.bind(this), 250)
     36│     this.state = {
     37│       recaptchaScale: 1.0
     38│     }
```

aliceriot: yeah just saw that on the CI build

aliceriot: seems like that's what we want to do!

aliceriot:
hm the travis build doesn't seem to be triggering for some reason

pdpinch:
👍

PR #1396, opened 2018-10-18
Storybook updates for post and comment

noisecapella:
WIP

#### Pre-Flight checklist

  - [ ] Desktop screenshots 
  - [ ] tag @ferdi or @pdpinch for review  
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

(Required)

#### What's this PR do?
(Required)

(Required)

(Optional)

(Optional)

(Optional)

(Optional)


static/js/storybook/index.js
```diff
@@ -135,3 +191,153 @@ storiesOf("Links and Buttons", module)
...
+          showPinUI={showPinUI}
+          removePost={action("remove post")}
+          ignorePostReports={action("ignore post reports")}
+          reportPost={action("report post")}
+          togglePinPost={action("pin")}
+          isModerator={isModerator}
+        />
+      </StoryWrapper>
+    )
+  })
```
aliceriot: I might split this out into a few stories, like `url post`, `compact post`, `pinned post`, etc. that way storybook can start to function as documentation of the different UI states we support for the component.

to make that easier it might make sense to write a function which takes a few parameters and returns a story. I didn't do that for the simpler stories I wrote (for the link and so on) because I didn't think it was worth it, but for something more complicated like this it might make sense to do so.

noisecapella: Do you have an idea what should be a knob and what should be split into separate stories?

aliceriot: I think if we have components where we have a few different designed UI states then those should be different stories, like link posts vs text posts, etc. It's a bit of a judgment call, but that's where I'd say the line is :woman_shrugging:

aliceriot:
this isn't related to changes specific to this PR, per se, but thinks look different in Storybook than they do in the whole app. on Storybook:

onstorybook

in the whole app:

notonstorybook

I'm not exactly sure why, but some styles are applying in different orders in storybook. But, in any case, it exposes that we have some redundant styles, especially for `<a>` tags, in `post.scss` which we should get rid of. Do you think it makes sense to do that in this PR?

noisecapella:
I noticed that too... I'll investigate

noisecapella:
@aliceriot What about this line? Take it out too? https://github.com/mitodl/open-discussions/blob/master/static/scss/post.scss#L205

aliceriot:
if it doesn't make any difference in the display then yeah, I think so

static/scss/post.scss
```diff
@@ -287,7 +279,7 @@
     line-height: 29px;
 
     a {
-      color: rgba(0, 0, 0, 0.87);
+      color: $font-black;
     }
```
aliceriot: based on [zeplin](https://app.zeplin.io/project/5b181d3bf71408ae4d406b94/screen/5ba4e7e0a4dc98a7608ee6b5) the `.post-title a` element should be the `$navy` color, not black. Probably though we should just put the `.navy` class on the link, instead of having a color set here manually.

noisecapella:
@aliceriot I "fixed" the CSS ordering issue. It looks like because the google links were added dynamically instead of being in the `base.html` template, they were applied after our own CSS was. What do you think of this workaround? https://github.com/mitodl/open-discussions/pull/1396/files#diff-52e2c2a29669824bd796521fbdbbaf57R36

PR #1409, opened 2018-10-24
Add indexing user as first moderator to every channel if not already a moderator

### LINKED ISSUE #1369: Channels should be owned by the indexing user
Right now channels are owned by the user who created the channel. If they were created through MicroMasters this is the indexing user, a special user who manages channels for indexing and other purposes. However if the channel was created through the discussions UI the channel is owned by the user who created it. This prevents us from properly managing the channel. We actually don't even record this user anywhere so if these channels are private we have to iterate through all users to see who has access to it.

Instead we should create all channels with the indexing user, then assign the creating user as a moderator right after. We should also make a script which will fix existing channels on production:
 - For each channel figure out who the current owner is (for private channels where we don't know the owner, this may require looping through all users and testing out access)
 - Record the list of moderators and their creation date (this determines who has seniority).
 - Add the indexing user as a moderator.
 - Remove all other users as moderators, making sure to remove the owner last.
 - As the indexing user, add all other moderators back in the order of seniority

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #1369

#### What's this PR do?
- Creates every channel as the indexing user, then adds the user who is submitting the new channel as a moderator for that channel.
- Adds a management command `add_index_user_to_channels` that will iterate over every `Channel` object and insert the indexing user into the moderators list if not already present.

- On the master branch, while logged in as anyone besides the indexing user, create some channels at `/manage/c/new`.  Add another user or two as moderators to some of the channels.
- Switch to this branch and run `docker-compose run web python manage.py add_index_user_to_channels`.  You should see a bunch of messages like `Channel <name> SUCCESS: Indexing user added or already present`
- There may be errors if you have `Channel` objects that don't actually exist in reddit, or channels that have 0 moderators in reddit.  I don't think there's any way to fix those.
- Every channel where the indexing user was missing should now have the indexing user as the most senior moderator, and all the other moderators should still be present in their original order.

[channels.management.commands.add_index_user_to_channels.py](https://github.com/mitodl/open-discussions/pull/1409/files#diff-4943cf61ee09cb70ba70d15b18b56884)

[channels.serializers.ChannelSerializer.create](https://github.com/mitodl/open-discussions/pull/1409/files#diff-7e5cbd3bde192b46904dbf56f630dc13R137)

### DIFF
--- a/channels/management/commands/add_index_user_to_channels.py
+++ b/channels/management/commands/add_index_user_to_channels.py
@@ -1,0 +1,68 @@
+"""Management command to add the indexing user account as a moderator to all channels"""
+import sys
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from django.contrib.auth.models import User
+from prawcore import Forbidden
+
+from channels.api import Api
+from channels.models import Channel
+
+
+class Command(BaseCommand):
+    """Add indexing user as moderator to all channels"""
+
+    help = "Add indexing user as moderator to all channels"
+
+    def handle(self, *args, **options):
+        """Adds the indexing user as the moderator on all channels"""
+
+        index_api = Api(User.objects.get(username=settings.INDEXING_API_USERNAME))
+        failed = False
+        for channel in Channel.objects.values_list("name", flat=True):
+            processed = False
+            for user in User.objects.filter(is_staff=True):
+                user_api = Api(user)
+                try:
+                    existing_mods = sorted(
+                        user_api.list_moderators(channel),
+                        key=lambda moderator: moderator.date,
+                    )
+                    if settings.INDEXING_API_USERNAME not in [mod.name for mod in existing_mods]:
+                        if existing_mods and existing_mods[0] != user.username:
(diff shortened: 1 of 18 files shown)


channels/management/commands/add_index_user_to_channels.py
```diff
@@ -0,0 +1,78 @@
...
+                            )
+                        )
+                    processed = True
+                    break
+                except Forbidden:
+                    # Just try the next user
+                    pass
+                except Exception as exc:  # pylint: disable=broad-except
+                    self.stderr.write(
+                        "Channel '{}' ERROR: {}".format(channel, str(exc))
```
noisecapella: This will output the exception message but not the stack trace. Can you use `traceback.format_exc`?

noisecapella:
Functionality works great

channels/management/commands/add_index_user_to_channels.py
mbertrand: Switched to `traceback.format_exc`

PR #1415, opened 2018-10-25
Query database to get lists of channels, posts, comments for indexing

### LINKED ISSUE #1368: Missing posts in the elasticsearch index
We need to change our reindexing code to reference our database models to get the post and channel information, rather than asking reddit for it since reddit cuts off posts at about 1000 posts per channel

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #1368 

#### What's this PR do?
- Queries the database instead of reddit to get the channels, posts, and comments to index.
- Adds factories for Django models (Channel, Post, Comment) and refactors existing factory class names for the reddit equivalents to differentiate between them.

- Run `docker-compose run web python manage.py recreate_index`
- Query Elasticsearch to verify that all posts and comments were indexed.

### DIFF
--- a/search/serializers.py
+++ b/search/serializers.py
@@ -166,12 +166,12 @@ class ESCommentSerializer(ESSerializer):
         }
 
 
-def _serialize_comment_tree_for_bulk(comments):
+def _serialize_comment_tree_for_bulk(comment_ids):
     """
-    Serialize an iterable of Comment and their replies for a bulk API request
+    Serialize comments by an iterable of their ids for a bulk API request
 
     Args:
-        comments (iterable of str): An iterable of comment ids
+        comment_ids (iterable of str): An iterable of comment ids
 
     Yields:
         dict: Documents suitable for indexing in Elasticsearch
@@ -179,7 +179,7 @@ def _serialize_comment_tree_for_bulk(comments):
     from channels.api import Api
 
     api = Api(user=User.objects.get(username=settings.INDEXING_API_USERNAME))
-    for comment_id in comments:
+    for comment_id in comment_ids:
         yield serialize_comment_for_bulk(api.get_comment(comment_id))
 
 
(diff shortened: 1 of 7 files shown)


search/serializers.py
```diff
@@ -162,14 +171,16 @@ def _serialize_comment_tree_for_bulk(comments):
     Serialize an iterable of Comment and their replies for a bulk API request
 
     Args:
-        comments (iterable of Comment): An iterable of comments
+        comments (iterable of str): An iterable of comment ids
```
noisecapella: Can we change `comments` to `comment_ids`?

search/serializers.py
```diff
@@ -162,14 +171,16 @@ def _serialize_comment_tree_for_bulk(comments):
     Serialize an iterable of Comment and their replies for a bulk API request
```
noisecapella: Can you update this? In particular we should remove `and their replies`

search/tasks.py
```diff
@@ -44,6 +43,10 @@ def wrap_retry_exception(*exception_classes):
     try:
         yield
     except Exception as ex:
+        if isinstance(ex, NotFound):
+            # No corresponding reddit post/comment found for django model object.  Log error and continue indexing.
+            log.exception(str(ex))
```
noisecapella: You can omit the `str(ex)` here, the whole stack trace will be logged automatically

mbertrand: ✔️

search/serializers.py
mbertrand: ✔️

search/tasks.py
```diff
@@ -45,7 +45,7 @@ def wrap_retry_exception(*exception_classes):
     except Exception as ex:
         if isinstance(ex, NotFound):
             # No corresponding reddit post/comment found for django model object.  Log error and continue indexing.
-            log.exception(str(ex))
+            log.exception()
```
noisecapella: Actually it looks like this requires an argument. Can you add a brief message? The exception stacktrace will be logged before it automatically

mbertrand: ✔️

Issue #1316, opened 2018-10-05
Channel Search


pdpinch:
As a user I'd like to be able search within channels. I expect to see relevant posts, comments and user profiles in my search results. 

InVision: [some results](https://invis.io/9EN908ZQPAX#/322898017_Channel_ClimateX_Search_Channel_New) and [no results](https://invis.io/9EN908ZQPAX#/322898016_Channel_ClimateX_Search_Channel_No_Results) and [mobile results](https://invis.io/9EN908ZQPAX#/326140814_Mobile_ChannelSearch)
Zeplin: [desktop](https://zpl.io/VxlD5eg) and [mobile](https://zpl.io/aN9ByMk)

Note that these designs include some changes to the channel header, which come under the epic #1327.

### acceptance criteria
- [x] index posts, comments and profiles
- [x] management command for reindexing all posts, comments and profiles
- [x] incrementally update index when posts, comments and profiles are added or edited
- [x] REST API 
- [x] default sort is relevance 
- [x] filter on all (default), posts, comments and profiles
- [x] anonymous users get the same results, for public channels
- [x] anonymous users get no results for private channels 
- [x] users who are not a member of private channel get no results when searching that channel
- [x] result page uses new channel header, implemented in #1327

### Questions
- What are sorting options? "Active" "Top" and "New" aren't relevant for profiles
  **A** for this issue, we'll only support sort by relevance

pdpinch:
I added the design for mobile results and a some acceptance criteria to the description. Hopefully it's nothing unexpected.

pdpinch:
Added Zeplin link for mobile design

pdpinch:
@noisecapella @mbertrand what else needs to be done here?

noisecapella:
The only thing left is sorting. Should we do that here or as another issue later?

pdpinch:
The plan for now is to only do relevance sorting, since that's the only one we can do consistently across posts, comments and profiles. 

I just answered the question in the description.

pdpinch:
We can close this now. 🎉

We will have some follow-up issues for styling, and how to deal with the various buttons on the post summary and comment cards.

PR #1422, opened 2018-10-26
Add page for channel search

noisecapella:
#### Pre-Flight checklist

  - [ ] Desktop screenshots 
  - [ ] tag @ferdi or @pdpinch for review  
- [x] Testing
  - [x] Code is tested
- [ ] Settings

Part of #1316 

#### What's this PR do?
Implements channel search

Run `./manage.py recreate_index` to ensure the index exists then run a search on the channel. You should be able to search the title and text for a post and only the text for a comment. You should never see any profile results and the search filter should only include options for comments or posts. Selecting 'All' should only show some combination of the two, no profile results.


search/api.py
```diff
@@ -90,5 +90,8 @@ def execute_search(*, user, query):
     search = Search(index=index, using=get_conn())
     search.update_from_dict(query)
     channel_names = [channel.display_name for channel in Api(user).list_channels()]
-    search = search.filter("terms", channel_name=channel_names)
+    channels_filter = Q("terms", author_channel_membership=channel_names) | Q(
```
mbertrand: Is dealing with public channels going to be in another PR?  This line will prevent searching in any channel that the user is not subscribed to (ie public channels, or conceivably a private channel which the user is a contributor/moderator on but not a subscriber?).

noisecapella: I'm going to change this query to restrict to only the channel being viewed, or no channels if the user does not have permission to view that channel.

noisecapella: Actually I changed it to use the models for contributor and moderator membership. It would be a significant amount of work to filter only on the user and since this is still behind a feature flag I think we can have a PR to add the public channel part of it before it is open to the public

noisecapella:
Ok, I think that's finally all of the initial feedback

search/api.py
```diff
@@ -84,11 +91,18 @@ def execute_search(*, user, query):
...
     search.update_from_dict(query)
-    channel_names = [channel.display_name for channel in Api(user).list_channels()]
-    search = search.filter("terms", channel_name=channel_names)
+    channel_names = list(
+        ChannelGroupRole.objects.filter(
+            group__user=user, role__in=(ROLE_CONTRIBUTORS, ROLE_MODERATORS)
+        )
+        .values_list("channel__name", flat=True)
+        .distinct()
+    )
```
mbertrand: This will prevent subscribers of restricted channels from getting search results (checkbox 2 in #1461).  Using `list(profiles.api.get_channels(user))` will include the channel names that the user subscribes to.

noisecapella: I think the only way we can support that is to add the `channel_type` information to the `Channel` model and then update the filter to permit all searches to those channels. I'm going to do that in the next PR and then update this logic to fix the filtering

mbertrand: OK, I tried `list(profiles.api.get_channels(user))` and that does work for searching within a restricted channel when logged in as a subscribed user, but I'm fine with solving it another way in a future PR.

PR #1424, opened 2018-10-26
Fix login page button label

### LINKED ISSUE #1423: Login button label says "Sign Up"
This is a regression from updating the signup form, which the login page also uses.

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing

Fixes #1423 

#### What's this PR do?
Parameterizes the form submit button label with the old default of "Next", parameterizes as "Sign Up" for that page.

Go to /login and /signup and verify you see appropriate labels.

### DIFF
--- a/static/js/components/auth/AuthEmailForm.js
+++ b/static/js/components/auth/AuthEmailForm.js
@@ -10,7 +10,9 @@ import { validationMessage } from "../../lib/validation"
 import type { EmailForm } from "../../flow/authTypes"
 import type { FormProps } from "../../flow/formTypes"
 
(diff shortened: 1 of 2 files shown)


aliceriot:
labels look good. I also noticed, though, that the signup card now doesn't get as big as the login in and other cards on desktop (I think because of the change to fix the captcha width sizing). idk if that's something we want to address, I think before there was a desire for all the auth cards to be the same width but idk.

aliceriot:
ah looks like there is a fmt and flow issue on the build though

PR #1445, opened 2018-10-31
Updated README and added references to common web app guide

### LINKED ISSUE #41: Centralize documentation for installation
We currently have some of this information in this README, some information in `reddit-config` and some in `reddit` itself (for example setting up `reddit.local` in `/etc/hosts`). We should put all of this in a `Getting Started` section at the top of this README

gsidebo:
#### Pre-Flight checklist

- [x] Testing

Closes #41 

#### What's this PR do?
Overhauls the README to correct various setup steps and adds references to the [common web app guide](https://github.com/mitodl/handbook/blob/master/common-web-app-guide.md)

If you _really_ want to test this, run through the full README from a completely clean slate and check that it works.

I'm going to create a separate issue to move the giant code block for setting up a new channel into a management command.


.env.example
```diff
@@ -1,45 +1,22 @@
 CELERY_TASK_ALWAYS_EAGER=True
 DJANGO_LOG_LEVEL=INFO
 LOG_LEVEL=INFO
+OPEN_DISCUSSIONS_DEFAULT_SITE_KEY=
+OPEN_DISCUSSIONS_BASE_URL=http://od.odl.local:8063
```
pdpinch: elsewhere in the docs it says to set this to `OPEN_DISCUSSIONS_BASE_URL=http://docker.for.mac.localhost:8063/` (On Mac)

This settings here seems correct to me. Maybe the other one should be changed?

.env.example
```diff
@@ -1,45 +1,22 @@
...
-# SAML IDP
-SOCIAL_AUTH_SAML_IDP_ENTITY_ID=
-SOCIAL_AUTH_SAML_IDP_URL=
-SOCIAL_AUTH_SAML_IDP_ATTRIBUTE_PERM_ID=
-SOCIAL_AUTH_SAML_IDP_ATTRIBUTE_NAME=
-SOCIAL_AUTH_SAML_IDP_ATTRIBUTE_EMAIL=
-SOCIAL_AUTH_SAML_IDP_X509=
-
-USE_X_FORWARDED_PORT=
-USE_X_FORWARDED_HOST=
```
pdpinch: Will the application start without these forwarded settings?

pdpinch:
I don't think I can finish a proper review of this today, so this will wait the week after next.

.env.example
gsidebo: It certainly does when running locally. In production environments it might not, which is why I made the comment above asking if the `.env.example` file has any special importance to devops.

blarghmatey:
@gsidebo WRT .env.example we make no use of that for our purposes so feel free to modify it as you need. Hopefully that will eventually get replaced by pillar data so it will be a moot point :)

.env.example
pdpinch: It's mentioned currently at https://github.com/mitodl/open-discussions/blob/master/README.md#L175 in the section https://github.com/mitodl/open-discussions/blob/master/README.md#6-integration-with-micromasters

pdpinch:
I don't have time to test this all out by hand, but it reflects the notes that I made when I recently tried to install. 

If you can resolve my [first comment](https://github.com/mitodl/open-discussions/pull/1445/files#r230480737) I think this is good to merge.

pdpinch:
I had a chance to try more of this out, and it seems like there's something not quite right about the public channels. They don't display for an anonymous user. 

I think there's something that may need to be run on reddit to make them work?

.env.example
gsidebo: @pdpinch The `docker.for.mac.localhost` value is only set on `MICROMASTERS_BASE_URL`, not `OPEN_DISCUSSIONS_BASE_URL`. Those two values are and should be different. I have `OPEN_DISCUSSIONS_BASE_URL=http://od.odl.local:8063` and `MICROMASTERS_BASE_URL=http://docker.for.mac.localhost:8079/` in my own `.env` file and it works

gsidebo:
@pdpinch We have not yet come up with a reliable way to configure an anonymous-user-viewable public channel in open-discussions running locally. I think Nathan has, but last I checked, no one else had. He gave me some steps to try and it didn't work out. You indicated that we will always have a public channel configured in production, so it didn't seem worth it to invest the time to troubleshoot the local setup at the time. It probably makes sense to create a separate issue for adding details about the correct configuration for public channels (and, indeed, finding out what those details actually are). What do you think?

pdpinch:
Ok then, I opened #1490.

this is 👍

Issue #1420, opened 2018-10-26
keep track of channel followers


pdpinch:
In discussing #1249 we learned that reddit does not have an API for getting a list of subreddit subscribers ("channel followers" in discussions). In order to be able to count and list followers, we will need to store this relationship in django. 

### acceptance criteria
- [x] model for storing user/channel subscription status
- [x] update the subscribe API in discussions to store the subscription state of the user
- [x] whenever the subscription state is updated, update the ES profile index
- [x] update the `backpopulate_default_channel` management command to also store subscriptions
- [x] api for getting a list of followers of a channel 

### related issues
- button and api for following/unfollowing channels

mbertrand:
@pdpinch does it make sense to add 'Update the Elasticsearch profile doc (`author_channel_membership`) when channel subscription status changes' to this issue, or create a new issue for it?

mbertrand:
@pdpinch is it correct that a channel follower is not necessarily a moderator or contributor?  I don't see any option in the UI to follow a channel (just posts), but the `Channel` object returned by the API includes the following attributes:
```
 user_is_contributor
 user_is_moderator
 user_is_subscriber
```
I'm assuming that a user is a follower if `user_is_subscriber` is true?  And we should store all the above in Django: a channel's moderators, contributors, and subscribers?

Ferdi:
> is it correct that a channel follower is not necessarily a moderator or contributor?
Yes. 
Private channels have:  contributors, moderators, and subscribers
Public channels have :  Moderators and subscribers

"Subscribe" just affects your front page feed .  So for example, you can be a member of a private channel but you unsubscribe because the channel is noisy and you dont want that channel to dominate your feed.  But you are still a member and can access the channel if you wanted.

mbertrand:
@pdpinch I checked the last checkbox, ok to close?

PR #1449, opened 2018-11-02
Store channel memberships (subscriber, moderator, contributor) in django

### LINKED ISSUE #1451: Profiles should update channel membership when they are subscribed or unsubscribed

### LINKED ISSUE #1453: author_channel_membership should be indexed as an array instead of a comma delimited string

mbertrand:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #1451
Closes #1453
Partially addresses #1420:
- [x] model for storing user/channel subscription status
- [x] update the subscribe API in discussions to store the subscription state of the user
- [x] whenever the subscription state is updated, update the ES profile index
- [x] update the backpopulate_default_channel management command to also store subscriptions (automatic).

#### What's this PR do?
- Adds a `ChannelGroupRole` model that will store the `Group` associated with each role for a channel.
- Adds a `ChannelSubscription` model that will store a user's subscription to a channel.
- Adds an API function `sync_channel_user_role_model` that is called whenever a channel moderator, or contributor is added or removed.
- Adds an API function `sync_channel_user_subscription_model` that is called whenever a channel subscriber is added or removed.
- Calls `search.tasks.update_author` to update the ES profile doc whenever a user's subscriber, moderator, or contributor status is changed.
- Adds a management command `backpopulate_subscriptions_roles` to iterate through all channels/users and create `ChannelGroupRole` and `ChannelSubscription` objects for them.
- Adds a `profiles.api.get_channels` function to return all channel names that a user is either moderator, contributor, or subscriber on.
- Uses the profiles.api.get_channels` function to populate `author_channel_membership` in the ES index.

- Run `docker-compose run web python manage.py migrate`
- Run `docker-compose run web python manage.py backpopulate_subscriptions_roles`
- Confirm that there is a `ChannelSubscription` object for every channel where the user is a subscriber.
- Confirm that users belong to the group associated with their role (moderators, contributors) for a channel, via `ChannelGroupRole`.
- Add and remove users as contributors or moderators for a channel.  Verify that the groups the user belongs to is updated after each change, and the profile ES doc's `author_channel_membership` field is updated.
- Run `docker-compose run web python manage.py recreate_index` and verify that `author_channel_membership` is populated correctly for profile docs.


Ferdi:
@mbertrand @rhysyngsun do you think we should store users subscriptions to channels  separate from roles ?

rhysyngsun:
Subscriber isn't really a role, it just means you can a subscription to the channel so it shows up in your frontpage. So we should store that in a separate `ChannelSubscription` model.

I think it also might be worth looking at something like [django-guardian](https://django-guardian.readthedocs.io/en/stable/), built on django's builtin permissions, which also hooks nicely into DRF: https://www.django-rest-framework.org/api-guide/permissions/#djangoobjectpermissions that way we don't have to chew off building something like that ourselves.

Ferdi:
> So we should store that in a separate ChannelSubscription model
That's what I was thinking as well. 
It looks like this PR deals with roles only. We can tackle permissions some other time - from what I remember we have used django guardian in another project in the past.

rhysyngsun:
Fair enough, then I'd suggest we lay the groundwork for that here by having the roles defined in terms of `django.auth.contrib.models.Group` and have a group for moderators and a group for contributors per channel, rather than defining our own models for those.

mbertrand:
Is this making things more complicated than they really need to be?  I thought the purpose of this epic was not to reimplement reddit's permissions system, but just to have a way of tracking which users are moderators, contributors, or subscribers for a channel in Django instead of having to query reddit for that information.

Ferdi:
@mbertrand ok, what do you think about this part though ?

> Subscriber isn't really a role, it just means you can a subscription to the channel so it shows up in your frontpage. So we should store that in a separate ChannelSubscription model.

?

mbertrand:
@Ferdi I guess it depends on how this will be used.  If we want to regularly query for all channels that a user is any of a subscriber, moderator, or contributor on, it seems more efficient to have all that info in one model instead of two.  If we are going to regularly query for channel subscriptions regardless of moderator/contributor status or vice versa, then logically it makes more sense to have them in separate models.

mbertrand:
@Ferdi @rhysyngsun how does this look:
```python
class ChannelSubscription(TimestampedModel):
    """
    Keep track of channel subscribers
    """

    channel = models.ForeignKey(Channel, on_delete=models.CASCADE)
    user = models.ForeignKey(User, on_delete=models.CASCADE)

    def __str__(self):
        return f"User {self.user.username} subscription for Channel {self.channel.name}"


class ChannelRole(TimestampedModel):
    """
    Keep track of channel contributors and moderators
    """
    group = models.ForeignKey(Group, on_delete=models.CASCADE)
    user = models.ForeignKey(User, on_delete=models.CASCADE)
    channel = models.ForeignKey(Channel, on_delete=models.CASCADE)

    def __str__(self):
        return f"User {self.user.username} role {self.group.name} for Channel {self.channel.name}"
```

```python
class Profile(models.Model)

    @property
    def channels(self):
        return set(
            list(
                ChannelSubscription.objects.filter(user=self.user).values_list(
                    "channel__name", flat=True
                )
            )
            + list(
                ChannelRole.objects.filter(user=self.user).values_list(
                    "channel__name", flat=True
                )
            )
        )
```

rhysyngsun:
@mbertrand that looks good to me

rhysyngsun:
Matt and I discussed that it's going to be useful to my work for #1459 to have a group for moderators on a per channel basis and be able to generically look up permissions (to be handled in #1459, not here) in the widget auth handler. That requires users being added to the groups, which is pretty standard for `django.contrib.auth`. So we want to evolve the `ChannelRole` table to a tuple of `("channel", "group", "role")` and add users to the group to enable the flexibility I'll need in the widgets work soon.

channels/api.py
```diff
@@ -350,6 +354,63 @@ def sync_comment_model(*, channel_name, post_id, comment_id, parent_id):
...
+
+def sync_channel_role_models(channel_name):
+    """
+    Get or create Channel roles
+
+    Args:
+        channel_name(str): The channel name
+
+    Returns:
+        list of ChannelGroupRole: the channel user role objects
```
rhysyngsun: Could you update this to reflect the `dict` return type?

channels/api.py
```diff
@@ -350,6 +354,63 @@ def sync_comment_model(*, channel_name, post_id, comment_id, parent_id):
...
+    """
+    Add or remove a user from a channel role's group
+
+    Args:
+        channel_name(str): The channel name
+        user(django.contrib.auth.models.User): The user
+        role(str): The role name (moderators, contributors)
+        remove(bool): If the user should be removed instead of added
+    """
+    roles = sync_channel_role_models(channel_name)
```
rhysyngsun: This will lead to 3 extraneous fetches from the database every time after the first when you just need the group. I think we need to move away from this sync_* pattern and instead ensure that these records are created when the channel is created and then in this situation simply fetch the group in one query. I've started doing this in my [PR for the article post](https://github.com/mitodl/open-discussions/pull/1452/files#diff-6b4e079385252444280613805ab05474R510), so that we can rely on that record being present if the channel is created. Since that all happens within the channel create api request, we can trust that we won't make any calls to add users before that is in place (if we do, then something is wrong).

channels/api.py
```diff
@@ -978,6 +1043,8 @@ def add_moderator(self, moderator_name, channel_name):
         except APIException as ex:
             if ex.error_type != "ALREADY_MODERATOR":
                 raise
+        update_user_role(channel_name, ROLE_MODERATORS, user)
```
rhysyngsun: I feel like it's not clear on reading this that it is adding the user, I think it might be better to be more explicit by having a method for to add and a method to remove

channels/api.py
```diff
@@ -350,6 +354,63 @@ def sync_comment_model(*, channel_name, post_id, comment_id, parent_id):
...
+
+    Args:
+        channel_name (str): The name of the channel
+        user (django.contrib.models.auth.User): The user
+
+    Returns:
+        ChannelSubscription: the channel user role object
+    """
+    with transaction.atomic():
+        channel = sync_channel_model(channel_name)
```
rhysyngsun: Can leave this for now as I mentioned in my other comment, I'll refactor the existing sync_* methods away in my other PR.

channels/models.py
```diff
@@ -185,3 +187,44 @@ class Comment(TimestampedModel):
...
+    """
+    Keep track of channel subscribers
+    """
+
+    channel = models.ForeignKey(Channel, on_delete=models.CASCADE)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = (("user", "channel"),)
+        index_together = (("user", "channel"),)
```
rhysyngsun: `unique_together` implies `index_together`, so you can remove the latter.

channels/models.py
```diff
@@ -185,3 +187,44 @@ class Comment(TimestampedModel):
...
+class ChannelGroupRole(TimestampedModel):
+    """
+    Keep track of channel moderators and contributors
+    """
+
+    channel = models.ForeignKey(Channel, on_delete=models.CASCADE)
+    group = models.ForeignKey(Group, on_delete=models.CASCADE)
+    role = models.CharField(
+        max_length=48,
+        choices=[
```
rhysyngsun: Nice trick to simplify:
```python
ROLE_CHOICES = (ROLE_MODERATORS, ROLE_CONTRIBUTORS)

choices=zip(ROLE_CHOICES,ROLE_CHOICES)
```

channels/tasks.py
```diff
@@ -82,3 +90,62 @@ def subscribe_user_range_to_default_channel(*, channel_name, usernames):
...
+
+@app.task(autoretry_for=(RetryException,), retry_backoff=True, rate_limit="600/m")
+def populate_user_subscriptions(user_ids):
+    """
+    Populate channel user roles for a list of user ids
+
+    Args:
+        user_ids(list of int): List of user ids
+    """
+    for user in User.objects.filter(id__in=user_ids):
```
rhysyngsun:
```suggestion
    for user in User.objects.filter(id__in=user_ids).iterator():
```

channels/api_test.py
```diff
@@ -775,14 +806,21 @@ def test_add_moderator_no_user(mock_client):
...
     moderator = UserFactory.create()
     client.remove_moderator(moderator.username, "channel_test_name")
     mock_client.subreddit.return_value.moderator.remove.assert_called_once_with(
         moderator
     )
+    assert (
+        not ChannelGroupRole.objects.get(
+            channel__name="channel_test_name", role=ROLE_MODERATORS
+        ).group
+        in moderator.groups.all()
```
rhysyngsun: Did you mean `X not in Y` rather than `not X in Y`

profiles/api.py
```diff
@@ -17,3 +18,20 @@ def ensure_profile(user, profile_data=None):
 
     profile, _ = Profile.objects.get_or_create(user=user, defaults=defaults)
     return profile
+
+
+def get_channels(user):
+    """
+    Get the list of channel names for which the user is a moderator, contributor, or subscriber
+
```
rhysyngsun:
```suggestion
Args:
    user (User): the user to fetch channel names for

```

channels/tasks_test.py
```diff
@@ -36,3 +50,84 @@ def test_subscribe_all_users_to_default_channel(settings, mocker, admin_user):
...
+    client_mock.return_value.list_moderators.return_value = [
+        redditor for redditor in redditors if is_moderator
+    ]
+    client_mock.return_value.list_contributors.return_value = [
+        redditor for redditor in redditors if is_contributor
+    ]
+    tasks.populate_user_roles.delay([channel.id for channel in channels])
+    for user in users:
+        for channel in channels:
+            tasks.sync_channel_role_models(channel.name)
```
rhysyngsun:
```suggestion
            tasks.sync_channel_role_models.delay(channel.name)
```

channels/tasks_test.py
```diff
@@ -36,3 +50,84 @@ def test_subscribe_all_users_to_default_channel(settings, mocker, admin_user):
...
+    client_mock = mocker.patch("channels.tasks.Api", autospec=True)
+    client_mock.return_value.list_channels.return_value = [
+        mocker.Mock(display_name=channel.name) for channel in channels if is_subscriber
+    ]
+    tasks.populate_user_subscriptions.delay([user.id for user in users])
+    for user in users:
+        for channel in channels:
+            assert (
+                ChannelSubscription.objects.filter(user=user, channel=channel).exists()
+                == is_subscriber
```
rhysyngsun:
```suggestion
                is is_subscriber
```

mbertrand:
@rhysyngsun comments addressed.

channels/api.py
```diff
@@ -350,6 +354,65 @@ def sync_comment_model(*, channel_name, post_id, comment_id, parent_id):
...
+        return ChannelSubscription.objects.update_or_create(channel=channel, user=user)[
+            0
+        ]
+
+
```
rhysyngsun:
```suggestion

@transaction.atomic
```

channels/api.py
```diff
@@ -350,6 +354,65 @@ def sync_comment_model(*, channel_name, post_id, comment_id, parent_id):
...
+        role(str): The role name (moderators, contributors)
+    """
+    get_role_model(channel_name, role).group.user_set.add(user)
+
+
+def remove_user_role(channel_name, role, user):
+    """
+    Remove a user from a channel role's group
+
+    Args:
```
rhysyngsun: Could you fix either the argument order or the order they appear in the docstring? Same for `add_user_role`.

channels/management/commands/backpopulate_subscriptions_roles.py
```diff
@@ -0,0 +1,29 @@
...
+from open_discussions.utils import now_in_utc
+
+
+class Command(BaseCommand):
+    """Create/update each user's channel subscriptions and moderator/contributor roles"""
+
+    help = "Create/update each user's ChannelSubscriptions and ChannelRoles (subscriber, moderator, contributor)"
+
+    def handle(self, *args, **options):
+        """Adds the indexing user as the moderator on all channels"""
```
rhysyngsun: Copy/paste

mbertrand: ✅

channels/api.py
mbertrand: ✅

channels/api.py
mbertrand: ✅

PR #1452, opened 2018-11-02
Nl/article post backend

### LINKED ISSUE #1436: Article Post Type: Backend
#### Posts
- Backend:
  - [ ] Extend our current posts API:
    - [ ] Don't allow creation of this post type if the channel settings don't enable it
    - [ ] Add shadow column for `post_type` to support this extended "article" post type. In reddit this will be stored as an `"any"` type, but the shadow column will override this value when returning and storing data.
    - [ ] Add a `article_text` field to post API backed by a `JSON` field in the django app
   - [ ] Add API to support multiple image uploads per article
   - [ ] Search:
     - [ ] Content of post should be indexed (need to determine how to best do this in ES, ideally balancing flexibility with complexity of implementation)

##### Channel Settings
  - [ ] Shadow and extend the supported post types to add the "article" type

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Part of #1436 

#### What's this PR do?
Adds a new post type "article".
- Adds an `Article` model that stores a JSON blob representing the article content. This will be managed by the frontend so it just needs to be valid JSON
- Moves much of the `sync_*` logic into transactions within the api itself, so we have guarantees of the creation timing of these objects.

I split the changes between fixture changes and actual code changes. The code is in commit https://github.com/mitodl/open-discussions/pull/1452/commits/14794ff3a33110a1b31adef66e041f3aa7d3199f.

Go to `/api/v0/channels/abc/posts/` and craft a POST request that includes a "`article_content`" field that is valid JSON. Verify that this request succeeds and the JSON is return in the reponse and that other post APIs still work for old post types and this new one.


channels/proxies.py
```diff
@@ -0,0 +1,61 @@
...
+        Avoid proxying __getattr__ but "proxy" to the wrapped object on any non-local fields
+        """
+        if name in ("update", "link_meta", "post_type", "article", "article_content"):
+            return super().__getattr(name)
+        else:
+            return getattr(self.__wrapped__, name)
+
+    def update(self, *, submission=None, post=None):
+        """
+        Return an updated proxy instnace by passing one or more of submission/post
```
mbertrand: typo for `instance`

channels/views/frontpage.py
```diff
@@ -36,6 +37,7 @@ def get(self, request, *args, **kwargs):  # pylint: disable=unused-argument
         pagination, posts = get_pagination_and_posts(paginated_posts, listing_params)
         users = lookup_users_for_posts(posts)
         posts = [post for post in posts if post.author and post.author.name in users]
+        posts = proxy_posts(posts)
```
mbertrand: Maybe combine with the above line, here and in similar code from `channels/views/posts.py`?
```python
posts = proxy_posts([post for post in posts if post.author and post.author.name in users])
```

channels/api.py
```diff
@@ -800,9 +828,19 @@ def create_comment(self, text, post_id=None, comment_id=None):
...
+            comment = self.get_post(post_id).reply(text)
         else:
-            return self.get_comment(comment_id).reply(text)
+            comment = self.get_comment(comment_id).reply(text)
+
+        _, link_id = get_kind_and_id(comment.link_id)
+
+        Comment.objects.create(
+            comment_id=comment.id,
+            post=Post.objects.get(post_id=link_id),
```
gsidebo: Couldn't this just be `post_id=link_id,`?

channels/migrations/0012_add_post_type_and_article.py
```diff
@@ -0,0 +1,40 @@
...
+                ('author', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to=settings.AUTH_USER_MODEL)),
+            ],
+            options={
+                'abstract': False,
+            },
+        ),
+        migrations.AddField(
+            model_name='post',
+            name='post_type',
+            field=models.CharField(choices=[('any', 'any'), ('link', 'link'), ('self', 'self'), ('article', 'article')], max_length=10, null=True),
```
gsidebo: Will we not need a migration step or a management command to backfill these?

search/indexing_api_test.py
```diff
@@ -249,12 +240,12 @@ def test_index_post_with_comments(
...
 
     serialize_post_mock.assert_any_call(post)
-    sync_post_mock.assert_any_call(serialize_post_mock.return_value)
-
     serialize_comments_mock.assert_any_call(post)
-    sync_comments_mock.assert_any_call(serialize_comments_mock.return_value)
 
     post.comments.replace_more.assert_called_once_with(limit=None)
+
+    print(bulk_mock.call_args_list)
```
gsidebo: print

search/indexing_api_test.py
```diff
@@ -249,12 +240,12 @@ def test_index_post_with_comments(
...
+    assert bulk_mock.call_count == 4
```
gsidebo: Can you include a comment about why we're expecting 4 calls? I'm assuming it's because of the 2 object types * 2 aliases each

channels/api.py
```diff
@@ -414,6 +356,34 @@ def remove_user_role(channel_name, role, user):
...
+
+    Args:
+        text (str): the post text
+        url (str): the post url
+        article_content (dict): the post article text data
+    Returns:
+        str: the link type
+    """
+    if (
+        len(list(filter(lambda val: val is not None, [text, url, article_content])))
```
gsidebo: Utility function would be nice here (`get_all_not_none` et al), but not important. You can take or leave that one

channels/api.py
rhysyngsun: `link_id` here is the reddit link id, not the postgres primary key, so the relation here is `post__post_id`.

channels/api.py
gsidebo: Side note: this would probably be more fitting to implement in a [`validate`](https://www.django-rest-framework.org/api-guide/serializers/#object-level-validation) method, though I realize this logic was already in `create` before this PR

channels/migrations/0012_add_post_type_and_article.py
rhysyngsun: I was planning on doing that in a separate PR as part of #1495 if that's alright, it's only written to here for new posts.

search/indexing_api_test.py
rhysyngsun: This was also missing an assertion on the two other calls so I added that.

channels/serializers.py
```diff
@@ -941,7 +949,10 @@ def get_comment(self, instance):
     def get_post(self, instance):
         """Returns the post if this report was for one"""
         if isinstance(instance, Submission):
-            return PostSerializer(instance, context=self.context).data
+            return PostSerializer(
+                PostProxy(instance, Post.objects.get(post_id=instance.id)),
```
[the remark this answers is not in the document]
rhysyngsun: I ended up making a super-lazy version of this so it won't hit the database unless a property that exists there is hit.

channels/proxies.py
```diff
@@ -59,3 +62,28 @@ def article(self):
...
+    posts = SimpleLazyObject(lambda: {
+        post.post_id: post
+        for post in Post.objects.filter(
+            post_id__in=[submission.id for submission in submissions]
+        )
+    })
+    return [
+        PostProxy(submission, SimpleLazyObject(lambda: posts[submission.id]))
+        for submission in submissions
+        if submission.id in posts
```
gsidebo: This guarantees that you would run the function passed into the `SimpleLazyObject` function, doesn't it? Seems to defeat the purpose of using `SimpleLazyObject` in the first place. I think this conditional is unnecessary anyway - the query above is already selecting for `Post`s that exist in the list of submission ids.

Also I still think it would be cleaner to (a) add a flag to the `PostProxy` constructor to fetch the `Post` for itself in a lazy object, or (b) create a `proxy_post` helper method to create a single `PostProxy`. Would  be cleaner than writing `proxy_posts([obj])[0]` every time we need a single `PostProxy`

channels/serializers.py
```diff
@@ -396,34 +394,39 @@ def validate_subscribed(self, value):
...
-        text = validated_data.get("text")
-        url = validated_data.get("url")
-
-        if text and url:
+        text = validated_data.get("text", None)
+        url = validated_data.get("url", None)
+        article_content = validated_data.get("article_content", None)
+
+        if (
+            len(list(filter(lambda val: val is not None, [text, url, article_content])))
```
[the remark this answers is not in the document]
rhysyngsun: I asked @noisecapella about this since he wrote the original, he agrees with that reasoning so I added a code comment to explain that.

channels/proxies.py
rhysyngsun: I opted for the latter solution, since it encapsulates construction of the proxies in that module.

rhysyngsun:
@mbertrand should be all set now

rhysyngsun:
@mbertrand added the test and also rewrote the logic a bit to fallback to the old check if `post_type` is None (which it will be until we backpopulate)

PR #1479, opened 2018-11-13
add validation for extremely long text posts

### LINKED ISSUE #1469: Post form: limit text post length to 40,000 characters
We saw the following error in production:

```APIException: TOO_LONG: 'this is too long (max: 40000)' on field 'text'```

We should prevent this from happening on the front end. I can only assume that the UX was not good in this case. 

https://sentry.io/mit-office-of-digital-learning/discussions/issues/757981611/

```
APIException: TOO_LONG: 'this is too long (max: 40000)' on field 'text'
(14 additional frame(s) were not displayed)
...
  File "channels/views/posts.py", line 75, in post
    serializer.save()
  File "rest_framework/serializers.py", line 214, in save
    self.instance = self.create(validated_data)
  File "channels/serializers.py", line 418, in create
    post = api.create_post(channel_name, title=title, **kwargs)
  File "search/task_helpers.py", line 42, in wrapped_api_func
    reddit_obj = func(*args, **kwargs)
  File "channels/api.py", line 563, in create_post
    return self.get_channel(channel_name).submit(title, selftext=text, url=url)

APIException: TOO_LONG: 'this is too long (max: 40000)' on field 'text'
```

### PR DESCRIPTION
**Pull request by aliceriot**
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #1469 

#### What's this PR do?

adds validation that the text post body isn't longer than the backend allows.

Go to the post create form, and try to create a text post. you should be able to go forward with creating a reasonably-sized text post no problem. then, try making a reeeeeally long text post. You can do this in the JavaScript console to get a nice long string:

```js
const longString = Array(40000).fill().map(() => "a").join("")
```

If the string is 40000 characters long you should be able to save it, but you should get a validation error for anything over that (see screenshot below).

waaaaaytoolongmessage


Ferdi:
@aliceriot Can we stop accepting input from users after the limit is reached ?   With just the message you have in thsi pr, as a user I'm not sure how much longer the post is than the allowed limit

Ideally -- As user approaches the limit , we print something like " 99 chars left " and then don't accept input after the limit.   If it's too much work, your PR is fine as is.

aliceriot:
We don't currently have a setup for validating on every input the user makes, so we'd need to add that it (we had a system for doing that in MM for the profiles, we'd need to implement a similar approach over here for that). It wouldn't be a huge amount of work but would probably be substantially more than what's here

Issue #1398, opened 2018-10-19
Profile: add social, web links


pdpinch:
As a user, I'd like to be able to add social links (and other links) to my profile for other users to discover. 

### Invision:
- [rendered profile (desktop)](https://invis.io/9EN908ZQPAX#/314904354_Profile_With_Posts)
- [profile form (dekstop)](https://invis.io/9EN908ZQPAX#/311617870_Edit_Profile)

### Zeplin:
- [rendered profile (desktop)](https://zpl.io/beLnmeN)
- [rendered profile (mobile)](https://zpl.io/bPvWRAy)
- [profile form (mobile)](https://zpl.io/bLxjXYm)

### acceptance criteria
- [ ] display two inputs by default, so users know they can add more than one
- [ ] allow for adding more web sites
- [ ] for recognized social sites, display an icon in the rendered view
  - svg icons are in Zeplin
- [ ] for unrecognized social sites, return a (validation) error (text coming soon)
- [ ] to discourage abuse (link spamming) these links should have `nofollow` attributes

pdpinch:
@gsidebo I asked Ferdi et al. you question about validating "social links" and I added a checkbox to the requirements for validating the social links. Let me know if you need more detail.

gsidebo:
@pdpinch Re: "allow for adding more web sites" - what does that mean exactly? From the [rendered profile](https://app.zeplin.io/project/5b181d3bf71408ae4d406b94/screen/5bc0d7424859890a3e0c1a69) I looks like we're just looking to show links for a single website and several social sites. Does "web site" in this case mean "personal web site or social site"?

PR #1485, opened 2018-11-13
Added backend support for adding/deleting user websites

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Provides the back end fro #1398 (a separate PR for the UI changes will be opened)

#### What's this PR do?
- Creates `UserWebsite` model
- Provides views for adding/deleting user websites

- Check that you can add and delete user websites via HTTP requests
- Check that you can't add multiple websites of the same 'type' for a user (e.g.: you can't submit 2 different sites with a 'facebook' domain, you can't submit 2 different sites with a domain that doesn't belong to one of our supported social sites since they will both be considered personal sites)
- Load your user's profile page and check the network tab - if you have created some user websites, they should be included in the `/api/v0/profiles/<your_username>/` response data

Code samples for the above using a Django shell:
```python
USER_EMAIL='<YOUR_USER_EMAIL_HERE>'
import json
from django.urls import reverse
from django.test.client import Client
user = User.objects.get(email=USER_EMAIL)
profile = user.profile
client = Client()
client.force_login(user)

# Create a new user website
websites_url = reverse('user_websites_api-list')
post_data = dict(
    profile=profile.id,
    url='http://www.facebook.com/someone'
)
resp = client.post(websites_url, data=json.dumps(post_data), content_type='application/json')

# Delete a user website
USER_WEBSITE_ID_TO_DELETE = 3
website_delete_url = reverse('user_websites_api-detail', args=[USER_WEBSITE_ID_TO_DELETE])
resp = client.delete(website_delete_url)
```


profiles/api.py
```diff
@@ -38,3 +45,20 @@ def get_channels(user):
...
+    """
+    Gets a site type (as defined in profiles.models) from the given URL
+
+    Args:
+        url (str): A URL
+
+    Returns:
+        str: A string indicating the site type
+    """
+    no_fetch_extract = tldextract.TLDExtract(suffix_list_urls=False)
```
mbertrand: [This uses the TLD snapshot](https://github.com/john-kurkowski/tldextract#note-about-caching), right?  Seems like the safest option since it shouldn't change much.

profiles/api.py
```diff
@@ -38,3 +45,20 @@ def get_channels(user):
...
+    extract_result = no_fetch_extract(url)
+    if extract_result.domain in SITE_TYPE_OPTIONS:
```
mbertrand: Change to `extract_result.domain.lower()`: if I enter 'HTTP://WWW.FACEBOOK.COM' as my website then it's assigned a `site_type` of 'personal'

profiles/api_test.py
```diff
@@ -41,3 +47,18 @@ def test_get_channels(user):
...
+
+
+@pytest.mark.parametrize(
+    "url,exp_site_type",
+    [
+        ("http://facebook.co.uk", FACEBOOK_DOMAIN),
+        ("http://twitter.com", TWITTER_DOMAIN),
+        ("https://www.linkedin.com", LINKEDIN_DOMAIN),
+        ("https://not.a.socialsite.ca", PERSONAL_SITE_TYPE),
+        ("bad_url", PERSONAL_SITE_TYPE),
```
mbertrand: Add test for a capitalized domain like `HTTP://TWITTER.com`

profiles/api.py
gsidebo: Great catch 👍

profiles/api.py
gsidebo: I'm passing in `suffix_list_urls=False` here because (1) I got permission errors in my docker containers when the library was trying to write to that cache file, and (2) my tests were failing because the library was attempting to make a live request for some information about a URL (https://github.com/john-kurkowski/tldextract/blob/master/tldextract/remote.py#L27). (2) could be solved by monkeypatching, but in general it seemed easiest just to forego the fetch and I didn't see any risks. Accurately matching a domain only matters for our social sites (facebook, twitter, linkedin), and I think we can be confident that there won't be any funky new TLDs for those sites. Could be wrong, though...

PR #1504, opened 2018-11-16
Add channel header to post detail and channel settings

### LINKED ISSUE #960: Add loading indication for members tab while contributors and moderators load

### LINKED ISSUE #1388: Add channel banner to post detail page & channel settings page
- [ ] Desktop post detail page, logged in: https://zpl.io/VQEJqKk  
- [ ] Desktop post detail page, not logged in: https://zpl.io/2ZzODOd
- [ ] Mobile post detail page: https://zpl.io/2vW869r
- [ ] Desktop channel settings page: https://zpl.io/V00ry1l

noisecapella:
#### Pre-Flight checklist

  - [x] Desktop screenshots 

Fixes #1388 
Fixes #960 

#### What's this PR do?
Adds `withChannelHeader`, refactors `withChannelSidebar` and uses them on the post list and channel settings page

Click between the various pages, especially the post list and channel settings page. Only part of the page should refresh. For the channel settings page you should not see the channel header navbar, only the settings navbar.


noisecapella:
Channel settings, desktop:
channel_settings_desktop

Mobile:
channel_settings_mobile

Post detail, desktop:
post_detail_header_desktop

Mobile:
post_detail_header_mobile

pdpinch:
Yuck, the navigation for channel settings on mobile looks pretty bad. That's out of scope for this issue, so I'll open a separate issue.

pdpinch:
Otherwise, the layout & styling looks good.

aliceriot:
I believe it already looks like that (I assume you're referencing the options flowing on to two lines) on master

pdpinch:
Yeah, that's what I meant and it is on master. I opened #1514 to address it, along with some other issues.

noisecapella:
@aliceriot This is ready for review now

aliceriot:
cool, will take a look today

aliceriot:
looking through, the functionality and so on looks good. I'm wondering whether this is the best approach overall though - we're adding a new HOC (to add the channel header) to each page where we want it. But all of these pages are nested under the same route:

- `/c/:channelName`
- `/c/:channelName/search`
- `/c/:channelName/:postID/:postSlug?`
- `/c/:channelName/:postID/:postSlug?/comment/:commentID`

I think we might be able to use react-routers nested routes here, so that we'd have the channel header and the channel information being fetched in one place, by the channel component, and then the other pages would be rendered as nested routes within that component. It would look something like this (pseudocode):

```js
class ChannelPage extends React.Component {
  render () {
    return (
      <React.Fragment>
        <ChannelHeader />
        <Route exact path={match.url} component={ChannelPostList} />
        <Route path={`${match.url}/search/`} component={SearchPage} />
        <Route path={`${match.url}/:postID/:postSlug?`} component={PostPage} />
      </React.Fragment>
    )
  )
}
```

where `ChannelPostList` is a component that does the work to fetch and render the list of posts for a channel. What do you think about that? I think this is probably more the way we should have built these pages initially (including putting the settings page in here too, instead of under `/manage`) but it would be a substantial refactor on top of what you've already done to do it. So we can definitely not go that route, I just was thinking that the existing approach we're using (here and across the app really) doesn't really take advantage of the things that react-router lets us do, like pulling functionality shared across a number of pages out into a component rendered at a top-level route, which all the separate pages are then nested under. Anyhow, just some thoughts I had! You can take or leave the suggestion, and if you don't want to go down this route now I might file a tech debt issue to refactor this later.

noisecapella:
That's interesting, I didn't think about how react-router could be used this way. I think I'll leave it as is just to keep the PR simple but I'd support a refactor down the line

static/js/hoc/withChannelSidebar.js
```diff
@@ -4,25 +4,27 @@ import R from "ramda"
 
 import ChannelSidebar from "../components/ChannelSidebar"
 import Sidebar from "../components/Sidebar"
-import { Grid, Cell } from "../components/Grid"
+import withTwoColumns from "./withTwoColumns"
 
 const withChannelSidebar = R.curry(
   (className: string, WrappedComponent: Class<React.Component<*, *>>) => {
+    const TwoColumnComponent = withTwoColumns(className, WrappedComponent)
```
aliceriot: IMO this is getting a bit too difficult to trace the code, when we have a HOC that ends up wrapping the component with another HOC. If I want to figure out where the class name (`main-content two-column channel-page`) set on the channel page container, for instance, is set I have to open and look at three different files. what do you think about just keeping things simple with a `Grid` and two `Cells` in this component?

static/js/containers/admin/ChannelModerationPage.js
```diff
@@ -155,6 +159,6 @@ export default R.compose(
   connect(mapStateToProps),
   withPostModeration,
   withCommentModeration,
-  withSingleColumn("edit-channel"),
-  withSpinnerLoading
+  withChannelHeader(false),
+  withSingleColumn("edit-channel")
```
aliceriot: not sure how much we are concerned about this, but this change removes the loading spinner from the moderation page, so when you navigate to it it says "No outstanding reports" for a second before the request to get moderation reports finishes, and then it flashes to show the reports. Before I think there was a loading spinner.

static/js/hoc/withChannelSidebar.js
noisecapella: Sounds good

static/js/containers/admin/ChannelModerationPage.js
noisecapella: I added back a spinner, what do you think of the refactor here?

noisecapella:
Any other comments?

aliceriot:
the spinner on the moderation page looks good, reading through code now

static/js/containers/admin/ChannelModerationPage.js
```diff
@@ -133,13 +145,21 @@ export class ChannelModerationPage extends React.Component<Props> {
...
+            LoadingComponent: Loading,
+            render:           () =>
+              reports.length === 0 ? (
+                <Card title="Reported Posts & Comments">
+                  <div className="empty-message">No outstanding reports</div>
+                </Card>
+              ) : (
+                reports.map(this.renderReport)
+              )
+          })}
```
aliceriot: it might make sense to pull this anonymous function into a `renderReports` instance method for readability

static/js/components/Loading.js
```diff
@@ -72,6 +72,36 @@ export const PostLoading = () => (
...
+  if (errored) {
+    return <div className="errored">Error loading page</div>
+  }
+
+  if (!loaded) {
+    return <LoadingComponent />
+  }
+
+  return <div className="loaded">{render()}</div>
+}
```
aliceriot: this always returns a react component, so I think in `withLoading` below and on `ChannelModerationPage` you could just call this component in JSX instead of as a function call.

`withLoadingRender` is not clearest name I think, maybe something like `LoadingRenderDeferrer` or `DelayedRenderer` or something like that? I'm not sure what is a great name for this, but something that captures more what it's doing would be good I think.

noisecapella:
How does that look?

PR #1529, opened 2018-11-26
Don't try to update profile index more than once on image change

### LINKED ISSUE #1522: Update By Query API request resulted in 1 version conflict(s) (alias: discussions_profile_default...
https://sentry.io/mit-office-of-digital-learning/discussions/issues/669606987/?referrer=github_plugin

```
Update By Query API request resulted in 1 version conflict(s) (alias: discussions_profile_default, query: {'query': {'bool': {'must': [{'match': {'author_id': '01CS1TFEMHQQM0MK92781Y0NG5'}}]}}})
```

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes #1522

#### What's this PR do?
Modifies the logic used to decide whether or not to update the profile index after changes are made to the `Profile` model object.  When the profile image is changed, `Profile.save` is called 3 times (additional two times are to save medium and small images), which caused the profile ES index doc to get 3 update requests.  

-On master branch, if you upload a new profile image, an error message of `Update By Query API request resulted in 1 version conflict(s)` should show up in your console.
-On this branch, that error message should not occur.
-The profile document in ES should still be updated with your new images (medium and small).

### DIFF
--- a/profiles/models.py
+++ b/profiles/models.py
@@ -54,6 +54,7 @@ class Profile(models.Model):
 
     __previous_name = None
     __previous_headline = None
+    __previous_bio = None
 
     user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
 
@@ -86,6 +87,7 @@ class Profile(models.Model):
         super(Profile, self).__init__(*args, **kwargs)
         self.__previous_name = self.name
         self.__previous_headline = self.headline
+        self.__previous_bio = self.bio
 
     @transaction.atomic
     def save(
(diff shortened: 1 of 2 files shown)


profiles/models.py
```diff
@@ -119,12 +121,14 @@ def save(
             or self.name != self.__previous_name
             or self.headline != self.__previous_headline
```
noisecapella: Do we need to add `bio` here too? Is `headline` something which needs to be updated for all posts but `bio` isn't?

mbertrand: `author_bio` isn't in the post/comment ES mapping, but `author_name` and `author_headline` are.

PR #1535, opened 2018-11-28
Add django-hijack

mbertrand:
#### Pre-Flight checklist
- [ ] Testing
  - [ ] Code is tested

Closes #

#### What's this PR do?
Adds django-hijack so admins can log in as another user.

Log in as someone with access to the admin console.  Go to `/admin/auth/user/`.  Next to each user should be a button `Log in as <username>`.  Click one of those buttons.  You should be able to view the channels that user has access to, create posts and comments as that user, and edit the user's profile.

Admin:
screen shot 2018-11-28 at 8 54 49 am

Logged in via Hijack:
screen shot 2018-11-28 at 8 53 58 am


rhysyngsun:
We might want to consider blocking access to the login / register pages and the social auth pipeline backend, because it's feasible that an account could be inadvertently linked to a different social auth user.

mbertrand:
@rhysyngsun when logged in via hijack, just the logout link is shown, the admin user would have to explicitly enter the login or register URL or maybe click on such a link from an email. 

If I login as another user via hijack, and in another tab go to MM and log in or register as someone else and click the 'discussions' link, I get an `AuthAlreadyAssociated at /complete/micromasters/` error.  If I then logout from discussions and try again from MM, I am logged in properly as the MM user.

pdpinch:
Does changing the base template cover all the pages, including the home page? When I try this I can hijack the sessions but the home page rendering still acts as though I haven't logged in. 

screen shot 2018-11-29 at 3 16 39 pm

rhysyngsun:
@mbertrand what if you go to touchstone as someone else for a touchstone account not currently associated?

mbertrand:
I can add something like this to the social-auth views, though not sure if it's necessary:
```
        if request.session.get('is_hijacked_user', False):
            return Response(status=status.HTTP_403_FORBIDDEN)
```

rhysyngsun:
@mbertrand yes, that should probably be enough, but I think you'll need to add it as a pipeline action (probably first in the list of actions) so it gets caught regardless of how the pipeline is called.

A few cases I could see this happening are:

- Direct navigation to /login
- A stale browser tab already on /login

authentication/pipeline/user.py
```diff
@@ -187,6 +187,9 @@ def require_touchstone_login(
         backend (social_core.backends.base.BaseAuth): the backend being used to authenticate
         flow (str): the type of flow (login or register)
     """
+    # As first step in pipeline, stop a hijacking admin from going any further
```
rhysyngsun: To keep things clean, can you split this out to it's own pipeline step instead?

rhysyngsun: Can you write a test for this too?

authentication/pipeline/user.py
```diff
@@ -187,6 +187,9 @@ def require_touchstone_login(
...
+    if strategy.session_get("is_hijacked_user", False) is True:
```
rhysyngsun: Shouldn't need `is True` here right?

authentication/pipeline/user.py
mbertrand: ✔️

authentication/pipeline/user.py
mbertrand: ✔️

authentication/pipeline/user.py
```diff
@@ -242,3 +239,19 @@ def require_micromasters_provider(
...
+
+    Args:
+        strategy (social_django.strategy.DjangoStrategy): the strategy used to authenticate
+        backend (social_core.backends.base.BaseAuth): the backend being used to authenticate
+        user (User): the current user
+        flow (str): the type of flow (login or register)
+    """
+    # As first step in pipeline, stop a hijacking admin from going any further
+    if strategy.session_get("is_hijacked_user"):
+        raise ValueError("You are hijacking another user, don't try to login again")
```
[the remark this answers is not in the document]
mbertrand: ✔️

PR #1545, opened 2018-11-30
Enable post voting in search results

### LINKED ISSUE #1539: Search: allow voting in search results
### acceptance critieria
- [x] allow up-votes on post summary cards
- [ ] allow up-votes and down-votes on comment cards

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Partially fixes #1539

#### What's this PR do?
Makes the upvote button work for posts in search results.  It does this by adding `toggleUpvote` and `upvotedPost` props to the `SearchResult` component.  `toggleUpvote` is the standard upvote function.  `upvotedPost` is the result returned after the `toggleUpvote` function is dispatched.  An `upvotedPosts` map has also been added to the state of `SearchPage` to keep track of all the posts that have been voted on.

This allows for the upvote buttons to be rendered correctly (black if upvoted, gray if upvote removed) with updated scores after they are clicked.  However, all the buttons will initially be rendered as gray because ElasticSearch has no way of knowing if the current user has already upvoted that post or not.  

- Do a channel-specific search and a global search.  
- For each search, upvote multiple posts by other authors (not your own).  The upvote buttons for those posts should all turn black.  The score may or may not increase depending on whether or not you already upvoted that post in the past.  
- Click the upvote button again on some of those posts to cancel your upvotes.  They should turn gray and the scores should decrease by 1.

### DIFF
--- a/static/js/containers/SearchPage.js
+++ b/static/js/containers/SearchPage.js
@@ -28,7 +28,6 @@ import { Cell, Grid } from "../components/Grid"
 import { toggleUpvote } from "../util/api_actions"
 
 type Props = {
-  dispatch: Dispatch<any>,
   location: Location,
   history: Object,
   channel: ?Channel,
@@ -46,13 +45,13 @@ type Props = {
   searchLoaded: boolean,
   searchProcessing: boolean,
   total: number,
-  upvotedPost: ?Post,
-  clearSearch: () => void
+  clearSearch: () => void,
+  toggleUpvote: () => void,
+  upvotedPosts: Map<string, Post>
 }
 type State = {
   text: string,
-  from: number,
-  upvotedPosts: Map<string, Post>
+  from: number
 }
 
 const shouldLoadChannel = (currentProps: Props, prevProps: ?Props) => {
@@ -79,9 +78,8 @@ export class SearchPage extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props)
     this.state = {
-      text:         qs.parse(props.location.search).q,
-      from:         0,
-      upvotedPosts: new Map()
+      text: qs.parse(props.location.search).q,
+      from: 0
     }
   }
 
@@ -169,10 +167,10 @@ export class SearchPage extends React.Component<Props, State> {
       searchProcessing,
       initialLoad,
       total,
-      dispatch,
-      upvotedPost
+      upvotedPosts,
+      toggleUpvote
     } = this.props
-    const { from, upvotedPosts } = this.state
+    const { from } = this.state
 
     if (searchProcessing && initialLoad) {
       return <PostLoading />
@@ -182,11 +180,6 @@ export class SearchPage extends React.Component<Props, State> {
       return (
         <div className="empty-list-msg">There are no results to display.</div>
       )
-    }
-
-    if (upvotedPost && upvotedPosts.get(upvotedPost.id) !== upvotedPost) {
-      upvotedPosts.set(upvotedPost.id, upvotedPost)
-      this.setState({ upvotedPosts })
     }
 
     return (
@@ -201,7 +194,7 @@ export class SearchPage extends React.Component<Props, State> {
           <SearchResult
             key={i}
             result={result}
-            toggleUpvote={toggleUpvote(dispatch)}
+            toggleUpvote={toggleUpvote}
             upvotedPost={
               result.object_type === "post"
                 ? upvotedPosts.get(result.post_id)
@@ -256,9 +249,8 @@ export class SearchPage extends React.Component<Props, State> {
 
 const mapStateToProps = (state, ownProps) => {
   const channelName = getChannelName(ownProps)
-  const { channels, search, postUpvotes } = state
+  const { channels, search, posts } = state
   const channel = channels.data.get(channelName)
-  const upvotedPost = postUpvotes.data
   const channelLoaded = channels.loaded
   const searchLoaded = search.loaded
   const searchProcessing = search.processing
@@ -271,6 +263,7 @@ const mapStateToProps = (state, ownProps) => {
     ? channels.error && channels.error.errorStatusCode === 403
     : false
   const { results, total, initialLoad } = search.data
+  const upvotedPosts = posts.data
 
   return {
     results,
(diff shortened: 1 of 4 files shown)


static/js/containers/SearchPage.js
```diff
@@ -171,6 +184,11 @@ export class SearchPage extends React.Component<Props, State> {
       )
     }
 
+    if (upvotedPost && upvotedPosts.get(upvotedPost.id) !== upvotedPost) {
+      upvotedPosts.set(upvotedPost.id, upvotedPost)
+      this.setState({ upvotedPosts })
+    }
```
aliceriot: I think with this approach it is possible to run into a race condition if the user upvotes / downvotes a whole bunch of posts very quickly on a slow connection. I'm not exactly sure what is causing it, but if I use chrome's connection throttler to slow down the connection a ton I can click upvote on like five or six posts a whole bunch and the UI sometimes (not always) gets out of sync with the actual upvoted / not upvoted state.

I think the source of the issue is that the `data` property on the post upvote reducer (which is used to get the `upvotedPost` value here) only seems to store the value of the most recently clicked one, I think when there are a bunch of requests going off simultaneously it's possible for this value to change in between re-renders of the react component, so one of the posts that should be added to the `upvotedPosts` map is dropped on the floor. At least that's what I think might be going on, I'm not sure.

I'm not sure the best way to get around this, or if we think it will be a huge issue. One solution might be to add an instance method here that fires off the request with the `toggleUpvote` function and then adds the post to the `upvotePosts` map.

One other thing, I'm not sure we want to be mutating the existing map by calling `.set` - we might want to make a new `Map` and then add the new post to that:

```js
const updatedPostMap = new Map(upvotedPosts)
updatedPostMap.set(upvotedPost.id, upvotedPost)
this.setState({ upvotedPosts: updatedPostMap })
```

something like that, just to avoid mutating an object we're holding in `this.state`.

what do you think?

static/js/containers/SearchPage.js
```diff
@@ -271,7 +301,8 @@ const mapDispatchToProps = (dispatch: Dispatch<*>, ownProps: Props) => ({
   getChannel: async () => {
     const channelName = getChannelName(ownProps)
     await dispatch(actions.channels.get(channelName))
-  }
+  },
+  dispatch: dispatch
```
aliceriot: I think you could just have `toggleUpvote: toggleUpvote(dispatch)` here if you wanted to simplify a little bit

static/js/containers/SearchPage_test.js
```diff
@@ -271,7 +279,8 @@ describe("SearchPage", () => {
     assert.deepEqual(inner.state(), {
       // Because this is non-incremental the previous from value of 7 is replaced with 0
       from: 0,
-      text: undefined
+      text: undefined,
+      upvotedPosts
```
aliceriot: if it's not too much trouble it might be good to add a test which just makes sure that the post we expect is added to the `upvotedPosts` map

static/js/containers/SearchPage.js
aliceriot: I don't think it would run into the same issue on the homepage and channel page. If you look at the implementation of the `toggleUpvote` function it, after dispatching the async call to the API, dispatches a `SET_POST_DATA` action which will put the updated post record (returned from the API call) into the post reducer. Then the UI on those pages just gets the data out of the post reducer. I don't think that setup is vulnerable to the same race condition, because the data update is done asynchronously after the API action completes for every upvote or downvote that we kick off. 

Here we're not using the post reducer, we're using a different piece of data (on the `postUpvotes` reducer). One way that we could fix things I think would be to look at the `post` reducer for our data instead of the `postUpdate` one, since the `toggleUpvote` function as we're using it here is already going to stick the updated post into the `post` reducer without us doing anything (if you look at the dispatched actions when voting on the search page you can see the `SET_POST_DATA` actions firing).

does that make sense?

static/js/containers/SearchPage.js
aliceriot: hmm, weird...I tried it out and it seemed to work ok:

```diff
diff --git a/static/js/containers/SearchPage.js b/static/js/containers/SearchPage.js
index cd0b0a17..e1c2b665 100644
--- a/static/js/containers/SearchPage.js
+++ b/static/js/containers/SearchPage.js
@@ -170,7 +170,8 @@ export class SearchPage extends React.Component<Props, State> {
       initialLoad,
       total,
       dispatch,
-      upvotedPost
+      upvotedPost,
+      toggleUpvote
     } = this.props
     const { from, upvotedPosts } = this.state
 
@@ -201,7 +202,7 @@ export class SearchPage extends React.Component<Props, State> {
           <SearchResult
             key={i}
             result={result}
-            toggleUpvote={toggleUpvote(dispatch)}
+            toggleUpvote={toggleUpvote}
             upvotedPost={
               result.object_type === "post"
                 ? upvotedPosts.get(result.post_id)
@@ -302,7 +303,7 @@ const mapDispatchToProps = (dispatch: Dispatch<*>, ownProps: Props) => ({
     const channelName = getChannelName(ownProps)
     await dispatch(actions.channels.get(channelName))
   },
-  dispatch: dispatch
+  toggleUpvote: toggleUpvote(dispatch)
 })
 
 export default R.compose(
```

it's kind of a nit though, so feel free to leave as-is

static/js/containers/SearchPage.js
mbertrand: I think so, thanks!

static/js/containers/SearchPage.js
```diff
@@ -271,6 +263,7 @@ const mapStateToProps = (state, ownProps) => {
     ? channels.error && channels.error.errorStatusCode === 403
     : false
   const { results, total, initialLoad } = search.data
+  const upvotedPosts = posts.data
```
aliceriot: I just noticed that this change has a nice side effect, which is that the `posts` reducer also holds any posts which the user has already loaded from the frontpage or channel APIs. So if I've upvoted a post, and it was already returned by the frontpage or the channel page API, then I'll see accurate information about whether it's upvoted when I look at the same post in search. Nice!

Issue #1460, opened 2018-11-06
Widgets: include channel widgets on channel page, post detail page


pdpinch:
As a user, I need to be able to view the widgets chosen by the channel moderator on the channel page and the post detail page. 

### related issues 
As a channel moderator, I also need to add, configure and remove widgets from my channel page. The management of widgets is covered in #1508 

Estimate: 16 hours

pdpinch:
@rhysyngsun is going to address the backend for this in his branch for adding the widget framework. 

Managing widgets has been moved to #1508 

The remainder of this issue is the front-end for displaying widgets lists in the channel page.

PR #1558, opened 2018-12-04
Add widget JS to open-discussions

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of #1460 

#### What's this PR do?
This copies over the widget frontend code, roughly from https://github.com/mitodl/open-widget-framework/pull/29

I tried to make very few changes otherwise, there will be other PRs for that

No UI yet


static/js/components/widgets/ListWrapper.js
```diff
@@ -0,0 +1,118 @@
...
+          className="edit-widget-list-bar btn-group card-header"
+          role="group"
+        >
+          <button
+            className={`btn btn-info col edit-widget-list-btn ${
+              editMode ? "active" : ""
+            }`}
+            onClick={this.toggleEditMode}
+          >
+            <Octicon name="pencil" />
```
rhysyngsun: Could you change this and the other usage to use material-icons classes like we do elsewhere so we don't need to pull in this extra dependency for a few icons? I think @aliceriot had advised that a few weeks ago but it got missed.

static/js/components/widgets/NewWidgetForm.js
```diff
@@ -0,0 +1,91 @@
...
+   *    the backend
+   *
+   * State:
+   *    widgetClasses: the classes of the widget available to create
+   *    widgetClassConfigurations: an object that maps the widget classes to those classes' configurations
+   *
+   * Props:
+   *    formProps (see makeFormProps in widget_list.js)
+   *    formWrapperProps (defined by custom FormWrapper or _defaultFormWrapper in defaults.js
+   */
```
aliceriot: maybe remove all these huge comments that don't really add any information?

noisecapella:
Ready for review again

PR #1571, opened 2018-12-11
Location for profiles

### LINKED ISSUE #1399: Profile: add location
As a user, I'd like to be able to share my location (as defined by me) so that others users "nearby" can find me. 

Design TBD, but we are thinking of something like Twitter. If you start typing a city, state or country, it helps you complete it. If you want to type your own string, that's also accepted. 

### acceptance criteria
- [ ] choose a placename API for helping with autocompletion and also converting to geo coordinates
  - in the past we had trouble with Google because it was blocked in some regions
  - other possibilities include https://nominatim.openstreetmap.org/ and something Ferdi recommended
  - need to review the terms for any API we choose, to consider cost and what data we can persist from the API
- [ ] extend profile model to include location -- as a string and also as geocoordinates 

### questions
- Should we add this data to elasticsearch?

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots 
- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #1399

#### What's this PR do?
- Adds an input field on the profile edit form for entering a city.
- Uses the Algolia Places API to look up and display possible matches for the user's input
- Saves the entire JSON object for the selected match to the `Profile` model
- Displays the 'value' attribute of the JSON object on the profile page.

- [Sign up for an Algolia account](https://www.algolia.com/users/sign_up) and create/get your app id and a search-only api key
- Set your `ALGOLIA_APP_ID` and `ALGOLIA_API_KEY` .env values
- Go to your profile page.  It should look the same as usual.
- Edit the profile.  There should be a new field with "Location (city)" as help text.
- Start typing in a city name.  You should soon see a list of suggestions.
- Choose a suggested city and submit the form.
- The city you chose should appear on the profile page underneath your profile image.
- Edit the profile again.  Clear out the location field and submit.
- The profile page should no longer show a location.
- Edit the profile again.  Enter a nonexistent city name with no match and submit.
- The profile page should display your fake city.

The location information returned from the API is being saved as a single JSON field in the `Profile` model for now, because it is still uncertain how much of the location data we will need and how we want to use it in the future.  Once that becomes clearer, it might make more sense to have a separate `Location` model with defined fields.

@pdpinch @Ferdi the `Built by Algolia` blurb under the suggestions list is mandatory according to the terms of service if we use the `places.js` library.  Also, the matched location text is not always consistent about the formatting it uses.  Most of the time it displays `<city>, <state/province>, <country>` but there are some notable exceptions such as New York City, which just displays `New York, United States of America`.

screen shot 2018-12-10 at 1 05 42 pm

screen shot 2018-12-10 at 1 05 50 pm

screen shot 2018-12-10 at 12 45 52 pm

screen shot 2018-12-10 at 12 46 26 pm


mbertrand:
@shaidar @blarghmatey will need an algolia app id and api key for ci, rc, and production environments when this is merged: https://www.algolia.com/users/sign_up.  Should be restricted by HTTP referrer since the keys will be exposed.

profiles/views_test.py
```diff
@@ -228,16 +233,22 @@ def test_patch_profile_by_user(client, logged_in_profile):
...
+    location_json = {"value": "Boston"}
     with make_temp_image_file(width=250, height=250) as image_file:
         # format patch using multipart upload
         resp = client.patch(
             url,
-            data={"bio": "updated_bio_value", "image_file": image_file},
+            data={
+                "bio": "updated_bio_value",
+                "image_file": image_file,
+                "location": json.dumps(location_json),
```
rhysyngsun: I think you should just be able to pass the nested object in without having to serialize to a string

static/js/components/LocationPicker.js
```diff
@@ -0,0 +1,75 @@
...
+  render() {
+    const { placeholder, initialLocation } = this.props
+
+    return (
+      <div>
+        <input
+          type="text"
+          placeholder={placeholder}
+          defaultValue={initialLocation}
+          ref={ref => {
```
rhysyngsun: Can you use the newer `React.createRef()` instead?

static/js/components/LocationPicker.js
```diff
@@ -0,0 +1,75 @@
...
+      ...options,
+      container: this.autocompleteElem
+    })
+
+    this.autocomplete.on("change", onChange)
+    this.autocomplete.on("clear", onClear)
+  }
+
+  shouldComponentUpdate() {
+    return false
```
rhysyngsun: Could you add a comment explaining why this is statically `false`? I can guess from the code, but I'm not sure I'm right or that we'll remember.

rhysyngsun:
Functionality looks good.

static/js/components/LocationPicker.js
mbertrand: I think I can actually remove this method altogether.

profiles/views_test.py
mbertrand: `multipart uploads do not support nested data` (multipart being used here for the simulated image upload).

static/js/components/LocationPicker.js
mbertrand: ✔️

Issue #1495, opened 2018-11-15
Article: support as a configurable post type in channels


rhysyngsun:
We want to allow moderators to configure the channel as allowing or disallowing article post types.

- Existing channels settings should not change
- New channels should default to only `link | text`

### Design

- In order to permit article post types, these are shadowed in reddit as a text (self post) with no content
- This means if a channel wants to be configured to only allow article post types, we need to permit text posts in reddit, but disallow them at the api layer.
- To accomplish this, we need to define a new source of truth in the database for `allowed_post_types`.
- Since we're moving validation of the link types to our API layer, we should update channel settings to permit "any" link type so we don't have to bother with the complexity required to translate our expanded post type settings into equivalent reddit settings.

###### Stage 1
- [ ] shadow `allowed_post_types` in `Channel`, update the API to set this value on subsequent channel API requests (creates + updates) and return it as a new field in `GET`s
- [ ] update to frontend to check both `link_type` and `allowed_link_types`, with the latter having precendence
- [ ] backpopulate `Channel.allowed_post_types` and `Post.post_type` given the above logic
###### Stage 2
- [ ] update the posts API to validate new posts against `Channel.allowed_post_types`
- [ ] default new channels to "any" link type
- [ ] update all channels' reddit settings to permit "any" link type
###### Stage 3 - _(permitting enough time from stage 1 deployment for clients to update)_
- [ ] update the UI to display articles as a post type option
- [ ] remove all logic around `link_type`

pdpinch:
@rhysyngsun I think this is all done?

rhysyngsun:
Yup

PR #1576, opened 2018-12-11
Added backend support for allowed post types

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [ ] Testing
  - [x] Code is tested

Part of #1495 

#### What's this PR do?
This updates the backend to accept a new configuration parameter for channels `allowed_posts_types`, which is a list of allowed types. The reason for this is to support the new "article" post type, while still integrating with native reddit post types and continuing to support the now-deprecated `link_type` field on channels.

- Test creating a channel in MM
- Test creating a channel in the discussions ui
- Test updating channel allowed post types in the ui
- Test that you can still create the allowed post types in those channels and can't create the ones that are disallowed


channels/tasks.py
```diff
@@ -131,3 +138,73 @@ def populate_subscriptions_and_roles(self):
...
+    Args:
+        channel_ids (list of int): list of post ids
+    """
+    client = Api(User.objects.get(username=settings.INDEXING_API_USERNAME))
+
+    for channel in Channel.objects.filter(id__in=channel_ids).iterator():
+        if channel.allowed_post_types:
+            continue
+
+        subreddit = client.get_subreddit(channel.display_name)
```
mbertrand: Use `channel.name` instead?  When this task runs via the `backpopulate_post_tasks` command, it throws an 
`AttributeError: 'Channel' object has no attribute 'display_name'`

channels/tasks.py
```diff
@@ -131,3 +138,73 @@ def populate_subscriptions_and_roles(self):
...
+
+        submission = client.get_submission(post.post_id)
+        if submission.url:
+            post.post_type = LINK_TYPE_LINK
+        elif post.article:
+            post.post_type = EXTENDED_POST_TYPE_ARTICLE
+        else:
+            post.post_type = LINK_TYPE_SELF
+
+        post.savE()
```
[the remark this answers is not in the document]
rhysyngsun: Probably more critically, a test would've caught this :woman_facepalming:

rhysyngsun:
@mbertrand addressed your feedback and added tests around the code. I also fixed the name generation on `ChannelFactory.create_batch()` to guarantee uniqueness since I ran into that.

channels/tasks.py
```diff
@@ -131,3 +139,73 @@ def populate_subscriptions_and_roles(self):
...
+    results = celery.group(
+        [
+            populate_post_post_type.si(ids)
+            for ids in chunks(
+                Post.objects.values_list("id", flat=True),
+                chunk_size=settings.ELASTICSEARCH_INDEXING_CHUNK_SIZE,
+            )
+        ]
+        + [
+            populate_channel_post_type.si(ids)
```
mbertrand: When run from `backpopulate_post_types` command, this prevents all existing channels from allowing creation of any new article posts (via `channels.api.get_allowed_post_types_from_link_type`).  Is this by design, so that channels have to be manually set to allow articles afterward?

rhysyngsun: Yes, that's intentional, because we want to deploy this with no functional or configuration change to production.

PR #1580, opened 2018-12-12
Fixed unverified user login bug

### LINKED ISSUE #1568: Attempting to log in prior to verifying email
When a user signs up with a non-MIT account and attempts to login with that account prior to verifying their account through the email sent to them, the following Sentry error is generated:

https://sentry.io/mit-office-of-digital-learning/discussions/issues/604126352/?query=is%3Aunresolved

```Received unexpected result: <HttpResponseRedirect status_code=302, "text/html; charset=utf-8", url="/">```

Also, the user does not get any notification about this issue as they continue clicking on the `Next` button.

### PR DESCRIPTION
**Pull request by gsidebo**
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1568 

#### What's this PR do?
Fixes login bug where a user who has registered but has not verified via email could go to the login page and attempt to log in unsuccessfully, but wouldn't see an error message.

1. Register a new email
1. Visit the login page without verifying using that email link
1. Enter that same email

You should see an error message indicating that the email doesn't exist

1. Ideally we would show an error message telling the user that they need to verify via the email link, but we can only determine that state if the user still has the same session active that they used to register the email. If they use a new browser, incognito window, or the session expires, we can't determine that they have registered but haven't verified. In order to keep the error messages consistent, I just opted to stick with the "doesn't exist" error in all cases
1. This fix is for new registrations. If a user in prod is still in this state and still has an active, unexpired session, this error may come up again, but it can only last as long as the lifespan of a session. Didn't seem worth it to add extra logic to handle this error that will be going away on its own soon.

### DIFF
--- a/authentication/serializers.py
+++ b/authentication/serializers.py
@@ -25,6 +25,9 @@ from authentication.utils import SocialAuthState
 from profiles.models import Profile
 from profiles.serializers import ProfileSerializer
 
+
+PARTIAL_PIPELINE_TOKEN_KEY = "partial_pipeline_token"
+
 log = logging.getLogger()
 
 User = get_user_model()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: authentication/serializers.py [hunk 1]
```diff
@@ -137,6 +140,7 @@ def save(self, **kwargs):
         """'Save' the auth request"""
         try:
             result = super().save(**kwargs)
+            # import pdb; pdb.set_trace()
```
rhysyngsun: Stray trace code

PR #1599, opened 2018-12-19
Cover image for articles - backend

### LINKED ISSUE #1593: Article: Cover Image backend
(from #1492)
- [ ] Add a cover_image file field to the Article
- [ ] Support setting the cover image via the API when the post is created
- [ ] Return the cover image in post APIs (listing and single-post)

### Notes:
- image is optional
- the image should be editable, i.e. the user should be able to choose a new image, or remove it
- we'd like to use a similar front end to the one used for Article images. This may obviate the need for image upload, since that's handled by the ckeditor image API.

mbertrand:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #1593 

#### What's this PR do?
- Add `cover_image` and `cover_image_small` fields to the `Article` model
- Updates the `PostSerializer` to handle the new fields (get/post/patch)
- Customizes `Article.save` method to autogenerate the thumbnail (`cover_image_small`) if the `cover_image` field is being updated.

There's no front-end yet but you can simulate creating an article post with cover image through the shell:

```python
import os
import json
from rest_framework.test import APIClient
from channels.models import Channel, Post, Article

client = APIClient()
client.force_login(User.objects.get(username='<username>'))
channel = Channel.objects.get(name="<channel_name>")
article_content = [{"key": "value", "nested": {"number": 4}}]
img_file = "static/images/mit-logo-micromasters.jpg"
url = reverse("post-list", kwargs={"channel_name": channel.name})
with open(img_file, "rb") as f:
    resp = client.post(
        url, 
        {"title": "foo", "article_content": json.dumps(article_content), "cover_image": f}, 
        format="multipart"
    )

post = Post.objects.get(post_id=resp.json()["id"])
print(post.article.cover_image.url, post.article.cover_image_small.url)
```

Then go to `http://od.odl.local:8063/api/v0/posts/<post_id>/` and verify that there are working URL's for `cover_image` and `thumbnail` that display the image file you used above.


profiles/utils.py
```diff
@@ -193,6 +193,22 @@ def banner_uri(instance, filename):
...
+    return generate_filepath(filename, instance.post.post_id, "_article", "article")
+
+
+def article_image_uri_small(instance, filename):
+    """
+    upload_to handler for Article.cover_image_small
+    """
+    return generate_filepath(
+        filename, instance.post.post_id, "_article_small", "article"
+    )
```
aliceriot: is it worthwhile to have some simple tests for these two functions?

channels/serializers.py
```diff
@@ -475,6 +498,13 @@ def update(self, instance, validated_data):
             elif validated_data["subscribed"] is False:
                 api.remove_post_subscription(post_id)
 
+        cover_image = validated_data.get("cover_image")
+        if instance and cover_image and hasattr(instance, "article"):
+            instance.article.cover_image.save(
+                f"article_image_{post_id}.jpg", cover_image, save=False
+            )
+            instance.article.save(update_fields=["cover_image"], update_image=True)
```
aliceriot: I'm wondering, should this logic for updating the article cover image maybe live in the `update_post` method on the `api` class? I'm just wondering because with this setup if you update an article post we're going to end up calling `.save` twice, once above in the call to `api.update_post` and again down here. Maybe it would be better to have this logic live in that method instead, so it's both all in one place and so that we can just call `.save` once?

channels/models.py
```diff
@@ -206,8 +212,30 @@ class Article(TimestampedModel):
...
+    )
+    cover_image_small = models.ImageField(
+        null=True, blank=True, max_length=2083, upload_to=article_image_uri_small
+    )
     content = JSONField()
 
+    def save(
+        self, *args, update_image=False, **kwargs
+    ):  # pylint: disable=arguments-differ
+        if update_image:
```
aliceriot: I'm wondering, could we also just do a check here to see if the update being made involves a change to the cover image, and use that fact to figure out whether we need to update the thumbnail and all? just wondering, because then I think we could, when we're updating this model, just pass in a new image or not and it would take care automatically of whether it needs to update the thumbnail or not, and we wouldn't need to know this detail of the `.save` method when we're calling it.

mbertrand: This current code is consistent with how images are being saved for channels (avatars, banners) and profiles: both the `Channel` model and `Profile` model have custom `save` methods using the same logic.

profiles/utils.py
mbertrand: ✔️

channels/serializers.py
mbertrand: ✔️

mbertrand:
@aliceriot comments addressed, I left the `Article.save(update_image=True)` method as is for now to be consistent with the other models, I think if that's going to be changed to use another approach it should be done for the other models as well.

channels/serializers.py
mbertrand: Oops, sorry, fixed now.

Issue #1604, opened 2018-12-21
Performance issues


pdpinch:
There's been a noticeable slowdown in the time for loading the front page and channel pages. 

I don't have before and after timing, but I can compare RC and production. 

On RC, it took me 870ms for the frontpage API to return. On production it took 4236ms. 

It's a little harder to compare channels across RC and prod, but here are two that have similar purposes, if not a similar number of posts:

393ms https://discussions-rc.odl.mit.edu/api/v0/channels/whatachannel/posts/
3817ms https://open.mit.edu/api/v0/channels/mit_odl_engineering/posts/

shaidar:
Based on info from New Relic, it looks like the following two calls are the ones taking quite a while:

PR #1605, opened 2018-12-21
Fixed article n+1 query

gsidebo:
#### Pre-Flight checklist

- [x] Testing

Related to #1604 

#### What's this PR do?
Fixes an n+1 query for `Article`s that was happening on channel pages/home page

Load one of the pages mentioned above and check your docker log output. You should not see a n+1 warning. You will still see as many HTTP requests as there are posts on the page (that can be addressed in a different PR)

### DIFF
--- a/channels/proxies.py
+++ b/channels/proxies.py
@@ -107,7 +107,7 @@ def proxy_posts(submissions):
         post.post_id: post
         for post in Post.objects.filter(
             post_id__in=[submission.id for submission in submissions]
-        )
+        ).select_related("article")
     }
     return [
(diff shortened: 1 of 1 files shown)


pdpinch:
Can you say more about the problem and how this solves it? It would help also to have the log line that you mention in the description.

gsidebo:
For posterity: pages that show a listing of posts (channel pages, front page) make a query for own own `Post` database objects that match a group of reddit id's. During the serialization of each post, the `article` property was being accessed, which resulted in a database query. Repeat that database query for each post being serialized

PR #1613, opened 2018-12-27
Fix other calls to fetch()

### LINKED ISSUE #1604: Performance issues

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1604 

#### What's this PR do?
This should wrap up any lingering performance issues relative to the combination of django model nullable relations, `ObjectProxy`, and `praw`'s fetch behavior.

Namely, this what the cause of extraneous requests to reddit on serializing the reponse:

1. `praw` does not set `Submission._fetched = True` for objects coming form listing generators
2. Check for a nullable django relation with `hasattr()`
3. Implementation of that attribute raises an `AttributeError`
4. This trigger's python's defualt behavior of calling the object's `__hasattr__` implementation, which is `ObjectProxy`'s one
5. `ObjectProxy.__hasattr__` proxies to praw's `_getattr__`, which calls `_fetch()` if `_fetched is False`

You'll not that I refactored relation attributes on `PostProxy` to perform the `hasattr()` checks and coerce to  `None` if the attribute doesn't exist, this avoids triggering the proxy behavior, although is non-standard as far as django goes so there's a small tradeoff here.

I also added a patch on `Submission.fetch()` to ensure we don't regress this issue in the future accidentally since `betamax` didn't help us catch this. 

Ensure that you can view and paginate the frontpage and channel listing pages.

### DIFF
--- a/channels/views/posts_test.py
+++ b/channels/views/posts_test.py
@@ -11,7 +11,10 @@ from profiles.utils import image_uri
 from channels.factories import LinkMetaFactory
 from channels.constants import VALID_POST_SORT_TYPES, POSTS_SORT_HOT
 from channels.models import Subscription, LinkMeta, Article
-from channels.views.test_utils import default_post_response_data
+from channels.views.test_utils import (
+    default_post_response_data,
+    raise_error_on_submission_fetch,
+)
 from open_discussions.constants import (
     NOT_AUTHENTICATED_ERROR_TYPE,
     PERMISSION_DENIED_ERROR_TYPE,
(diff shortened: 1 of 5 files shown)


channels/views/posts_test.py
```diff
@@ -24,6 +24,14 @@
 pytestmark = pytest.mark.betamax
 
 
+def raise_error_on_submission_fetch(mocker):
```
noisecapella: Can we import this from the other file?

rhysyngsun:
@noisecapella refactored that into `test_utils.py`

PR #1617, opened 2018-12-28
Implement mobile widget view

### LINKED ISSUE #1518: Widgets: mobile layout
In mobile, widgets are currently wrapped below the post summary cards in the channel list page. With infinite scroll, it's impossible to get to the widgets. 

In the design at https://zpl.io/bAE6Yy8 widget links are added to the tabs, with an overflow menu. When the user clicks on a link, the widget slides out over the page (see https://zpl.io/VknZJ0g) with a back link to return the user to the channel listing.

noisecapella:
#### Pre-Flight checklist

  - [x] Desktop screenshots 
- [x] Testing
  - [x] Code is tested

Fixes #1518 

#### What's this PR do?
Implements mobile view for widgets. In the mobile view widgets will be removed from the sidebar and shown on a separate page. The user should not see anything different in the desktop view.

Narrow the browser window to mobile width. Browse the channel page and verify that the navigation works as you expect.

Note that the react renderer will not update after you resize the browser window. You will need to refresh after any resize to test changes.

Desktop:
desktop

Mobile:
mobile_channel

mobile_2


static/js/containers/ChannelPage.js
```diff
@@ -111,9 +113,11 @@ export class ChannelPage extends React.Component<ChannelPageProps> {
       subscribedChannels,
       posts,
       location: { search },
-      renderPosts
+      renderPosts,
+      isModerator
     } = this.props
 
+    const isAbout = match.path === "/c/:channelName/about/"
```
rhysyngsun: This could be done more simply with nested `<Route/>` tags, but I don't think you can do that here without a non-trivial refactor of how the other pages under /c/:channelName/ are routed too. I'll open a ticket for it (#1623).

static/js/containers/ChannelPage.js
```diff
@@ -123,13 +127,26 @@ export class ChannelPage extends React.Component<ChannelPageProps> {
...
           </MetaTags>
-          <div className="post-list-title">
-            <PostSortPicker
-              updatePickerParam={updatePostSortParam(this.props)}
-              value={qs.parse(search).sort || POSTS_SORT_HOT}
-            />
-          </div>
-          {renderPosts()}
+          <Grid className="main-content two-column channel-page">
+            {!isAbout || !isMobileWidth() ? (
```
rhysyngsun: We might want to consider something like [react-media](https://www.npmjs.com/package/react-media) to handle hiding components responsively. Since this has the same limitation as the drawer currently does, you can take-it or leave-it.

static/js/containers/ChannelPage.js
```diff
@@ -123,13 +127,26 @@ export class ChannelPage extends React.Component<ChannelPageProps> {
...
+              <Cell width={8}>
+                <div className="post-list-title">
+                  <PostSortPicker
+                    updatePickerParam={updatePostSortParam(this.props)}
+                    value={qs.parse(search).sort || POSTS_SORT_HOT}
+                  />
+                </div>
+                {renderPosts()}
```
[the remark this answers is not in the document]
noisecapella: I think there would be some duplication as we will need to render the widget list in both the about page and the channel page for desktop view. Having the `ChannelPage` do both also solves the case where a desktop user gets a link to `/about` from a user who shares the page from their mobile device.

static/js/containers/ChannelPage.js
noisecapella: It might be useful but I'd like to deal with it in a separate PR so the drawer can use it too

static/js/containers/ChannelPage.js
noisecapella: Sounds good

static/js/containers/ChannelPage.js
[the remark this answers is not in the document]
rhysyngsun: Ok, as long as @pdpinch is fine with it, it's functional.

PR #1618, opened 2018-12-28
Reduce version conflict errors in ES when updating profiles

### LINKED ISSUE #1602: Update By Query API request resulted in 1 version conflict(s) (alias: discussions_profile_default...
We thought this was fixed by #1529, but there are still some conditions where it happens regularly. 

https://sentry.io/mit-office-of-digital-learning/discussions/issues/669606987/?referrer=github_plugin

```
Update By Query API request resulted in 1 version conflict(s) (alias: discussions_profile_default, query: {'query': {'bool': {'must': [{'match': {'author_id': '01CZ88RQVRP0KERXT87PBXTJ8A'}}]}}})
```

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #1602 

#### What's this PR do?
- Uses `update_document_with_partial` instead of `update_field_values_by_query` to update ES profile docs.  This eliminates most (but not all) of the version conflict errors
- Add `retry_on_conflict=1` parameter for the `update_author` call.  This seems to eliminate the errors entirely.

- On master branch, you should see numerous `Update By Query API request resulted in 1 version conflict(s)` errors in your celery log if you run the following:
```
from channels.api import *
from channels.models import *
from profiles.models import *

api = Api(User.objects.get(username=settings.INDEXING_API_USERNAME))
channel = "<your_channel_name>"

for profile in Profile.objects.exclude(user__username=settings.INDEXING_API_USERNAME):
  api.remove_contributor(profile.user.username, channel)
  api.remove_subscriber(profile.user.username, channel)
  api.add_contributor(profile.user.username, channel)
  api.add_subscriber(profile.user.username, channel)
```
- Switch to this branch and run the above again.  There should be no more errors.

I'm not sure why using the different ES update method reduces so many of the errors, my guess is that it runs much faster and that's what significantly reduces the probability of a conflict.  I was still seeing an occasional conflict when running 100's of updates at a time, until I added the `retry_on_conflict=1` parameter.


search/task_helpers.py
```diff
@@ -193,7 +193,12 @@ def update_author(user_obj):
     if user_obj.username != settings.INDEXING_API_USERNAME:
         profile_data = ESProfileSerializer().serialize(user_obj.profile)
         profile_data.pop("author_id", None)
-        update_fields_by_username(user_obj.username, profile_data, [PROFILE_TYPE])
+        update_document_with_partial.delay(
+            gen_profile_id(user_obj.username),
+            profile_data,
+            PROFILE_TYPE,
+            retry_on_conflict=1,
```
rhysyngsun: Is this worth putting behind an env variable width a default of 1 or are you fairly certain this value will fix things?

search/indexing_api.py
```diff
@@ -191,7 +191,9 @@ def update_field_values_by_query(query, field_dict, object_types=None):
             )
 
 
-def _update_document_by_id(doc_id, data, object_type, update_key=None):
+def _update_document_by_id(
+    doc_id, data, object_type, update_key=None, retry_on_conflict=0
```
rhysyngsun:
```suggestion
    doc_id, data, object_type, *, update_key=None, retry_on_conflict=0
```

mbertrand: ✔️

search/task_helpers.py
mbertrand: Added an env variable for it.

PR #1642, opened 2019-01-04
Search text input focus

### LINKED ISSUE #1538: Search: when clicking search icon, set focus to the search field
As a user, I expect to be able to click on the search icon and then immediately start typing. 

Currently. I have to click on search icon, and then click on the search field. Then I can start typing.

### acceptance criteria
- [ ] set focus in site search
- [ ] set focus in channel search

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes #1538

#### What's this PR do?
Sets the focus to the search textbox input when it is displayed.

- Click the search icon for a channel.  The focus should move to the search text input field after it is rendered.  Run a search, then click the X to clear the search text.  The focus should still be on the input field.
- Do the same for the global search icon.

### DIFF
--- a/static/js/components/SearchTextbox.js
+++ b/static/js/components/SearchTextbox.js
@@ -26,8 +26,12 @@ export default class SearchTextbox extends React.Component<Props> {
     this.focus()
   }
 
-  componentDidUpdate() {
-    this.focus()
+  componentDidUpdate(prevProps) {
+    // Focus again if input has been cleared
+    const { value } = this.props
+    if (!value && prevProps && prevProps.value !== value) {
+      this.focus()
+    }
   }
 
   render() {
--- a/static/js/components/SearchTextbox.js
+++ b/static/js/components/SearchTextbox.js
@@ -8,24 +8,49 @@ type Props = {
   value: string
 }
 
-const SearchTextbox = ({ onChange, onClear, onSubmit, value }: Props) => (
-  <div className="search-textbox">
-    <i className="material-icons search-icon" onClick={onSubmit}>
-      search
-    </i>
-    {value ? (
-      <i className="material-icons clear-icon" onClick={onClear}>
-        clear
-      </i>
-    ) : null}
-    <input
-      type="text"
-      className="underlined"
-      value={value}
-      onChange={onChange}
-      onKeyDown={event => (event.key === "Enter" ? onSubmit(event) : null)}
-    />
-  </div>
-)
+export default class SearchTextbox extends React.Component<Props> {
+  input: { current: null | React$ElementRef<typeof HTMLInputElement> }
 
-export default SearchTextbox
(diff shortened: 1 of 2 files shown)


static/js/components/SearchTextbox.js
```diff
@@ -8,24 +8,49 @@ type Props = {
...
+            clear
+          </i>
+        ) : null}
+        <input
+          ref={this.input}
+          type="text"
+          className="underlined"
+          value={value}
+          onChange={onChange}
+          onKeyDown={event => (event.key === "Enter" ? onSubmit(event) : null)}
```
aliceriot: just one quick comment - you may be able to do this by just setting the `autofocus` prop on this `<input>` tag

mbertrand: Using `autoFocus` will work initially, but if I click the X to clear out the search text, it doesn't retain focus.  The current approach will maintain the focus after clearing the text.

aliceriot: ahh gotcha :)

static/js/components/SearchTextbox.js
```diff
@@ -8,24 +8,49 @@ type Props = {
...
+      this.input.current.focus()
+    }
+  }
+
+  componentDidMount() {
+    this.focus()
+  }
+
+  componentDidUpdate() {
+    this.focus()
```
aliceriot: I noticed that one small side-effect of calling this `focus` method in `componentDidUpdate` is that any update to a prop that this component receives will cause the search input to grab focus for a second. this can be a bit odd, especially because clicking on the dropdown to change how the search results are sorted will cause that to happen.

I think on mobile this may cause the keyboard to come up when it happens, but I'm not sure (I don't have access to a simulator to test it out).

static/js/components/SearchTextbox_test.js
```diff
@@ -17,7 +17,8 @@ describe("SearchTextbox", () => {
   })
 
   const render = (props = {}) =>
-    shallow(
+    // mount is necessary to check focus
+    mount(
```
aliceriot: might be worth keeping the `shallow` based test renderer around for the tests that don't need mount - what do you think? you can quickly make a shallow renderer using the `configureShallowRenderer` test util btw.

static/js/components/SearchTextbox.js
mbertrand: I modified `componentDidUpdate` to focus only if the input value has changed and is empty.

static/js/components/SearchTextbox_test.js
mbertrand: There's only a few of them, but I reverted `render()` to use shallow as before and added a `renderMount` method for the tests that check on focus.

PR #1646, opened 2019-01-07
Remove text widget class and add wysiwyg widget field editor

### LINKED ISSUE #1639: Add a wysiwyg editor to the markdown widget
As a channel moderator, I'd like to have a simple wysiwyg editor for widgets. 

Because the existing "About this channel" widgets are stored in markdown, it will be easier to use prosemirror for this instead of ckeditor. 

Once we make this change, we should remove the text widget, since it's redundant to, and less functional than, the markdown widget. However, we may want to rename the markdown widget.

noisecapella:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Fixes #1639 

#### What's this PR do?
Adds a `markdown_wysiwyg` field type and updates the markdown widget to use it for the editor.

Create a text widget. Run migrations and verify that the widget instance is deleted.

Edit a markdown widget. You should be able to do anything the editor allows for and it should update the widget markdown accordingly.


pdpinch:
And change the button text from "Create widget" to "Add widget"

noisecapella:
I created https://github.com/mitodl/open-discussions/issues/1656 to cover the copy changes

widgets/serializers/markdown.py
```diff
@@ -3,13 +3,13 @@
...
 )
-from widgets.serializers.react_fields import ReactCharField
+from widgets.serializers.react_fields import ReactMarkdownWysiwygField
 
 
 class MarkdownWidgetSerializerConfigSerializer(WidgetConfigSerializer):
     """Serializer for MarkdownWidgetSerializer config"""
 
-    source = ReactCharField(help_text="Enter widget text")
+    source = ReactMarkdownWysiwygField(help_text="Enter widget text")
```
mbertrand: The help text doesn't display in the source editor.

noisecapella: I looked into adding placeholder text to the prosemirror input and it looks nontrivial. Do you mind if we skip it for now and add it as a separate issue?

PR #1666, opened 2019-01-09
Added title and channel_type to Channel to avoid many reddit requests

### LINKED ISSUE #1658: Frontpage / Channel listing performance improvement
Based on https://rpm.newrelic.com/accounts/1698630/applications/34990656/profiles/2027116

Both of these APIs are spending 55% of their request time fetching the subreddit in order to render channel name, title and type.

We already store the channel name in `Channel.name`, so if we start storing `title` and `channel_type` too, we can coalesce those 25 requests to reddit down to a singe database fetch (which we're already doing anyway).

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Fixes #1658 

#### What's this PR do?
Adds `channel_type` and `title` to `Channel` so that we can avoid N+1 trips to reddit triggered by accessing `Submission.subreddit`, since the frontpage and channel listing APIs reddit provides don't provide the channel title or type

- Ensure you can still load the frontpage and channel pages
- Verify you have channels with `None` set on these new fields:
```python
from channels.models import *
from django.db.models import Q
Channel.objects.filter(Q(title__isnull=True) | Q(channel_type__isnull=True) | Q(allowed_post_types__isnull=True)).count()
```
- Run `./manage.py backpopulate_channel_fields`
- Run the check script again and verify no channel match
- Retest the frontpage and channel pages and verify they work, you should see a noticeable performance improvement even locally


channels/serializers/posts.py
```diff
@@ -101,15 +101,15 @@ def get_created(self, instance):
 
     def get_channel_name(self, instance):
         """The channel which contains the post"""
-        return instance.subreddit.display_name
+        return instance.channel.name or instance.subreddit.display_name
```
noisecapella: `channel.name` would only be falsey if we didn't run the backpopulate yet, right?

rhysyngsun: Correct, it _might_ be possible to get an empty string in there somehow, but i'm pretty sure the UI at least doesn't allow it.

noisecapella: I looked into adding placeholder text to the prosemirror input and it looks nontrivial. Do you mind if we skip it for now and add it as a separate issue?

noisecapella: Whoops wrong issue

Issue #1667, opened 2019-01-09
split up lib/api.js into separate modules


aliceriot:
It's not like *huge* huge, but it's getting big, and especially the associated test file is getting quite long (~1000 loc). I think it would make sense to split it up into modules grouped by functionality (like posts, comments, channels, etc).

PR #1668, opened 2019-01-09
start splitting up `lib/api.js`

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

a start on #1667 

#### What's this PR do?

splits `lib/api.js` into modules grouped by functionality, which live in a new `lib/api/` directory.

No code has been added or removed, functions have just been moved into different modules (and their associated tests split out as well). So everything *should* just work as it did before, but it's definitely good to go around and exercise all of the major functions of the app and make sure everything is working.

### DIFF
--- a/static/js/lib/api/api.js
+++ b/static/js/lib/api/api.js
@@ -2,57 +2,30 @@
 /* global SETTINGS:false, fetch: false */
 // For mocking purposes we need to use "fetch" defined as a global instead of importing as a local.
 import R from "ramda"
-import "isomorphic-fetch"
 import qs from "query-string"
 import { PATCH, POST, DELETE } from "redux-hammock/constants"
 import { fetchJSONWithCSRF } from "redux-hammock/django_csrf_fetch"
 
-import { buildSearchQuery } from "./search"
+import { buildSearchQuery } from "../search"
 import {
   fetchJSONWithAuthFailure,
   fetchWithAuthFailure,
   fetchJSONWithToken
 } from "./fetch_auth"
-import { toQueryString } from "../lib/url"
-import { getPaginationSortParams } from "../lib/posts"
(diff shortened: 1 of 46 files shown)


static/js/lib/api/api.js
```diff
@@ -7,52 +7,26 @@ import qs from "query-string"
...
-const getPaginationSortQS = paramsToQueryString(getPaginationSortParams)
-
-const getCommentSortQS = paramsToQueryString(R.pickAll(["sort"]))
+  SocialAuth
+} from "../../flow/discussionTypes"
+import type { NotificationSetting } from "../../flow/settingsTypes"
+import type { WidgetListResponse } from "../../flow/widgetTypes"
+import type { SearchParams } from "../../flow/searchTypes"
 
 export function search(params: SearchParams): Promise<*> {
```
noisecapella: Was this function, `getComments` and other functions left in this file because they didn't seem to fit elsewhere?

static/js/lib/api/channels.js
```diff
@@ -0,0 +1,125 @@
+import R from "ramda"
```
noisecapella: Missing `// @flow`

static/js/lib/api/embedly.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import "isomorphic-fetch"
```
noisecapella: Can we move this into `fetch_auth` or do we need this in each file?

aliceriot: it seems we don't need it! tests pass and everything seems to work ok without. I'll remove it from all the API modules.

static/js/lib/api/api.js
noisecapella: Sounds good

PR #1675, opened 2019-01-10
Minimum search query length

### LINKED ISSUE #1507: Investigate requiring a minimum length for text when searching content
Currently the search will display all results, which may not be what we want

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #1507

#### What's this PR do?
- Sets a configurable minimum length for the search query parameter (default=2)
- Does not run the search if the query string is less than the minimum length.
- Shows an error message if the input is submitted and the query string is less than the minimum length.

- Under global and channel search, try submitting a search with 0, 1, and 2+ character search terms.
- You should see this with < 2-character searches:
screen shot 2019-01-10 at 10 44 19 am
- You should see results (or a `There are no results to display.` message depending on query) for 2+ character searches.


static/js/components/SearchTextbox.js
```diff
@@ -1,11 +1,13 @@
-// @flow
+// @flowarticle
```
noisecapella: Typo?

static/js/lib/validation.js
```diff
@@ -185,6 +185,14 @@ export const validateContentReportForm = validate([
   )
 ])
 
+export const validateSearchQuery = (text: ?string): ?string =>
```
noisecapella: Can you add a test for this function?

static/js/containers/SearchPage_test.js
```diff
@@ -107,6 +107,19 @@ describe("SearchPage", () => {
       )
     })
   })
+  ;["", "a"].forEach(query => {
+    it("doesn't run a search if initial search text is empty or < min characters", async () => {
```
noisecapella: Can we alter the test name based on what `query` is?

static/js/components/SearchTextbox.js
mbertrand: oops, yup

static/js/containers/SearchPage_test.js
mbertrand: ✔️

static/js/lib/validation.js
mbertrand: ✔️

noisecapella:
If I delete all text from the field all results disappear. I think we should preserve the results until the user clicks the search icon or presses enter

mbertrand:
@noisecapella the previous search results should continue to display now until input submission.

PR #1690, opened 2019-01-11
fix an issue with resetting the create post form

### LINKED ISSUE #1687: Post form: error when switch channel
As a user, when I create a post, I can currently set or change the channel after choosing the post type. 

However, not all channels allow all post types. This leads to an error when the user chooses a post type (say "article") and then chooses a channel that doesn't allow that post type.

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #1687

#### What's this PR do?

this fixes an issue with the form reset logic we have on the create post page. basically, there are a few cases where, when the user switches between channels, we want to reset the input values for the form, either because the current input isn't valid on the new channel or because the new channel offers the user an option which isn't present on the previous channel.

The particular issue this is fixing is a situation where we wanted to be resetting the form but we were not. The situation is:

- the user navigates to `/create_post/`
- the user clicks on 'Write an Article'
- the user fills in the article and title
- the user selects a channel which allows text and link posts, but not article posts
- the user submits the form

on `master` right now this will fail to clear the article post type and the article input from the form, and the user will get backend error because the post type is not supported by the channel they've selected. we should be clearing the form of input if the user makes a selection like this.

the problem is that the logic we have wasn't fully updated to support three post types, but implicitly assumes that the max is two.

Anyhow, I rewrote the `shouldFormReset` function to hopefully both make it more clear and straightforward and also to take account of this case.

I also discovered that many tests for the create post page were not running because of a typo (i.e. 1 out of 17 tests cases was being run). So I fixed that, and added some flow types which would have caught that.

- make sure you can reproduce the above case on master and that it results in an error
- ensure this branch fixes it
- ensure the form clearing behavior is otherwise as it should be, and that there are no regressions.

### DIFF
--- a/static/js/containers/CreatePostPage_test.js
+++ b/static/js/containers/CreatePostPage_test.js
@@ -32,6 +32,9 @@ import { newPostForm } from "../lib/posts"
 
 import type { CreatePostPayload } from "../flow/discussionTypes"
 
+type CDUTypeOne = [Array<string>, Array<string>, ?string, boolean, boolean]
+type CDUTypeTwo = [Array<string>, ?string, boolean, boolean]
+
 describe("CreatePostPage", () => {
   let helper,
     listenForActions,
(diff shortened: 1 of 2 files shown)


static/js/containers/CreatePostPage_test.js
```diff
@@ -492,7 +495,7 @@ describe("CreatePostPage", () => {
     //
     ;[
       // starting with ANY
-      ([ALL_LINK_TYPES, [LINK_TYPE_TEXT], LINK_TYPE_LINK, true, true],
```
noisecapella: Oh javascript. Is the comma operator ever useful to anyone?

PR #1693, opened 2019-01-14
Allow article cover images to be deleted

### LINKED ISSUE #1691: allow clearing the cover image on an article post
currently the API doesn't support removing an image from an article post, all you can do is replace your existing image with a new one.

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #1691

#### What's this PR do?
- Removes an article cover image if `article_cover` is not sent with a patch request.
- Prevents an exception from occurring if `article_cover` in the patch request is a URL instead of an uploaded file (which happens on master when editing an existing article but leaving the cover as is).

- Edit an article with a cover image.  Change the text but not the cover image.  Save.  You should not see an exception when the post detail page reloads, and the cover image should still exist.
- Edit the article again.  Remove the cover image.  Save.  You should not see an exception when the post detail page reloads, and the cover image should be gone.

### DIFF
--- a/channels/views/posts_test.py
+++ b/channels/views/posts_test.py
@@ -264,7 +264,7 @@ def test_patch_article_validate_cover_image(
     assert resp.status_code == status.HTTP_400_BAD_REQUEST
     assert resp.json() == {
         "error_type": "ValidationError",
-        "cover_image": [f"Expected cover image to be a file"],
+        "cover_image": [f"Expected cover image to be a file or url"],
     }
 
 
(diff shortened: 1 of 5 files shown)


aliceriot:
functionality looks great! taking a look at the code now.

channels/views/posts_test.py
```diff
@@ -783,6 +783,22 @@ def test_update_article_cover(
...
+    article = Article.objects.get(post__post_id=resp.json()["id"])
+    assert image_url.endswith(article.cover_image.url)
+
+    # Remove the image
+    resp = user_client.patch(url, {"article_content": "foo"}, format="json")
+    assert resp.status_code == status.HTTP_200_OK
+    article = Article.objects.get(post__post_id=resp.json()["id"])
+    assert article.cover_image.name == ""
+    assert article.cover_image_small.name == ""
+
```
aliceriot: do you think we should also add a test making sure you can issue a PATCH request with a URL string for the cover image?

mbertrand: That is tested on lines 786-793, but I can put it in a separate test if you prefer

aliceriot: gosh I should really read more closely! haha

Issue #1662, opened 2019-01-08
Refactor recreate_index to work off of our database


noisecapella:
For performance we should look at updating our indexing to work off of the shadow tables in our database and not contact the reddit instance at all. To do this we need to:

 - Add fields to the `Channel`, `Post` and `Comment` models corresponding to data we use to create the Elasticsearch documents, which we currently fetch from reddit
 - Change our indexing to use these fields instead of querying reddit

PR #1709, opened 2019-01-14
Update Post and Comment tables so they can be a source of truth for indexing

### LINKED ISSUE #1699: Model changes and backpopulate scripts for post and comment model fields for search
Add the fields search needs to reindex posts and comments and populate them from reddit.

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Fixes #1699 
Part of #1662 

#### What's this PR do?
Adds additional fields to shadow tables so they can be a source of truth for reindexing

Test the new code:
- Apply the migrations
- Create new posts and comments and verify they get data in the new fields
- Run `./manage.py backpopulate_post_and_comment_fields`
- Verify older posts and comments have been populated with data

Test backpopulate_missing_posts:
- Delete non-article posts:
```python
from channels.models import *
Post.objects.filter(article__isnull=True).delete()
```
- Run `./manage.py backpopulate_missing_posts`


rhysyngsun:
Just noticed that the migration here breaks the `backpopulate_missing_posts` command, so I'm going to introduce a quick refactor to bring that and the code in channels/api.py under common code to create those records.

gsidebo:
Functionality looks great so far

channels/test_utils.py
```diff
@@ -19,3 +19,18 @@ def no_ssl_verification():
         warnings.resetwarnings()
 
         requests.Session.request = old_request
+
+
+def assert_values_eq(obj, expected):
```
gsidebo: Nit: `assert_properties_eq` might be a more accurate name

channels/backpopulate_api.py
```diff
@@ -0,0 +1,108 @@
...
+        post (channels.models.Post): the post to backpopulate
+        submission (Submission): the reddit submission to source data from
+    """
+
+    if not submission.is_self:
+        post.post_type = LINK_TYPE_LINK
+        post.url = submission.url
+        if post.link_meta is None and settings.EMBEDLY_KEY:
+            post.link_meta = utils.get_or_create_link_meta(submission.url)
+    elif hasattr(post, "article") and post.article:
```
gsidebo: `elif getattr(post, "article", None):`?

channels/api.py
```diff
@@ -968,19 +1057,32 @@ def update_post(
...
-                    f"article_image_{post.id}.jpg", cover_image, save=False
-                )
-        elif post.article.cover_image:
-            post.article.cover_image = None
-            post.article.cover_image_small = None
-        post.article.save(update_image=(cover_image is not None))
+        with transaction.atomic():
+            if text:
+                Post.objects.filter(post_id=post_id).update(text=text, edited=True)
+                return proxy_post(post.edit(text))
```
gsidebo: Related to my earlier comment: seems like we will also want a helper on `PostProxy` to handle text updates.

This would end up looking something like this:
```python
if text:
  post_proxy = proxy_post(post)
  return post_proxy.edit(text)
```

channels/api.py
```diff
@@ -683,7 +687,76 @@ def update_channel(self, name, title=None, channel_type=None, **other_settings):
...
+                # end handle upvote
+                # --------
+                # start handle downvote
+                # adds the downvote
+                -1 if downvote and not is_downvoted else 0,
+                # removes an opposing upvote, if present
+                -1 if downvote and is_upvoted else 0,
+                # clear an existing downvote, but only count this once if upvoting also
+                1 if downvote is False and is_downvoted and not upvote else 0,
+                # end handle downvote
```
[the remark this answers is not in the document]
rhysyngsun: Reason for this is because `vote_action` erases a critical piece of information we need here: "does an upvote need to increment by 1 or is there a current downvote and we need to increment by 2?"

channels/api.py
rhysyngsun: Same thing as me feedback above.

rhysyngsun:
@gsidebo feedback addressed/responded to

rhysyngsun:
@gsidebo I've fixed the missing writes

PR #1721, opened 2019-01-15
Create and update course documents in Elasticsearch

### LINKED ISSUE #1717: Create and populate a search index with mappings for courses

### LINKED ISSUE #1718: Add triggers/methods to update the ES index when existing courses are updated

mbertrand:
Depends on and includes code from #1753

#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #1717
Closes #1718

#### What's this PR do?
- Adds ES serializer and mapping for courses
- Creates a course index

First, import some courses (ask me or devops for values of `EDX_API_URL`, `EDX_API_CLIENT_ID`, `EDX_API_CLIENT_SECRET`):
```python
from course_catalog.tasks import get_edx_data
get_edx_data()
```

Then run `docker-compose run web python manage.py recreate_index`

Post to `http://localhost:9101/discussions_local_all_default/_search` with `{"query": {"bool": {"must": [{"match": {"object_type": "course"}}]}}}`, you should get some results.


search/serializers.py
```diff
@@ -86,6 +87,48 @@ def postprocess_fields(self, discussions_obj, serialized_data):
         return {"author_channel_membership": sorted(get_channels(discussions_obj.user))}
 
 
+class ESCourseSerializer(ESSerializer):
```
rhysyngsun: Could we possibly have this be a direct model serializer instead like `PostSerializer` and `CommentSerializer` are now?

search/serializers.py
```diff
@@ -86,6 +87,48 @@ def postprocess_fields(self, discussions_obj, serialized_data):
...
+        "language",
+        "semester",
+        "year",
+        "level",
+        "start_date",
+        "end_date",
+        "enrollment_start",
+        "enrollment_end",
+    ]
+    rename_keys = {"title": "course_title", "image_src": "course_image"}
```
rhysyngsun: Is there a rationale behind renaming these particular fields? It seems like `short_description` could arguably be prefixed by "course_" too, so I'm trying to understand if there's a functional reason to this.

search/serializers.py
```diff
@@ -86,6 +87,48 @@ def postprocess_fields(self, discussions_obj, serialized_data):
...
+
+        return CourseSerializer
+
+    def postprocess_fields(self, discussions_obj, serialized_data):
+        return {
+            "instructors": [
+                "{} {}".format(i.first_name, i.last_name)
+                for i in discussions_obj.instructors.iterator()
+            ],
+            "topics": [t.name for t in discussions_obj.topics.iterator()],
```
rhysyngsun: This could just be a `values_list()` and is the expected result set big enough an iterator needed for this and the other two fields here?

search/serializers.py
```diff
@@ -86,6 +87,48 @@ def postprocess_fields(self, discussions_obj, serialized_data):
...
+            "prices": [
```
rhysyngsun: `"prices": discussions_obj.prices.values("price", "mode")` gives you the same result in less code.

search/serializers.py
```diff
@@ -269,3 +312,27 @@ def serialize_comment_for_bulk(comment_obj):
...
+
+
+def serialize_bulk_courses(ids):
+    """
+    Serialize courses for bulk indexing
+
+    Args:
+        ids(list of int): List of course id's
+    """
+    for course in Course.objects.filter(id__in=ids).iterator():
```
rhysyngsun: An iterator is probably unnecessary overhead here: the batch size of `ids` will always be smaller than the batch size of the iterator (default is 2000) so it ends up incurring the overhead of a DB cursor and never ends up using it.

search/task_helpers.py
```diff
@@ -349,3 +352,44 @@ def update_indexed_score(instance, instance_type, vote_action=None):
         incr_amount=vote_increment,
         object_type=instance_type,
     )
+
+
+@if_feature_enabled(INDEX_UPDATES)
+def index_new_course(course_obj):
```
rhysyngsun: This doesn't seem to be used anywhere aside from tests

search/task_helpers.py
```diff
@@ -349,3 +352,44 @@ def update_indexed_score(instance, instance_type, vote_action=None):
...
+
+    Args:
+        course_obj (course_catalog.models.Course): A Course object
+    """
+    data = ESCourseSerializer().serialize(course_obj)
+    create_document.delay(gen_course_id(course_obj.course_id), data)
+
+
+@if_feature_enabled(INDEX_UPDATES)
+def update_course(course_obj):
```
rhysyngsun: Same here, this is unused anywhere outside of tests.

rhysyngsun:
Needs a rebase too.

search/serializers.py
mbertrand: These two just seemed a bit too short & generic to me, I thought it might be good to more specifically differentiate them from similar fields like `post_title` for example.  I can keep the original field names if you prefer.

search/serializers.py
mbertrand: Switched to `ESProxySerializer` like `ESProfileSerializer`

search/task_helpers.py
mbertrand: Usage must have been inadvertently removed during a previous rebase, added back (ditto for update_course)

mbertrand:
Comments addressed, sorry for the mess, I seem to have wiped out some changes while resolving earlier rebase conflicts.

course_catalog/task_helpers.py
```diff
@@ -260,6 +263,7 @@ def digest_ocw_course(
     if "PROD/RES" in course_prefix:
         course_fields["learning_resource_type"] = ResourceType.ocw_resource.value
 
+    index_func = update_course if course_instance else index_new_course
```
rhysyngsun:
```suggestion
    index_func = update_course if course_instance is not None else index_new_course
```

mbertrand:
@rhysyngsun made one more change, added `course.id` to the course ES serializer and index

search/serializers.py
```diff
@@ -339,3 +395,29 @@ def serialize_comment_for_bulk(comment_obj):
...
+
+def serialize_bulk_courses(ids):
+    """
+    Serialize courses for bulk indexing
+
+    Args:
+        ids(list of int): List of course id's
+    """
+    for course in Course.objects.filter(id__in=ids).prefetch_related(
+        "topics", "instructors", "prices"
```
rhysyngsun: I'm not 100% certain about this, but I think the only prefetch that provides benefit here is `instructors` because you walk `instructors.all()` in the serializer. `get_topics` and `get_prices` actually generate _new_ queries, so including the other fields in this prefetch counter-intuitively adds overhead.

There are probably two options here:

- Don't worry about trying to prefetch those other fields and we'll worry about it if/when it becomes a bad pain point
- Use [`Prefetch`](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#prefetch-objects) to set attrs as for `topics` and `prices`. An example would be:

```python
from django.db.models import Prefetch

topic_names = CourseTopic.objects.values_list("name", flat=True)
prefetch_topic_names = Prefetch("topics", queryset=topic_names, to_attr="topic_names")
Course.objects.filter(id__in=ids).prefetch_related(prefetch_topic_names)
```

And then it can just be a simple `ListField` on the serializer.

The major downside to this approach is it very tightly couples the serializer implementation to the queryset it is passed.

I'd probably go for the former option, but I just wanted to play out the alternative.

search/api.py
```diff
@@ -58,6 +60,20 @@ def gen_profile_id(profile_id):
...
+    """
+    Generates the Elasticsearch document id for a course
+
+    Args:
+        course_id (str): The course_id of a Course object
+
+    Returns:
+        str: The Elasticsearch document id for this object
+    """
+    safe_id = re.sub(r"\W+", "", course_id)
```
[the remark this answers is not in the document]
mbertrand: Number of MITx courses imported matches number of MITX courses in ES (632) but I can replace with this just in case.

search/serializers.py
mbertrand: Went with the former

mbertrand:
@rhysyngsun newest comments addressed

PR #1723, opened 2019-01-16
use the `post_type` field for post-type specific UI

### LINKED ISSUE #1652: Refactor front-end to recognize article post type
Currently, the front-end users the presence or absence of post text to determine whether a post is a text post or an article post (see https://github.com/mitodl/open-discussions/issues/1550#issuecomment-450967579). This approach prevents us from adding other posts types in the future. 

Instead, we should use the post type as returned by the posts API (PR #1633) to determine the post type in the front end.

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #1652 

#### What's this PR do?

Basically this just uses the relatively-new `post_type` value to decide what UI to show or which logic to use in a few places where we need to do different things based on the post type. I went through and changed everywhere that I could think of, but if I missed somewhere where it would make sense to do this refactor as well plz point it out.

Everything should be working as it was before.

Additionally, 'title only' posts (essentially empty text posts) should now be editable, whereas before they were not (because we used the value of `post.text` to figure out if a post should be editable or not).

### DIFF
--- a/static/js/components/CompactPostDisplay.js
+++ b/static/js/components/CompactPostDisplay.js
@@ -25,6 +25,7 @@ import {
   POST_PREVIEW_LINES
 } from "../lib/posts"
 import { userIsAnonymous } from "../lib/util"
+import { LINK_TYPE_LINK } from "../lib/channels"
 
 import type { Dispatch } from "redux"
 import type { Post } from "../flow/discussionTypes"
@@ -134,11 +135,11 @@ export class CompactPostDisplay extends React.Component<Props> {
               </div>
             </div>
           </div>
-          {post.url || post.thumbnail ? (
+          {post.post_type === LINK_TYPE_LINK || post.thumbnail ? (
             <div
               className={`column2 ${post.thumbnail ? "link-thumbnail" : ""}`}
             >
-              {post.url ? (
+              {post.post_type === LINK_TYPE_LINK ? (
                 <React.Fragment>
                   <div className="top-right">
                     <a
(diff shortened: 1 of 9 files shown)


static/js/components/CompactPostDisplay.js
```diff
@@ -134,11 +135,11 @@ export class CompactPostDisplay extends React.Component<Props> {
               </div>
             </div>
           </div>
-          {post.url || post.thumbnail ? (
+          {post.post_type === LINK_TYPE_LINK || post.thumbnail ? (
```
gsidebo: Take or leave since it isn't directly related to this PR: it really seems like this 'or' clause is unnecessary. The only thing necessarily in common with link-type posts and posts that have a `thumbnail` value is the wrapper div with the `column2` class, and even then they differ in that a thumbnail-ed post will have an additional class set. Seems like it should just be 2 separate if's.

static/js/components/ExpandedPostDisplay.js
```diff
@@ -119,7 +134,8 @@ export default class ExpandedPostDisplay extends React.Component<Props> {
         </div>
         <div className="right">
           {SETTINGS.username === post.author_id &&
-          (post.text || post.article_content) ? (
+          (post.post_type === LINK_TYPE_TEXT ||
+            post.post_type === LINK_TYPE_ARTICLE) ? (
```
gsidebo: Also take or leave: I'm in favor of the other cases where we explicitly check the `post_type` value, but here I think it would be a little more readable and reusable to create a helper function. `isEditablePostType` or something like that

aliceriot: sure, then we don't have to edit every place we might want to have some sort of editing interface later.

PR #1786, opened 2019-01-30
Implement embedly widget

### LINKED ISSUE #1510: Widgets: embed widget
See the design screencast at https://drive.google.com/open?id=1-IlF8Ew63nvUdhqmzGWrK_WGYU7_Zq_4

The embed widget built into the framework does simple iframe embedding, which doesn't work with many sites (e.g. vimeo.com). We should update the widget to use the embed.ly API which can handle vimeo and other sites. 

### acceptance criteria
- [ ] update the form components, labels and default text
- [ ] preview the results through the embedly API
  when the API doesn't return an iframe (e.g. from Wikipedia), we can display a linked image 
- [ ] copy:
  >Title: “Embed widget”
  >Box 1: “Title”
  >Box 2: “URL” (needs to be capitalized)

noisecapella:
#### Pre-Flight checklist

  - [x] Desktop screenshots 
- [x] Testing
  - [x] Code is tested

Fixes #1510 

#### What's this PR do?
Uses the embedly card API to render embedly widgets. The actual rendering is done by the embedly platform; this PR just injects HTML into the DOM with the appropriate CSS class.

Edit widgets, make sure they look fine. Try out many different embedly sources. Comment if you see any source which produces a strange looking UI.

Widget dialog
screenshot from 2019-01-30 12-26-32

Widget rendering in sidebar
screenshot from 2019-01-30 12-26-16

### DIFF
--- a/static/js/components/EmbedlyCard.js
+++ b/static/js/components/EmbedlyCard.js
@@ -3,8 +3,7 @@
 import React from "react"
 import isURL from "validator/lib/isURL"
 
-import { loadEmbedlyPlatform } from "../lib/embed"
-import ReactDOMServer from "react-dom/server"
+import { loadEmbedlyPlatform, renderEmbedlyCard } from "../lib/embed"
 
 type Props = {
   url: string
@@ -21,16 +20,6 @@ export default class EmbedlyCard extends React.Component<Props> {
       return null
     }
 
-    const html = ReactDOMServer.renderToStaticMarkup(
-      <a
-        data-card-chrome="0"
-        data-card-controls="0"
-        data-card-key={SETTINGS.embedlyKey}
-        href={url}
-        className="embedly-card"
-      />
-    )
-
-    return <div dangerouslySetInnerHTML={{ __html: html }} />
+    return <div dangerouslySetInnerHTML={{ __html: renderEmbedlyCard(url) }} />
   }
 }
--- a/static/js/components/EmbedlyCard.js
+++ b/static/js/components/EmbedlyCard.js
@@ -1,0 +1,36 @@
+/* global SETTINGS: false */
+// @flow
+import React from "react"
+import isURL from "validator/lib/isURL"
+
+import { loadEmbedlyPlatform } from "../lib/embed"
+import ReactDOMServer from "react-dom/server"
+
+type Props = {
+  url: string
+}
+export default class EmbedlyCard extends React.Component<Props> {
+  componentDidMount() {
+    loadEmbedlyPlatform()
+  }
+
+  render() {
+    const { url } = this.props
+
+    if (!isURL(url, { allow_underscores: true })) {
+      return null
+    }
+
+    const html = ReactDOMServer.renderToStaticMarkup(
+      <a
+        data-card-chrome="0"
+        data-card-controls="0"
+        data-card-key={SETTINGS.embedlyKey}
+        href={url}
+        className="embedly-card"
+      />
+    )
+
+    return <div dangerouslySetInnerHTML={{ __html: html }} />
+  }
+}
--- a/widgets/serializers/react_fields.py
+++ b/widgets/serializers/react_fields.py
@@ -18,7 +18,7 @@ class ReactField:
 
     def _get_input_type(self):
         """Returns the field's input type"""
-        return "text"
+        raise NotImplementedError()
 
     def get_field_spec(self):
         """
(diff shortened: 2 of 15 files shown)


Ferdi:
@noisecapella looks great.

static/js/components/EmbedlyCard.js
```diff
@@ -0,0 +1,36 @@
...
+
+    const html = ReactDOMServer.renderToStaticMarkup(
+      <a
+        data-card-chrome="0"
+        data-card-controls="0"
+        data-card-key={SETTINGS.embedlyKey}
+        href={url}
+        className="embedly-card"
+      />
+    )
```
aliceriot: We do essentially the same thing in the ckeditor for rendering inline embeds there (rendering a div with these `data-` attributes and so on to static HTML).

maybe it would be good to pull this out into a shared function, in `lib/embed.js`?

static/js/components/EmbedlyCard.js
```diff
@@ -0,0 +1,36 @@
...
+import React from "react"
+import isURL from "validator/lib/isURL"
+
+import { loadEmbedlyPlatform } from "../lib/embed"
+import ReactDOMServer from "react-dom/server"
+
+type Props = {
+  url: string
+}
+export default class EmbedlyCard extends React.Component<Props> {
```
aliceriot: if you wanted, this could just be a stateless component. `loadEmbedlyPlatform` is idempotent (if the thing is already there it just won't do anything) so it would be safe to just call it on every render of the component.

static/js/components/EmbedlyCard.js
aliceriot: something with a signature like `(url: string) => html: string`

widgets/serializers/react_fields.py
```diff
@@ -74,12 +74,25 @@ def _get_props(self):
...
+        """Returns the field's input type"""
+        return "url"
+
     def _get_props(self):
         """Returns the field properties"""
         return {
             **super()._get_props(),
             "max_length": self.max_length or "",
             "min_length": self.min_length or "",
+            "show_embed": self.show_embed,
```
aliceriot: what is this value used for?

static/js/components/EmbedlyCard.js
noisecapella: How about `(url: string) => React.Component` for the div? Or do you think it would be better to separate the two?

static/js/components/EmbedlyCard.js
noisecapella: My main concern is that the React docs recommend putting this kind of DOM manipulation in `componentDidMount` and leaving it out of `render` for clarity's sake.

widgets/serializers/react_fields.py
noisecapella: The RSS widget also accepts a URL input but I want to communicate to the frontend not to render it with embedly haha

static/js/components/EmbedlyCard.js
noisecapella: Looking at the other example, it doesn't use render with React so it does need to be `=> html: string`

widgets/serializers/react_fields.py
aliceriot: ok cool, was just wondering because I didn't see it used anywhere

static/js/components/EmbedlyCard.js
aliceriot: well if we had a function that returns a react component, in both of these places we're going to go ahead and immediately pass that react component to `ReactDOM.renderToStaticMarkup`, so to me it would make sense to write a function like this:

```js
const getEmbedlyCardMarkup = (url: string): string =>
  ReactDOM.renderToStaticMarkup(
    <a
      data-card-chrome="0"
      data-card-controls="0"
      data-card-key={SETTINGS.embedlyKey}
      href={url}
      className="embedly-card"
    />
  )
```

then we can call that function here and also in the `ArticleEditor`.

widgets/serializers/react_fields.py
noisecapella: :thinking: If it's not in the frontend code anywhere then it's probably something I missed actually...

noisecapella: Yeah I missed it. I just pushed a commit to check the flag.

noisecapella:
Comments addressed

Issue #1778, opened 2019-01-28
Channel about page


pdpinch:
As a channel moderator, I'd like to have an about page where I can explain the channel at length, using the same features as an article. 

As a user, I'd like to be able to click an about tab on the channel listing page and find out more about the channel. 

### acceptance criteria:
- [x] add a field to the channel model for storing ckeditor data (similar to article) 
- [x] display an "about" tab on the channel listing page, next to the "Home" tab. 
- [x] when clicked, display the rendered contents of the "about" field
- [x] display an edit button on the about page to channel moderators, similar to the edit button on the post detail page for articles
- [x] the editor interface should function similarly to the article editor interface

aliceriot:
one question here - should the about page have a cover image, similar to the cover image for article type posts?

pdpinch:
It doesn't need a cover image, because it doesn't need a thumbnail. 

I presume that, like an article, users will be able to embed images in the page. Just not at the top.

aliceriot:
ok cool. and yeah, we'll get image embedding out of the box just from using the ckeditor.

pdpinch:
There was a lingering question about how the user creates the about page initially. We've decided to avoid this by just displaying the about tab at all times. We know that means there will be channels with no content. Laura or I will take care of populating it after the feature goes live.

aliceriot:
ok, what I was planning to do was to always show the about tab to users who have permission to edit it, and conditionally show / hide it based on the presence of the about field for user who do not. I can just go ahead and show it all the time though, it's slightly easier that way.

pdpinch:
Oh, that would be nice: conditionally show / hide it based on the presence of the about field for user who do not have permission to edit. If it's not too much work, that would be better.

aliceriot:
sure, it's not that complicated and we have helper functions already for a lot of those permissions checks

aliceriot:
@pdpinch how does this look for a super-basic editing UI? I can try to whip up something more fancy, but since this is just going to be shown to moderators I was thinking we might be able to get away with something simple:

![basicedit](https://user-images.githubusercontent.com/6207644/52584569-5c9a3b00-2e00-11e9-9699-c2775764f778.png)
![basicsave](https://user-images.githubusercontent.com/6207644/52584588-6885fd00-2e00-11e9-8f92-d1b0cadbdad2.png)

pdpinch:
I thought we were just going to reuse the article editor UI?

aliceriot:
[...]

PR #1788, opened 2019-01-30
add about field to channel model

aliceriot:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

part of #1778 

#### What's this PR do?

this adds a `JSONField` called `about` to the `Channel` object, and then also adds it to the serializer for the `Channel`.

Make sure that you can set the value of the `about` field if you use the DRF view for the API (`/api/v0/channels/${channelName}`). It should only let you set the value to be a list (what will be using for the ckeditor document) and it should give you an error if you try to set it to be something like `{foo: "bar"}` (but `[{foo: "bar" }]` should be good).


channels/serializers/channels.py
```diff
@@ -126,6 +127,21 @@ def validate_allowed_post_types(self, value):
 
         return {"allowed_post_types": value}
 
+    def get_about(self, channel):
+        """get the value for the about field"""
+        channel_obj = channel._self_channel  # pylint: disable=protected-access
```
rhysyngsun: We want to have this kind of in `channels.proxies.ChannelProxy` instead of the serializer so that when we change the object type being serialized from that to an actual `Channel` object the serializer doesn't need to be updated.

So I'd suggestion putting an `@property` in `ChannelProxy` for `about` and just making this a `JSONField`. That'll give you full JSON validation too so you'll be able to drop both your `get_about` and `validate_about` implementations in one fell swoop too.

channels/serializers/channels_test.py
```diff
@@ -72,6 +79,7 @@ def test_serialize_channel(
         if allowed_post_types
         else get_allowed_post_types_from_link_type(link_type),
         "widget_list_id": channel.widget_list_id,
+        "about": channel._self_channel.about,  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
aliceriot: sure

channels/serializers/channels.py
aliceriot: I added `"about"` to the list of fields to look up on `self._self_channel` in the `__getattr__` method on `ChannelProxy`, does that accomplish the same thing?

aliceriot: It seems to work ok in the API when I test it

channels/serializers/channels.py
```diff
@@ -194,4 +195,9 @@ def update(self, instance, validated_data):
             channel_obj.banner.save(f"channel_banner_{name}.jpg", banner, save=False)
             channel_obj.save(update_fields=["banner"], update_image=True)
 
+        about = validated_data.get("about")
+        if about:
```
rhysyngsun: Is there ever going to be a need to null out the about value? Or would you just end up setting it to `[]`?

rhysyngsun:
Functionality generally looks good just a few small things above

channels/serializers/channels.py
aliceriot: hmm I suppose we might want users to be able to null it out, in case a channel wanted to no longer have an about page.

aliceriot: ok, I rewrote this logic to check if there is a value for `about` in the validated data and, if so, save that as the new value for about. I figure if we're going to delete the value of the field we shouldn't just look for its absence from the request, but require the user to make a request with `about: null` in it.

rhysyngsun:
Code looks good, I'll give it a quick test when I'm in.

PR #1789, opened 2019-01-30
Fix backpopulate not adding comments

### LINKED ISSUE #1787: backpopulate_api.populate_comments does not create the comment if it is missing
The query going on here is a SQL `UPDATE` so it is not creating the record if it does not exist. I think the original reasoning behind this was to not stomp on any writes `channels.api.Api` was doing on comment creation, but it actually caused the new backpopulate command to not work as expected.

This should probably be an `update_or_create()` instead.

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1787 

#### What's this PR do?
Switches how we create comments to address the link issue plus any potential write conflicts due to race conditions.

Delete a `Comment` record, run `./manage.py backpopulate_posts` and it should be recreated.


channels/api.py
```diff
@@ -506,6 +506,59 @@ def create_channel_groups_and_roles(channel):
...
+        text=comment.body,
+        score=comment.score,
+        edited=comment.edited if comment.edited is False else True,
+        removed=comment.banned_by is not None,
+        deleted=comment.body == DELETED_COMMENT_OR_POST_TEXT,
+        created_on=datetime.fromtimestamp(comment.created, tz=timezone.utc).isoformat(),
+    )
+
+
+def create_comment(post, comment, author):
```
noisecapella: Can we change to `create_comment(*, post, comment, author)` to require keyword args for clarity?

channels/backpopulate_api.py
```diff
@@ -59,32 +58,16 @@ def backpopulate_post(post, submission):
...
-        score=comment.score,
-        edited=comment.edited if comment.edited is False else True,
-        removed=comment.banned_by is not None,
-        deleted=comment.body == DELETED_COMMENT_OR_POST_TEXT,
-        created_on=datetime.fromtimestamp(comment.created, tz=timezone.utc).isoformat(),
-    )
-
-
-def backpopulate_comments(submission):
+def backpopulate_comments(post, submission):
```
noisecapella: Can we use  `*` here to enforce keyword args?

channels/api.py
rhysyngsun: I'm not sure I understand why they need to be keyword args?

rhysyngsun:

PR #1797, opened 2019-02-01
Show related posts on the post detail page

### LINKED ISSUE #1748: related content card on post detail page
As a user, I'd like to know about content on MIT Open that's similar to what I'm currently viewing on a post detail page. 

See https://zpl.io/2ZzODOd

### acceptance criteria
- [ ] behind a feature flag; when flag is false, only visible to django staff
- [ ] use elastic search to find similar text, link and article posts (not users, channels or courses)
- [ ] configurable set of results, defaulting to 4
- [ ] Display post title, author and channel for all posts
  - [ ] link title to post detail
  - [ ] link author to author profile
  - [ ] link channel to channel 
- [ ] display link icon and off-site link for link posts

### nice to have
- [ ] boot posts with higher votes and comments

gsidebo:
#### Pre-Flight checklist

  - [x] Desktop screenshots 
- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #1748

#### What's this PR do?
Shows related posts on the post detail page

- Set `FEATURE_RELATED_POSTS_UI=True`
- Visit a post detail page
- Confirm that 4 related posts are shown in a sidebar with correct links, and that there seems to be a relationship between the similarity of those posts & how high they appear on the list

Additional testing:
- Make sure everything works as expected when that feature flag is set to false
- Make sure you can't see private channel posts in the related sidebar if your logged-in user is not a member, and make sure that you can see them in the opposite case

[More Like This query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-mlt-query.html)

For this feature we're querying ES for related content based on a specific document (i.e.: the one you're looking at on a post detail page) rather than by a string or set of strings. The fields that we're matching on: `"plain_text", "post_title", "author_id", "channel_name"`. If a post has similar content in one or more of those fields, its score will be boosted in the "more like this" query results. It's obvious that posts with similar titles and similar text content are boosted, but we're also choosing to boost content with the same author and posted in the same channel in order to "fill out" the results in case there aren't any matches based on title/content.

A note about link posts: URL values cannot be partially matched in ES by default – they are stored as a single "word". This means that a link post with the URL "mit.edu/academic_calendar" and another with the URL "mit.edu/academics" will not match (in other words, they will not be considered related). The `post_link_url` field was left off of the "more like this" fields list for that reason. We are considering the possibility of indexing summary content that we get from embedly, which would give us something more meaningful to compare between link posts. Until then, they will only match based on their titles, authors, and channels.

ss 2019-02-01 at 13 50 50
ss 2019-02-01 at 13 51 34


gsidebo:
@pdpinch Screenshot review ^^^^

pdpinch:
LGTM, thanks

Ferdi:
@gsidebo Related posts are only those from the same channel as the given post , right ?
In that case "in channel 18 .. " is not necessary    @pdpinch

gsidebo:
@Ferdi Here are two screenshots - the first from a moderator/staff, the second from a normal user with just 1 channel subscription. Same article, different related posts due to ES filters we apply. Also note that both of them have posts from different channels since the titles are similar or identical

ss 2019-02-01 at 15 56 48
ss 2019-02-01 at 15 56 55

Ferdi:
Great, how are "link" post type matched ?

static/js/containers/PostDetailSidebar_test.js
```diff
@@ -0,0 +1,100 @@
...
+import { makePostResult } from "../factories/search"
+
+describe("PostDetailSidebar", () => {
+  let helper, render, post, relatedPostData
+
+  beforeEach(() => {
+    post = makePost()
+    relatedPostData = R.times(makePostResult, 4)
+    helper = new IntegrationTestHelper()
+    helper.getRelatedPostsStub.returns(Promise.resolve(relatedPostData))
```
gsidebo: @aliceriot This is the test file that I mentioned earlier. The `PostDetailSidebar` component executes a `post` on the `relatedPosts` redux hammock endpoint. In the app, the `postSuccessHandler` is invoked, but for whatever reason I cannot get it to run in these tests. As a result I have to stub this API method to return the value that `postSuccessHandler` is actually returning. Obviously I'd prefer not to do that, so if you have any ideas about what might be going wrong, let me know

aliceriot:
one thing I'm seeing is that, even with the feature flag off, there is an extra spinner on the post page:

loadingtingngn

gsidebo:
@aliceriot I can't reproduce that. I set `FEATURE_RELATED_POSTS_UI=False` in `.env`

aliceriot:
I got that when I didn't have a value set in the `.env` file for the feature flag yet

gsidebo:
@aliceriot Yeah we default all feature flags to `True` for the dev environment. If you didn't get any similar posts in that case, that's worth discussing since this query is designed to almost always return something. Let's chat IRL if that's the case

aliceriot:
ah, I forgot that was set in the `docker-compose`. carry on!

I think the issue that I was seeing was because the flag was enabled (because of `OPEN_DISCUSSIONS_FEATURES_DEFAULT`) but the request to the related posts API was returning nothing because my elasticsearch was messed up.

gsidebo:
@aliceriot found a bug where text posts with just a title and no text are not returning any related posts. This is something of an edge case but it's weird enough that I really think it's worth digging into

pdpinch:
Can we address that as a follow-up issue and get this merged?

I wonder if those type of posts are findable in the post/comment search?

Ferdi:
@gsidebo 
> text posts with just a title and no text are not returning any related posts

Why is this happening ?  I think this is an important use case especially in private channels where lost of content is questions that might not have additional text

gsidebo:
@Ferdi @pdpinch @aliceriot: This is a known bug and appears to be fixed in ES 6.4 (our ES image is for version 6.2.4) - https://github.com/elastic/elasticsearch/issues/30148. I'll create a separate issue. I've confirmed that those title-only posts can be found via normal search. This appears limited to More Like This queries

static/js/containers/PostDetailSidebar_test.js
aliceriot: yeah, I'm kind of stumped on this one tbh, I've looked through and it seems like there's no reason why it shouldn't run. My only thought is that is has to do with using the HOC renderer, it doesn't use a full redux store (it's a mock store instead) and I'm not sure if it fully supports the redux-thunk middleware. But that's just sort of a guess, I'm really not sure why it's not working.

static/js/factories/channels.js
```diff
@@ -25,24 +25,25 @@ export const makeChannel = (privateChannel: boolean = false): Channel => {
...
+    channel_type:           privateChannel ? "private" : "public",
+    description:            casual.description,
+    public_description:     casual.description,
+    num_users:              casual.integer(0, 500),
+    user_is_contributor:    casual.coin_flip,
+    user_is_subscriber:     casual.coin_flip,
+    user_is_moderator:      casual.coin_flip,
+    membership_is_managed:  casual.coin_flip,
+    allow_widgets_ui:       casual.coin_flip,
+    allow_related_posts_ui: casual.coin_flip,
```
aliceriot: is this supposed to be added to the settings object? the `allow_widgets_ui` on the line above looks out of place as well.

gsidebo: It is in the settings object [here](https://github.com/mitodl/open-discussions/pull/1797/files#diff-2792129aef83d6b2961cc36842c2df4eR24). Am I missing a different settings object where this needs to be? Also I'm not sure what you mean by `allow_widgets_ui` being out of place

aliceriot: I was just wondering why it's been added to the channel factory here

aliceriot: given that it isn't included in the API responses from the channel API

gsidebo: Just saw your latest comment. 100% correct, `allow_widgets_ui ` shouldn't have been there in the first place. I removed it and `allow_related_posts_ui`

aliceriot: haha, probably could have been more clear with my initial comment on that one :D

PR #1803, opened 2019-02-04
People widget

### LINKED ISSUE #1655: People widget
The People widget combines elements of the members widget (#1054) and the moderators widget (#1053). 

There is a UX video at https://drive.google.com/file/d/1-PBt3UO4wpVuRlL5gFZMS9kQgJGjUOHi/view?usp=sharing

### MVP acceptance criteria:
- [ ] widget name is "People" not "Members" as used in the video & Zeplin
- [ ] only implement the "Select Members" option
- [ ] search for members by email address (name would be nice)
- [ ] trash can icon removes chosen member
- [ ] back button so user can change the configuration before adding it to the list

### nice to have
- [ ] reordering of selected members (use same icon as widget reorder icon)

### out of scope
- preview the widget before adding it to the list
- styling changes to "Add a widget" form
- "All moderators" option
- "All members" option
- link to "All members"

noisecapella:
#### Pre-Flight checklist

  - [x] Desktop screenshots 
- [x] Testing
  - [x] Code is tested

Most of #1655 

#### What's this PR do?
Adds a new widget for listing people. The widget allows an admin to use a search textbox to list the profiles matching the name and to populate the widget.

Start managing widgets. Add a people widget. Type in a name in the box and you should see at least one suggestion for the profile with that name. Click 'Add' and you should see the suggestion list disappear and the item added below the text field. Click 'create widget' and then 'Done' and you should see the widget with a list of people in it.

You should also be able to edit the widget to add or delete people from it.

people_widget

Dialog:
people_widget_edit

people_widget_selector


noisecapella:
@pdpinch This PR does not cover searching via email address, just by name, headline and bio. I can do that next but this PR is already pretty big.

Ferdi:
Looks great !

static/js/components/widgets/PeopleSelector.js
```diff
@@ -0,0 +1,186 @@
...
+
+  componentDidMount() {
+    const { clearSearch } = this.props
+    clearSearch()
+  }
+
+  onSuggestionsFetchRequested = ({
+    value,
+    reason
+  }: {
```
rhysyngsun: This is a bit hard to read here, I had to read it a few times before I realized it was the type definition of the argument and not the return type. Might be worth extracting this to a top-level `type` definition for an easier read.

static/js/components/widgets/PeopleSelector.js
```diff
@@ -0,0 +1,186 @@
...
+    fetchSuggestions(value)
+  }
+
+  renderInputComponent = (inputProps: Object) => (
+    <SearchTextbox {...inputProps} />
+  )
+
+  onChangeText = (
+    event: Event,
+    { newValue, method }: { newValue: string, method: string }
```
rhysyngsun: Same feedback here as the one above, might be worth throwing them in their own file under `js/flow`. Looks like there's a few other cases of this but I won't comment on every one.

static/js/components/widgets/PeopleSelector.js
```diff
@@ -0,0 +1,186 @@
...
+          addProfile={null}
+          deleteProfile={this.deleteProfile}
+          onSortEnd={this.onSortEnd}
+          useDragHandle={true}
+        />
+      </React.Fragment>
+    )
+  }
+}
+const mapStateToProps = (state, ownProps) => {
```
rhysyngsun: Nit: some white space between this and the class def above would be nice.

rhysyngsun:
A few things about functionality/usability I noticed:

- The number of people that can be added doesn't appear to have a limit, so it's possible to add enough that you can't close the dialog. This might be an edge case because they can simply remove people and it takes ~20 people to trigger this.
- The alignment of the names isn't consistent if you have a mix of people with/without headlines (e.g. the name is top-aligned with a headline and middle-aligned if not):
screenshot_2019-02-12 evil dinosaurs 2 mit open learning
- Are the names in the widget supposed to link anywhere? The original spec or video mockup doesn't indicate it should, but I'm also not sure how helpful this widget would be without it.

pdpinch:
1. I'm not worried about too many names to fit this until/unless it becomes an issue for a user who can't figure out how to shorten their list. 

2. I think I'm missing something about the alignment. It seems like the one or two lines of text are vertically centered, which seems an OK compromise for handling the two different states. 

3. Yes! Can we link the names to profile pages (same as search results)?

static/js/components/widgets/PeopleSelector.js
noisecapella: Will do

static/js/components/widgets/PeopleSelector.js
noisecapella: What do you mean by other cases?

noisecapella:
I added the link and made the changes to types mentioned

static/js/components/widgets/PeopleItem.js
```diff
@@ -25,7 +27,9 @@ export const PeopleItem: StatelessFunctionalComponent<Props> = ({
   <div className="person">
     <ProfileImage imageSize={PROFILE_IMAGE_MICRO} profile={profile} />
     <div className="description">
-      <span className="name">{profile.name}</span>
+      <Link to={profileURL(profile.username)} className="name navy">
```
pdpinch: I just want to double-check that this will have the same underline-on-hover styling as other author links, e.g. in the post summary card, or the search results.

PR #1822, opened 2019-02-06
upgrade to react 16.8

aliceriot:

https://reactjs.org/blog/2019/02/06/react-v16.8.0.html :tada: 

#### What's this PR do?

upgrades to react 16.8 - we were already on 16.7 so it's not a very big upgrade.

make sure everything is working!


aliceriot:
it's currently using `entry.link` as a key - what do you think we should use instead?

noisecapella:
I think we should probably just use the index value in the list for the `RssWidget`. I don't think any part of an RSS entry is guaranteed to be unique in a list, and we aren't really doing any modification of the RSS entries after the initial render so using the index should be fine there.

aliceriot:
ok cool, I'll make that change

aliceriot:
@noisecapella good to merge?

noisecapella:
:+1:

PR #1827, opened 2019-02-07
Remove usages of channel description and allow_widget_ui

### LINKED ISSUE #1802: Remove description field under basic channel settings
With widgets turned on in all envs, this field no longer shows up anywhere so we should probably remove the field in the admin settings to reduce confusion. I think this was meant to be dealt with conditionally based on the widgets being turned on, but it doesn't appear to have been.

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1802 

#### What's this PR do?
Removes references to the `channel.description` field since this is no longer used. Particularly the channel creation/edit forms. This also eliminated any usefulness the widget ui feature flag provided, so I removed that too, which did impact quite a few tests.

The application should not break


static/js/flow/discussionTypes.js
```diff
@@ -47,7 +45,6 @@ export type ChannelForm = {
 
 export type ChannelAppearanceEditValidation = {
   title:              string,
-  description:        string,
```
aliceriot: kind of odd that we even had this field here, since the validator didn't have a check for `description`, but oh well!

aliceriot:
there's a test in `EditChannelAppearancePage_test` called `should set the description` that I think we may want to rename. It's not actually testing anything to specific to the description field, just basically that a good `onUpdate` handler is being passed down to the `EditChannelAppearanceForm`.

rhysyngsun:
@aliceriot done

aliceriot:
one other thing I noticed! this test was passing as-is but should probably be edited like this:

```diff
diff --git a/static/js/lib/api/channels_test.js b/static/js/lib/api/channels_test.js
index c1db6000..32a2c10b 100644
--- a/static/js/lib/api/channels_test.js
+++ b/static/js/lib/api/channels_test.js
@@ -118,7 +118,6 @@ describe("Channels API", () => {
         method: PATCH,
         body:   JSON.stringify({
           title:              input.title,
-          description:        input.description,
           public_description: input.public_description,
           channel_type:       input.channel_type
         })
```

aliceriot:
that it the test called `updates a channel`

rhysyngsun:
Oh the joys of JS `undefined`...

Issue #1749, opened 2019-01-22
Investigate caching RSS feed results


noisecapella:
Right now when a user loads the channel page it will cause an RSS feed fetch on the backend, which slows down page load and adds extra load on the RSS feed server. If feasible we should investigate implementing a simple cache for the RSS feed results.

Ferdi:
Do you think we should also try to load selected widget types asynchronously from the main page?

PR #1835, opened 2019-02-08
RFC: Caching system for third party data

rhysyngsun:
[Rendered](https://github.com/mitodl/open-discussions/blob/rfc/0001-caching-system/docs/rfcs/0001-caching-system.md)

RFC for #1749 

#### What's this PR do?
Adds an RFC for how to approach caching third party data

### DIFF
--- a/docs/rfcs/0001-caching-system.md
+++ b/docs/rfcs/0001-caching-system.md
@@ -1,0 +1,83 @@
+## Caching System for Third-Party Data
+
+### Table of Contents
+
+- [Abstract](#abstract)
+- [Architecture Changes](#architecture-changes)
+- [Security Considerations](#security-considerations)
+  - [Authentication](#authentication)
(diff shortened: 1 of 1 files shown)


gsidebo:
This definitely seems to be a very common Redis use case, we have already some familiarity with Redis (at least on the Ops side), it's easy enough to add another Redis service to our docker compose config (though I'm not sure how much that would slow things down while developing), and I think it would be pretty straightforward for Ops to write salt formulas for running 2 redis servers on the same machine or on separate machines. I'm no caching expert but my gut says go with Redis

Ferdi:
A simple option for both these use cases is to use our postgres db as a cache store.

Do you have in mind a time based caching ?

aliceriot:
when I was initially implementing the embedly API I set up a memcached-based solution to cache requests to that API. It was relatively straightforward to set up, we might want to look into that?

Issue #1811, opened 2019-02-04
Add embedly link preview content to the index


gsidebo:
Link posts cannot be found via search based on their URL nor the contents of the site being linked to. Example: if I have a link post for the url "news.mit.edu", I cannot find it by entering "mit.edu" into the search bar, nor can I find it using any text that exists in the body of that page. In addition to that, link posts can only be considered related to a given post based on the post title, channel, and author (related posts first implemented in #1748, PR #1797). Indexing the web page contents from embedly will give us a way to search for these link posts in a more meaningful way.

Tech notes: 
1. We might be able to use `plain_text` to index this content, but we should think through the implications of that
2. Obviously the contents of a web page can change, and therefore the indexed contents of a site can become stale. We'll need to decide how much we care about that staleness

EDIT (02/05/2019): For the time being we're not going to worry about updating the site preview content from embedly after we first index it.

aliceriot:
I think from embedly we'll usually get back a "description" field, which might be useful for this

gsidebo:
@pdpinch @Ferdi - I'm indexing the preview description block that embedly gives us, but we might also want to think about adding some other values they provide to the text that we index. Here's some partial JSON documents we get back (with their respective URLs above them):

```json
// news.mit.edu
{
  "description": "MIT News is dedicated to communicating to the media and the public the news and achievements of the students, faculty, staff and the greater MIT community.",
  "title": "MIT News | Massachusetts Institute of Technology",
  "provider_name": "Mit",
}
// https://www.theatlantic.com/science/archive/2019/02/how-to-stop-mosquito-bites/582190/
{
  "description": "NPY acts by sticking to receptor proteins, and the pharmaceutical industry has tried to develop appetite-controlling drugs that target these receptors. Duvall got her hands on some of these drugs five years ago, and fed them to Aedes mosquitoes.",
  "title": "A New Way to Keep Mosquitoes From Biting",
  "provider_name": "The Atlantic"
}
```

I could see `title` and `provider_name` being relevant to searches and for fetching related posts. If both of those values were added to the text that we index, it might look like this for the second example: `The Atlantic - A New Way to Keep Mosquitoes From Biting - NPY acts by sticking to receptor proteins, and the pharmaceutical industry...`. Let me know what you think

pdpinch:
@gsidebo can you do a quick RFC PR for this? Since we're talking about changing the search index, I think it would be worth some wider discussion. 

Questions I have: 
[...]

PR #1846, opened 2019-02-12
RFC: Embedly link preview indexing

gsidebo:
[Rendered](https://github.com/mitodl/open-discussions/blob/c9a2fc6bf95bafb417801490ac57b36c30726594/docs/rfcs/0002-indexing-embedly-link-preview.md)

RFC for #1811 

#### What's this PR do?
Adds an RFC for indexing preview content from Embedly to make link posts searchable.

### DIFF
--- a/docs/rfcs/0002-indexing-embedly-link-preview.md
+++ b/docs/rfcs/0002-indexing-embedly-link-preview.md
@@ -1,0 +1,50 @@
+## Indexing link preview content from Embedly to make link posts searchable 
+
+Read the description of [#1811](https://github.com/mitodl/open-discussions/issues/1811) for a description of the problem we're solving here. This RFC is to address some decisions about how we go about indexing content from Embedly responses and the index structure in general.
+The goal of this RFC is to call attention to some of the finer points of implementation.
+
+#### 1) Using a more descriptive Embedly API endpoint
+
+Here's a direct comparison between two Embedly endpoints using the same URL:
+
+- ['oembed' endpoint](https://embed.ly/docs/explore/oembed?url=https%3A%2F%2Fwww.theatlantic.com%2Fscience%2Farchive%2F2019%2F02%2Fhow-to-stop-mosquito-bites%2F582190%2F)
+- ['extract' endpoint](https://embed.ly/docs/explore/extract?url=https://www.theatlantic.com/science/archive/2019/02/how-to-stop-mosquito-bites/582190/)
+
+`extract` potentially gives us much more text content to index in the `content` field than `oembed` does in the `description` field (this isn't [always true](https://embed.ly/docs/explore/extract?url=https%3A%2F%2Fwww.mit.edu) though). 
+
+*Proposal*: Use the `extract` endpoint instead of `oembed`; if `content` exists, strip the HTML and save that as the link preview text since it's more complete/descriptive; if `content` doesn't exist, use the `description` value. 
+
+*Drawbacks*: Larger responses both from Embedly and from our own API (since ES documents for link posts will be larger), but that's very unlikely to have any significant impact on performance. Text/article posts are already liable to have a similar amount of raw content.
+
+#### 2) Including additional content in the text that we index for a link
+
+Here's a partial document from an embedly response:
+
+```json
+{
+  "content": "<div><p>NPY acts by sticking to receptor proteins, and the pharmaceutical industry ...",
+  "title": "A New Way to Keep Mosquitoes From Biting",
+  "provider_name": "The Atlantic",
+  ...
+}
+```
+
+The `title` and `provider_name` are potentially significant in terms of the search-ability and relatability of these document (example: if I search "Atlantic", it makes some sense to include this link post from the publication called "The Atlantic" in the results). 
+
+*Proposal*: Prepend `provider_name` and `title` to the text that we index for link posts, and separate the values by dashes (et al). Using the example above, the text that we index would look like this: `"The Atlantic - A New Way to Keep Mosquitoes From Biting - NPY acts by sticking to receptor proteins, and ..."`
+
+#### 3) Using the existing `text` model field for storing link preview text
+
+*Proposal*:  Save the preview text for a link post in the `channels.models.Post.text` field (we wouldn't attempt to save text for the corresponding reddit object)
+
+*Potential Issues*: There might still be some application logic on the back- _or_ front-end that considers a post with non-blank `text` value to be a self/text post. This would break that logic.
+
+#### 4) (Possible) Getting rid of either the `plain_text` or `text` field in ES documents
+
+After reviewing this a bit, it became clear that we don't seem to be using the ES documents' `text` field. We only care about the plain text content of these posts (stored in the `plain_text` field). 
+
+*Proposal*: (a) Get rid of the `text` field in ES documents, or (b) get rid of `plain_text`, and instead store only plain text content in the `text` field.
+
+*Drawbacks*: For (a) there may still be some application logic that makes use of that value. For (b) it could be jarring to have two properties with the same name on the model object and its corresponding ES document with different values.
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/0002-indexing-embedly-link-preview.md
```diff
@@ -0,0 +1,50 @@
...
+  "content": "<div><p>NPY acts by sticking to receptor proteins, and the pharmaceutical industry ...",
+  "title": "A New Way to Keep Mosquitoes From Biting",
+  "provider_name": "The Atlantic",
+  ...
+}
+```
+
+The `title` and `provider_name` are potentially significant in terms of the search-ability and relatability of these document (example: if I search "Atlantic", it makes some sense to include this link post from the publication called "The Atlantic" in the results). 
+
+*Proposal*: Prepend `provider_name` and `title` to the text that we index for link posts, and separate the values by dashes (et al). Using the example above, the text that we index would look like this: `"The Atlantic - A New Way to Keep Mosquitoes From Biting - NPY acts by sticking to receptor proteins, and ..."`
```
Ferdi: @gsidebo I still have the same question:  Do we currently search for the term in the "Title" field of the doc ?

docs/rfcs/0002-indexing-embedly-link-preview.md
```diff
@@ -0,0 +1,50 @@
...
+
+#### 3) Using the existing `text` model field for storing link preview text
+
+*Proposal*:  Save the preview text for a link post in the `channels.models.Post.text` field (we wouldn't attempt to save text for the corresponding reddit object)
+
+*Potential Issues*: There might still be some application logic on the back- _or_ front-end that considers a post with non-blank `text` value to be a self/text post. This would break that logic.
+
+#### 4) (Possible) Getting rid of either the `plain_text` or `text` field in ES documents
```
Ferdi: @gsidebo Is this related to saving content from embedly ?

docs/rfcs/0002-indexing-embedly-link-preview.md
gsidebo: Yes we do currently search for the title field, and it's also incorporated into related post queries

docs/rfcs/0002-indexing-embedly-link-preview.md
gsidebo: This affects any scenario where we index some text content for a post, which includes indexing this embedly content. It's not specific to embedly. I figured that this probably wasn't a big enough consideration for a completely new RFC and that I could gather feedback on it while we're already thinking about indexing text content. I didn't intend to do this as part of the PR fro #1811. I'll clarify that

docs/rfcs/0002-indexing-embedly-link-preview.md
```diff
@@ -0,0 +1,50 @@
...
+
+*Potential Issues*: There might still be some application logic on the back- _or_ front-end that considers a post with non-blank `text` value to be a self/text post. This would break that logic.
+
+#### 4) (Related, but separate issue) Getting rid of either the `plain_text` or `text` field in ES documents
+
+After reviewing this a bit, it became clear that we don't seem to be using the ES documents' `text` field. We only care about the plain text content of these posts (stored in the `plain_text` field). 
+
+*Proposal*: (a) Get rid of the `text` field in ES documents, or (b) get rid of `plain_text`, and instead store only plain text content in the `text` field.
+
+*Drawbacks*: For (a) there may still be some application logic that makes use of that value. For (b) it could be jarring to have two properties with the same name on the model object and its corresponding ES document with different values.
```
rhysyngsun: The frontend definitely expects `text` to be the markdown version and `plain_text` to be what you expect. [See this utility](https://github.com/mitodl/open-discussions/blob/master/static/js/lib/search.js#L57) that converts ES results to the frontend `Post` type. We need to be careful here because there's a few dimensions to this problem:

a) when code is deployed that indexes with the document structure
b) when a full reindex actually occurs
c) when frontend code is deployed that handles a) and b)

So I think we'll need to follow some kind of deprecation process

docs/rfcs/0002-indexing-embedly-link-preview.md
Ferdi: Great, than I don't think to do anything for "provider_name" . Title and body/content would be enough

docs/rfcs/0002-indexing-embedly-link-preview.md
```diff
@@ -0,0 +1,50 @@
...
+}
+```
+
+The `title` and `provider_name` are potentially significant in terms of the search-ability and relatability of these document (example: if I search "Atlantic", it makes some sense to include this link post from the publication called "The Atlantic" in the results). 
+
+*Proposal*: Prepend `provider_name` and `title` to the text that we index for link posts, and separate the values by dashes (et al). Using the example above, the text that we index would look like this: `"The Atlantic - A New Way to Keep Mosquitoes From Biting - NPY acts by sticking to receptor proteins, and ..."`
+
+#### 3) Using the existing `text` model field for storing link preview text
+
+*Proposal*:  Save the preview text for a link post in the `channels.models.Post.text` field (we wouldn't attempt to save text for the corresponding reddit object)
```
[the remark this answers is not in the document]
gsidebo: 👍 Easy enough

docs/rfcs/0002-indexing-embedly-link-preview.md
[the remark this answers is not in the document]
gsidebo: The `title` value that I'm referring to is the one that we get from Embedly, which is the actual HTML document title of the page at the URL that you provide. The title that we show in search results and in post summary/detail pages is the title that the user submits. That title value would be unaffected. I'm suggesting that we add the title value to the summary text for the document in ES. We do not currently show that summary text (i.e.: `plain_text`) to the user under any circumstances for link posts, and this issue doesn't include any design changes for link posts in search results

docs/rfcs/0002-indexing-embedly-link-preview.md
gsidebo: Agreed, will need to be done in stages. I'll add here that the `text` field is never used after converting an ES post result to a frontend `Post` object.

docs/rfcs/0002-indexing-embedly-link-preview.md
```diff
@@ -0,0 +1,50 @@
+## Indexing link preview content from Embedly to make link posts searchable 
+
+Read the description of [#1811](https://github.com/mitodl/open-discussions/issues/1811) for a description of the problem we're solving here. This RFC is to address some decisions about how we go about indexing content from Embedly responses and the index structure in general.
+The goal of this RFC is to call attention to some of the finer points of implementation.
```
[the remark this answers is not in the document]
gsidebo: @Ferdi Worth getting your input on this. I'm assuming the answer is "no" for scoping reasons, but if we want this in the future it's good to consider it now. If we want to index the embedly preview for links that a user has included in their article post, I imagine we would either (a) add a new field to the index (`link_content` or something), or (b) intersperse the full preview text in the positions where the user included the links. Either way, we would have documents that could match a search based on text that the user wrote _or_ text from an external site. One case where that would make things interesting is if we ever want to show search results that include highlighted text that matched the search term. We should also consider the fact that an article post can include any number of links.

docs/rfcs/0002-indexing-embedly-link-preview.md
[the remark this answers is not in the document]
gsidebo: Looks like this won't be needed anyway, so disregard

docs/rfcs/0002-indexing-embedly-link-preview.md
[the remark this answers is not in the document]
Ferdi: @gsidebo I suggest we keep it simple for now

PR #1848, opened 2019-02-12
Remove automatic focus from dialog radio buttons

### LINKED ISSUE #1814: in add widget pop-up there shouldn't be a default highlight
Currently, the default state of the add widget pop-up make is look like the first item is selected even when it isn't. 

screen shot 2019-02-05 at 9 54 17 am

The current functionality is fine, with no default choice. Just the highlight should be removed.

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1814 

#### What's this PR do?
Sets focus to the submit button so that the default focus is not on the radio button.

Start managing widgets and then click '+ Add Widget'. You should not see any radio buttons highlighted at all.


aliceriot:
I think this makes it easier to navigate with the keyboard too, since after you make the selection the 'next' button is automatically selected.

PR #1851, opened 2019-02-13
Add url to Course model and helper method to determine its value

### LINKED ISSUE #1834: add url field to course model
As a user searching courses, I'd like to be able to link to them 

#### Designs and Mockups

part of #1820 

#### Acceptance Criteria:

- [ ] for OCW courses, we can construct a complete url by prepending `https://ocw.mit.edu/` to the url field in JSON files
- [ ] for MITx courses, we should use the `marketing_url` which _should_ contain a link to the "About" page on edX, which is the preferred landing page. 

#### Out of Scope

- it appears that some courses have incorrect data for the marketing URL, see https://github.com/mitodl/open-discussions/issues/1820#issuecomment-461784055

mbertrand:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #1834 

#### What's this PR do?
- Adds `url` to the `Course` model
- Adds `task_helpers.get_course_url` to determine what that value should be, which is unfortunately not very straightforward (see #1834 for details).

- Run the new migration.  If you already have courses imported, they should have url values afterward.
- If you do not already have courses imported, run `recreate_index`, set `OCW_UPLOAD_IMAGE_ONLY=True` in your `.env` file (and set other `EDX_` and `OCW_` variables) and then `backpopulate_edx_data` and `backpopulate_ocw_data` - you can interrupt the ocw import after a few minutes, it will take hours to import all.  Verify that the courses have url's.


course_catalog/constants.py
```diff
@@ -3,6 +3,10 @@
 """
 from enum import Enum
 
+OCW_BASE_URL = "http://ocw.mit.edu/"
+MITX_BASE_URL = "https://www.edx.org/course/"
+MITX_ALT_URL = "https://courses.edx.org/courses/"
```
rhysyngsun: Might be a good idea to move these to `settings.py` as env vars with these as defaults just in case.

course_catalog/migrations/0019_course_url.py
```diff
@@ -0,0 +1,30 @@
...
+from course_catalog.task_helpers import get_course_url
+
+
+def populate_urls(apps, schema_editor):
+    """
+    Calculates url's for courses
+    """
+    Course = apps.get_model("course_catalog", "Course")
+    # At the point the migration runs
+    for course in Course.objects.all():
```
rhysyngsun: Just to be safe, could you use `iterator()` here?

course_catalog/task_helpers.py
```diff
@@ -396,3 +404,25 @@ def generate_course_prefix_list(bucket):
...
                 ocw_courses.add("/".join(key_pieces[:-2]) + "/")
     return list(ocw_courses)
+
+
+def get_course_url(course_id, course_json, platform):
+    """
+    Get the url for a course if any
+    """
+    if platform == PlatformType.ocw.value:
+        if course_json is not None:
```
rhysyngsun: This can be cleaned up a bit by putting a `if course_json is None: return None` at the top so we only perform that check once upfront.

course_catalog/task_helpers.py
```diff
@@ -396,3 +404,25 @@ def generate_course_prefix_list(bucket):
             if "/".join(key_pieces[:-2]) != "":
                 ocw_courses.add("/".join(key_pieces[:-2]) + "/")
     return list(ocw_courses)
+
+
+def get_course_url(course_id, course_json, platform):
+    """
+    Get the url for a course if any
```
rhysyngsun: Could you add `Args:` and `Returns:` here?

course_catalog/task_helpers_test.py
```diff
@@ -455,3 +264,51 @@ def test_safe_load_bad_json(mocker):
...
+            PlatformType.ocw.value,
+            "http://ocw.mit.edu/someurl",
+        ],
+        ["e9387c256bae4ca99cce88fd8b7f8272", {"url": ""}, PlatformType.ocw.value, None],
+        ["e9387c256bae4ca99cce88fd8b7f8272", {}, PlatformType.ocw.value, None],
+    ],
+)
+def test_get_course_url(course_id, course_json, platform, expected):
+    """ Test that url's are calculated as expected """
+    assert get_course_url(course_id, course_json, platform) == expected
```
rhysyngsun: This will compare `None == None` on the last two tests and it should be `is expected` for those ones.

course_catalog/task_helpers.py
```diff
@@ -245,8 +251,9 @@ def digest_ocw_course(
         is_published (Bool): Flags OCW course as published or not
         course_prefix (String): (Optional) String used to query S3 bucket for course raw JSONs
```
rhysyngsun: Since we're modifying this function, could we add `Returns:` while we're at it? Also some of the args types are not valid python types (`Bool` -> `bool`, `String` -> `str`)

course_catalog/task_helpers.py
mbertrand: It's actually possible to  get a mitx url even if the JSON is None (just need the course id)

rhysyngsun: Hmmm, right. Guess there's not much that can be improved in readability here...

course_catalog/task_helpers.py
mbertrand: This function doesn't return anything, but I updated the types

mbertrand:
@rhysyngsun comments addressed

PR #1852, opened 2019-02-13
add validation to post edit form

### LINKED ISSUE #1682: Article edit form needs validation
You can edit an article to have a blank body, whereas the post creation form does not allow that. We need to add the same validation behavior and UI to the edit form. **NOTE:** This was noticed on PR #1676. When that merges, in addition to validating the article body, we should include error behavior/UI for failed image uploads like the post creation page already has.

Post creation w/ validation message:
ss 2019-01-10 at 15 57 24

Post detail with article body set to blank:
ss 2019-01-10 at 15 56 37

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #1682

#### What's this PR do?

This adds validation to the post editing on the post detail page. You should be able to trigger a validation message showing up by either:

- editing an article post to be empty
or
- making a verrrryyy long (more than 40000 characters) post

Editing a post should work as before. Creating posts should work normally and there should be no regressions there. All flows for creating posts should be tested: all post types, switching between post types, switching between channels which disallow different post types, etc.

(Optional)

(Optional)
artfjfjffj
toooooolong

### DIFF
--- a/static/js/lib/validation.js
+++ b/static/js/lib/validation.js
@@ -68,7 +68,13 @@ const validateIfPostType = (
   postType: PostFormType,
   validator: Function
 ) => (postForm: { value: PostForm }) => {
-  if (R.isEmpty(postForm) || postForm.value.postType !== postType) {
+  if (R.isEmpty(postForm)) {
+    return S.Nothing
+  }
+
+  const formPostType = postForm.value.postType || postForm.value.post_type
+
+  if (formPostType !== postType) {
     return S.Nothing
   } else {
     return validator(postForm)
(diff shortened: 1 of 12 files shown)


pdpinch:
I'm a little confused about the mix of "Article" and "Post." Does this valid both article and text posts? If it's just article posts, can we change "This post is too long" to "This article is too long" ?

Also, just curious: where does the 40000 character limit come from?

aliceriot:
The length validation is just for text posts. I believe the restriction is to enforce the max length that Reddit allows, but I don't remember exactly why we added it.

pdpinch:
Ah thanks. That answers my question and moots my comment.

static/js/lib/validation.js
```diff
@@ -91,12 +97,32 @@ export const postURLValidation = validateIfPostType(
...
+
+  // if it's an empty array that's easy
+  if (R.equals(article, [])) {
+    return true
+  }
+
+  // sometimes the editor will leave one empty tag in it
+  if (article.length === 1) {
+    const [obj] = article
+    if ((R.equals(obj.attributes, {}), R.equals(obj.children, []))) {
```
noisecapella: Was the intent to use the comma operator here? It will ignore the left item and only return the result of the right item

aliceriot: ah no that should have been `||`, which I'm pretty sure I wrote...wondering if the formatter had other ideas

aliceriot:
@noisecapella I fixed that issue

PR #1859, opened 2019-02-14
Adds functionality to export edx courses into json format and upload it to s3

alizag:
Addresses this issue ["export edX discovery API response to S3"](https://github.com/mitodl/course-catalog/issues/51)

#### What's this PR do?
Adds functionality to export edx courses into json format and upload it to s3

`course_catalog/tasks.py`

### DIFF
--- a/course_catalog/tasks.py
+++ b/course_catalog/tasks.py
@@ -44,11 +44,12 @@ def sync_and_upload_edx_data(force_overwrite=False):
         return
     url = settings.EDX_API_URL
     access_token = get_access_token()
-    error_occurred = False
+    should_upload_to_s3 = True
 
     edx_data = {"count": 0, "catalog_url": settings.EDX_API_URL, "results": []}
 
     while url:
+        log.info("Syncing edX data...")
         response = requests.get(url, headers={"Authorization": "JWT " + access_token})
         if response.status_code == 200:
             for course_data in response.json()["results"]:
@@ -59,14 +60,16 @@ def sync_and_upload_edx_data(force_overwrite=False):
                     log.exception(
                         "Error encountered parsing MITx json for %s", course_data["key"]
                     )
+                    should_upload_to_s3 = False
         else:
             log.error("Bad response status %s for %s", str(response.status_code), url)
-            error_occurred = True
+            should_upload_to_s3 = False
             break
 
         url = response.json()["next"]
 
-    if not error_occurred:
+    if should_upload_to_s3:
+        log.info("Uploading edX courses data to S3")
         edx_data["count"] = len(edx_data["results"])
         raw_data_bucket = boto3.resource(
             "s3",
--- a/course_catalog/tasks.py
+++ b/course_catalog/tasks.py
@@ -3,6 +3,7 @@ course_catalog tasks
 """
 import logging
 
+import json
 import requests
 import boto3
 from django.conf import settings
@@ -35,16 +36,26 @@ def get_edx_data(force_overwrite=False):
         settings.EDX_API_URL
         and settings.EDX_API_CLIENT_ID
         and settings.EDX_API_CLIENT_SECRET
+        and settings.OCW_LEARNING_COURSE_BUCKET_NAME
+        and settings.OCW_LEARNING_COURSE_ACCESS_KEY
+        and settings.OCW_LEARNING_COURSE_SECRET_ACCESS_KEY
     ):
         log.warning("Required settings missing for get_edx_data")
         return
     url = settings.EDX_API_URL
     access_token = get_access_token()
 
+    edx_data = {
+        "count": 0,
+        "catalog_url": "",
+        "results": [],
+    }
+
     while url:
         response = requests.get(url, headers={"Authorization": "JWT " + access_token})
         if response.status_code == 200:
             for course_data in response.json()["results"]:
+                edx_data["results"].append(course_data)
                 try:
                     parse_mitx_json_data(course_data, force_overwrite)
                 except:  # pylint: disable=bare-except
@@ -54,6 +65,13 @@ def get_edx_data(force_overwrite=False):
             break
 
         url = response.json()["next"]
+    edx_data["count"] = len(edx_data["results"])
+    raw_data_bucket = boto3.resource(
+        "s3",
+        aws_access_key_id=settings.OCW_CONTENT_ACCESS_KEY,
+        aws_secret_access_key=settings.OCW_CONTENT_SECRET_ACCESS_KEY,
+    ).Bucket(name=settings.OCW_CONTENT_BUCKET_NAME)
+    raw_data_bucket.put_object(Key="edx_courses.json", Body=json.dumps(edx_data), ACL="public-read")
 
 
 @app.task
--- a/course_catalog/tasks.py
+++ b/course_catalog/tasks.py
@@ -44,6 +44,7 @@ def sync_and_upload_edx_data(force_overwrite=False):
         return
     url = settings.EDX_API_URL
     access_token = get_access_token()
+    error_occurred = False
 
     edx_data = {"count": 0, "catalog_url": settings.EDX_API_URL, "results": []}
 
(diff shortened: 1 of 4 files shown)


course_catalog/tasks.py
```diff
@@ -35,16 +36,26 @@ def get_edx_data(force_overwrite=False):
...
+    }
+
     while url:
         response = requests.get(url, headers={"Authorization": "JWT " + access_token})
         if response.status_code == 200:
             for course_data in response.json()["results"]:
+                edx_data["results"].append(course_data)
                 try:
                     parse_mitx_json_data(course_data, force_overwrite)
                 except:  # pylint: disable=bare-except
```
kyfantaz: @mbertrand do you know why L50 is a bare `except` instead of it being catching a base exception i.e. `except Exception`? It's changed from the course-catalog repo and with the pylint disable, I assume it's intentional.

course_catalog/tasks.py
```diff
@@ -35,16 +36,26 @@ def get_edx_data(force_overwrite=False):
...
+        and settings.OCW_LEARNING_COURSE_ACCESS_KEY
+        and settings.OCW_LEARNING_COURSE_SECRET_ACCESS_KEY
     ):
         log.warning("Required settings missing for get_edx_data")
         return
     url = settings.EDX_API_URL
     access_token = get_access_token()
 
+    edx_data = {
+        "count": None,
```
mbertrand: Why is `count` set to `None`?  Doesn't seem like it ever gets updated to another value.

course_catalog/tasks.py
```diff
@@ -54,6 +65,13 @@ def get_edx_data(force_overwrite=False):
             break
 
         url = response.json()["next"]
+    edx_data["count"] = len(edx_data["results"])
+    raw_data_bucket = boto3.resource(
+        "s3",
+        aws_access_key_id=settings.OCW_CONTENT_ACCESS_KEY,
+        aws_secret_access_key=settings.OCW_CONTENT_SECRET_ACCESS_KEY,
+    ).Bucket(name=settings.OCW_CONTENT_BUCKET_NAME)
+    raw_data_bucket.put_object(Key="edx_courses.json", Body=json.dumps(edx_data))
```
mbertrand: Can you add a unit test that mocks `put_object` and verifies that it is called with the correct arguments here (`Key="edx_courses.json", Body=json.dumps(edx_data)`)

course_catalog/tasks.py
mbertrand: That is what's typically used for celery tasks in this project (for example most of `search/tasks.py`.

kyfantaz: Sure. It'll help ensure the service is kept running, though it introduces the risk of catching exceptions like `KeyboardInterrupt` and `SystemExit ` that we may sometimes want to halt a program.

Would it be helpful if more detail was added to the next line `log.exception("Error encountered parsing MITx json")`, perhaps with something from `course_data`?

course_catalog/tasks.py
kyfantaz: No particular reason other than lack of other sensible defaults. It's populated to be an `int` later. We can change it to be `0` if you like.

course_catalog/tasks.py
mbertrand: sure, sounds good

course_catalog/tasks.py
mbertrand: I think you might be able to leave it out here entirely, and it will get included later when it's assigned the count number.

kyfantaz: My opinion is that it may be clearer for the key to exist with some default value in the initial dictionary declaration such that I can glance at the dictionary declaration and expect what keys are there; of course, dictionaries are mutable, so it's not exactly a promise on content schema, but it can serve as a useful signpost for someone skimming the code. 

I'll remove it, unless you change your mind. Ultimately, it's your call.

mbertrand: ok, how about leaving it in there but initializing it to 0

kyfantaz: :+1:

course_catalog/tasks.py
kyfantaz: Added.

course_catalog/tasks_test.py
```diff
@@ -104,19 +104,43 @@ def setup_s3(settings):
...
+    conn = boto3.client(
+        "s3",
+        aws_access_key_id=settings.OCW_LEARNING_COURSE_BUCKET_NAME,
+        aws_secret_access_key=settings.OCW_LEARNING_COURSE_ACCESS_KEY,
+    )
+    json_is_uploaded = conn.list_objects(
+        Bucket=settings.OCW_LEARNING_COURSE_BUCKET_NAME, Prefix="edx_courses.json"
+    )
+    # check that pub_object call to create edx_courses.json succeeded
+    assert json_is_uploaded
```
mbertrand: This will be `True` even if the file doesn't exist.  Can you retrieve the body of the object and verify it contains expected course results.

```python
conn.list_objects(
    Bucket=settings.OCW_LEARNING_COURSE_BUCKET_NAME, 
    Prefix="totally_fake_filename.json"
)

{'EncodingType': 'url',
 'IsTruncated': False,
 'Marker': '',
 'MaxKeys': 1000,
 'Name': 'open-learning-course-data-ci',
 'Prefix': 'totally_fake_filename.json',
 'ResponseMetadata': {'HTTPHeaders': {'content-type': 'application/xml',
   'date': 'Fri, 15 Feb 2019 21:53:39 GMT',
   'server': 'AmazonS3',
   'transfer-encoding': 'chunked',
   'x-amz-bucket-region': 'us-east-1',
   'x-amz-id-2': 'rBWLBWdwFYrqJETvlgmQMi2f8hfYd8WVcyCnGActq7WPcyoCrcu2Epm3Wv3/t4vPPV1Y/np7qB0=',
   'x-amz-request-id': '5BEB8BB6BDF74530'},
  'HTTPStatusCode': 200,
  'HostId': 'rBWLBWdwFYrqJETvlgmQMi2f8hfYd8WVcyCnGActq7WPcyoCrcu2Epm3Wv3/t4vPPV1Y/np7qB0=',
  'RequestId': '5BEB8BB6BDF74530',
  'RetryAttempts': 0}}
```

kyfantaz: Good point. Updated the unit test.

gsidebo:
Hey Zagaran folks 👋  Matt's out this week so I'm taking over this PR

course_catalog/tasks.py
```diff
@@ -35,26 +36,44 @@ def get_edx_data(force_overwrite=False):
...
 
+    edx_data["count"] = len(edx_data["results"])
+
+    raw_data_bucket = boto3.resource(
+        "s3",
+        aws_access_key_id=settings.OCW_LEARNING_COURSE_ACCESS_KEY,
+        aws_secret_access_key=settings.OCW_LEARNING_COURSE_SECRET_ACCESS_KEY,
+    ).Bucket(name=settings.OCW_LEARNING_COURSE_BUCKET_NAME)
+    # saves edx API response results in S3 as one file
+    raw_data_bucket.put_object(Key="edx_courses.json", Body=json.dumps(edx_data))
```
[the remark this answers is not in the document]
pdpinch: Thanks for thinking about resiliency. Definitely in cases 1 and 3 we don't want to write over any existing file.

For case 2, I could go either way. The OCW parsing of the edX data won't have the same requirements as the open-discussions parsing, so the upload doesnt _need_ to fail. But if it's easier to lump everything together, I'm ok with that too. 

For context, this OCW data hasn't been updated in more than a year. So we can tolerate a failed update for some time before it becomes critical.

course_catalog/tasks.py
```diff
@@ -35,26 +36,46 @@ def get_edx_data(force_overwrite=False):
...
+                        "Error encountered parsing MITx json for %s", course_data["key"]
+                    )
         else:
             log.error("Bad response status %s for %s", str(response.status_code), url)
+            error_occurred = True
             break
 
         url = response.json()["next"]
 
+    if error_occurred:
```
gsidebo: Don't know if you're still working on this, but a few comments:
1. This is the reverse of what we want. We only want to upload if there wasn't an error
1. It would be clearer to call this something like `should_upload_to_s3` and initialize it to `True`. If `parse_mitx_json_data`, that's technically an error, so it's confusing as to why you wouldn't set `error_occurred` to `True` when it fails.
1. A couple `log.info` lines would be helpful in this task. One to indicate that we're fetching and syncing edx data, and another one to indicate that we're uploading to S3. You would only need to log the latter if there wasn't an error

kyfantaz: @gsidebo Thanks for setting in to wrap up this PR!
1) Thanks---I saw that immediately but got pulled away from pushing right away. I also took the opportunity to add a new check to the unit tests in the same commit.
2) Happy to rename it...renamed!The reason I didn't set the flag to not upload on a  `parse_mitx_json_data` error is because you mentioned you think we should upload in your scenario case #2 and Peter said he could go either way. If we expect there to be at least one parsing error frequently due to a bad course, we may not want to block the upload of data for all the other courses. I'll add the flag for now to this case though, and it can be removed later if need be.
3) Added!

Let me know if you have any other PR comments.

course_catalog/tasks.py
```diff
@@ -35,26 +36,49 @@ def get_edx_data(force_overwrite=False):
...
+        log.warning("Required settings missing for sync_and_upload_edx_data")
         return
     url = settings.EDX_API_URL
     access_token = get_access_token()
+    should_upload_to_s3 = True
+
+    edx_data = {"count": 0, "catalog_url": settings.EDX_API_URL, "results": []}
 
     while url:
+        log.info("Syncing edX data...")
```
gsidebo: Can you move this so it runs before the `while` loop, not inside of it?

kyfantaz: Definitely.

aliceriot:
in the future please squash down to a single commit before merging so we don't clutter the git history

PR #1860, opened 2019-02-14
Cache RSS widget feed data

### LINKED ISSUE #1749: Investigate caching RSS feed results

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Fixes #1749 
Based on RFC #1835 

#### What's this PR do?
This PR adds caching to RSS feed data. It does so on a per-url basis while allowing multiple RSS widget instances with the same URL to utilize the same cache because they may only differ by number of items shown.

- Run the app, ensure the `external_asset_cache` table is created
- Hit a page with an RSS feed, ensure a record shows up in the table for it
- Put a `raise Exception()` in `_fetch_rss_entries` and hit that page again. The exception should not get raised
- Edit the widget and change the rss feed url. Ensure after saving that you see the new feed items.
- Create a new RSS widget and ensure everything works for that too

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,4 @@
+release: scripts/heroku-release-phase.sh
 web: bin/start-nginx bin/start-pgbouncer-stunnel newrelic-admin run-program uwsgi uwsgi.ini
 worker: newrelic-admin run-program celery -A open_discussions.celery:app worker -B -l $OPEN_DISCUSSIONS_LOG_LEVEL
 extra_worker: newrelic-admin run-program celery -A open_discussions.celery:app worker -l $OPEN_DISCUSSIONS_LOG_LEVEL
--- a/widgets/serializers/rss.py
+++ b/widgets/serializers/rss.py
@@ -1,6 +1,9 @@
 """RSS widget"""
+import logging
 import time
 
+from cache_memoize import cache_memoize
+from django.conf import settings
 import feedparser
 
 from widgets.serializers.widget_instance import (
@@ -13,6 +16,8 @@ from widgets.serializers.react_fields import ReactURLField, ReactIntegerField
 ISOFORMAT = "%Y-%m-%dT%H:%M:%SZ"
 MAX_FEED_ITEMS = 10
 
+log = logging.getLogger()
+
 
 class RssFeedWidgetConfigSerializer(WidgetConfigSerializer):
     """Serializer for RssFeedWidget config"""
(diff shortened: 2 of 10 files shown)


widgets/serializers/rss.py
```diff
@@ -33,17 +56,16 @@ class RssFeedWidgetSerializer(WidgetInstanceSerializer):
 
     def get_json(self, instance):
         """Obtains RSS feed data which will then be provided to the React component"""
-        rss = feedparser.parse(instance.configuration["url"])
-        if not rss:
-            return {"title": "RSS", "entries": []}
+        rss = _fetch_rss(instance.configuration["url"])
+        entries = rss.entries if rss else []
```
gsidebo: `getattr(rss, 'entries', [])`?

scripts/release.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
gsidebo: This is only being run from the Procfile, which means it's only relevant to Heroku, right? If that's right, the name `release.sh` seems misleading. I would think this script creates a new release or something like that. Any reason we wouldn't name it something like `heroku-db-cache-setup.sh`? If there's more that might go in here, maybe something more general like `heroku-setup.sh`?

scripts/release.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -eo pipefail
+indent() {
+    RE="s/^/       /"
+    [ $(uname) == "Darwin" ] && sed -l "$RE" || sed -u "$RE"
+}
+
+MANAGE_FILE=$(find . -maxdepth 3 -type f -name 'manage.py' | head -1)
+MANAGE_FILE=${MANAGE_FILE:2}
```
gsidebo: Nit: can you include a comment that this is just trimming the `./` from the path?

scripts/release.sh
rhysyngsun: I'll name it `heroku-release-phase.sh` since that's specifically what it's for.

rhysyngsun:
@gsidebo feedback addressed

Procfile
```diff
@@ -1,3 +1,4 @@
+release: scripts/release.sh
```
gsidebo: Need to update the name here

widgets/serializers/rss.py
```diff
@@ -63,3 +85,18 @@ def get_json(self, instance):
...
+    def save(self, **kwargs):
+        """Saves the widget settings"""
+        instance = super().save(**kwargs)
+
+        try:
+            url = instance.configuration["url"]
+
+            # force a cache refresh immediately after a successful save by invalidating and then loading it
+            _fetch_rss.invalidate(url)
+            _fetch_rss(url)
```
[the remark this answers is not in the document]
rhysyngsun: It can, and I originally had this, but `pylint` chokes on it so the line ended up needing to be:

```python
_fetch_rss(url, _refresh=True)  ## pylint: disable=unexpected-keyword-arg
```

So I ended up going with the more verbose version here over disabling the pylint check.

rhysyngsun:
@gsidebo all set

PR #1867, opened 2019-02-15
Fixed anonymous create post page bug

### LINKED ISSUE #1710: don't allow anonymous user to access /create_post
When an anonymous users clicks on the "Compose a post" we display a dialog asking them to log in or create an account. 

However, if they somehow manage to navigate directly to /create_post, there is no message and there are no channels available in the drop-down. 

I think we should redirect anonymous users to /create_post to the /login page.

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1710 

#### What's this PR do?
- Fixes bug where anonymous users could access the create post page
- Adds a new `Route` component that auto-redirects if a user is not logged in (`PrivateRoute`)

As an anonymous user, try to navigate to `/create_post/` and `/create_post/<some_channel_name>`. Both should redirect you to the login page with the right `next` param

- `PrivateRoute` implementation inspired by [this example](https://reacttraining.com/react-router/web/example/auth-workflow)
- `/create_post/<some_channel_name>` already redirected before this PR, but `/create_post/` didn't. That was kind of by accident. If a channel name existed in the URL, you would get redirected to login since the API call to fetch channel info is implemented with `fetchJSONWithAuthFailure`. If you didn't have the channel name in the URL, that fetch wouldn't be attempted, so the route wasn't protected

### DIFF
--- a/static/js/components/auth/PrivateRoute.js
+++ b/static/js/components/auth/PrivateRoute.js
@@ -1,0 +1,25 @@
+// @flow
+import React from "react"
+
+import { userIsAnonymous } from "../../lib/util"
+import { generateLoginRedirectUrl } from "../../lib/auth"
+
+import { Route, Redirect } from "react-router-dom"
+
+export default function PrivateRoute({
+  component: Component,
+  ...routeProps
+}: Object) {
+  return (
+    <Route
+      {...routeProps}
+      render={props => {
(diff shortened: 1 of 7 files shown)


aliceriot:
functionality works great! reading the code now.

PR #1878, opened 2019-02-21
Hide embedly title for embedly widgets

### LINKED ISSUE #1856: Investigate removing embedly widget title
If possible we should see if we can remove the title from the embedly widget. This is inside an iframe so it may be impossible, I forget what we can do with CSS and iframes. There could be some way to configure the embedly platform to omit it.

If we determine this to be difficult to do, for example, if we require switching away from the embedly card API, then @pdpinch says we should just skip it.

#### Designs and Mockups
screenshot from 2019-02-13 16-55-36

#### Acceptance Criteria:

- [ ] the "wikipedia" text in the image above and logo is not visible

#### Out of Scope

- Not going to investigate switching away from the embedly card API to solve this

noisecapella:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Fixes #1856 

#### What's this PR do?
Adds an embedly hook to add CSS to hide a title for embedly card iframes in widgets

Add an embedly widget. You should not see the title (see images below)

#### Screenshot
Before:
screenshot from 2019-02-21 16-29-17

After:
screenshot from 2019-02-21 16-28-33

### DIFF
--- a/static/js/lib/embed.js
+++ b/static/js/lib/embed.js
@@ -60,8 +60,7 @@ export const loadEmbedlyPlatform = () => {
     // $FlowFixMe
     script.parentNode.insertBefore(el, script)
 
-    // $FlowFixMe
-    embedly("on", "card.rendered", function(iframe) {
+    window.embedly("on", "card.rendered", function(iframe) {
       // IE does not provide Element.closest, but it's not essential functionality
       if (iframe.closest && iframe.closest(".no-embedly-title")) {
         const head = iframe.contentDocument.head
--- a/static/js/lib/embed.js
+++ b/static/js/lib/embed.js
@@ -40,6 +40,22 @@ export const hasIframe = R.memoizeWith(R.identity, (html: string) => {
   return !!div.querySelector("iframe")
 })
 
+export const onCardRendered = (iframe: HTMLIFrameElement): void => {
+  // IE does not provide Element.closest, but it's not essential functionality
+  if (iframe.closest && iframe.closest(".no-embedly-title")) {
+    const head = iframe.contentDocument.head
+    const style = iframe.contentDocument.createElement("style")
+    style.setAttribute("type", "text/css")
+    // hide the title
+    style.innerText = ".hdr { display: none; }"
+
+    if (head) {
+      // for flow
+      head.appendChild(style)
+    }
+  }
+}
+
 export const loadEmbedlyPlatform = () => {
   const id = "embedly-platform"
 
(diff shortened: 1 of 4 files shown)


static/js/lib/embed.js
```diff
@@ -59,6 +59,19 @@ export const loadEmbedlyPlatform = () => {
     const script = document.getElementsByTagName("script")[0]
     // $FlowFixMe
     script.parentNode.insertBefore(el, script)
+
+    // $FlowFixMe
+    embedly("on", "card.rendered", function(iframe) {
```
aliceriot: I think you can remove the `$FlowFixMe` if you do `window.embedly` instead (worked locally for me anyhow)

static/js/lib/embed_test.js
```diff
@@ -48,4 +58,59 @@ describe("embed utils", () => {
...
+    it("loads embedly", () => {
+      loadEmbedlyPlatform()
+      assert.equal(document.querySelectorAll("script").length, 2)
+      const firstScript = document.querySelectorAll("script")[0]
+      assert.equal(
+        firstScript.getAttribute("src"),
+        "http://cdn.embedly.com/widgets/platform.js"
+      )
+    })
+    ;[true, false].forEach(hidesTitle => {
```
aliceriot: nit: can you put a `//` line above tests like this, so there's a blank line b/tw tests?

static/js/lib/embed_test.js
```diff
@@ -48,4 +58,59 @@ describe("embed utils", () => {
...
+
+        const container = document.createElement("div")
+        document.body.append(container)
+        if (hidesTitle) {
+          container.setAttribute("class", "no-embedly-title")
+        }
+        const iframe = document.createElement("iframe")
+        container.appendChild(iframe)
+
+        const callback = embedlyStub.firstCall.args[2]
```
aliceriot: it might be just a little easier to follow what's going on here if the callback were defined as a separate function in `lib/embed.js`, so you don't have to grab it out of the stub here

noisecapella: I think I would need to export the function then reimport it in order to be able to stub it as a separate function. Does that sound right?

aliceriot: not sure I follow - I was proposing that the function defined inline on line 63 of `lib/embed.js` be defined in that file as a separate function just to make testing a little easier. I don't think that would require any changes to how you're stubbing things, unless I'm missing something.

noisecapella: Ah, I think I misunderstood. I made the change in the last commit, let me know if that's what you had in mind.

noisecapella:
I think that should cover everything

PR #1880, opened 2019-02-25
add livestream widget to home page

### LINKED ISSUE #1767: livestream "widget" on front page
As a user, I'd like to see what's happening at MIT right now. 

The MVP department webcasts dozens of videos every week, using livestream.com. We can embed a simple "What's happening now" from livestream.com on the front page.

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #1767 

#### What's this PR do?

This basically adds a little widget to the homescreen which will showcase an event which is currently being livestreamed by MIT video's livestream account.

This involves:

- adding code on the backend for storing API credentials and using them to get information about the relevant account
- adding an API which makes that information available to the frontend
- adding to the frontend a new component which makes a request to that API and, if there is a currently-live event, uses an iframe to embed the event (allowing the user to watch it right there).

message me for configuration stuff. then test that, with things configured, the homepage works normally. you should see a new request being made to the livestream API on the homepage, which should not result in anything being displayed.

in order to do an end-to-end test we need to have a currently live event. I have my phone set up so I can do this on our test account, so if you're testing it just lmk and I'll set up an event so it will come back as a "live" event and you can test the whole thing.

here's what it looks like when there is an event:

basicdisplay


Ferdi:
Great, what happens when you click the video ?

aliceriot:
it will play inline, similar to how youtube videos work in the 'embed' widget

livestream/api_test.py
```diff
@@ -0,0 +1,14 @@
...
+from livestream.api import get_upcoming_events
+
+
+def test_get_upcoming_events(settings, mocker):
+    """test get upcoming events"""
+    settings.LIVESTREAM_ACCOUNT_ID = 392_239
+    settings.LIVESTREAM_SECRET_KEY = "secret key"
+    requests_patch = mocker.patch("requests.get", autospec=True)
+    get_upcoming_events()
+    requests_patch.assert_called_once_with(
```
noisecapella: Can you also test the return value? `assert get_upcoming_events() == requests_patch.return_value` I think should work

livestream/views.py
```diff
@@ -0,0 +1,20 @@
+"""Livestream Views"""
+from rest_framework import status
+from rest_framework.decorators import api_view, permission_classes
+from rest_framework.response import Response
+
+from django.conf import settings
```
noisecapella: This should be grouped with the `rest_framework` imports since they're both 3rd party libraries

livestream/views.py
```diff
@@ -0,0 +1,20 @@
...
+
+from livestream.api import get_upcoming_events
+from open_discussions.permissions import AnonymousAccessReadonlyPermission
+
+
+@api_view()
+@permission_classes([AnonymousAccessReadonlyPermission])
```
noisecapella: If we're only using `GET` for this API and everyone is allowed to read from it then we can just pass an empty list here and pass `["GET"]` as an argument to `api_view`

livestream/views_test.py
```diff
@@ -0,0 +1,39 @@
...
+    resp = client.get(reverse("livestream"))
+    get_stub.assert_called_once()
+    assert resp.json() == {
+        "total": 2,
+        "data": [{"first": "event"}, {"second": "event"}],
+    }
+    assert resp.status_code == status.HTTP_200_OK
+
+
+def test_get_events_no_vars(client, user, settings):
```
noisecapella: Can we parameterize this to also test if one or the other is missing? For example

    @pytest.mark.parametrize("has_account_id,has_secret_key", [
        [True, False],
        [False, True],
        [True, False],
    ])
    def test_get_events_no_vars(client, user, settings, has_account_id, has_secret_key):
        """we should get a 503 if the settings aren't present"""
        settings.LIVESTREAM_ACCOUNT_ID = "account_id" if has_account_id else None
        settings.LIVESTREAM_SECRET_KEY = "secret_key" if has_secret_key else None
        client.force_login(user)
        resp = client.get(reverse("livestream"))
        assert resp.json() == {}
        assert resp.status_code == status.HTTP_503_SERVICE_UNAVAILABLE

static/js/components/LiveStream.js
```diff
@@ -0,0 +1,67 @@
...
+    if (liveEvent && SETTINGS.livestream_ui_enabled) {
+      const { ownerAccountId, id } = liveEvent
+
+      return (
+        <Card className="widget">
+          <div className="title-row">
+            <span className="title">Live Now @MIT</span>
+          </div>
+          <iframe
+            id="ls_embed_1550773899"
```
noisecapella: What's the significance of this id?

static/js/components/LiveStream.js
```diff
@@ -0,0 +1,67 @@
...
+  render() {
+    const { liveEvent } = this.props
+
+    if (liveEvent && SETTINGS.livestream_ui_enabled) {
+      const { ownerAccountId, id } = liveEvent
+
+      return (
+        <Card className="widget">
+          <div className="title-row">
+            <span className="title">Live Now @MIT</span>
```
noisecapella: Is the title something we'll want to parametrize?

pdpinch:
This is behind a feature flag, right?

livestream/api_test.py
aliceriot: sure

static/js/components/LiveStream.js
aliceriot: I think it was there when I copied out their example iframe and I just forgot to remove it, I'll try deleting it and see if that messes anything up

aliceriot: yup doesn't seem necessary

static/js/components/LiveStream.js
aliceriot: for now I don't think so, if we want to add that later it's not so complicated

aliceriot:
@pdpinch yep, `LIVESTREAM_UI`

livestream/views.py
aliceriot: cool, will do

aliceriot:
@noisecapella ok I think all comments should be addressed

PR #1892, opened 2019-02-28
Display all topics in CompactCourseDisplay, make clickable

### LINKED ISSUE #1888: Course Search: display all Subject areas on Course card
As a search users, it would make more sense to see all subject areas for a given course in the search results. 

Currently, we only show one of the subject areas on the Course Card. 

#### Designs and Mockups

none

#### Acceptance Criteria:

- [ ] Find out which course(s) have the maximum number of Subject areas and include them in the screenshot of the PR
- [ ] Clicking a course subject on the Course card in the search results should link to a search with that filter applied.

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes #1888

#### What's this PR do?
- Displays all subject areas on the compact course card instead of just one.
- Makes each subject area on the card clickable which updates the search filters.

- Run a search
- On one of the cards, click a subject area.  That subject area's checkbox should get checked, the URL should get updated to include it, and the search should refresh with that filter applied.
- Uncheck the recently checked facet, it should be removed from the URL and the search results should refresh again without that filter applied.

I ran into a strange problem where unchecking a facet checkbox that was recently activated via clicking on a CompactCourseCard would not trigger an `onChange` event.  This was resolved by altering the code to use `onClick` instead, though that does trigger a warning message in the console. 

screen shot 2019-03-04 at 12 33 37 pm

screen shot 2019-03-04 at 12 33 58 pm

### DIFF
--- a/static/js/components/CompactCourseDisplay.js
+++ b/static/js/components/CompactCourseDisplay.js
@@ -46,7 +46,7 @@ export class CompactCourseDisplay extends React.Component<Props> {
             <div className="row topics-row">
               {_.sortBy(course.topics).map((topic, i) => (
                 <div
-                  className="grey-surround flexless"
+                  className="grey-surround facet topic"
                   key={i}
                   onClick={() => this.onToggleFacet("topics", topic.name)}
                 >
@@ -55,16 +55,14 @@ export class CompactCourseDisplay extends React.Component<Props> {
               ))}
             </div>
           </div>
-          <div className="row preview-footer" onClick={this.setCourseForDrawer}>
-            <div className="course-info">
-              <span className="course-availability grey-surround flexless">
-                {courseAvailability(course)}
-              </span>
-              <span className="course-platform grey-surround flexless">
-                {course.platform.toUpperCase()}
-              </span>
+          <div className="row preview-footer">
+            <div className="course-availability">
+              {courseAvailability(course)}
             </div>
-            <div className="course-price grey-surround">{maxPrice(course)}</div>
+            <div className="course-platform">
+              {course.platform.toUpperCase()}
+            </div>
+            <div className="course-price">{maxPrice(course)}</div>
           </div>
         </div>
         {course.image_src ? (
--- a/static/js/components/CompactCourseDisplay.js
+++ b/static/js/components/CompactCourseDisplay.js
@@ -16,7 +16,7 @@ import type { Dispatch } from "redux"
 type Props = {
   course: Course,
   dispatch: Dispatch<*>,
-  toggleFacet: ?Function
+  toggleFacet?: Function
 }
 
 export class CompactCourseDisplay extends React.Component<Props> {
--- a/static/js/components/CompactCourseDisplay.js
+++ b/static/js/components/CompactCourseDisplay.js
@@ -2,6 +2,7 @@
 /* global SETTINGS:false */
 import React from "react"
 import { connect } from "react-redux"
+import _ from "lodash"
 
 import Card from "./Card"
 import { setShowCourseDrawer } from "../actions/ui"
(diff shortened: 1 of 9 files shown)


mbertrand:
@pdpinch @Ferdi I've updated the screenshots with some more styling changes.   Decided to remove the `grey-surround` style for the availability, platform, and price since it made them look like buttons but clicking on them won't do anything.

Ferdi:
Great !

aliceriot:
functionality looks great!

static/js/components/CompactCourseDisplay.js
```diff
@@ -23,31 +25,44 @@ export class CompactCourseDisplay extends React.Component<Props> {
     dispatch(setShowCourseDrawer({ courseId: course.id }))
   }
 
+  onToggleFacet = async (name: string, value: string) => {
+    const { toggleFacet } = this.props
+    if (toggleFacet) {
```
aliceriot: will this prop ever be null? I was just looking through and I didn't see anywhere where it would be null.

static/js/containers/CourseSearchPage.js
```diff
@@ -84,6 +93,12 @@ export class CourseSearchPage extends React.Component<Props, State> {
     this.runSearch()
   }
 
+  componentDidUpdate(prevProps: Object, prevState: Object) {
+    if (shouldRunSearch(prevState, this.state)) {
+      this.runSearch()
+    }
+  }
```
aliceriot: do you think we need to add a test for this little bit of logic?

static/js/components/CompactCourseDisplay.js
mbertrand: When I initially started on this, my assumption was that `CompactCourseDisplay` would also be used on the course index page, and in that case the function would be null.  But it looks like the index page will need a bigger card, so will need something like a `RegularCourseDisplay` component.  So assuming it's never used outside of search, it should never be null.

static/js/containers/CourseSearchPage.js
mbertrand: will add it

static/js/components/CompactCourseDisplay.js
mbertrand: Changed props code slightly to `toggleFacet?: Function` to be consistent with other function props, this is due to the way `SearchResult` is designed, it may or may not have certain props depending on whether the result is for posts, comments, or courses.  And so I left the `if (toggleFacet)` condition there to make flow happy and avoid having to replace it with a `// $FlowFixMe` line.

static/js/containers/CourseSearchPage.js
mbertrand: There were already tests to verify a search was run if the text or a subject facet changed, I added another test to verify that no search is run if no search parameters are changed.

static/js/components/CompactCourseDisplay.js
aliceriot: ah, I didn't realize that the `SearchResult` component was shared between all the different search interfaces. that's not optimal I think, but for now we can stick with this.

aliceriot: looking at the course search page - is there anything we get out of using the `SearchResult` component there? It seems like in the course search use-case it is basically just passing props down to the `CompactCourseDisplay` without adding any functionality. For simplicity if we can get away with just using the `CompactCourseDisplay` directly on the `CourseSearchPage` I think we should do that (and avoid putting extra components on the page that don't add any functionality)

mbertrand: There are a couple differences between the fields present in the Elasticsearch result and the Course model fields, which is handled via `searchResultToCourse`, similar to `searchResultToProfile`, `searchResultToPost`, etc.  I could change the `CompactCourseDisplay` to expect a `CourseResult` object in props instead of `Course`, but that assumes we never want to use `CompactCourseDisplay` outside of a search page.  That's true now, but not sure if that will always be the case?

aliceriot: could you just inline the `searchResultToCourse` call into the JSX for the `CompactCourseDisplay`, like this?

```js
<CompactCourseDisplay course={searchResultToCourse(result)} toggleFacet={toggleFacet} />
```

mbertrand: okay, I can do this and remove the course-related code from `SearchResult`, but this assumes we won't want at some point want a search page to return all possible doc types (post, comment, profile, and search).

mbertrand:
@aliceriot one thing that concerned me was needing to change `onChange` to `onClick` on the facet checkboxes for the reason mentioned in the PR description, have you ever run into that before?

static/js/components/CompactCourseDisplay.js
mbertrand: This change works functionally but ends up causing 60 flow errors.  Since the code in question is not new to this PR, I think it might be best to leave it as is and open a tech debt issue to change it at some point in the future.

aliceriot: I was able to do this and get no flow errors:

```diff
diff --git a/static/js/containers/CourseSearchPage.js b/static/js/containers/CourseSearchPage.js
index dbddd5ed..a5fa5f53 100644
--- a/static/js/containers/CourseSearchPage.js
+++ b/static/js/containers/CourseSearchPage.js
@@ -22,17 +22,19 @@ import {
   withLoading
 } from "../components/Loading"
 import SearchTextbox from "../components/SearchTextbox"
-import SearchResult from "../components/SearchResult"
+import CompactCourseDisplay from "../components/CompactCourseDisplay"
+
 import { actions } from "../actions"
 import { clearSearch } from "../actions/search"
 import { COURSE_AVAILABILITIES } from "../lib/constants"
 import { SEARCH_FILTER_COURSE } from "../lib/picker"
 import { preventDefaultAndInvoke, toArray } from "../lib/util"
+import { searchResultToCourse } from "../lib/search"
 
 import type {
   SearchInputs,
   SearchParams,
-  Result,
+  CourseResult,
   CourseFacetResult
 } from "../flow/searchTypes"
 
@@ -44,7 +46,7 @@ type Props = {
   isModerator: boolean,
   match: Match,
   runSearch: (params: SearchParams) => Promise<*>,
-  results: Array<Result>,
+  results: Array<CourseResult>,
   searchLoaded: boolean,
   loaded: boolean,
   searchProcessing: boolean,
@@ -53,6 +55,7 @@ type Props = {
   facetChoices: CourseFacetResult,
   facetChoiceProcessing: boolean
 }
+
 type State = {
   text: string,
   topics: Array<string>,
@@ -205,9 +208,9 @@ export class CourseSearchPage extends React.Component<Props, State> {
         loader={<Loading className="infinite" key="loader" />}
       >
         {results.map((result, i) => (
-          <SearchResult
+          <CompactCourseDisplay
             key={i}
-            result={result}
+            course={searchResultToCourse(result)}
             toggleFacet={this.toggleFacet}
           />
         ))}
```

I think that's cleaner and more straightforward, you can take it or leave it since it's working as is.

aliceriot: you know, nvm, I think it's fine as is

aliceriot:
I haven't run into the `onChange` issue before, I'm not too worried about using `onClick` instead but I'm a little curious why it didn't work...sometimes React has some mysterious corners, haha

PR #1899, opened 2019-03-05
populates and indexes new Course field `availability`

### LINKED ISSUE #1898: Course search: add "availability field" to course model
As a user searching for courses, I'd like to be able to see at a glance whether courses are available now, in the future, or in an archive. I'd also like to be able to filter my search results by availability. 

Currently, we calculate whether a course is available when building the index. but we found that we're missing some necessary dates for this calculation. 

Instead of using a calculation, we can rely on the `availability` field in the edX course API. All OCW course can be "Current" 

#### Acceptance Criteria:

- [ ] For MITx courses, import `availability` values from the course feed
- [ ] For OCW courses, all courses are "Current" 
- [ ] Add availability to the ES index

kyfantaz:
Closes #1898 in conjunction with PR #1900.
Note: this PR merges code into the branch that PR #1900 is asking merging into `master`.
Code is split into two separate PRs on Peter's request.

#### What's this PR do?
Adds content population for newly added `availability` Course field on import and includes the content for Elasticsearch indexing via `ESCourseSerializer`

One question for the reviewer:
Should values for the `Course.availability` field be enforced at the database-level via a "choices" field parameter populated by the values of`course_catalog/constants/AvailabilityType` enum? 
Position of the author: I would typically prefer enforcement from an enum (and thus giving me a guarantee on the range of values I as a developer can expect from the database), but I ultimately opted to not enforce it at the database-level because we don't technically know all the possible values. [The edX documentation](https://course-catalog-api-guide.readthedocs.io/en/latest/course_catalog/catalog.html) doesn't seem to have any mention of availability or possible values returned by the API. Instead, this code trusts the edX API to return consistent values and accepts whatever the edX API returns (which appears to be only three values: `"Current"`, `"Upcoming"`, and `"Starting Soon"`). If the reviewer agrees, no action is needed. (note: the values for the enum are based on what I've seen returned from the API upon manual inspection; the enum exists to serve as a constant for when our code is using it)
 
You may call the sync functions manually to check if the new field is being populated correctly:
```
course_catalog/tasks/sync_and_upload_edx_data()
course_catalog/tasks/digest_ocw_course()
```
You can call the index function to check if the field exists in the ES index:
```
search/task_helpers/index_new_course()
```

### DIFF
--- a/course_catalog/factories.py
+++ b/course_catalog/factories.py
@@ -3,7 +3,7 @@ import factory
 from factory.django import DjangoModelFactory
 from factory.fuzzy import FuzzyChoice
 
-from course_catalog.constants import PlatformType
+from course_catalog.constants import PlatformType, AvailabilityType
 from course_catalog.models import Course, CourseInstructor, CourseTopic, CoursePrice
 
 # pylint: disable=unused-argument
@@ -43,6 +43,13 @@ class CourseFactory(DjangoModelFactory):
 
     course_id = factory.Sequence(lambda n: "COURSE%03d.MIT" % n)
     platform = FuzzyChoice((PlatformType.mitx.value, PlatformType.ocw.value))
+    availability = FuzzyChoice(
+        (
+            AvailabilityType.current.value,
+            AvailabilityType.upcoming.value,
+            AvailabilityType.starting_soon.value,
+        )
+    )
 
     @factory.post_generation
     def instructors(self, create, extracted, **kwargs):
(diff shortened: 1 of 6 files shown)


course_catalog/factories.py
```diff
@@ -43,6 +43,13 @@ class CourseFactory(DjangoModelFactory):
 
     course_id = factory.Sequence(lambda n: "COURSE%03d.MIT" % n)
     platform = FuzzyChoice((PlatformType.mitx.value, PlatformType.ocw.value))
+    availability = FuzzyChoice(
+        (
+            AvailabilityType.current.value,
+            AvailabilityType.upcoming.value,
+            AvailabilityType.starting_soon.value,
```
mbertrand: `AvailabilityType.archived.value`

kyfantaz:
> `availability` also should be added to `search.indexing_api.COURSE_OBJECT_TYPE`:
thanks

PR #1900, opened 2019-03-05
adds new field `availability` to Course model

### LINKED ISSUE #1898: Course search: add "availability field" to course model

kyfantaz:
Closes #1898 in conjunction with PR #1899. 
Note: This PR can be reviewed and approved independently, but PR #1899 should be merged first.

#### What's this PR do?
Adds new `availability` field to `course_catalog/Course` model.

You could run the newly added migration and check if the `Course` table has the appropriate column added in Postgres via manual inspection, and that the values match the `raw_json`.
```python
for c in Course.objects.all(): print(c.platform, c.availability)
```


mbertrand:
@kyfantaz running `docker-compose run web black course_catalog` should fix the formatting issue.

course_catalog/task_helpers.py
```diff
@@ -431,3 +434,28 @@ def get_course_url(course_id, course_json, platform):
...
+    """
+    Gets the attribute `availability` for a course if any
+
+    Args:
+        course: Course model instance
+
+    Returns:
+        str: The url for the course if any
+    """
+    if course.platform == PlatformType.ocw.value:
```
mbertrand: Can you add unit tests to cover this function? ocw vs edx, edx with and without course runs

kyfantaz: :+1:

mbertrand:
PS Code should be [squashed before merging into master](https://github.com/mitodl/handbook/blob/master/development-process.md#step-3-merging)

kyfantaz:
:+1: thanks, I remember Alice's reminder on the last PR!

Issue #1299, opened 2018-10-03
Change styling and rendering approach for popups/tooltips


gsidebo:
There are a few issues with our tooltips (referred to in the codebase as 'popups') that prompt anonymous users to log in. Their positioning depends on where they are rendered, so you can't render a single tooltip at, say, the bottom of the page, and use it in multiple places on the page. You instead need to render a separate tooltip for each link/button that needs to reveal one. This is problematic especially on pages with lots of comments (each 'reply' link reveals a tooltip). Another bad side effect is that since the position cannot be determined dynamically, the tooltips can cause some styling issues when they're revealed (e.g.: https://github.com/mitodl/open-discussions/issues/1296)

Overall goal of this issue: 
1. Allow for a single tooltip to be rendered on a page if multiple components make use of it
1. Display the tooltips in a way that doesn't alter the rest of the UI, located near the element that reveals them when clicked
1. (strongly recommended but I guess not crucial) Create an HOC that lets us add tooltips to a component without having to rewrite the rendering and state updating logic over and over
1. (optional but seriously please do this) Rename all variables/classes/components with 'popup' in their name to instead have 'tooltip' in their name. These components are much more accurately described as tooltips

We may want to look at what material design suggests: https://material.io/design/components/tooltips.html#behavior

aliceriot:
yeah, I think we should do this. I think it won't be too hard to throw together a nice solution using the react context API.

gsidebo:
NOTE: The 'Share' tooltip (`static/js/components/SharePopup.js`) also needs to be updated for the same reasons mentioned above. Since it's a separate component, it might be best to work on that in a separate issue for the sake of keeping the diff small (cc @aliceriot)

aliceriot:
oh yeah, I think it should definitely be updated to use the same system

PR #1904, opened 2019-03-07
upgrade flow to @latest

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

I was working on #1299 and I ran into an issue where a newer feature of React wasn't supported by the version of flow we are currently using, so I went ahead and upgraded it.

The newer version had a big change (see here: https://medium.com/flow-type/asking-for-required-annotations-64d4f9c1edf8) where in many cases type parameters that should have been required were not, and lots of typings were silently failing. 

Upgrading lets us also use newer libdefs for things like `react-redux` - in particular, with the new version of that `connect` is no longer a type black hole!

Now instead of just having the return value of `connect` be more or less totally untyped (as far as flow knows anyhow) it accepts a series of parameters, and it will actually check that props that the component is supposed to receive from parents are passed down.

To do this you need to declare new types, `StateProps`, `OwnProps`, and `DispatchProps`. `StateProps` is the return value of the `mapStateToProps` function, and includes any props that come from the redux state. `OwnProps` is props which the component receives from it's parents, and `DispatchProps` are the props which come from `mapDispatchToProps`. Then the type which the unconnected component itself uses is

```js
type Props = {|
  ...OwnProps,
  ...StateProps,
  ...DispatchProps
|}
```

then you annotate `connect` like

```js
export default connect<Props, OwnProps, _, DispatchProps, _,_>(mapStateToProps)(Component)
```

the `_` syntax tells Flow to infer those types (you can read the react-redux typedef docs to see more).

Anyway, this is really nice! Unfortunately it still isn't fully supported in the places where we use `R.compose` for this purpose, since the ramda libdef isn't compatible easily with our usage (I'm talking 7000 type errors, lol), but for more simple components it's a good thing imo.

#### What's this PR do?

upgrades to flow `0.94.0`, and fixes all the type errors that came up. I also bumped the libdefs for a few libraries up to the latest one (we might want to more comprehensively update those soon, but they can be a big pain).

The app should function as normal. There are very few actual code changes, the vast majority of the changes are just flow things, so I would say read through the code and make sure it makes sense, and where there is a code change make sure it's good. Also probably good to just do a basic smoke test of main functions of the app.

### DIFF
--- a/static/js/containers/auth/RegisterConfirmPage.js
+++ b/static/js/containers/auth/RegisterConfirmPage.js
@@ -34,15 +34,27 @@ const partialTokenFromLocation = R.compose(
 
 const isInvalid = R.pathEq(["data", "state"], STATE_INVALID_EMAIL)
 
-type Props = {
+type OwnProps = {|
   match: Match,
-  confirmCode: Function,
   location: Object,
-  history: Object,
-  invalid: boolean
-}
+  history: Object
+|}
 
-export class RegisterConfirmPage extends React.Component<Props, *> {
+type StateProps = {|
+  invalid: boolean
+|}
+
+type DispatchProps = {|
+  confirmCode: Function
+|}
+
+type Props = {|
+  ...OwnProps,
+  ...StateProps,
+  ...DispatchProps
+|}
+
+export class RegisterConfirmPage extends React.Component<Props> {
   componentDidMount() {
     const { confirmCode, location, history } = this.props
     const code = confirmationCodeFromLocation(location)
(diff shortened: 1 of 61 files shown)


static/js/containers/auth/RegisterConfirmPage.js
```diff
@@ -88,13 +100,14 @@ export class RegisterConfirmPage extends React.Component<Props, *> {
...
 
-const mapStateToProps = state => {
+const mapStateToProps = (state: Object): StateProps => {
   const invalid = isInvalid(state.auth)
   return {
     invalid
   }
 }
 
+// $FlowFixMe: just tired of it flow, stop
```
rhysyngsun: :joy:

Issue #1887, opened 2019-02-26
Course Seach: Subject area facet


pdpinch:
As a search user, I'd like to have a full-featured Subject Area facet

#### Designs and Mockups

similar to the "Country" facet of this Searchkit demo: http://demo.searchkit.co/imdb

e.g. 
![image](https://user-images.githubusercontent.com/430126/53446383-d6c1e680-39e0-11e9-8271-d124e887988d.png)

and 
![image](https://user-images.githubusercontent.com/430126/53446400-e17c7b80-39e0-11e9-9654-ee1ff3e4533b.png)

#### Acceptance Criteria:

- [ ] Display the hit count next to each subject filter
- [ ] Don't display a subject for which there are zero hits in the current result set
- [ ] Sort the Subject area filters by hit count
- [ ] Only display the first 5 Subject area filters (we may want to try different numbers for this)
- [x] When the checkbox next to a filter is checked, add it to the search filter
- [ ] When a search filter is added, display it above the search results -- as a reminder to user of which filters are active and so that they can be removed.

mbertrand:
An `availability` field should be added to the `Course` model and search index first, in order to display facet counts properly for `Current`, 'Prior`, and `Upcoming` facets - as described in #1861

mbertrand:
@pdpinch @ferdi: Is this supposed to apply to facets in general (including platforms, availabilities, etc)? Or only subject areas?

Assuming all facets, a consequence of only displaying facet choices contained in the current result set is that it is no longer possible to select more than one facet choice for platform and availability - as soon as one option is selected for either of those, the counts for all other options will be 0 in the results.  For example, selecting "Upcoming" for availability => 0 results with Prior, Current, and Coming Soon, so a user won't be able to select both "Current" and "Upcoming".

PR #1906, opened 2019-03-07
Search facet improvements

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Partially addresses #1887

#### What's this PR do?
Implements the following front-end changes to the course search facets:
- [x] Display the hit count next to each facet
- [x] Don't display a facet for which there are zero hits in the current result set
- [x] Sort the facets by hit count
- [x] Only display the first 5 facet filters, with 'View more/less' functionality if there's > 5

Changes the search API call to include aggregations with each response, assuming facets are included with the request. 

Removes the separate aggregations request which used to be called on initial page load to get a static list of all facet choices.

- if you've already imported courses before, run `recreate_index` (there was a recent addition of 'availability' to the course docs).  If you haven't, run `./manage.py backpopulate_edx_data` and `./manage.py backpopulate_ocw_data` (interrupt the latter after about 10 minutes or it will continue for hours).
- Go to `/courses/search`
- Facets for subject area, platform, and availability should appear.  Only the first 5 subject areas should be visible along with a 'View more' link.
- Clicking 'View more' should show all the subjects and a 'View less' link, which if clicked just shows 5 again.
- Check one of the subject facets.  The search results and URL should change accordingly,  along with the facets and their counts.  The facets in each category should be sorted by count in descending order.
- Check one of the availability facets.  The other availability facets should disappear because they are not found in the results.
- Check one of the platform facets.  The other platform facet should disappear because it is not found in the results.
- Go to the regular search page and verify you can still search for posts, comments, and profiles.

#### Screenshots
screen shot 2019-03-07 at 4 26 03 pm

screen shot 2019-03-07 at 4 26 16 pm


mbertrand:
@Ferdi @pdpinch I added screenshots to the description

mbertrand:
@ferdi I updated the screenshots with counts aligned to the right.

Ferdi:
@mbertrand I think that looks great !

aliceriot:
so which facets are shown is based on which ones are present in the currently selected facets? 

to me it's a little confusing how it works currently, especially with the subject matter facets. my expectation would be that a list of options on a search interface like this would show all the options possible, and I could select and unselect them as I wanted, and the order / presence of these options wouldn't change. That's for instance how options presented on Amazon searches tend to work:

fofdofdofdf

I've selected the 'arabic' and 'italian' language options here, so obviously books with 'german' as their language aren't going to appear in the result set, but the option is still there.

Does that make sense? idk, I just bring it up because I found the behavior of the facet options changing order and disappearing / reappearing to b e a surprising and somewhat jarring behavior.

mbertrand:
@aliceriot only showing facets that are included in search results is one of the issue requirements ("Don't display a facet for which there are zero hits in the current result set")

aliceriot:
yeah I saw that, maybe more a question for @pdpinch, I don't think it's the most intuitive behavior so I wanted to raise whether that's really what we want

aliceriot:
also based on looking at this demo: http://demo.searchkit.co/imdb

it seems that the list of countries (which the issue seems to be based on) is populated based on which countries are found in the initial results for the text search. selecting, say, 'United Kingdom' as a facet does not cause other countries to be removed from the list, but changing the text search to a different query does change which countries show up in the list.

mbertrand:
I discussed with @pdpinch and @Ferdi yesterday, because one of the consequences is that you can't select both "Current" and "Upcoming" for availability, for example: as soon as you select one of these, the other availability facet choices disappear (because a course only has one availability, but can have multiple subjects).  The decision was to leave it as the issue described, with only facets included in  search results being shown.

aliceriot:
one other thing I noticed is that you can construct queries with the URL that result in no facet UI being shown, for me locally if I go to http://localhost:8063/courses/search?a=Current&p=mitx&t=Language&type=course I get no search UI:

fasdfasdfjfjf

maybe that's not such a big deal because users typically don't search for things via URL params, haha.

static/js/components/SearchFacet_test.js
```diff
@@ -0,0 +1,101 @@
...
+    )
+  }
+
+  beforeEach(() => {
+    helper = new IntegrationTestHelper()
+    onUpdateStub = helper.sandbox.stub()
+    dispatchStub = helper.sandbox.stub().returns({ type: "action" })
+    results = new Map(Object.entries(makeSearchFacetResult())).get("topics")
+    // $FlowFixMe: buckets not missing here
+    facet = facet = results.buckets[0]
```
aliceriot: typo?

static/js/containers/CourseSearchPage.js
```diff
@@ -329,6 +294,5 @@ export default R.compose(
   connect(
     mapStateToProps,
     mapDispatchToProps
-  ),
-  withLoading(CourseSearchLoading)
+  )
```
aliceriot: since you removed the `withLoading` component here you can also get rid of the `R.compose` call

Ferdi:
@mbertrand @aliceriot @pdpinch maybe I misunderstood something about this.  I went to the searchkit demo and selected writer=Joe Sweet.   The list of countries in the "countries" got updated to just USA. Is that what's happening in our case as well ? 

http://demo.searchkit.co/imdb?writers[0]=Joe%20Sweet&runtimeMinutes[0]=0_20

aliceriot:
yeah, it seems that the searchkit UI is a little inconsistent - I noticed that as well, but if you select a country the list of countries doesn't change. Maybe that has to do with how the index is set up? I glanced quickly at the source code and the author and country fields are rendered there with the same react component, so I'm not sure.

Anyway, if we're good with the present behavior that's good with me, I just raised it because I found that it did not behave in the way that I expected it would.

static/js/lib/search.js
```diff
@@ -285,44 +208,28 @@ export const buildSearchQuery = ({
...
-    const availabilityClause = buildAvailabilityClause(availabilities)
-
-    const courseClauses =
-      type === "course"
-        ? [...topicClauses, ...platformClauses, ...availabilityClause]
-        : []
+              }
+            ]
+            : []
+        facetClauses.push(...facetClause)
```
aliceriot: might be a teensy bit clearer here to just say

```js
facets.forEach((values, key) => {
  if (values && values.length > 0) {
    facetClauses.push({
      bool: ...
    })
  }
})

**Comment by mbertrand** [comment]
@aliceriot comments addressed.

File: static/js/components/SearchFacet_test.js [hunk 4]
```diff
@@ -23,7 +23,10 @@ describe("SearchFacet", () => {
         title={title}
         onUpdate={onUpdateStub}
         dispatch={dispatchStub}
+        // $FlowFixMe: test results are fine
```
aliceriot: :)

PR #1913, opened 2019-03-08
Different toolbar and no channel drawer for courses

### LINKED ISSUE #1889: Course Search: Hide channel navigation drawer while in course search
As a user, I find channel listings confusing when I'm doing a course search. it should be hidden. 

We probably want a unique template for the course search page. In particular, this would allow us to have some sub-branding for this page (e.g. "MIT Open | Courses" or similar) 

#### Designs and Mockups

None yet

#### Acceptance Criteria:

- [ ] hide the drawer when doing a course search

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes #1889

#### What's this PR do?
Modifies `App` to use a different toolbar and no channel drawer for url's starting with `/courses`

- Go to `/courses/search`.  There should be no drawer at any width, no menu icon to display the drawer, and the toolbar should display the title "MIT Open | Courses" This title should consist of two links - one to the root url, the other to `/courses/` (which will be the course index page in another PR).  The toolbar should still have a working profile icon.
- Go to any other page, the channel drawer and toolbar should look and behave the same as always.

I'm guessing that the course index page URL will be `/courses/`, if not then this PR will need to be tweaked to the correct value.

#### Screenshots

screen shot 2019-03-08 at 12 32 36 pm

screen shot 2019-03-08 at 12 33 17 pm

### DIFF
--- a/static/js/components/CourseToolbar.js
+++ b/static/js/components/CourseToolbar.js
@@ -1,24 +1,22 @@
 // @flow
 /* global SETTINGS: false */
 import React from "react"
-import { NavLink } from "react-router-dom"
+import { Link } from "react-router-dom"
 import { MDCToolbar } from "@material/toolbar/dist/mdc.toolbar"
 
+import MITLogoLink from "./MITLogoLink"
 import UserMenu from "./UserMenu"
-import HamburgerAndLogo from "../components/HamburgerAndLogo"
-
-import { SITE_SEARCH_URL } from "../lib/url"
+import { COURSE_URL } from "../lib/url"
 
 import type { Profile } from "../flow/discussionTypes"
 
 type Props = {
   profile: ?Profile,
   showUserMenu: boolean,
-  toggleShowDrawer: Function,
   toggleShowUserMenu: Function
 }
 
-export default class Toolbar extends React.Component<Props> {
+export default class CourseToolbar extends React.Component<Props> {
   toolbarRoot: HTMLElement | null
   toolbar: Object
 
@@ -32,12 +30,6 @@ export default class Toolbar extends React.Component<Props> {
     }
   }
 
-  toggleShowDrawer = (e: Event) => {
-    const { toggleShowDrawer } = this.props
-    e.preventDefault()
-    toggleShowDrawer()
-  }
-
   render() {
     const { toggleShowUserMenu, showUserMenu, profile } = this.props
 
@@ -46,19 +38,14 @@ export default class Toolbar extends React.Component<Props> {
         <header className="mdc-toolbar" ref={div => (this.toolbarRoot = div)}>
           <div className="mdc-toolbar__row">
             <section className="mdc-toolbar__section mdc-toolbar__section--align-start">
-              <HamburgerAndLogo onHamburgerClick={this.toggleShowDrawer} />
+              <MITLogoLink />
+              <span className="mdc-toolbar__title">
+                <Link to="/">MIT Open</Link>
+                {" | "}
+                <Link to={COURSE_URL}>Courses</Link>
+              </span>
             </section>
             <section className="mdc-toolbar__section mdc-toolbar__section--align-end user-menu-section">
-              {SETTINGS.allow_search ? (
-                <NavLink
-                  exact
-                  to={SITE_SEARCH_URL}
-                  activeClassName="active"
-                  className="search-link navy"
-                >
-                  <i className="material-icons">search</i>
-                </NavLink>
-              ) : null}
               <UserMenu
                 toggleShowUserMenu={toggleShowUserMenu}
                 showUserMenu={showUserMenu}
--- a/static/js/components/CourseToolbar.js
+++ b/static/js/components/CourseToolbar.js
@@ -17,11 +17,18 @@ type Props = {
 }
 
 export default class CourseToolbar extends React.Component<Props> {
-  toolbarRoot: HTMLElement | null
+  toolbarRoot: { current: null | React$ElementRef<typeof HTMLElement> }
   toolbar: Object
 
+  constructor(props: Props) {
+    super(props)
+    this.toolbarRoot = React.createRef()
+  }
+
   componentDidMount() {
-    this.toolbar = new MDCToolbar(this.toolbarRoot)
+    if (this.toolbarRoot.current) {
+      this.toolbar = new MDCToolbar(this.toolbarRoot.current)
+    }
   }
 
   componentWillUnmount() {
@@ -35,7 +42,7 @@ export default class CourseToolbar extends React.Component<Props> {
 
     return (
       <div className="navbar">
-        <header className="mdc-toolbar" ref={div => (this.toolbarRoot = div)}>
+        <header className="mdc-toolbar" ref={this.toolbarRoot}>
           <div className="mdc-toolbar__row">
             <section className="mdc-toolbar__section mdc-toolbar__section--align-start">
               <MITLogoLink />
--- a/static/js/components/Toolbar.js
+++ b/static/js/components/Toolbar.js
@@ -19,11 +19,18 @@ type Props = {
 }
 
 export default class Toolbar extends React.Component<Props> {
-  toolbarRoot: HTMLElement | null
+  toolbarRoot: { current: null | React$ElementRef<typeof HTMLElement> }
   toolbar: Object
 
+  constructor(props: Props) {
+    super(props)
+    this.toolbarRoot = React.createRef()
+  }
+
   componentDidMount() {
-    this.toolbar = new MDCToolbar(this.toolbarRoot)
+    if (this.toolbarRoot.current) {
+      this.toolbar = new MDCToolbar(this.toolbarRoot.current)
+    }
   }
 
   componentWillUnmount() {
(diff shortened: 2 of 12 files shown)


static/js/containers/App.js
```diff
@@ -177,18 +178,30 @@ class App extends React.Component<Props> {
...
+            showUserMenu={showUserMenu}
+            profile={profile}
+          />
+        ) : (
+          <React.Fragment>
+            <Toolbar
+              toggleShowDrawer={this.toggleShowDrawer}
+              toggleShowUserMenu={this.toggleShowUserMenu}
+              showUserMenu={showUserMenu}
+              profile={profile}
```
aliceriot: why not just pass the `isCourseUrl` boolean into `Toolbar` and conditionally render things in there? I think you'll end up writing less code that way

mbertrand:
@ferdi @pdpinch screenshot is in the PR description

static/js/components/Toolbar.js
```diff
@@ -39,17 +40,41 @@ export default class Toolbar extends React.Component<Props> {
...
+                    />
+                  </a>
+                  <span className="mdc-toolbar__title">
+                    <Link to="/">MIT Open</Link>
+                    {" | "}
+                    <Link to={coursesURL()}>Courses</Link>
+                  </span>
+                </React.Fragment>
+              ) : (
+                <HamburgerAndLogo onHamburgerClick={this.toggleShowDrawer} />
```
rhysyngsun: I think we should refactor this `Toolbar` into an HOC since the logic is starting to get complicated. These variants of the `Toolbar` (e.g. one for the course page and one for the rest of the site), would be the two concrete versions of it, similar to what we're already doing with the `Sidebar` components.

static/js/containers/App.js
```diff
@@ -177,6 +177,8 @@ class App extends React.Component<Props> {
       return <Redirect to={AUTH_REQUIRED_URL} />
     }
 
+    const isCourseUrl = pathname.startsWith(`/courses`)
```
rhysyngsun: `react-router` should be able to handle this kind of logic, I think also because we're now starting to really branch out the UI between channel-related UI and non-channel UI, we should start trying to address some of this now before it necessitates an even bigger refactor. You could then do a `<Switch>` with the first route being for `/courses` and the second one being a default match for the rest of the site and put the corresponding toolbars I described above in each. `<Drawer/>` below would also go in the default match and then you don't need this kind of logic here anymore.

static/js/lib/url.js
```diff
@@ -71,6 +71,7 @@ export const postPermalink = (post: Post): string =>
   absolutizeURL(postDetailURL(post.channel_name, post.id, post.slug))
 
 export const siteSearchURL = (): string => "/search/"
+export const coursesURL = (): string => "/courses/"
```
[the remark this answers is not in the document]
mbertrand: Should I change `siteSearchURL` while I'm at it?  I didn't write it as a function initially until I noticed that a function is used for siteSearchURL, so I thought I should try to be consistent with that.

pdpinch:
> @pdpinch any concerns about the fact that removal of the drawer means there is no copyright message or links to "contact us", TOS, or community guidelines for course pages?

We need to rethink the left hand nav in this context and it's easier for now to remove it than figure it all out. When we add it back, it should have copyright and related links.

static/js/components/Toolbar.js
mbertrand: I initially started this as a separate `CourseToolbar` component, @aliceriot [suggested modifying the existing `Toolbar` instead with a `isCourseURL` property](https://github.com/mitodl/open-discussions/pull/1913#discussion_r263889003) to determine what should be rendered, so I rewrote it that way, which does result in less code, especially taking unit test files into account, but is arguably maybe a tiny bit more complex.  Personally I'm fine with either approach, they all get the job done.  @aliceriot what do you think of an HOC toolbar with two versions?

aliceriot: I'm not a huge fan of two different HOCs as a solution.

I think one way to implement this that would be cleaner than passing in the `isCourseURL` property or using two different HOCs would be to have a `<Switch>` component with routes in the relevent parts of the `Toolbar` component itself. Then the `<Route>` components there would control whether the course-specific or rest-of-the-site-specific UI is rendered, and the external API of the `Toolbar` component wouldn't need to change.

I recently refactored the channel header rendering along this sort of lines. Previously there was a `withChannelHeader` HOC that added the component to a page with an indirect method of specifying what the children should be. I instead changed it so that the `ChannelHeader` is rendered for all routes under `/c/`, and then used `<Routes>` in the `ChannelHeader`'s children to control what was rendered inside of it. I thought that was a cleaner approach that leveraged react-router more and made the code easier to follow.

mbertrand: I don't see those changes involving `<Route>s` to `ChannelHeader` or it's children in master, but I gave a shot at implementing it that way as I understand it.  Seems kind of overly complex, requiring two `Switch` tags:
```python
    return (
      <div className="navbar">
        <header className="mdc-toolbar" ref={div => (this.toolbarRoot = div)}>
          <div className="mdc-toolbar__row">
            <section className="mdc-toolbar__section mdc-toolbar__section--align-start">
              <Switch>
                <Route
                  path={`${match.url}courses/`}
                  render={() => (
                    <React.Fragment>
                      <MITLogoLink />
                      <span className="mdc-toolbar__title">
                        <Link to="/">MIT Open</Link>
                        {" | "}
                        <Link to={COURSE_URL}>Courses</Link>
                      </span>
                    </React.Fragment>
                  )}
                />
                <Route
                  render={() => (
                    <HamburgerAndLogo
                      onHamburgerClick={this.toggleShowDrawer}
                    />
                  )}
                />
              </Switch>
            </section>
            <section className="mdc-toolbar__section mdc-toolbar__section--align-end user-menu-section">
              <Switch>
                <Route path={`${match.url}courses/`} />
                <Route
                  render={() =>
                    SETTINGS.allow_search ? (
                      <NavLink
                        exact
                        to={SITE_SEARCH_URL}
                        activeClassName="active"
                        className="search-link navy"
                      >
                        <i className="material-icons">search</i>
                      </NavLink>
                    ) : null
                  }
                />
              </Switch>
              <UserMenu
                toggleShowUserMenu={toggleShowUserMenu}
                showUserMenu={showUserMenu}
                profile={profile}
              />
            </section>
          </div>
        </header>
      </div>
```

At this point I'm inclined to use a hybrid of all these approaches: revert back to having two separate toolbar components (`Toolbar` and `CourseToolbar`), minus the HOC (also seems to add unnecessary complexity - still need separate toolbar components, plus the HOC), and use `Switch->Route` in `App.js` to determine which toolbar to use:

```python
        <Switch>
          <Route path={`${match.url}courses/`}>
            <CourseToolbar
              toggleShowUserMenu={this.toggleShowUserMenu}
              showUserMenu={showUserMenu}
              profile={profile}
            />
          </Route>
          <Route render={() => (
            <React.Fragment>
              <Toolbar
                toggleShowDrawer={this.toggleShowDrawer}
                toggleShowUserMenu={this.toggleShowUserMenu}
                showUserMenu={showUserMenu}
                profile={profile}
              />
              <Drawer/>
            </React.Fragment>
          )} />
```

aliceriot: yeah, that seems like a reasonable approach to me

static/js/lib/url.js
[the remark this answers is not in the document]
mbertrand: Changed both (`coursesURL()` -> `COURSE_URL`, `siteSearchURL()` -> `SITE_SEARCH_URL`)

static/js/components/Toolbar.js
```diff
@@ -39,17 +40,41 @@ export default class Toolbar extends React.Component<Props> {
...
 
     return (
       <div className="navbar">
         <header className="mdc-toolbar" ref={div => (this.toolbarRoot = div)}>
           <div className="mdc-toolbar__row">
             <section className="mdc-toolbar__section mdc-toolbar__section--align-start">
-              <HamburgerAndLogo onHamburgerClick={this.toggleShowDrawer} />
+              {isCourseUrl ? (
+                <React.Fragment>
+                  <a href="http://www.mit.edu" className="mitlogo">
```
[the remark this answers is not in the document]
mbertrand: Added a `MITLogoLink` component used by both toolbars

static/js/containers/App.js
mbertrand: Modified `<Switch><Route>` in `App.js` to determine what to render (`Toolbar` or `CourseToolbar`, `Drawer` or `null`)

static/js/components/Toolbar.js
rhysyngsun: Sounds good to me

static/js/components/CourseToolbar.js
```diff
@@ -0,0 +1,60 @@
...
+      this.toolbar.destroy()
+    }
+  }
+
+  render() {
+    const { toggleShowUserMenu, showUserMenu, profile } = this.props
+
+    return (
+      <div className="navbar">
+        <header className="mdc-toolbar" ref={div => (this.toolbarRoot = div)}>
```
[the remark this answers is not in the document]
mbertrand: Sure, just committed this change.

PR #1916, opened 2019-03-12
Buttons to clear facets

### LINKED ISSUE #1887: Course Seach: Subject area facet

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes #1887

#### What's this PR do?
Adds buttons below the text input field for each active facet filter, along with a 'Clear all filters' link.
Each button should remove the specified facet from the search filter.
The 'Clear all filters' link should remove all facets and text input from the search.

- Run a search with some facets selected and some search text.
- There should be buttons below the text input field for each selected facet.
- Click the 'x' on some of these buttons.  The appropriate facets should be unchecked and the search results should update.
- Click 'Clear all filters'.  All the facets should be unchecked and the text input should be cleared.  Search results should update accordingly.

Screen Shot 2019-03-12 at 4 55 36 PM

Screen Shot 2019-03-12 at 4 56 16 PM

### DIFF
--- a/static/js/containers/CourseSearchPage.js
+++ b/static/js/containers/CourseSearchPage.js
@@ -14,6 +14,7 @@ import Card from "../components/Card"
 import { Cell, Grid } from "../components/Grid"
 import { Loading, PostLoading } from "../components/Loading"
 import SearchFacet from "../components/SearchFacet"
+import SearchFilter from "../components/SearchFilter"
 import SearchTextbox from "../components/SearchTextbox"
 import SearchResult from "../components/SearchResult"
 import { actions } from "../actions"
@@ -29,7 +30,6 @@ import type {
   Result,
   FacetResult
 } from "../flow/searchTypes"
-import SearchFilter from "../components/SearchFilter"
 
 type OwnProps = {|
   dispatch: Dispatch<any>,
--- a/static/js/containers/CourseSearchPage.js
+++ b/static/js/containers/CourseSearchPage.js
@@ -69,6 +69,12 @@ type State = {
   error: ?string
 }
 
+const facetDisplayMap = [
+  ["topics", "Subject Area", null],
+  ["availability", "Availability", null],
+  ["platform", "Platform", _.upperCase]
+]
+
 const shouldRunSearch = R.complement(
   R.allPass([R.eqProps("text"), R.eqProps("activeFacets")])
 )
@@ -214,11 +220,6 @@ export class CourseSearchPage extends React.Component<Props, State> {
   render() {
     const { match, facets, total, loaded, processing } = this.props
     const { text, error, activeFacets } = this.state
-    const facetDisplayMap = [
-      ["topics", "Subject Area", null],
-      ["availability", "Availability", null],
-      ["platform", "Platform", _.upperCase]
-    ]
 
     return (
       <React.Fragment>
--- a/static/js/containers/CourseSearchPage.js
+++ b/static/js/containers/CourseSearchPage.js
@@ -29,6 +29,7 @@ import type {
   Result,
   FacetResult
 } from "../flow/searchTypes"
+import SearchFilter from "../components/SearchFilter"
 
 type OwnProps = {|
   dispatch: Dispatch<any>,
@@ -43,7 +44,7 @@ type OwnProps = {|
 type StateProps = {|
   initialLoad: boolean,
   results: Array<Result>,
-  facets: Map<string, ?FacetResult>,
+  facets: Map<string, FacetResult>,
   loaded: boolean,
   processing: boolean,
   total: number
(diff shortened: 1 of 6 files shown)


mbertrand:
@ferdi @pdpinch I've styled these to look about the same as in Micromasters for now.

mbertrand:
@ferdi @pdpinch the color for these is the same as in Micromasters, but I feel it's a bit hard to distinguish from the background.  What do you think?

Ferdi:
Looks great

static/js/containers/CourseSearchPage.js
```diff
@@ -62,33 +63,28 @@ type Props = {|
...
-      topics:         qs.parse(props.location.search).t,
-      platforms:      qs.parse(props.location.search).p,
-      availabilities: qs.parse(props.location.search).a,
-      from:           0,
-      error:          null
+      text:         qs.parse(props.location.search).q,
+      activeFacets: new Map([
+        ["platform", toArray(qs.parse(props.location.search).p) || []],
+        ["topics", toArray(qs.parse(props.location.search).t) || []],
+        ["availability", toArray(qs.parse(props.location.search).a) || []]
```
aliceriot: you might not want to address this now, but I was just wondering: would it make sense to store the selected options in the different facets (platform, topic, etc) as a `Set` instead of an `Array`? We can achieve the same thing roughly I suppose by using the `_.union` function and so on, but using a `Set` might make it more clear how it's supposed to be used, and make it so that we don't have to worry about de-duping and so on ourselves.

aliceriot: then `activeFacets` could be a `Map<string, Set<string>>`

static/js/containers/CourseSearchPage.js
```diff
@@ -216,7 +213,12 @@ export class CourseSearchPage extends React.Component<Props, State> {
 
   render() {
     const { match, facets, total, loaded, processing } = this.props
-    const { text, error, topics, platforms, availabilities } = this.state
+    const { text, error, activeFacets } = this.state
+    const facetDisplayMap = [
+      ["topics", "Subject Area", null],
+      ["availability", "Availability", _.capitalize],
+      ["platform", "Platform", _.upperCase]
+    ]
```
aliceriot: unless this object is going to be mutated at runtime it would be good to pull it out as a separate constant (so we don't redeclare it on every render call)

static/js/containers/CourseSearchPage.js
```diff
@@ -234,6 +236,27 @@ export class CourseSearchPage extends React.Component<Props, State> {
                 validation={error}
               />
             </div>
+            <div className="search-filters-row">
+              {facetDisplayMap.map(([name, title, labelFunction]) =>
+                (activeFacets.get(name) || []).map((facet, i) => (
```
aliceriot: do we need the `|| []` check here? just wondering because in the constructor we `setState` to always have a `[]` for these values, so we shouldn't run into a case where it's `undefined` or something (as far as I can tell anyhow, haha)

static/js/containers/CourseSearchPage.js
mbertrand: Good idea, I'll do that

static/js/containers/CourseSearchPage.js
mbertrand: ✔️

static/js/containers/CourseSearchPage.js
mbertrand: This only here to prevent flow from complaining about it, I can replace with `$FlowFixMe` instead if you prefer.

static/js/containers/CourseSearchPage.js
mbertrand: Actually that seems to cause issues with using `.map(...)`, maybe best to leave as is for now.

mbertrand: I just added another use of `_.union()` to ensure that there are no dupes, if you enter a URL of `../courses/search?t=Engineering&t=Engineering&type=course`, the duplicate `t=Engineering` parameter will be removed.

mbertrand:
@aliceriot comments addressed

static/js/containers/CourseSearchPage.js
aliceriot: ah, yeah I think `.get` always has a return type of `?value` so it makes sense flow would complain :/

static/js/containers/CourseSearchPage.js
aliceriot: yeah `Set` object do not have a `.map` method, if you want to map across them you need to first turn them into an array by doing something like `[...set]`. I don't really know why they don't have that method defined on them, seems like something you'd want to me but :woman_shrugging:

PR #1920, opened 2019-03-13
Show min price instead of max price for courses

### LINKED ISSUE #1901: Course search: when a course is auditable, don't show the price
### Steps to Reproduce

Search results that include MOOCs with verified certificates display the verified price instead of the audit price. This is misleading because the course is available for free. 

### Expected Behavior

For now, display "free" similar to OCW courses. (We may qualify this in some way in the future with a tooltip or similar)

### Screenshot or Screencast

image

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #1901

#### What's this PR do?
Shows the minimum price ("Free" if 0) instead of the maximum price.

- Search for courses.  Almost all of them should show "Free" for price in the compact and expanded course views.
- Search for "Basic Circuit Analysis" with the "Current" Availability facet checked.  The minimum price for one of the top results should be $9.99

The `maxPrice` function in `static/js/lib/courses.js` is no longer used, but I left it there in case we decide in the future to show a price range instead.

### DIFF
--- a/static/js/lib/courses.js
+++ b/static/js/lib/courses.js
@@ -24,3 +24,8 @@ export const maxPrice = (course: Course) => {
   const price = Math.max(...R.map(R.view(R.lensProp("price")), course.prices))
   return price > 0 ? `$${price}` : "Free"
 }
+
+export const minPrice = (course: Course) => {
+  const price = Math.min(...R.map(R.view(R.lensProp("price")), course.prices))
+  return price > 0 && price !== Infinity ? `$${price}` : "Free"
+}
--- a/static/js/lib/courses.js
+++ b/static/js/lib/courses.js
@@ -1,6 +1,5 @@
 //@flow
 import moment from "moment"
-import R from "ramda"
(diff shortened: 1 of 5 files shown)


static/js/lib/courses.js
```diff
@@ -24,3 +24,8 @@ export const maxPrice = (course: Course) => {
   const price = Math.max(...R.map(R.view(R.lensProp("price")), course.prices))
   return price > 0 ? `$${price}` : "Free"
 }
+
+export const minPrice = (course: Course) => {
+  const price = Math.min(...R.map(R.view(R.lensProp("price")), course.prices))
```
aliceriot: could this just be `Math.min(...course.prices.map(price => price.price))` or `Math.min(...course.prices.map(R.prop('price')))` ? might be just a little bit more succinct, unless there's some reason we need to use the lens

mbertrand: Thanks, done

PR #1922, opened 2019-03-14
Label course availability by model field instead of dates in UI

### LINKED ISSUE #1861: Some prior courses are missing end dates so are shown as "Available Now"
### Steps to Reproduce

Search courses with query "Advanced Introductory Classical Mechanics" - a bunch of results will return.  Most are from years past and are no longer available on MITX, but will be listed as "Available Now" because the courses have empty end dates.

### Expected Behavior
Not sure - Is it correct to assume that a course is still available if it has no end date, and this is a bad data issue?  Should end date be estimated as 4-5 months beyond the start date if missing?  

### Screenshot or Screencast
screen shot 2019-02-14 at 11 52 11 am

### LINKED ISSUE #1910: Update availability on the course detail view and search result cards
The availability shown on the course detail view and search result cards are still calculated from start and end dates, they should be changed to use the new `availability` field now that it has been added to the model.

Related to #1898

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #1910
Closes #1861

#### What's this PR do?
Displays course availability based on the `Course` model's availability field, rather than start and end dates.  In particular, changes the `courseAvailability` function and renames it to `availabilityLabel`.  This function will now replace "Archived" with "Prior", "Current" with "Available Now", and display other availabilities as is.

Go to `/courses/search`.  The search results and facet choices should show these availabilities: Available Now, Prior, Upcoming, Starting Soon.  Select the 'Available Now' or 'Prior' facet and verify that they still work correctly.

### DIFF
--- a/static/js/factories/search.js
+++ b/static/js/factories/search.js
@@ -4,6 +4,7 @@ import casual from "casual-browserify"
 import R from "ramda"
 
 import { LINK_TYPE_LINK, LINK_TYPE_TEXT } from "../lib/channels"
+import { COURSE_ARCHIVED, COURSE_CURRENT, platforms } from "../lib/constants"
 
 import type {
   CommentResult,
@@ -11,7 +12,6 @@ import type {
   PostResult,
   ProfileResult
 } from "../flow/searchTypes"
-import {COURSE_ARCHIVED, COURSE_CURRENT, platforms} from "../lib/constants"
 
 export const makeProfileResult = (): ProfileResult => ({
   author_avatar_medium: casual.url,
@@ -89,7 +89,8 @@ export const makeCourseResult = (): CourseResult => ({
   topics:            [casual.word, casual.word],
   prices:            [{ mode: "audit", price: casual.number }],
   object_type:       "course",
-  availability:      casual.random_element[COURSE_ARCHIVED, COURSE_CURRENT, "Upcoming"]
+  availability:
+    casual.random_element[(COURSE_ARCHIVED, COURSE_CURRENT, "Upcoming")]
 })
 
 export const makeSearchResult = (type: ?string) => {
--- a/static/js/factories/search.js
+++ b/static/js/factories/search.js
@@ -11,7 +11,7 @@ import type {
   PostResult,
   ProfileResult
 } from "../flow/searchTypes"
-import { platforms } from "../lib/constants"
+import {COURSE_ARCHIVED, COURSE_CURRENT, platforms} from "../lib/constants"
 
 export const makeProfileResult = (): ProfileResult => ({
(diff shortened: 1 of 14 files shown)


aliceriot:
functionality looks good!

static/js/factories/search.js
```diff
@@ -88,7 +88,9 @@ export const makeCourseResult = (): CourseResult => ({
   instructors:       ["instuctor 1", "instructor 2"],
   topics:            [casual.word, casual.word],
   prices:            [{ mode: "audit", price: casual.number }],
-  object_type:       "course"
+  object_type:       "course",
+  availability:
+    casual.random_element[(COURSE_ARCHIVED, COURSE_CURRENT, "Upcoming")]
```
aliceriot: would it make sense to declare `"Upcoming"` as a constant in `constants.js` as well?

static/js/lib/courses.js
```diff
@@ -1,23 +1,23 @@
...
-  course.platform === platforms.OCW
-    ? COURSE_AVAILABLE_NOW
-    : moment(course.start_date).isAfter(moment())
-      ? _.capitalize(COURSE_UPCOMING)
-      : moment(course.end_date).isBefore(moment())
-        ? _.capitalize(COURSE_PRIOR)
-        : COURSE_AVAILABLE_NOW
+import type { Course } from "../flow/discussionTypes"
+
+export const availabilityLabel = (availability: ?string) => {
```
aliceriot: just a question, for similar little helper functions for the course UI we generally have the signature `Course => string`, do you think it would make sense to keep the same API here for consistency? You could add a line right below this one like

```js
const { availability } = course
```

to grab the availability value out of the course object

static/js/factories/search.js
mbertrand: I had thought about it, but this is just for tests, and could be anything like "blahblah" for testing purposes, so decided against it.  Happy to change it if you prefer.

static/js/lib/courses.js
mbertrand: This was the original signature, but I changed it because it's now being used by `SearchFacet` as the `labelFunction`, which does not get passed the full `Course` object.

mbertrand: https://github.com/mitodl/open-discussions/pull/1922/files#diff-f3d2dc8b6c7c97d4acc6f60b8e663a61R77

aliceriot: ah ok, that makes sense

static/js/factories/search.js
aliceriot: not necessary, it just seemed like it was popping up in a few different places, haha

PR #1928, opened 2019-03-18
Modify facet behavior within/between groups

### LINKED ISSUE #1917: When selecting facets within a group, only update facet choices/counts in other groups
As a user, when selecting facets within "Availability" for example, the facet choices and counts within the "Platform" and "Subject area" groups should be updated to reflect the current results, but not within "Availability".  This will allow the user to select both "Current" and "Upcoming" for "Availability".  Currently, the user can only select one facet choice within "Availability" because the other choices disappear (since they are not in returned results).

#### Acceptance Criteria:

- [ ] When a user selects a facet within a group, the facet choices and counts are updated only for other groups.

### LINKED ISSUE #1923: course search page responsiveness needs improvement
### Steps to Reproduce

Just typing in a search term often leads to perceptible lag in the interface, because the search is updated and the UI is re-rendered on every keystroke. To me it's a noticeable lag locally (in development) and also on the RC environment.

### Expected Behavior

I think that we should debounce the search updates so that we only fire off the request and re-render when the user pauses for a second in typing (like a 250ms debounce would probably do it).

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Fixes #1917 
Fixes #1923

#### What's this PR do?
Adds a `currentFacetGroup` property to `CourseSearchPage` to track which group of facets was most recently changed and what facets are contained therein, along with a new `mergeFacetOptions` method to determine which facets should be sent to the `SearchFacet` component.

- Within the 'Availability' group, select "Starting Soon".  The other Availability facets should continue to display without changes in counts, but "OCW" should disappear from the "Platform" group.
- Select "Available Now" in the "Availability" group.  The other Availability facets should continue to display without changes in counts, and "OCW" should reappear in the "Platform" group.
- Choose both "OCW" and "MITx" in the "Platform" group.  Then unselect "Available Now" from "Availability".  "OCW" should continue to display but with a count of 0, because it is still a selected facet.
- Unselect "OCW" from Platform, it should continue to display.  Only "Starting Soon" should appear in "Availability".
- Unselect "Starting Soon".  4 facet choices should appear again for "Availability". 
- Unselect "MITX".  "OCW" should appear again for "Platform"
- Start typing in the text input field.  No new search should be performed until you click 'Enter' (#1923)

facets

### DIFF
--- a/static/js/containers/CourseSearchPage.js
+++ b/static/js/containers/CourseSearchPage.js
@@ -69,7 +69,8 @@ type State = {
   text: ?string,
   activeFacets: Map<string, Array<string>>,
   from: number,
-  error: ?string
+  error: ?string,
+  currentFacetGroup: ?Map<string, FacetResult>
 }
 
 const facetDisplayMap = [
@@ -78,9 +79,7 @@ const facetDisplayMap = [
   ["platform", "Platform", _.upperCase]
 ]
 
-const shouldRunSearch = R.complement(
-  R.allPass([R.eqProps("text"), R.eqProps("activeFacets")])
-)
+const shouldRunSearch = R.complement(R.eqProps("activeFacets"))
 
 export class CourseSearchPage extends React.Component<Props, State> {
   constructor(props: Props) {
@@ -95,8 +94,9 @@ export class CourseSearchPage extends React.Component<Props, State> {
           _.union(toArray(qs.parse(props.location.search).a) || [])
         ]
       ]),
-      from:  0,
-      error: null
+      from:              0,
+      error:             null,
+      currentFacetGroup: null
     }
   }
 
@@ -121,8 +121,43 @@ export class CourseSearchPage extends React.Component<Props, State> {
         ["platform", []],
         ["availability", []],
         ["topics", []]
-      ])
+      ]),
+      currentFacetGroup: null
     })
+  }
+
+  mergeFacetOptions = (group: string) => {
+    const { facets } = this.props
+    const { activeFacets, currentFacetGroup } = this.state
+    const resultFacets = facets ? facets.get(group) : null
+
+    // add any selected facets not in results to the group
+    if (resultFacets && resultFacets.buckets) {
+      // $FlowFixMe: this is not undefined
+      activeFacets.get(group).map(facet => {
+        if (!_.find(resultFacets.buckets, { key: facet })) {
+          resultFacets.buckets.push({
+            key:       facet,
+            doc_count: 0
+          })
+        }
+      })
+    }
+
+    // if this is the currentFacetGroup, add any new facets in that group from search results
+    if (currentFacetGroup && currentFacetGroup.keys().next().value === group) {
+      const mergedFacetGroup = _.clone(currentFacetGroup).get(group)
+      if (mergedFacetGroup && resultFacets) {
+        resultFacets.buckets.map(facet => {
+          if (!_.find(mergedFacetGroup.buckets, { key: facet.key })) {
+            mergedFacetGroup.buckets.push(facet)
+          }
+        })
+      }
+      return mergedFacetGroup
+    }
+
+    return resultFacets
   }
 
   loadMore = async () => {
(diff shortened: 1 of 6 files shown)


mbertrand:
@ferdi @pdpinch an animated screenshot is attached to the description

aliceriot:
I think the functionality looks good, generally if I'm using a single facet group the options aren't disappearing unexpectedly and it generally works how I would expect.

static/js/containers/CourseSearchPage.js
```diff
@@ -121,10 +121,45 @@ export class CourseSearchPage extends React.Component<Props, State> {
...
+          if (!_.find(mergedFacetGroup.buckets, { key: facet.key })) {
+            mergedFacetGroup.buckets.push(facet)
+          }
+        })
+      }
+      return mergedFacetGroup
+    }
+
+    return resultFacets
+  }
```
aliceriot: this logic to store the facet options in the `currentFacetGroup` value and merge with the changed facets on `this.props` strikes me overall as pretty complicated and hard to follow. Do you think there are any opportunities to simplify it? I was wondering in particular about a utility function for merging these `FacetResults` objects. If you had a `mergeFacets` function which did that work for you, then I think this function could look something like:


```js
if (!facets) {
  return null
}

const emptyActiveFacets = { buckets: activeFacets.map(facet => ({ key: facet, doc_count: 0 })) }

if (currentFacetGroup.get(group)) {
  return mergeFacetResults(
    facets.get(group),
    emptyActiveFacets,
    currentFacetGroup.get(group)
  )
} else {
  return mergeFacetResults(
    facets.get(group),
    emptyActiveFacets
  )
}
```
 
IMO it would be easier to read and follow the code this way, and there wouldn't be this duplicated logic to merge the new result facets with the active facets and the 'current' facets.

I think a merge function could look something like this:

```js
const mergeFacetResults = (...args: Array<FacetResult>) => ({
  buckets: args
    .map(R.prop("buckets"))
    .reduce((buckets, acc) => R.unionWith(R.eqBy(R.prop("key")), buckets, acc))
})
```
assuming I've understood the data structure :smile:

mbertrand:
@aliceriot thanks, I made your recommended changes

static/js/containers/CourseSearchPage.js
```diff
@@ -222,12 +219,15 @@ export class CourseSearchPage extends React.Component<Props, State> {
...
     const updatedFacetGroup =
-      currentFacetGroup && currentFacetGroup.get(name)
+      currentFacetGroup && currentFacetGroup.group === name
         ? currentFacetGroup
-        : new Map([[name, facets.get(name)]])
+        : {
+          group:  name,
+          result: facets.get(name)
+        }
+    // $FlowFixMe: nothing undefined here
```
aliceriot: I think there is an issue in the logic here. In particular, if we have the case that `currentFacetGroup.group === name` I don't think we want to just set `updatedFacetGroup` to be `currentFacetGroup`. It is possible, for instance, that there are new options for the facet present in `facets.get(name)` which aren't saved in `currentFacetGroup` it leads to the very behavior that this PR is trying to eliminate. Here's how I can reproduce the issue:

1. start with a full page refresh on `/courses/search`
2. select 'OCW' for platform
3. select 'Available Now' for 'availability'
4. de-select 'OCW' for platform'
5. de-select 'Available Now' for 'availability

From this point if I select and de-select 'Available Now' the options shown in the facet change - if I have nothing selected I see all the options, but selecting 'Available Now' causes the other options to go away so that I only see the selected 'Available Now' option.

Here's a gif that shows the issue:

issue


This is happening because, in the above scenario, when I de-select 'Available Now' the `currentFacetGroup` is set to be the 'availability' facet with just the 'available now' option present. Then the search is run (fired by the `componentDidUpdate` hook) and more facet values come back in the response, since at this point I have no facets selected and all values are returned. This means that in the `mergeFacetOptions` method when we call 

```js
mergeFacetResults(currentFacetGroup.result, emptyActiveFacets, facets.get(group))
```

we're going to get all options displaying in the UI because they're all present in `facets.get(group)`. But then, if I re-select 'Available Now' then `currentFacetGroup` isn't updated, because `currentFacetGroup.group === name` when this `toggleFacet` method is called. So then when we call `mergeFacetOptions` to render the `availability` facet we only get the 'Available Now' option in the UI, because when we make the above call to `mergeFacetOptions` we have got only the 'Available Now' option in both `currentFacetGroup.result` _and_ in `facets.get(group)`. 

Does that make sense?

I believe this can be address by changing the `toggleFacet` method to merge `facets.get(name)` with `currentFacetGroup.results` when `currentFacetGroup.group === name`, something like this:


```js
this.setState({
  activeFacets:      updatedFacets,
  currentFacetGroup: {
    group:  name,
    result:
      currentFacetGroup && currentFacetGroup.group === name
        ? mergeFacetResults(currentFacetGroup.result, facets.get(name))
        : facets.get(name)
  }
})
```

static/js/lib/search.js
```diff
@@ -247,3 +248,9 @@ export const buildSearchQuery = ({
   }
   return builder.build()
 }
+
+export const mergeFacetResults = (...args: Array<FacetResult>) => ({
+  buckets: args
+    .map(R.prop("buckets"))
+    .reduce((buckets, acc) => R.unionWith(R.eqBy(R.prop("key")), buckets, acc))
```
aliceriot: I think it will work ok without it, but you probably want to add a `[]` here as the starting value for the `.reduce` call

static/js/containers/CourseSearchPage.js
mbertrand: Thanks, updated

static/js/lib/search.js
mbertrand: Do you mean this: `.reduce((buckets, acc) => R.unionWith(R.eqBy(R.prop("key")), [], buckets, acc))`? That led to some strange behavior so I left it as is.

mbertrand:
Also updated `componentDidMount` to always `runSearch()`, to handle a problem with it not being triggered when clicking the back button in the browser from another URL to return to that page.

static/js/lib/search.js
aliceriot: ah sorry no, `.reduce` takes two arguments, a function and a value that will be the value of `acc` the first time the function is called. I was just suggesting adding `[]` as the initial value for clarity, but it's optional so feel free to leave as-is.

PR #1937, opened 2019-03-29
Addresses issue #1930

### LINKED ISSUE #1930: Omit MITx "DELETE" courses from course database
### Steps to Reproduce

1. Go to course search, populated with all the courses in the MITx feed. 
2. Search on DELETE
3. A number of "junk" courses are in the search results. 

### Expected Behavior

No courses with a title of DELETE should show up in the results. Just an OCW course on Syntax. 

### Actual Behavior

I see six "DELETE" courses in the results. 

### Screenshot or Screencast

image

kyfantaz:
#### Pre-Flight checklist
**As part of merging this PR, an manual action should be performed: 
Remove the Course instances for the six courses of interest as well as their Elasticsearch index objects. (alternatively, set their instance Course.published to False)**

Closes #1930 

#### What's this PR do?
This PR prevent "deleted" Courses by being ingested again at the next data sync with a code change to course_catalog.task_helpers

To test, after the code change, we can perform a sync and see if the courses of interest get synced again.

### DIFF
--- a/course_catalog/task_helpers.py
+++ b/course_catalog/task_helpers.py
@@ -61,6 +61,10 @@ def parse_mitx_json_data(course_data, force_overwrite=False):
 
     # Parse each course run individually
     for course_run in course_data.get("course_runs"):
+
+        if should_skip_course(course_run.get("title")):
+            continue
+
         course_run_key = course_run.get("key")
 
         # Get the last modified date from the course run
@@ -459,3 +463,27 @@ def get_course_availability(course):
         for run in course_runs:
             if run.get("key") == course.course_id:
                 return run.get("availability")
+
+
+def should_skip_course(course_title):
+    """
+    Returns True if '[delete]', 'delete ' (note the ending space character)
+    exists in a course's title or if the course title equals 'delete' for the
+    purpose of skipping the course
+
+    Args:
+        course_title (str): The course.title of the course
+
+    Returns:
+        bool
+
+    """
+    title = course_title.strip().lower()
+    if (
+        "[delete]" in title
+        or "(delete)" in title
+        or "delete " in title
+        or title == "delete"
+    ):
+        return True
+    return False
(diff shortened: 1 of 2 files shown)


course_catalog/task_helpers.py
```diff
@@ -459,3 +463,22 @@ def get_course_availability(course):
...
+
+    Args:
+        course_title (str): The course.title of the course
+
+    Returns:
+        bool
+
+    """
+    title = course_title.lower()
+    if ("[delete]" in title) or ("delete " in title) or (title == "delete"):
```
mbertrand: This works perfectly with the current data set.  But is it possible that in the future there may be a course with a slightly different title like "(Delete)"?  Or maybe a valid course with a title like "[Delete hidden viruses](https://www.udemy.com/how-to-delete-hidden-virus/)"?  @pdpinch if you think that is unlikely then I'll approve this PR.

kyfantaz: I'd be happy to change the rules as you see fit, including adding `(delete)` and/or a more broad clause of `"delete" in title`. I can also add a whitespace strip `.strip()` to strip whitespace at the ends for edge cases that may include that.

pdpinch: Honestly, it's impossible to predict. I don't really know why these courses appear in the feed at all. Does this implementation catch the course whose title starts with "[DELETE]"?

kyfantaz: It does. It's also one of the test cases.

kyfantaz: A note on my decision: I'm opting to lean towards conservatism. I looked into this and the substring "delete" doesn't really show up elsewhere in English language other than in words related to or with the suffix of "delete". Despite that, I wanted to avoid using the board clause of "delete" in string, which would look for the substring "delete" anywhere in the title string. That is, it would match "deleterious" if that word happened to be in the title. So we have a rule that mostly captures the only cases we're seeing and variations on them.

pdpinch: Leaning towards conservatism is the right choice. 👍

PR #1953, opened 2019-04-17
Added redirect for handling themove.mit.edu

blarghmatey:
N/A

#### What's this PR do?
Adds a redirect for themove.mit.edu to a discussions channel

Testing locally, add an `/etc/hosts` entry for the app that points themove.mit.edu at localhost


shaidar:
👍

PR #1957, opened 2019-04-18
Repair posts if they don't appear in the hot posts list

### LINKED ISSUE #1955: Async task to ensure posts appear in hot listing
Some posts occasionally don't appear in the hot listings, our current best guess is that reddit's cached queries in cassandra are failing to update on occasion. We have a current manual workaround which is to upvote the post at which point the cache appears to be regenerated and the post appears in the hot listing. The easiest way to fix this would be to automate this repair process via a celery task:

- Run a celery task for the post id that checks the hot posts for that subreddit and ensures the post is present
- If the post is present, complete the task
- If the post is not present, execute these operations as the indexing user:
  - `submission.upvote()` followed by `submission.clear_vote()`
  - Recheck the top posts list:
    - If the post is present, complete the task
    - If the post is not present, retry the task after some configurable interval (e.g. 30s), only retry up to some configurable limit at which point raise an error

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings

Fixes #1955 

#### What's this PR do?
Repairs posts that don't show up in the hot listing by having the indexing user upvote then clear the vote.

Create a post, ensure you don't get any errors. It's impractical to reproduce the actual production scenario so this fix is largely based on using existing known-to-work api methods.


channels/tasks.py
```diff
@@ -376,3 +378,53 @@ def send_invitation_email(channel_invitation_id):
...
+        new_posts_ids = set(new_posts_by_id.keys())
+
+        hot_posts = admin_api.list_posts(channel_name, SORT_HOT_LISTING_PARAMS)
+        hot_posts_ids = {post.id for post in islice(hot_posts, limit)}
+
+        missing_post_ids = new_posts_ids - hot_posts_ids
+
+        print(new_posts_ids)
+        print(hot_posts_ids)
+        print(missing_post_ids)
```
aliceriot: were these `print` statements left in unintentionally?

channels/tasks.py
```diff
@@ -376,3 +378,53 @@ def send_invitation_email(channel_invitation_id):
...
+    submission.upvote()
+    submission.clear_vote()
+
+    # check again and if it's missing still, just report an error so we can investigate
+    if _find_missing_submission() is not None:
+        log.error(
+            "Failed to repair submission %s missing from hot posts in channel %s",
+            reddit_post_id,
+            channel_name,
+        )
```
aliceriot: so `_find_missing_submission` is going to look at the first page of `new` and `hot` submissions and will return the submission if a post with the id we're looking for is in `new` but not in `hot`. this makes sense to me, it tells us generally if we need to repair it or not. I think it's possible though that the post is on the second or third page of the `hot` results, so we might end up repairing a post in a situation where we dont absolutely need to, but that's ok I think.

I'm wondering though about here where we use `_find_missing_submission` again to check whether the repair was successful. In the case where the repair was successful but the post is on the second page of the `hot` listings, wouldn't `_find_missing_submission` still return a post object, causing us to raise this error incorrectly?

admittedly, right now we might not have enough traffic on any of our channels that this is a concern, but still, what do you think? I'm not sure there is another good way to exhaustively check that the repair was successful, without trawling through every page of the `hot` listing to exhaustively check if it's _somewhere_ in there.

rhysyngsun: Yeah, it's possible that it may get pushed onto the second page, but it's unlikely. We'd probably need to be getting some combination of tens of posts per minute and hundreds of upvotes per minute. This fix definitely falls in the category of "stopgap" and that's also why I put in a configuration for how many posts back into the history to check because that will allow us to increase that (at the expense of site performance).

aliceriot: makes sense, I think if we are ever dealing with lots of traffic like that we will hopefully already have migrated away from reddit anyhow...

Issue #1972, opened 2019-04-24
Upload OCW master json to S3


sesterki:
OCW would like to use the master json files generated during course import for integration with other services. 

Update course catalog to upload the OCW master_json file to S3.

PR #1973, opened 2019-04-24
Zagaran/upload master json

sesterki:
#1972 

#### What's this PR do?
Updates requirements to use ocw-data-parser 0.2.6, and modifies course catalog to take advantage of the upload master json feature.

Use course catalog to process courses and then check that files with the name [uid]_master.json show up in the S3 bucket.

### DIFF
--- a/course_catalog/tasks.py
+++ b/course_catalog/tasks.py
@@ -10,6 +10,7 @@ from django.conf import settings
 from ocw_data_parser import OCWParser
 
 from open_discussions.celery import app
+from course_catalog.constants import PlatformType
 from course_catalog.models import Course
 from course_catalog.task_helpers import (
     get_access_token,
@@ -172,7 +173,7 @@ def get_ocw_data(upload_to_s3=True):  # pylint:disable=too-many-locals,too-many-
                 if settings.OCW_UPLOAD_IMAGE_ONLY:
                     parser.upload_course_image()
                 else:
-                    parser.upload_all_media_to_s3()
+                    parser.upload_all_media_to_s3(upload_master_json=True)
 
             digest_ocw_course(
                 parser.get_master_json(),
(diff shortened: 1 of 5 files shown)


course_catalog/tasks_test.py
```diff
@@ -206,6 +206,17 @@ def test_get_ocw_data(settings, mock_course_index_functions):
...
+        "s3",
+        aws_access_key_id=settings.OCW_LEARNING_COURSE_BUCKET_NAME,
+        aws_secret_access_key=settings.OCW_LEARNING_COURSE_ACCESS_KEY,
+    )
+    # The filename was pulled from the uid 1.json in the TEST_JSON_PATH files.
+    with pytest.raises(botocore.exceptions.ClientError):
+        s3.Object(
+            settings.OCW_LEARNING_COURSE_BUCKET_NAME,
+            "16197636c270e1ab179fbc9a56c72787_master.json",
+        ).load()
```
mbertrand: This just seems to be testing that an error is raised when trying to retrieve an object from S3?  I think it would be better if possible to mock `OCWParser.upload_all_media_to_s3` and verify that it is called with argument `upload_master_json=True`

course_catalog/tasks.py
```diff
@@ -172,7 +172,7 @@ def get_ocw_data(upload_to_s3=True):  # pylint:disable=too-many-locals,too-many-
                 if settings.OCW_UPLOAD_IMAGE_ONLY:
                     parser.upload_course_image()
                 else:
-                    parser.upload_all_media_to_s3()
+                    parser.upload_all_media_to_s3(upload_master_json=True)
```
mbertrand: I think this will only be run for courses that have changed since the last check or are have not yet been imported.  I could manually delete all the OCW courses, but it might be useful to add a `force_overwrite` parameter to this function, that if True would run through all courses even if they haven't changed.  But either way, all the media would have to be uploaded again.  Another alternative is to write a django management command to upload the `raw_json` field of each existing OCW course to S3 as the master_json file, assuming it is the same content.  @pdpinch what do you think?

pdpinch: I do need all the files, so I'll need someway to recreate them. A management command seems most convenient to me, but I'm flexible.

sesterki: I think a management command makes sense in this case. Should this command check if the file already exists or just naively redo everything every time (though the command may only ever be needed once)?

mbertrand: I'm fine with naively redoing all of them, since it will probably just be run once when none exist.

sesterki: Should I do that as part of this PR or as a separate one?

mbertrand: You can add it in this PR

sesterki: I have added the management command.

course_catalog/tasks_test.py
sesterki: We are now properly testing that the file was uploaded.

sesterki:
I am calculating the s3 folder from `course.url` although I'm not sure this is 100% foolproof. The OCW parser uses `course_prefix` which comes from the s3 bucket the original files are stored in, while `url` comes from the actual json. There is a correlation though.

PR #1984, opened 2019-04-26
Fix RSS widgets for invalid urls and add admin ui for them

### LINKED ISSUE #1982: RSS widget error crashes the entire widget list for a channel
ValueError: I/O operation on closed file.

https://sentry.io/organizations/mit-office-of-digital-learning/issues/978826391/?project=216201&referrer=github_plugin

```
ValueError: I/O operation on closed file.
(21 additional frame(s) were not displayed)
...
  File "rest_framework/serializers.py", line 560, in data
    ret = super(Serializer, self).data
  File "rest_framework/serializers.py", line 262, in data
    self._data = self.to_representation(self.instance)
  File "rest_framework/serializers.py", line 527, in to_representation
    ret[field.field_name] = field.to_representation(attribute)
  File "rest_framework/fields.py", line 1893, in to_representation
    return method(value)
  File "widgets/serializers/rss.py", line 59, in get_json
    rss = _fetch_rss(instance.configuration["url"])

ValueError: I/O operation on closed file.
```

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1982 

#### What's this PR do?
- Fixes the RSS widget so it doesn't cause the entire API to fail if an invalid RSS url is input or the url becomes invalid after a successful save.
- Adds `WidgetList`and friends to django admin

- Verify django admin works within reason (`position` needs to be maintained manually and there is no validation of the configuration)
- Verify that if you enter an invalid url (through django admin) that the widget list returns, but with an empty list.

### DIFF
--- a/widgets/serializers/rss.py
+++ b/widgets/serializers/rss.py
@@ -5,6 +5,7 @@ import time
 from cache_memoize import cache_memoize
 from django.conf import settings
 import feedparser
+from rest_framework.serializers import ValidationError
 
 from widgets.serializers.widget_instance import (
     WidgetConfigSerializer,
@@ -56,8 +57,15 @@ class RssFeedWidgetSerializer(WidgetInstanceSerializer):
 
     def get_json(self, instance):
         """Obtains RSS feed data which will then be provided to the React component"""
-        rss = _fetch_rss(instance.configuration["url"])
-        entries = getattr(rss, "entries", [])
+        try:
+            rss = _fetch_rss(instance.configuration["url"])
+            entries = getattr(rss, "entries", [])
+        except:  # pylint: disable=bare-except
+            # log an exception and coerce this to an empty list so the feed and UI don't crash
+            log.exception(
+                "Error trying to refresh cached RSS feed for widget id: %s", instance.id
+            )
+            entries = []
 
         timestamp_key = (
             "published_parsed"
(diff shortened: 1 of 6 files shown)


aliceriot:
not sure if I'm going about testing this the right way, but I did the following:

- verify that I can make a change to a random field on the widget in the admin
- changed a previously-working RSS widget url to a non-working value (I changed 'https' to 'httsdfp')
- saved the list

and now when I try to load the channel corresponding to that widget list I get a bunch of errors on startup:

validationerrorthing

the error crashes the app.

rhysyngsun:
@aliceriot I can't reproduce that behavior, are you sure you don't have an invalidly configured url widget that's causing this since I think that's the only widget that actually uses `EmbedlyCard`?

aliceriot:
yeah, I think that's what was going on

aliceriot:
ok I sorted that out, can now confirm that this works!

widgets/serializers/rss.py
```diff
@@ -56,8 +57,13 @@ class RssFeedWidgetSerializer(WidgetInstanceSerializer):
...
     def get_json(self, instance):
         """Obtains RSS feed data which will then be provided to the React component"""
-        rss = _fetch_rss(instance.configuration["url"])
-        entries = getattr(rss, "entries", [])
+        try:
+            rss = _fetch_rss(instance.configuration["url"])
+            entries = getattr(rss, "entries", [])
+        except:  # pylint: disable=bare-except
+            # log an exception and coerce this to an empty list so the feed and UI don't crash
+            log.exception("Error trying to refresh cached RSS feed")
```
aliceriot: would it be good to put the objects ID in the log message here, so we can track it down easily if we see it in the logs?

rhysyngsun: Done, also updated the log message when we save to be more informative too

Issue #1976, opened 2019-04-25
include bootcamps in course search


pdpinch:
As a user, I'd like to find out about bootcamps when I search for courses

#### Designs and Mockups

tbd

#### Acceptance Criteria:

- [x] static json file of bootcamp metadata, from https://github.com/mitodl/bootcamps.json
- [x] periodically import bootcamps and add them to the database and search index
- [ ] bootcamps should show up in search results 

#### Out of Scope

- we should have similar metadata for bootcamps as other courses, so they should be able to use the same design. We may want to add some different branding, or change the facets later

pdpinch:
I've been thinking about this, and I think it would be better to merge the bootcamps.json file into this repo. It won't change very frequently and if it's part of this app, we'll have the possibility of running travis against any changes.

pdpinch:
closed via #2001

PR #2001, opened 2019-05-03
Parse bootcamps json data into Course model

sesterki:
#1976 

#### What's this PR do?
This PR adds a framework for the bootcamps.json data, as well as functions to parse that data into the course_catalog Course table.

`course_catalog.tasks.get_bootcamp_data` can be run manually. You can then check the database to see if all 3 existing bootcamps were properly loaded.


pdpinch:
nit: can you run the code formatter on the json sample: 

```
node node_modules/.bin/prettier-eslint --write test_json/bootcamps.json
```

sesterki:
Done

app.json
```diff
@@ -43,6 +43,10 @@
       "description": "S3 Bucket name.",
       "required": false
     },
+    "BOOTCAMPS_URL": {
+      "description": "URL to retrieve bootcamps data",
+      "required": false
+    },
```
mbertrand: Peter [suggested including the bootcamp JSON file in the repo instead of retrieving it via a URL](https://github.com/mitodl/open-discussions/issues/1976#issuecomment-486814949), if that's still the case then this can be deleted and the JSON file can be added/used instead.

course_catalog/tasks_test.py
```diff
@@ -320,3 +335,14 @@ def test_upload_ocw_master_json(settings, mocker):
...
+
+def test_process_bootcamps(mock_get_bootcamps):
+    """
+    Test that bootcamp json data is properly parsed
+    """
+    get_bootcamp_data()
+    assert Course.objects.count() == 3
+
+    get_bootcamp_data()
+    assert Course.objects.count() == 3
```
mbertrand: Can you also test that some of the fields are correctly populated, including `platform`?
Speaking of `platform`, it might be good to add `Bootcamps` to `course_catalog.constants.PlatformType` though it's not necessary yet:

```
class PlatformType(Enum):
    """
    Enum for platforms
    """

    ocw = "ocw"
    mitx = "mitx"
    bootcamps = "Bootcamps"
```

app.json
sesterki: There was another discussion more recently and we decided it was best for the JSON to live in its own repo, so we don't need a whole deploy in order to update courses.

course_catalog/tasks_test.py
sesterki: Should I create a `ResourceType` value for bootcamps as well?

mbertrand: Might be more a question for @pdpinch.  I think the default "course" gets assigned to `learning_resource_type`, not sure if bootcamps should have a different type or not.

pdpinch: ‘Learning_resource_type’ is a field from OCW (I think there are only two possible values, “course” and “resource”). If we can’t leave it bull, then I guess we should set it to “course.”

PR #2027, opened 2019-05-28
New course catalog data models

### LINKED ISSUE #2028: New data models for course catalog
As a user, I would like to be able to create different lists of  learning objects like “my favorites”, “My suggested list of courses to take if you want to take your blank knowledge at the next level”, etc,.  Each list can be private or public and the elements in the list should be orderable. 
As a user, I'd like to be able to see Courses and Bootcamps organized into official Program lists.   

- Course
- Program
- Bootcamp
- Learning path

#### Designs and Mockups

Google doc spec:
Invision designs:

#### Acceptance Criteria:

Create a new model for bootcamps (for clarity)
Create a new model for programs ( Programs can contain courses)

#### Out of Scope
- Not going to do Video LearningResource
- Create a new API for xpro platform to get the list of courses and programs

### LINKED ISSUE #2029: TimestampedModel is defined a second time in the course models
This model is partially defined here and only used for those models:

https://github.com/mitodl/open-discussions/blob/e63dbf3df041b7babb32cfdb3bc73d236b012e21/course_catalog/models.py

But we already have a more full fledged implementation here: 

https://github.com/mitodl/open-discussions/blob/a243f8b338f3d68cb79b3f31217eab731e2cdc1a/open_discussions/models.py#L44

This is both a tech-debit issue and a source of potential bugs.

kyfantaz:
Closes #2028
Closes #2029

#### What's this PR do?
This PR adds the new catalog models for Bootcamp, Program, and LearningPath and associated base models discussed in-person on 5/28. 

#### Open Questions (seeking answers): 
@pdpinch 
1) What Program-specific fields do we want to add?
2) What LearningPath-specific fields do we want to add?
3) What other metadata should Lists and ListItems be tracking?
4) What other metadata is common to all our LearningResources?
5) Does CourseTopic suffice as subjects for all our LearningResources?

#### Things Left to do for this PR:
 - [ ] Modify `parse_bootcamp_json_data` to use the new Bootcamp model

#### At Release:
- [ ] Ensure `get_bootcamp_data()` runs at release to repopulate Bootcamp


course_catalog/models.py
```diff
@@ -74,12 +88,105 @@ class Course(TimestampedModel):
     url = models.URLField(null=True, max_length=2048)
     program_type = models.CharField(max_length=32, null=True, blank=True)
     program_name = models.CharField(max_length=256, null=True, blank=True)
-    location = models.CharField(max_length=128, null=True, blank=True)
```
rhysyngsun: I think we should keep this for now, I'll explain in my overall feedback.

course_catalog/models.py
```diff
@@ -74,12 +88,105 @@ class Course(TimestampedModel):
...
+    author = models.ForeignKey(User, on_delete=models.PROTECT)
+    privacy_level = models.CharField(max_length=32, null=True, blank=True)
+
+
+class LearningPathItem(ListItem):
+    """
+    ListItem model for LearningPaths
+    """
+
+    list = models.ForeignKey(
```
rhysyngsun: Should rename this field to not be a python builtin.

course_catalog/models.py
```diff
@@ -74,12 +88,105 @@ class Course(TimestampedModel):
...
+
+    pass
+
+
+class ProgramItem(ListItem):
+    """
+    ListItem model for Programs
+    """
+
+    list = models.ForeignKey(Program, related_name="items", on_delete=models.CASCADE)
```
rhysyngsun: Should rename this field to not be a python builtin.

course_catalog/models.py
```diff
@@ -74,12 +88,105 @@ class Course(TimestampedModel):
...
+
+    course_id = models.CharField(max_length=128, unique=True)
+    language = models.CharField(max_length=128, null=True, blank=True)
+    year = models.IntegerField(null=True, blank=True)
+    full_description = models.TextField(null=True, blank=True)
+    start_date = models.DateTimeField(null=True, blank=True)
+    end_date = models.DateTimeField(null=True, blank=True)
+    enrollment_start = models.DateTimeField(null=True, blank=True)
+    enrollment_end = models.DateTimeField(null=True, blank=True)
+    image_src = models.URLField(max_length=400, null=True, blank=True)
```
rhysyngsun: This and other image-fields should be an `ImageField` so it can take advantage of our S3 image hosting and cloudfront distributions (look at `channels.models.Channel` for examples)

course_catalog/models.py
kyfantaz: I had the same thought initially, but after taking a look at `ESCourseSerializer`, it didn't seem to be the case that `location` field is getting indexed for search right now (which makes some sense since the code responsible for syncing bootcamps was only recently merged in). Additionally, the data scope is very small. There are only 3 bootcamps in total, and only 1 of them has a location (bootcamp data is being sourced from here: https://github.com/mitodl/bootcamps.json/blob/master/bootcamps.json), so if it were the case that location was being indexed, this would only affect one bootcamp course. Location is null for all other courses. It's a field only for bootcamps that we recently added to the Course model because we squished bootcamps into Course. I'm happy to defer to you if you still feel strongly; we can totally do the model-change/migration in follow-up PR. Just let me know.

course_catalog/models.py
kyfantaz: I had written down the same concern in the spec! Any ideas for proposed names?

course_catalog/models.py
kyfantaz: I had written down the same concern in the spec! Any ideas for proposed names?

course_catalog/models.py
kyfantaz: Will do :+1:

kyfantaz: Question: should we maybe just use `ImageField` for user-created `LearningPaths` for now, instead of doing the migration of images/thumbnails for all Courses and Bootcamps to S3 right now as well? Right now we're getting URLs for hosted images directly from where we're importing Courses (e.g. edx, OCW). These image files are hosted elsewhere, but they may have CDN infrastructure behind them already. In any case, does it make sense to do this migration in another PR later?

course_catalog/models.py
kyfantaz: I've changed this from `list` to `learning_path` for now

course_catalog/models.py
kyfantaz: I've changed this from `list` to `program` for now

course_catalog/models.py
kyfantaz: Do `program_image_upload_uri` and `learning_path_image_upload_uri` look about right to you?

rhysyngsun: Oh, are these image urls provided in the upstream data as opposed to being a piece of content uploaded within discussions? If yes, ignore this piece of feedback.

rhysyngsun: Although a comment on the field would probably be helpful

kyfantaz: For user-created `LearningPaths`, we discussed users can pick a image from a selection if they make their list public. We haven't yet discussed whether we want to let a user to upload a custom image for their `LearningPaths` . @pdpinch would you like to weigh in?

course_catalog/models.py
kyfantaz: @rhysyngsun can you confirm?

course_catalog/models.py
Ferdi: Custom image would be great

kyfantaz: Then the change I made in LearningPath to use `ImageField` will allow that (@rhysyngsun, yes?)
And we want to keep `URLField` for the image urls we are getting from Course and Bootcamp import sources for now @Ferdi ?

rhysyngsun: That sounds right to me - `ImageField` is for image assets that discussions handles and `URLField` is for externally managed assets we're referencing.

kyfantaz:
While waiting for approval on this PR, @sesterki can branch off this code to update the bootcamp sync code and its unit tests to use the new Bootcamp model. Let us know if there are objections to that plan. Also let us know if there any objections to this PR.

course_catalog/models.py
rhysyngsun: Looks like that's correct

course_catalog/tasks_test.py
```diff
@@ -355,7 +355,7 @@ def test_process_bootcamps(mock_get_bootcamps):
     course = Course.objects.get(course_id="Bootcamp3")
     assert course.platform == PlatformType.bootcamps.value
     assert course.title == "MIT Sports Entrepreneurship Bootcamp"
-    assert course.location == "Hoffenheim, Germany"
+    # assert course.location == "Hoffenheim, Germany"
```
rhysyngsun: You can just remove this line altogether since we won't be adding the field back in

course_catalog/utils.py
```diff
@@ -0,0 +1,18 @@
+""" Utils for course catalog """
+from profiles.utils import generate_filepath
+
+
+def learning_path_image_upload_uri(instance, filename):
+    """
+    upload_to handler for user-created Learning Path image
+    """
+    return generate_filepath(
```
rhysyngsun: Since this PR pulls this into a different module, can you move `generate_filepath` and its tests into `open_discussions/utils.py` so it's obvious this is a more generic method?

course_catalog/migrations/0023_adds_new_learning_resources_and_lists_20190528.py
```diff
@@ -0,0 +1,236 @@
+# Generated by Django 2.1.7 on 2019-05-28 23:21
+
+from django.conf import settings
+from django.db import migrations, models
+import django.db.models.deletion
+
+from course_catalog.constants import PlatformType
+
+
+def delete_bootcamps(apps, schema_editor):
```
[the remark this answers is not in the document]
sesterki: Removed.

course_catalog/tasks_test.py
sesterki: Removed

course_catalog/utils.py
sesterki: Moved into `open_discussions/utils.py`

profiles/utils.py
```diff
@@ -12,10 +11,10 @@
 
 from PIL import Image
 
-from open_discussions.utils import now_in_utc
-
 
 # Max dimension of either height or width for small and medium images
+from open_discussions.utils import generate_filepath
```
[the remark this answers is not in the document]
sesterki: I made this change. No idea why its not reflected on github.

PR #2033, opened 2019-06-04
fix course search page loading behavior

### LINKED ISSUE #1924: course search infinite scroll doesn't work as expected
### Steps to Reproduce

1. visit `/courses/search`
2. scroll down to the bottom of the page

### Expected Behavior

I would expect there to be a loading indicator shown at the bottom of the course list which is eventually replaced with  the newly fetched courses (as the infinite scroll works on, for instance, the post lists we show on the home and channel pages).

### Actual Behavior

When scrolling down on the source search page when the loading of a new page of results is triggered the whole UI re-renders to show a loading state, instead of just showing a loading indicator at the bottom where the new posts will be inserted. This leads to a stange and disconcerting flash for the user.

### Screenshot or Screencast

here's a gif I recorded with network throttling enabled, just to make it a little clearer:

loadingindicator

### LINKED ISSUE #1991: Flaky CreatePostsPage test
Error output from a travis build:

```
  1) CreatePostPage
       attempts to clear form and load channels on mount:
     AssertionError: expected 'menuMicroMasters arrow_drop_downhomeaddMADOSDNBNAEAcloseCreate a PostClear Post Content?This will clear any input you have made, are you sure you want to continue?CancelYesSelect a channel:Select a channelMollitia alias ametRatione fugitNon beataeDolore odit porroSit deserunt architectoDolorem expedita exercitationemPorro maioresNatus architecto etEt amet officiisShort postShare a linkCancelPost' to include 'Ut natus reprehenderit'
      at Context._callee4$ (static/js/containers/CreatePostPage_test.js:4:1994)
      at tryCatch (node_modules/@babel/polyfill/node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/@babel/polyfill/node_modules/regenerator-runtime/runtime.js:296:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/@babel/polyfill/node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (static/js/containers/CreatePostPage_test.js:1:79705)
      at _next (static/js/containers/CreatePostPage_test.js:1:80448)
```

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #1924 
also closes #1991 (I got too frustrated by build failures haha)

#### What's this PR do?

this fixes an issue with the infinite scroll on the course search page. basically, the issue was that we had two different types of loading states, one which was triggered by scrolling down and one triggered by changing the search text or the selected facets. In the case of the latter we want to reload the whole result UI, clear the old results, and start from scratch, but in the former case we want to show the already-fetched results while the infinite scroll component takes care of the loading indicator at the bottom. But our existing code only dealt with the latter case, and didn't distinguish the former adequately (there is more explanation of this on the issue if this is not clear).

So to fix it I basically just added a value, `incremental: boolean`, to the state on the course search page component. When we're rendering our results we can inspect this value to determine whether the most recent request that was fired off was an incremental reload (i.e. triggered by the infinite scroll component) or a full reload (i.e. triggered by a change in the search text, etc). This lets us determine when to show a full loading UI and when to let the infinite scroll component do its thing.

Look at the course search page on master and confirm the issue. Basically, when you scroll down the whole page should be taken over by a loading indicator, instead of just showing you a small loading indicator at the bottom of the post list.

then check out this branch and confirm that it fixes the issue. if you change facets or search text, though, you should still see the overall loading indicator.

### DIFF
--- a/static/js/containers/CourseSearchPage_test.js
+++ b/static/js/containers/CourseSearchPage_test.js
@@ -91,6 +91,8 @@ describe("CourseSearchPage", () => {
       )
     })
   })
+
+  //
   ;["", "a"].forEach(query => {
     it(`still runs a search if initial search text is '${query}'`, async () => {
       await renderPage({
@@ -131,7 +133,23 @@ describe("CourseSearchPage", () => {
     })
     // from is 5, plus 5 is 10 which is == numHits so no more results
     assert.isFalse(inner.find("InfiniteScroll").prop("hasMore"))
+    assert.deepEqual(inner.state(), {
+      // Because this is non-incremental the previous from value of 7 is replaced with 0
+      text:         "text",
+      activeFacets: new Map(
+        Object.entries({
+          platform:     [],
+          topics:       [],
+          availability: []
+        })
+      ),
+      from:              5,
+      error:             null,
+      currentFacetGroup: null,
+      incremental:       true
+    })
   })
+
   it("searches with parameters", async () => {
     SETTINGS.search_page_size = 5
     await renderPage(
@@ -162,6 +180,8 @@ describe("CourseSearchPage", () => {
       )
     })
   })
+
+  //
   ;[0, 5].forEach(from => {
     it(`InfiniteScroll initialLoad ${shouldIf(
       from > 0
@@ -172,6 +192,8 @@ describe("CourseSearchPage", () => {
       assert.equal(infiniteScroll.prop("initialLoad"), from === 0)
     })
   })
+
+  //
   ;[
     [true, false, false],
     [false, true, true],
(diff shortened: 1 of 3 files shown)


static/js/containers/CourseSearchPage_test.js
```diff
@@ -91,6 +91,8 @@ describe("CourseSearchPage", () => {
       )
     })
   })
+
+  //
```
rhysyngsun: :raised_hands:

PR #2036, opened 2019-06-04
Se/use bootcamps model

### LINKED ISSUE #2041: Update bootcamp data ingestion to double write to `Course` and `Bootcamp` tables
Update bootcamp data ingestion to double write to `Course` and `Bootcamp` tables so that we don't remove data from existing indexing strategies while populating the new table on our periodic updates.

sesterki:
#2028 
closes #2041 

#### What's this PR do?
This PR changes the bootcamp parsing code to use the new Bootcamp database model.

Open up the shell, import and run `get_bootcamp_data` from `course_catalog.tasks`. Then check to see that the Bootcamp model is being properly populated with bootcamp data.

### DIFF
--- a/course_catalog/task_helpers.py
+++ b/course_catalog/task_helpers.py
@@ -20,8 +20,14 @@ from course_catalog.constants import (
     ResourceType,
     AvailabilityType,
 )
-from course_catalog.models import Course, CourseTopic, CourseInstructor, CoursePrice
-from course_catalog.serializers import CourseSerializer
+from course_catalog.models import (
+    Course,
+    CourseTopic,
+    CourseInstructor,
+    CoursePrice,
+    Bootcamp,
+)
+from course_catalog.serializers import CourseSerializer, BootcampSerializer
 from search.task_helpers import update_course, index_new_course
 
 log = logging.getLogger(__name__)
@@ -535,7 +541,7 @@ def get_micromasters_data():
     return requests.get(settings.MICROMASTERS_COURSE_URL).json()
 
 
-def parse_bootcamp_json_data(course_data, force_overwrite=False):
+def parse_bootcamp_json_data_course(course_data, force_overwrite=False):
     """
     Main function to parse bootcamp json data for one course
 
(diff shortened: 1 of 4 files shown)


course_catalog/task_helpers.py
```diff
@@ -535,73 +541,79 @@ def get_micromasters_data():
     return requests.get(settings.MICROMASTERS_COURSE_URL).json()
 
 
-def parse_bootcamp_json_data(course_data, force_overwrite=False):
+def parse_bootcamp_json_data(bootcamp_data, force_overwrite=False):
```
rhysyngsun: To follow up on my write for the new issue #2041, I think what we want here is actually two copies of this method:

- One that writes to `Course`
- One that writes to `Bootcamp`

And update the caller of `parse_bootcamp_json_data` to call both of those operations.

BYou'll almost definitely want to pass each of these methods a `copy.deepcopy` of the original data since the implementation is mutating the passed object via `pop()` and friends.

sesterki: Added a call to each function. The functions duplicate a ton of code but one will be deleted in a subsequent PR, and I am working on a refactor to the whole task_helpers file which will be another PR.

course_catalog/task_helpers.py
```diff
@@ -605,3 +611,81 @@ def parse_bootcamp_json_data(course_data, force_overwrite=False):
             )
             course.prices.add(course_price)
         index_func(course)
+
+
+def parse_bootcamp_json__data_bootcamp(bootcamp_data, force_overwrite=False):
```
rhysyngsun: Could you fix the name to avoid the double underscore?

sesterki: Fixed

course_catalog/tasks_test.py
```diff
@@ -342,19 +348,16 @@ def test_process_bootcamps(mock_get_bootcamps):
     Test that bootcamp json data is properly parsed
     """
     get_bootcamp_data()
-    assert Course.objects.count() == 3
+    assert Bootcamp.objects.count() == 3
```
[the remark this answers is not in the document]
sesterki: Added

PR #2054, opened 2019-06-13
introduce Formik and use it for the comment creation / editing form

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

no ticket

#### What's this PR do?

introduces Formik and uses it to refactor out a `<CommentForm />` component, which lets us delete a bunch of yucky code. We were previously sharing from code from the predecessor of this component with the post editing form component - I'll swing back at some point to switch that over to Formik as well, but for now I'll leave it as-is.

Make sure you can make and edit comments, and reply to comments, without any problems. Editing a post should work as before as well.


static/js/components/CommentForm.js
```diff
@@ -0,0 +1,223 @@
...
+              disabled={
+                isSubmitting || userIsAnonymous() || !values.commentText
+              }
+            >
+              Submit
+            </button>
+            {closeReply ? (
+              <button
+                onClick={preventDefaultAndInvoke(closeReply)}
+                className="cancel"
```
rhysyngsun: Should this be disabled when `isSubmitting` too?

static/js/components/CommentForm.js
```diff
@@ -0,0 +1,223 @@
...
+            }`
+          )
+        )
+      }
+    }
+  }
+})
+
+const CommentForm = connect<Props, OwnProps, _, DispatchProps, _, _>(
+  () => ({}),
```
rhysyngsun: I think you can just pass `null` here and avoid the extra func/call overhead.

static/js/components/CommentTree.js
```diff
@@ -135,13 +155,8 @@ export default class CommentTree extends React.Component<Props> {
...
               beginEditing={e => {
-                if (beginEditing) {
-                  beginEditing(
-                    replyToCommentKey(comment),
-                    getCommentReplyInitialValue(comment),
-                    e
-                  )
-                }
+                e.preventDefault()
+                this.addCommentToState("replying", comment)
```
rhysyngsun: Nit: define constants for the state key

static/js/components/CommentTree.js
```diff
@@ -90,14 +101,25 @@ export default class CommentTree extends React.Component<Props> {
     )
   }
 
+  removeCommentFromState = (key: string, comment: CommentInTree) => {
+    const newSet = new Set(this.state[key])
+    newSet.delete(comment.id)
+    this.setState({ [key]: newSet })
+  }
+
+  addCommentToState = (key: string, comment: CommentInTree) => {
```
rhysyngsun: Nit: Type `key` as `"replying" | "editing"`

static/js/components/CommentForm.js
aliceriot: yeah, that makes sense I think

aliceriot:
@rhysyngsun Ok those things should be addressed

PR #2000, opened 2019-05-03
add animation to the drawer opening and closing

aliceriot:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [x] Tag @ferdi or @pdpinch for review
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?

closes #1994
closes #1998

#### What's this PR do?

This takes care of two things. The first is adding a css transition to animate the change between the expanded and compact state (for the desktop drawer). 

For this I just did a css transition of 500ms on the `width` property. I also changed the styling slightly to deal with things getting a little squished while the animation was planing. Basically, before we had the `.mdc-drawer--persistent` div, and the `.mdc-drawer__drawer` inside of that, and both of those changed their width when the open / closed state was toggled. I changed it around so that `.mdc-drawer__drawer` is always the full width, but `.mdc-drawer--persistent` has `overflow: hidden` so we don't see the full width of that div when the drawer is compacted.

I also changed how the navigation subtitles are rendered, so that they always occupy the same amount of vertical space, even if the drawer is compacted. this should make it so there is no visual jumping up and down between the compact and expanded state.

#### How should this be manually tested?

play around with the drawer and make sure it feels nice to use. also confirm that the mobile drawer is unaffected.

#### Screenshots (if appropriate)

this shows the spacers:

closesfodpff
closes

PR #2055, opened 2019-06-14
drawer open/close animation tweak

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

no ticket, but follow-on from #2000 

#### What's this PR do?

This modifies the drawer again to be more in-line with the behavior that exists for gmail. Basically, this changes a few things about how the hover-over version works:

- now only opening / closing with the hamburger button causes the rest of the page to move. if you hover over the drawer expands, but the content on the rest of the page doesn't move.
- I also added a slight delay only for the hover-over open. I noticed that gmail has this - it prevents the animation from starting if you just quickly mouse in and then out, which can prevent some unsightly animation flashing. you can see how this looks in the gif below.
- I also added a very slight box-shadow to the right edge of the desktop drawer, just to add a slight visual indicator for where the edge of the drawer is. without the main content shrinking it looked a little weird to have it run into the content at certain narrower widths.

I think that's it?

check out this branch and make sure that all the above described things work as advertised. there shouldn't be any weirdness! one thing to be sure to test is to make sure it works the same everywhere on the app - I ran into a weird bug where it worked differently on the channel and post pages from the home page, I fixed that but there could be a few more strange little edge cases lurking out there.

the mobile drawer should be unchanged, but best to check that too.

drawer

### DIFF
--- a/static/js/components/NavigationItem.js
+++ b/static/js/components/NavigationItem.js
@@ -5,15 +5,22 @@ export const NavigationExpansion = React.createContext<boolean>(false)
 
 type Props = {
   badge?: Function,
-  whenExpanded: Function
+  whenExpanded: Function,
+  fading?: boolean
 }
 
-const NavigationItem = ({ badge, whenExpanded }: Props) => (
+const NavigationItem = ({ badge, whenExpanded, fading }: Props) => (
   <NavigationExpansion.Consumer>
     {expanded => (
       <React.Fragment>
         {badge ? badge() : null}
-        {expanded ? whenExpanded() : null}
+        {fading ? (
+          <div className={expanded ? "fading-item expanded" : "fading-item"}>
+            {whenExpanded()}
+          </div>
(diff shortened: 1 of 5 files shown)


aliceriot:
Yeah, I had to do it that way because you can't do a css transition animation on elements being added to the dom, so I had to have them just pop in.

I just had an idea though - maybe I could have them there all the time but have their opacity initially set so that they're transparent and then have them fade in at the same time as the rest of the drawer. I'll give that a try, cause if that could work it would be nice.

aliceriot:
@mbertrand ok I fiddled with it a bit more - I wasn't able to come up with a good solution for the "compose" button but I think the other ones feel nicer now. lmk what you think! I'll probably come back to fix the compose button as well at some point, but at the moment I'm not sure the best way to do so.

PR #2056, opened 2019-06-14
updates indexing code for course catalog models

### LINKED ISSUE #2042: Update indexing code in the `search` app to index `Bootcamp` table
We need to ensure that if a reindex is performed that we index documents in the `Bootcamp` table:

  - Update `search.tasks.start_recreate_index` to index bootcamps separately from `Course`, because those indexed documents are uniqued on `Course.course_id` and we're losing the integrity guarantee that having all data in `Course` gave us
    - You'll need to add corresponding `index_bootcamps`, indexing serializer methods and class as well, it'll probably more or less be a copy of the current search indexing, but it needs to get indexed under a new `bootcamp` type.
    - in `search.api._apply_general_query_filters`, line 124, add the new bootcamp type

kyfantaz:
_This is PR is intended to start the necessary conversation about the exact behavior we want from the indexing code of the new learning resource objects._

Closes #2042

#### What's this PR do?
Updates indexing-related code for the new learning resource models, namely for Bootcamps. In addition to indexing related code, this PR adds analog functions for creating, updating, and deleting associated index objects for Bootcamps, Programs, and UserLists, where we are creating, updating, deleting Bootcamps, Programs, and UserLists; however, they are not used any where yet. There are no Programs of UserLists currently. There do exists 3 Bootcamps.

run ES indexing code, check for the created ES indices, search for those objects
1) Create index:
```
docker-compose  run web python manage.py recreate_index
```
2) Hit this endpoint to run a search query for bootcamp: `http://localhost:9101/discussions_local_all_default/_search`
with a POST request:
Method: `POST`
Content-Type: `application/json`
Body: `{"query":{"term":{"object_type":"bootcamp"}}}`


search/api.py
```diff
@@ -73,6 +81,59 @@ def gen_course_id(course_id):
...
+    Returns:
+        str: The Elasticsearch document id for this object
+    """
+    safe_id = (
+        urlsafe_b64encode(program_obj.title.encode("utf-8"))
+        .decode("utf-8")
+        .rstrip("=")
+        .replace(" ", "_")
+    )
+    return "prog_{}".format(safe_id)
```
kyfantaz: What should be the id for a Program? Will titles for Programs be unique (in which case, we can add a database-enforced unique constraint to the model)?

search/api.py
```diff
@@ -73,6 +81,59 @@ def gen_course_id(course_id):
...
+        str: The Elasticsearch document id for this object
+    """
+    learning_path_id = str(learning_path_obj.author.username + learning_path_obj.title)
+    safe_id = (
+        urlsafe_b64encode(learning_path_id.encode("utf-8"))
+        .decode("utf-8")
+        .rstrip("=")
+        .replace(" ", "_")
+    )
+    return "lp_{}".format(safe_id)
```
kyfantaz: What should be the id for a Learning Path? Should we enforce uniqueness of titles for user-created Learning Paths? Should we allow same titles of LearningPaths across users and create a LearningPath id using a user's username and a title?

search/api.py
```diff
@@ -123,10 +184,27 @@ def _apply_general_query_filters(search, user):
 
     course_filter = Q("term", published=True) | ~Q("terms", object_type=[COURSE_TYPE])
 
+    bootcamp_filter = Q("term", published=True) | ~Q(
+        "terms", object_type=[BOOTCAMP_TYPE]
+    )
+
+    program_filter = ~Q("terms", object_type=[PROGRAM_TYPE])
```
kyfantaz: How should we query for Programs?

search/api.py
```diff
@@ -123,10 +184,27 @@ def _apply_general_query_filters(search, user):
...
+
+    learning_path_filter = Q("term", privacy_level=PrivacyLevel.public.value) | ~Q(
+        "terms", object_type=[LEARNING_PATH_TYPE]
+    )
```
kyfantaz: How should we query for Learning Paths?

search/tasks.py
```diff
@@ -209,6 +209,63 @@ def index_courses(ids):
...
+    except (RetryException, Ignore):
+        raise
+    except:  # pylint: disable=bare-except
+        error = f"index_programs threw an error"
+        log.exception(error)
+        return error
+
+
+@app.task(autoretry_for=(RetryException,), retry_backoff=True, rate_limit="600/m")
+def index_learning_paths(ids):
```
kyfantaz: `index_*` functions like `index_bootcamps`, `index_programs`, `index_learning_paths` , etc can be refactored into a shared function. Should we do so, or keep it separate as `index_{profile,comment,post,course}` are currently?

search/serializers.py
```diff
@@ -303,6 +317,82 @@ class Meta:
...
+        """
+        return list(program.topics.values_list("name", flat=True))
+
+    class Meta:
+        model = Program
+        fields = ["id", "short_description", "title", "image_src", "topics"]
+
+        read_only_fields = fields
+
+
```
kyfantaz: How do we want to search for Programs and LearningPaths? Currently, end users can only search for them by title, description, and topics. We can't search for courses in these Lists models. How we answer this question will affect `ESProgramSerializer` and `ESLearningPathSerializer`, which is currently only serializing what's shown.

search/api.py
mbertrand: Another concern besides uniqueness is - what if the program title needs to be modified?  Ideally the id would be something that would never change.

search/tasks.py
mbertrand: It might be easier to keep separate, since they all used different serializers/models

search/tasks.py
```diff
@@ -255,6 +314,27 @@ def start_recreate_index(self):
...
             ]
+            + [
+                index_bootcamps.si(ids)
+                for ids in chunks(
+                    Bootcamp.objects.order_by("id").values_list("id", flat=True),
+                    chunk_size=settings.ELASTICSEARCH_INDEXING_CHUNK_SIZE,
+                )
+            ]
+            + [
+                index_courses.si(ids)
```
mbertrand:
```python
index_programs.si(ids)
```

search/tasks.py
```diff
@@ -255,6 +314,27 @@ def start_recreate_index(self):
...
+                for ids in chunks(
+                    Program.objects.order_by("id").values_list("id", flat=True),
+                    chunk_size=settings.ELASTICSEARCH_INDEXING_CHUNK_SIZE,
+                )
+            ]
+            + [
+                index_courses.si(ids)
```
mbertrand:
```python
index_learning_paths.si(ids)
```

search/tasks.py
kyfantaz: Fixed.

search/tasks.py
kyfantaz: Fixed.

search/api.py
kyfantaz: Indeed. What else could we use?

search/tasks.py
kyfantaz: :+1:

search/api.py
pdpinch: For now, we'll only query the Program metadata, i.e. program description and program title. We won't query across courses that are _in_ the program list.

search/api.py
kyfantaz: As discussed in-person, we'll be using the autogenerated object ID.

search/api.py
kyfantaz: Users can see their own `private` LearningPaths, too.

search/api.py
kyfantaz: As discussed in-person, we'll be using the autogenerated object ID. No uniqueness for titles.

search/serializers.py
kyfantaz: As mentioned above, only metadata, not objects within the lists.

course_catalog/constants.py
```diff
@@ -35,9 +35,19 @@ class AvailabilityType(Enum):
...
 
 
+class ListType(Enum):
+    """
+    Enum for UserLists. User-created lists that are not Learning Paths are displayed
+    to the user as "Your Lists"
+    """
+
+    learning_path = "Learning Path"
+    list = "List"
```
rhysyngsun: Change to not use a python builtin as a constant. Uppercasing the constants would fix that.

search/api.py
```diff
@@ -73,6 +81,48 @@ def gen_course_id(course_id):
...
+    """
+    Generates the Elasticsearch document id for a bootcamp
+
+    Args:
+        bootcamp_id (str): The course_id of a Bootcamp object
+
+    Returns:
+        str: The Elasticsearch document id for this object
+    """
+    safe_id = urlsafe_b64encode(bootcamp_id.encode("utf-8")).decode("utf-8").rstrip("=")
```
rhysyngsun: Is there a reason for stripping the `=` here? I think you can leave it in so we can reverse it.

kyfantaz: I don't know the range of possible contents for a Bootcamp's `course_id`, so I don't know for sure. The code copies the logic of `gen_course_id`, as bootcamps are course-like. If you think there's a good reason we should keep it for `gen_course_id` but not for `gen_bootcamp_id`, we can remove the strip for `=` here.

course_catalog/constants.py
kyfantaz: 👍

search/api.py
rhysyngsun: I didn't see that. In that case, keep it here and add it to `gen_program_id` and `gen_user_list_id` since those dropped the `rstrip()` call

kyfantaz:
> More generally: is there a reason this PR contains a lot of datamodel changes too when it's just supposed to be indexing changes?
The model changes fell from the in-person discussion yesterday (about this PR) that these indexing changes will somewhat depend on. We could carve out the changes into a separate PR that merged into this branch, if you prefer. This PR is getting big.

rhysyngsun:
Yes, let's pull the model changes out into their own PR and have that merge first since model changes are out of scope for the issue this PR was originally for.

search/api.py
kyfantaz: We discussed yesterday that it would be fine to use the autogenerated Django object id for Programs and UserLists, since they don't have a reliably generated unique id like `course_id` that we source. As such, the inputs will likely just be `ints`, and probably don't need the `rstrip()` calls. It's confusing, and I'm going to add comments. Do you think we should still put it in?

kyfantaz:
@rhysyngsun will do. Pulling out the model changes, how does this PR otherwise look?

search/api.py
rhysyngsun: yeah, if that's the case then the id can just go directly into the generated id and not be encoded.

rhysyngsun:
I'd like to see the model change PR first and then this PR would need to be rebased and I'll take another look at it then.

kyfantaz:
@rhysyngsun with #2061 merged into master and pulled on this PR, could you take another look for additional requested changes on this PR?

search/serializers.py
```diff
@@ -427,3 +525,72 @@ def serialize_course_for_bulk(course_obj):
...
+def serialize_bulk_user_lists(ids):
+    """
+    Serialize user_lists for bulk indexing
+
+    Args:
+        ids(list of int): List of user_list id's
+    """
+    for user_list in UserList.objects.filter(id__in=ids):
```
rhysyngsun: I think this and all other `LearningResource` subclasses should `prefetch_related("topics")`

kyfantaz: Sure, I can make that change, but for it to have the intended effect (saving a database call), I'll have to change the downstream Serializer code containing `list(course.topics.values_list("name", flat=True))` to something like [topic.name for topic in course.topics.all()], as `values_list` call will put the burden on the database to collate the `name` field, thus making another database call despite the prefetching.

kyfantaz: Handled in https://github.com/mitodl/open-discussions/pull/2056/commits/70e5b4bcb79946a316496229befc67b58c1cc8f5

kyfantaz:
Anything further requests, @rhysyngsun?

Issue #2060, opened 2019-06-19
model to store a learner's Favorites


kyfantaz:
As an learner/end user, I'd like to favorite courses, programs, learningpaths, etc to review later or save for other purposes

kyfantaz:
Closed by #2061

PR #2061, opened 2019-06-19
adds new catalog model FavoriteItem; renames LearningPath model

kyfantaz:
As requested by @rhysyngsun, this PR pulls out the model updates from #2056. This PR should be merged in before #2056 

#2060

#### What's this PR do?
adds new catalog model FavoriteItem; renames LearningPath model to UserList


course_catalog/models.py
```diff
@@ -187,3 +185,21 @@ class ProgramItem(ListItem):
...
+    """
+
+    user = models.ForeignKey(User, on_delete=models.PROTECT)
+    content_type = models.ForeignKey(
+        ContentType,
+        limit_choices_to={"model__in": ("Course", "Bootcamp", "UserList", "Program")},
+        on_delete=models.CASCADE,
+    )
+    object_id = models.PositiveIntegerField()
+    item = GenericForeignKey("content_type", "object_id")
```
rhysyngsun: There should probably be a unique constraint across the tuple of `('user', 'content_type', 'object_id')` to ensure an item can't be favorited multiple times.

course_catalog/models.py
```diff
@@ -149,25 +146,26 @@ class Meta:
...
+    UserList is a user-created model tracking a restricted list of LearningResources.
     """
 
     author = models.ForeignKey(User, on_delete=models.PROTECT)
     privacy_level = models.CharField(max_length=32, null=True, blank=True)
     image_src = models.ImageField(
-        null=True, max_length=2083, upload_to=learning_path_image_upload_uri
+        null=True, max_length=2083, upload_to=user_list_image_upload_uri
     )
+    list_type = models.CharField(max_length=128)
```
rhysyngsun: This looks like it has a `default` value in the migration but not here, could you reconcile that?

kyfantaz: The migration file asked for a default because it has to assume there already exists instances that are missing values for this new column. In reality, there aren't any instances yet, and instead of adding a default to the model declaration making a explicit declaration optional, I think I would prefer the list_type to be a required field that is explicitly declared. Do you disagree?

course_catalog/models.py
kyfantaz: Sure thing.

course_catalog/models.py
rhysyngsun: ah, ok, that makes sense

PR #2064, opened 2019-06-20
Add endpoints for users to favorite and view favorited items

sesterki:
_This is PR is intended to start the conversation around how the frontend and backend will interact to create the feature of users being able to favorite learning resource objects and displaying search results as favorited or not. Ideally, a user can favorite and unfavorite an item in real-time and have that be quickly reflected in their Favorites list_

Questions:
- How can we display whether a Search result is favorited or not?
- How do we populate the "My Favorites" list on the home page? Does the content from Elasticsearch, or does it come from an backend endpoint? If course/etc data comes from Elasticsearch, do we make ES queries for each favorited object?
- Would endpoint that takes in `object_id, object_type` pairs and returns true/false whether it is a favorite of the current user be helpful?

A starting proposal: 
Right now, what each favorites is not being indexed. Instead, we could do something akin to this: when the frontend loads and displays search results, it makes a request with a userid to a Favorites API that returns ids of what objects have been favorited, stored by the frontend. If the Id matches what comes up in a search result, the frontend displays the object’s card to be visually marked as favorited. 

Note that anonymous users (not logged in) don’t get to favorite things (otherwise we would have to come up with a scheme to store temporary IDs of what people favorite in their cookie or similar)

A user needs a way to favorite and unfavorite various course_catalog models. They also need to see which objects they favorited. 

#### What's this PR do?
This PR adds a viewset to view all FavoriteItem objects for a user. It also adds an is_favorite field to the serializers for Bootcamps, Courses, Programs, and UserLists.

Manually create some Bootcamps, Courses, Programs, and UserLists in the backend. Also create a user. Then create FavoriteItem objects linking the user to the other objects. Check the relevant endpoints to see that the `is_favorite` flag populates correctly, and then check `/api/v0/favorites` to see that all of the favorited items show up.

### DIFF
--- a/course_catalog/serializers_test.py
+++ b/course_catalog/serializers_test.py
@@ -18,6 +18,8 @@ from course_catalog.serializers import (
     BootcampSerializer,
     ProgramItemSerializer,
     FavoriteItemSerializer,
+    UserListSerializer,
+    ProgramSerializer,
 )
 from open_discussions.factories import UserFactory
 
@@ -97,11 +99,11 @@ def test_favorites_serializer():
 
     favorite_item = FavoriteItem(user=user, item=user_list)
     serializer = FavoriteItemSerializer(favorite_item)
-    assert serializer.data.get("content_data") == UserListFactory(user_list).data
+    assert serializer.data.get("content_data") == UserListSerializer(user_list).data
 
     favorite_item = FavoriteItem(user=user, item=program)
     serializer = FavoriteItemSerializer(favorite_item)
-    assert serializer.data.get("content_data") == ProgramFactory(program).data
+    assert serializer.data.get("content_data") == ProgramSerializer(program).data
 
     favorite_item = FavoriteItem(user=user, item=course_topic)
     serializer = FavoriteItemSerializer(favorite_item)
--- a/course_catalog/serializers_test.py
+++ b/course_catalog/serializers_test.py
@@ -89,19 +89,19 @@ def test_favorites_serializer():
 
     favorite_item = FavoriteItem(user=user, item=course)
     serializer = FavoriteItemSerializer(favorite_item)
-    assert serializer.data.get("content_data").get("id") == course.id
+    assert serializer.data.get("content_data") == CourseSerializer(course).data
 
     favorite_item = FavoriteItem(user=user, item=bootcamp)
     serializer = FavoriteItemSerializer(favorite_item)
-    assert serializer.data.get("content_data").get("id") == bootcamp.id
+    assert serializer.data.get("content_data") == BootcampSerializer(bootcamp).data
 
     favorite_item = FavoriteItem(user=user, item=user_list)
     serializer = FavoriteItemSerializer(favorite_item)
-    assert serializer.data.get("content_data").get("id") == user_list.id
+    assert serializer.data.get("content_data") == UserListFactory(user_list).data
 
     favorite_item = FavoriteItem(user=user, item=program)
     serializer = FavoriteItemSerializer(favorite_item)
-    assert serializer.data.get("content_data").get("id") == program.id
+    assert serializer.data.get("content_data") == ProgramFactory(program).data
 
     favorite_item = FavoriteItem(user=user, item=course_topic)
     serializer = FavoriteItemSerializer(favorite_item)
--- a/course_catalog/serializers_test.py
+++ b/course_catalog/serializers_test.py
@@ -10,13 +10,16 @@ from course_catalog.factories import (
     CourseInstructorFactory,
     BootcampFactory,
     ProgramFactory,
+    UserListFactory,
 )
-from course_catalog.models import ProgramItem
+from course_catalog.models import ProgramItem, FavoriteItem
 from course_catalog.serializers import (
     CourseSerializer,
     BootcampSerializer,
     ProgramItemSerializer,
+    FavoriteItemSerializer,
 )
+from open_discussions.factories import UserFactory
 
 pytestmark = pytest.mark.django_db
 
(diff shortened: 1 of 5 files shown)


Ferdi:
> How can we display whether a Search result is favorited or not?

can we have an array field favorited_by in ES , with IDs of users who have favorited the given learning resource.  If current users is in the list , we show it as 'favorited" 

> How do we populate the "My Favorites" list on the home page? Does the content from Elasticsearch, or does it come from an backend endpoint?

From backend endpoint,  I'd think.

mbertrand:
I think it would be best to have a backend API call for retrieving the "My Favorites" list, and you could create a "favorited_by" field in ES for search results like Ferdi mentioned.  This list could get pretty big though for popular courses.  Another option would be to retrieve the id's of the user's favorites via an initial API call on the search page. Then as search results are returned, cross-check against this list to determine which ones should be displayed as "Favorited".

kyfantaz:
@mbertrand So which option would you like to go with? 
1) a "favorited_by" field in ES that contains list of all user ids that have favorited the item
2) retrieve the id's of the user's favorites via an initial API call on the search page

mbertrand:
My guess is option 2 might be better because it would require fewer ES updates and avoid potential issues for popular courses (long lists and maybe conflicting updates).  Might be worth asking for another opinion from @rhysyngsun too.

Ferdi:
Yes option 1 might get us into performance issues. But it seems to me that option 2 solves just one small use case : in the list search how to identify items favorited by the current user.
If you look at the designs, the need is to return items that are favorited by user AND are in Computer science and offered by ocw. How do we do that?

kyfantaz:
@Ferdi I think the idea is that the user searches against ES for Computer Science offered by OCW courses, and of the returned results/items, we show all of them, but check against a list of IDs of favorited items (obtained from an independent call to the API) whether we display each course card as starred or not (favorited or not). The " My Favorites" list, can is displayed on the home search page or appear as an card in My Lists, could potentially be populated from the API alone, as long as the frontend understands and can obtain all the content it needs from the API response.

Ferdi:
There must be a better way.

rhysyngsun:
I think option 1 is not at all viable. An additional problem with that approach is that we'd be leaking who favorited which items in the search API, which poses both privacy issues and performance issues by potentially returning a huge object back to the client that it has no need for.

I think the only option to satisfy the requirements defined here and avoid the extra API is to add a field `favorited_by` that is a single value - either null or a user id. By default, we'd index objects with a null value. When a user favorites an object we insert a copy of that into the index with the value set to that user's id. Unfavoriting removes that record. We'd probably want to specify a suffix for the document id

When searching, if the user is not filtering by favorited items, we'd explicitly filter for `favorited_by=None`, and when they are filtering for their favorited items, we filter by `favorited_by=user.id`.

This avoids the api call and avoids a costly array on one object (which). It's also far more idiomatic to how you're supposed to store objects in ES. It does mean we'd storing multiple copies of objects, and the index update story gets a little more complex, but ES' [Update By Query](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html) is built for this type of use case.

kyfantaz:
@rhysyngsun I think there are two potential requirements that may not be captured currently by your proposal, 1) is that the designs indicate that the returned search results are mixed, some some results are favorited and some not (a search with no filter on favorites), and 2) each course/etc. card has a clickable star for the user to be able to favorite/unfavorite it, with an expectation that clicking on the star would just fill/unfill it as solid and not change my current search results. 
Could we modify your proposal to also satisfy the above requirements?

rhysyngsun:
You might be able to use a top hits aggregation to prioritize the favorited results over non-favorited results then: https://stackoverflow.com/questions/26509045/filter-elasticsearch-results-to-contain-only-unique-documents-based-on-one-field

kyfantaz:
Ah, in that case, is having two results returned for a favorited course is OK? If a user searches for a particular course, its OK for the results to be two cards for the same course, but the first/prioritized one is favorited, and second one is not.

kyfantaz:
@mbertrand FYI, unit tests added

rhysyngsun:
No, it would just return one record per unique course/program/bootcamp, with the favorited ones taking priority.

kyfantaz:
Ah, OK, so using top hits aggregation, we can aggregate on a shared unique id field to deduplicate original objects and a user's favorited copy. For Courses and Bootcamps, this would presumably be `course_id`. But what about for Programs, UserLists (learning paths, etc)? I don't think we have a stable unique id field that would be shared between the original object and the favorited copy currently. (In the indexing code, the original discussion landed on not wanting to rely on title uniqueness or an unchanging title).

Ferdi:
@kyfantaz Can you guys have a look at this https://www.spacevatican.org/2012/6/3/fun-with-elasticsearch-s-children-and-nested-documents/   -- Parent & Child  and  "Building Lists" paragraphs

kyfantaz:
Setting our Elasticsearch strategy on the side for a moment, if there's some consensus that we'll need an API for at least one purpose, whether to help populate the "My Favorites" list on the main page or otherwise, we could accept this PR (which adds a simple, unit-tested API without privacy/leak issues since it only returns the logged-in user's favorites) and then all meet to formalize the exact interaction and implementation with Elasticsearch during a less busy time in July as an option for next steps. Thoughts? @Ferdi @rhysyngsun @mbertrand

Ferdi:
sure !

course_catalog/serializers_test.py
```diff
@@ -71,3 +74,36 @@ def test_generic_foreign_key_serializer():
...
+    user = UserFactory()
+    course = CourseFactory()
+    bootcamp = BootcampFactory()
+    user_list = UserListFactory(author=user)
+    program = ProgramFactory()
+    course_topic = CourseTopicFactory()
+
+    favorite_item = FavoriteItem(user=user, item=course)
+    serializer = FavoriteItemSerializer(favorite_item)
+    assert serializer.data.get("content_data").get("id") == course.id
```
mbertrand: Might as well compare the whole thing:
```python
assert serializer.data.get("content_data") == CourseSerializer(course).data
```

Ditto for the other favorites below.

kyfantaz: 👍

PR #2066, opened 2019-06-20
add to README and docstring

### LINKED ISSUE #2067: rename availability labels
As a user, I'd like to easily know the availability of courses or bootcamps when looking at their cards.

Rename current Availability values to new ones:
```
Current -> Available now
Upcoming (no change)
Starting Soon (no change)
Archived -> Prior
```
Note that existing Courses/Bootcamps should be migrated to new values, so this requires a small custom migration.

kyfantaz:
#### What's this PR do?
This is a really small PR. It does two simple things:
- Renames previous availability values (user front-facing) to new ones.
- Adds to README about how to search course catalog elasticsearch index.

closes #2067

Manually run migrations and see if all availability values have been renamed.
Follow the README instructions.


README.md
```diff
@@ -243,6 +243,18 @@ To enable channel widgets, run through these steps:
...
+To enable searching the course catalog on elasticsearch, run through these steps:
+1. Start the services with `docker-compose up`
+2. With the above running, run this management command, which kicks off a celery task, to create an elasticsearch index:
+```
+docker-compose  run web python manage.py recreate_index
+```
+If there is an error running the above command, observe what traceback gets logged in the celery service.
+3. Once created and with `docker-compose up`  running, hit this endpoint in your browser to see if the index exists: `http://localhost:9101/discussions_local_all_default/_search`
+4. If yes, to run a specific query, make a `POST` request (using `curl`, [Postman](https://www.getpostman.com/downloads/), Python `requests`, etc.) to the above endpoint with a `json` payload. For example, to search for all courses, run a query with Content-Type as `application/json` and with a body `{"query":{"term":{"object_type":"course"}}}`
+
```
kyfantaz: Is this a good start? @mbertrand

mbertrand: 👍

course_catalog/constants.py
```diff
@@ -26,13 +26,13 @@ class ResourceType(Enum):
...
-    All OCW courses should be set to "Current"
+    All OCW courses should be set to "Available Now"
     """
 
-    current = "Current"
+    available_now = "Available Now"
     upcoming = "Upcoming"
     starting_soon = "Starting Soon"
-    archived = "Archived"
+    prior = "Prior"
```
mbertrand: Some courses from mitx will have availabilities of "Current" and "Archived" when imported.  If the point of this PR is to display them differently, there is some front-end JS code to do that [here](https://github.com/mitodl/open-discussions/blob/e710b62e1f39d02ec1dc065d745a87be6551b89b/static/js/lib/courses.js#L11-L20).  Or you could transform the values during the [course data import](https://github.com/mitodl/open-discussions/blob/e710b62e1f39d02ec1dc065d745a87be6551b89b/course_catalog/task_helpers.py#L320).

kyfantaz: Ah, thanks for the data context. Given that, to minimize change, I think we just keep the Availability constants on the backend as is and simply make a note that the values are coming in that way, what the mapping is, and the conversion happening in the frontend.

PR #2077, opened 2019-07-08
some JS dependency upgrades

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

no ticket

#### What's this PR do?

a few small dependency upgrades. I also did a small tweak that removes a warning from 'browserslist' that we were getting on travis-ci.

the upgraded packages are only used when testing, so I think that as long as the build is passing and the app seems to work I think we should be good.


rhysyngsun:
Noticed that this upgrades redux, I'm guessing that you wrote the PR before that, so I'll just give it a quick smoke test locally but I'm not expecting much.

aliceriot:
oh yeah I did upgrade redux as well, I forgot to update the PR description. I was going to upgrade a few other packages but ran into some incompatibilities so I stopped there for now - I want to upgrade react-redux soon though to get the new [hooks-based api](https://react-redux.js.org/api/hooks) but upgrading the package broke a bunch of tests so I'll have to circle back on that.

PR #2079, opened 2019-07-09
Fix search query for anonymous users

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
Fixes broken search for anonymous users

Search for courses and search for posts/comments, while logged in and anonymously.  There should be no errors either way.

### DIFF
--- a/search/api.py
+++ b/search/api.py
@@ -183,27 +183,26 @@ def _apply_general_query_filters(search, user):
     # Search programs
     program_filter = ~Q("terms", object_type=[PROGRAM_TYPE])
 
+    # Search public lists (and user's own lists if logged in)
+    user_list_filter = (
+        Q("term", privacy_level=PrivacyLevel.public.value)
+        | ~Q("terms", object_type=[USER_LIST_TYPE])
+    )
+    if not user.is_anonymous:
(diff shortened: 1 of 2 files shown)


search/api.py
```diff
@@ -161,40 +162,49 @@ def _apply_general_query_filters(search, user):
...
         .filter(content_filter)
         .filter(course_filter)
         .filter(bootcamp_filter)
         .filter(program_filter)
-        .filter(user_list_filter)
     )
 
+    if not user.is_anonymous:
+        # Search the user's lists
+        user_list_filter = (
```
rhysyngsun: Should this actually be:

```python
user_list_filter = (
    Q("term", privacy_level=PrivacyLevel.public.value)
    | ~Q("terms", object_type=[USER_LIST_TYPE])
)
if not user.is_anonymous:
    user_list_filter = user_list_filter | Q("term", author_id=user.id)
```

and can then be moved back to its original location so that public lists are still searchable but private lists are only searchable if logged in?

mbertrand: Right, I misread the query as just returning the user's own lists, which would be irrelevant if anonymous, will change.

PR #2098, opened 2019-07-18
switch course search to use grid design and course card

### LINKED ISSUE #2089: Change course search results from list to grid
The search results for courses are currently in a list layout and this needs to be a grid per the mockups

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #2089

#### What's this PR do?

this switches over the course search UI to use a grid (instead of a vertical list of cards) and to use the new CourseCard component to render the courses.

I also tweaked the CourseCard slightly so that it now calls the callback to open the course drawer only when the course cover image or title are clicked, instead of when the whole card is clicked. I also ported over the behavior we had previously on the CompactCourseDisplay component, where clicking on one of the topics on the course would add that topic to the current selected search facets.

make sure that everything renders as it should on the course search. you should be able to search for things, scroll down on the infinite scroll, etc without any problems.

you should also be able to open the course detail drawer by clicking on the photo or the title for a course card, and clicking on one of the course's topics should add that topic to your search.

I think that's about it? the course carousel should also work as it did before, with no regressions.

computersearchfff
computersearch
computersearchmobbb

note that there isn't a mobile design yet, but the cards are at least functional.


noisecapella:
It looks like the infinite scroll sometimes shows only one or two cards on a row rather than revealing a new row all at once. I'm happy leaving it that way since it seems difficult to fix given the responsive grid but commenting here in case you had an idea how to address that.

static/js/containers/CourseSearchPage.js
```diff
@@ -365,6 +377,9 @@ const mapDispatchToProps = (dispatch: Dispatch<*>) => ({
     dispatch(actions.search.clear())
     await dispatch(clearSearch())
   },
+  setShowCourseDrawer: ({ courseId }: { courseId: string }) => {
```
noisecapella: It looks like this is uncovered, can you add a quick test for it?

aliceriot:
I think it's because the search page size isn't set to be a multiple of three

aliceriot:
I could change the default to be 6 instead of five - is there any reason to worry about doing so?

noisecapella:
@aliceriot I don't think changing it to 6 would affect anything negatively

aliceriot:
word, I'll just change it to 6 then so we're covered on all environments. I checked on prod and it looks like we're using the default there currently.

aliceriot:
@noisecapella ok I added a test for that line

Issue #2091, opened 2019-07-12
UI for adding and removing a course search result as a favorite


rhysyngsun:
As a user searching for courses, I'd like to be able to favorite results so I can easily find them later

#### Designs and Mockups

https://projects.invisionapp.com/d/main#/console/17477251/367253747/preview

#### Acceptance Criteria:

- [ ] As a logged in user, I should be able to add a search result to my favorites
- [ ] As a logged in user, I should be able to remove a search result from my favorites

#### Out of Scope

- For a first pass, we should just focus on the toggle of favorite on/off and not address the popup/user list UI

rhysyngsun:
@pdpinch @abdulkdawson Do we have an idea what the experience should be for anonymous users since they won't be able to toggle favorites? Would it be the same signup/login popup that anonymous users get when they try to create a post?

aliceriot:
one other question: currently for the course cards that we show on the home page clicking anywhere on the card opens the side drawer, which shows details for the course. Based on the mockups it seems that clicking on the star in the corner should favorite / unfavorite a course.

my question is, how should we change how we open the course details drawer? we could potentially change it so that only clicking the title, for instance, opens the detail drawer, or make it so that clicking anywhere except on the 'favorite' star will open the drawer.

abdulkdawson:
@aliceriot Yes The entire card should not activate the drawer. Clicking the course title or thumbnail should open the drawer.

abdulkdawson:
@rhysyngsun since the favorites are for signed in users I feel we should be prompting users to sign-up/login in order to utilize favorites. That to me is the best course of action.

PR #2102, opened 2019-07-23
Search for bootcamps

### LINKED ISSUE #2043: Update frontend code to additionally search for `bootcamp` record types
Support the new LearningResource types on the frontend
- [ ] Add facets for the `LearningResource` types
- [ ] Allow filtering on courses
- [ ] allow filtering on bootcamps

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Fixes #2043 

#### What's this PR do?
- Modifies the frontend search query code to handle bootcamps and courses.
- Adds a search facet for 'learning resource' that includes courses and bootcamps
- Adds CompactBootcampDisplay and ExpandedBootcampDisplay components to display Bootcamp results
- renames `CourseDrawer` to `LearningResourceDrawer` and modifies it to handle different object types instead of just courses.

- Run `docker-compose run web python manage.py recreate_index`
- Import some bootcamps if you haven't already done so:
```python
get_bootcamp_data()
```
- Import some course data if you haven't already done so:
```
sync_and_upload_edx_data()
```
- Go to `/courses/search`.  There should be a new facet for `Learning Resource` that includes checkboxes for 'Course' and 'Bootcamp'.  Try these out and ensure that you can see both Course and Bootcamp records returned, and that clicking on a result opens the right-side drawer containing the expanded display for the course/bootcamp.

**Learning Resource Facet**
Screen Shot 2019-07-23 at 2 18 53 PM

**Learning Resource Facet, mobile**
Screen Shot 2019-07-23 at 2 19 35 PM

**Bootcamp facet selected, bootcamp card displayed**
Screen Shot 2019-07-23 at 2 19 06 PM

** Bootcamp expanded display **
Screen Shot 2019-07-23 at 2 32 20 PM


mbertrand:
@abdulkdawson  @Ferdi @pdpinch screenshots added

static/js/components/CourseCard.js
```diff
@@ -41,7 +42,7 @@ const CourseCard = ({
             alt={`cover image for ${course.title}`}
           />
         </div>
-        <div className="row course-title" onClick={showCourseDrawer}>
+        <div className="row course-title" onClick={showResourceDrawer}>
```
aliceriot: Is there some reason we need a separate card type for bootcamps? I think if we can do it it would be better to rename this `CourseCard` into `LearningResourceCard` and then modify it to support both courses and bootcamps, since we're displaying basically the same stuff for them. What do you think?

static/js/components/ExpandedBootcampDisplay.js
```diff
@@ -0,0 +1,117 @@
...
+
+const BOOTCAMP_IMAGE_DISPLAY_HEIGHT = 239
+const BOOTCAMP_IMAGE_DISPLAY_WIDTH = 440
+const entities = new AllHtmlEntities()
+
+type Props = {
+  bootcamp: Bootcamp
+}
+
+export default class ExpandedBootcampDisplay extends React.Component<Props> {
```
aliceriot: I have the same question here - if it's possible to have an `ExpandedLearningResourceDisplay` instead of two different very similar components I think that would be preferable

static/js/components/CourseCard.js
mbertrand: They are very similar but some of the info is different - for example bootcamps do not have a `semester` or `platform`.  I initially considered adding some conditional statements for these to `CourseCard` but ultimately decided to create separate card components in case they ever diverged further, and also because the `objectType` parameter sent to `LearningResourceDrawer` from the card needs to be different.  But if you think it's better to do the conditional approach I can change it.

aliceriot: I think if they need to diverge further in the future we can create separate components then, but unless we need to do that it's better to add a few small conditionals and avoid writing extra code.

static/js/components/CourseCard.js
```diff
@@ -63,16 +65,20 @@ const CourseCard = ({
...
+          <div className="row platform">
+            <img
+              className="course-platform"
+              // $FlowFixMe: only courses will end up here
+              src={platformLogoUrls[object.platform]}
+              // $FlowFixMe: only courses will end up here
+              alt={`logo for ${object.platform}`}
+            />
+          </div>
+        ) : null}
```
aliceriot: fyi in the designs we do want to have the logo show up for bootcamp items here

mbertrand: @abdulkdawson can you send me the logo image for bootcamps?

aliceriot: sorry - should have clarified, it's already added, if you look at the `platformLogoUrls` object in `constants.js` you can find the URL for it there

mbertrand: thanks, added the logo for bootcamps

static/js/components/ExpandedBootcampDisplay.js
mbertrand: ✔️

static/js/components/SearchResult.js
```diff
@@ -83,22 +89,42 @@ const ProfileSearchResult = ({ result }: ProfileProps) => {
...
+  )
+}
+
+const BootcampSearchResult = ({
+  result,
+  toggleFacet,
+  setShowResourceDrawer
+}: LearningResourceProps) => {
+  // $FlowFixMe: this should only be used for bootcamps
+  const bootcamp = searchResultToBootcamp(result)
```
aliceriot: one other approach I was thinking that we could take here is to, instead of having a separate `searchResultToBootcamp` we could modify the existing `searchResultToCourse` to be `searchResultToLearningResource` to sort of follow-along with how the UI is organized. That function could set `platform` to be `bootcamps` when the `object_type` is `bootcamp`, and then that value could be used throughout the UI to control whether we show bootcamp-specific UI or not.

Does that make sense? if we did it that way we'd avoid the need for a separate "bootcamp" search result, we wouldn't need to pass down the "objectType" parameter, etc because anywhere we need to figure out how to render the thing we could just look at the `platform` value.

static/js/containers/CourseIndexPage.js
```diff
@@ -123,7 +122,7 @@ const mapStateToProps = (state: Object): StateProps => ({
 })
 
 const mapDispatchToProps = {
-  setShowCourseDrawer
+  setShowResourceDrawer: setShowResourceDrawer
```
aliceriot: if you want to you can leave out the colon here, it's called a 'shorthand property name' https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015

static/js/components/CourseCard_test.js
```diff
@@ -1,49 +1,56 @@
...
-    })
+    renderCourseCard = ({ ...props }) =>
+      shallow(
+        <CourseCard
+          object={course}
+          objectType="course"
+          setShowResourceDrawer={setShowResourceDrawerStub}
+          {...props}
+        />
+      )
```
aliceriot: any reason why you refactored away from the `configureShallowRenderer` here?

static/js/factories/resources.js
```diff
@@ -0,0 +1,76 @@
+//@flow
```
aliceriot: might be good to call this file `learning_resources` so it's more immediately clear what it's about

static/js/lib/search.js
```diff
@@ -192,7 +227,11 @@ export const buildSearchQuery = ({
     builder.sort(field, option)
   }
 
-  const types = type ? [type] : ["comment", "post", "profile"]
+  const types = type
+    ? Array.isArray(type)
+      ? type
+      : [type]
+    : ["comment", "post", "profile"]
```
aliceriot: if possible it might be good to refactor away from a nested ternary, just a bit hard to follow

static/js/lib/util.js
```diff
@@ -148,3 +148,11 @@ export const flatZip = R.compose(
...
   R.zip
 )
+
+export const capitalize = R.converge(R.concat(), [
+  R.compose(
+    R.toUpper,
+    R.head
+  ),
+  R.tail
+])
```
aliceriot: what does this function do differently than `_.capitalize`? I tried it out in a node repl and `_.capitalize` seems to return the same thing

static/js/reducers/bootcamps.js
```diff
@@ -0,0 +1,21 @@
...
+  return update
+}
+
+export const bootcampsEndpoint = {
+  name:              "bootcamps",
+  verbs:             [GET],
+  initialState:      { ...INITIAL_STATE, data: new Map() },
+  getFunc:           (id: number) => bootcampAPI.getBootcamp(id),
+  getSuccessHandler: updateBootcampHandler
+}
```
aliceriot: I think it's not necessary for this PR, but if you want to use redux-query for managing this API instead of redux-hammock I think that would be a good idea, we're generally migrating away from redux-hammock and towards redux-query. I've been using it for any new APIs in Open for a little bit.

aliceriot: If you don't want to take the time now though that's alright, this is a pretty simple API so we can easily change it out later.

static/js/lib/util.js
mbertrand: In an earlier PR review, Nathan said I should avoid using lodash for new functionality and use ramda instead, so this is the ramda equivalent.

static/js/components/SearchResult.js
mbertrand: I can make a `searchResultToLearningResource` that then calls the existing `searchResultToCourse` or `searchResultToBootcamp` function depending on the `object_type`.  The `Bootcamp` type does not include `platform` or `level` or `semester`), and not sure if `platform` makes sense anyway for the upcoming `LearningPath` results, which will be very different than bootcamps and courses.

static/js/lib/util.js
aliceriot: ah ok, I was just wondering because I saw some other uses of `_.capitalize` around as well.

static/js/components/SearchResult.js
aliceriot: yeah, I was thinking instead of even having separate `Course` and `Bootcamp` types why not have a single `LearningResource`? the two `Course` and `Bootcamp` types you have now have the majority of their fields in common already.

mbertrand: Don't I need different types since they have different fields, even if currently it's just a few, otherwise flow will start complaining? I could have a `LearningResource` that both the `Course` and `Bootcamp` types inherit from with common fields, but once I introduce `Program` and `LearningPath` the common fields will go down substantially.

mbertrand: Maybe I'm misunderstanding but I'm trying to emulate the current code pattern where for example there are separate types for `Post` and `Comment` (though they have `AuthoredContent` type for common fields) and separate `searchResultToPost` and `searchResultToComment` functions to convert the ES search results to their respective types.

aliceriot: to get around flow complaining fields can be typed as optional. yeah, I'm not sure exactly how we do it, I'm just seeing how much code is being added just to add one more type of result (bootcamps) and that has me worried about dealing with tons of similar code when we add programs and learning paths as well, so I think if we can get out in front of it by trying to re-use as much as possible we'll be in a better position going forward.

it seems to me that the display we want to have in the learning resource card is mostly the same for all of these types of resources, and then there are differences in how we want to display them in the learning resource drawer. to me the way to handle that that makes things easiest is to treat these things the same way everywhere except where we actually need to render something different, like the list of things in the drawer display for a learning path or similar. it's a bit analogous to how we don't worry about whether a post is a link post or a text post unless we're actually at the point of rendering it's content.

anyhow, if you don't agree that's fine, we can keep this as-is for now. I just want to make sure we don't end up with a lot of very similar code for handling these different things that's hard to maintain later.

mbertrand: I think I might understand now, the `LearningResource` type would be independent of the existing `Bootcamp` and `Course` types (still needed for the expanded display in the drawer) and have only the fields displayed in the compact card?

static/js/lib/search.js
mbertrand: ✔️

static/js/reducers/bootcamps.js
mbertrand: I think I'd rather work on that for the next PR when I introduce the API's for learning paths and programs.  Will probably need to add `reselect` for redux-query to package.json & yarn.lock, plus rewrite the course API code, and this PR is already pretty big.

static/js/components/SearchResult.js
mbertrand: I did the above

static/js/reducers/bootcamps.js
aliceriot: sounds good, I'm adding reselect and fiddling a bunch with the courses API currently anyhow in my work for #2091

PR #2112, opened 2019-07-26
Renamed course_catalog.task_helpers to course_catalog.api

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
The methods in this module are more in line with what we put in api.py, so I've renamed it

Tests should pass

### DIFF
--- a/course_catalog/api_test.py
+++ b/course_catalog/api_test.py
@@ -1,5 +1,5 @@
 """
-Test course_catalog task_helpers
+Test course_catalog.api
 """
 import copy
 import json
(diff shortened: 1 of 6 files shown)


course_catalog/api_test.py
```diff
@@ -13,7 +13,7 @@
 from course_catalog.factories import CourseFactory
```
aliceriot: github won't let me comment on the relevant line, but the docstring up at the top of this file says 'Test course_catalog task_helpers'

PR #2121, opened 2019-07-30
Added CourseRun model

### LINKED ISSUE #2114: Data model for course runs
We need a data model for course runs such that a course can have multiple course runs. This models should have most (if not all) of the fields that are currently on `Course`. There are also a number of fields that are common already between `Bootcamp` and `Course` that we'll likely need here, so we should also pull those out into an abstract model to tamp down on the duplicate code now that we've got a more solid idea how we're modeling this data.

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations

Closes #2114 

#### What's this PR do?
Adds a data model for `CourseRun`. I kept the majority of the ones that were on `course`, except those that were OCW-specific.

You should be able to run the app and the migrations.


rhysyngsun:
I'll remove the `raw_json` in this PR.

rhysyngsun:
@mbertrand done

course_catalog/models.py
```diff
@@ -72,45 +67,70 @@ class Course(LearningResource):
...
+
+
+class Course(AbstractCourse):
+    """
+    Course model for courses on all platforms
+    """
+
+    course_id = models.CharField(max_length=128, unique=True)
+
+    level = models.CharField(max_length=128, null=True, blank=True)
```
mbertrand: `level` is also defined for each EdX course run, they're probably going to be the same nearly all the time but it can't be guaranteed.  So this field should probably be included for `CourseRun` too.

course_catalog/models.py
```diff
@@ -72,45 +67,70 @@ class Course(LearningResource):
...
+
+
+class CourseRun(AbstractCourse):
+    """
+    Model for course runs
+    """
+
+    course_run_id = models.CharField(max_length=128, unique=True)
+    semester = models.CharField(max_length=20, null=True, blank=True)
+    platform = models.CharField(max_length=128)
```
mbertrand: `platform` should be the same for every run in a `Course`, so it might not be needed here?

rhysyngsun:
@mbertrand all set

Issue #2124, opened 2019-07-31
Update course drawer and search results to use an appropriate course run


rhysyngsun:
The changes in #2115, #2116, and #2122 mean that the data used by the drawer has moved into a nested `course_run` list. This means for MITX courses we need to pick an appropriate `CourseRun` out of this list to open the drawer for. 

The appropriate `CourseRun` differs depending on the user's current search criteria on the Availability facet (note the changes to the Availability facet in #2123):

- **No filter** the first `CourseRun` found in this priority order:
  - The current course run (as defined below)
  - The first future course run (as defined below)
  - The most recent past course run
- **Current** - the first `CourseRun` found in this priority order:
  - A course run with `enrollment_start` in the past and `enrollment_end` in the future
  - The most recent course run with `enrollment_start` and `enrollment_end` in the past
- ** Future**
  - The course run with the lowest `enrollment_start` set in the future

We should open the drawer with that course run displayed.

We also want a dropdown to switch between the other course runs on the course.

mbertrand:
@abdulkdawson is there a design for this that includes a dropdown for selecting an alternate run?

abdulkdawson:
@mbertrand Yes it is being tweaked as we speak.

mbertrand:
[In the design](https://projects.invisionapp.com/d/main#/console/17477251/382578136/preview), semester and year are shown in the course run dropdown.  What should the alternative be for runs that have neither course nor year?  There are about 12 of them, and 6 of those are also lacking any start/enrollment date.  Show the start date if available (or convert it to a semester/year), and if start date is also missing show the course run id instead?

```
'course-v1:MITx+6.002x+2013',
 'course-v1:MITx+6.002x+MITx_2012_Alumni',
 'course-v1:MITx+CTL.CFx+2T17U8',
 'course-v1:MITx+CTL.CFx+1T18MU',
 'course-v1:MITx+CTL.CFx+1T18U1',
 'course-v1:MITx+CTL.CFx+1T18U5',
 'course-v1:MITx+CTL.CFx+1T18U8',
 'course-v1:MITx+HR3+2017_Q3',
 'course-v1:MITx+CTL.CFx+2T17MU',
 'course-v1:MITx+CTL.CFx+2T17U1',
 'course-v1:MITx_PRO+ENx+InternalOnly',
 'course-v1:MITx+CTL.CFx+2T17U5'
```

mbertrand:
edx, xpro - use start date if available, else best available date
[...]

Issue #2116, opened 2019-07-29
Index courses and course runs


rhysyngsun:
We need to index courses and course runs coming form MITX appropriately. We'll need to refactor our current course indexing datastructure to support nested documents that are course runs because we want each result to logically be one course and that's really the only way to do that kind of grouping.

ES documentation on nested documents: https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html

Here's an example of how we did this in Micromasters for enrollments in different course runs: https://github.com/mitodl/micromasters/blob/master/search/indexing_api.py#L190-L199

The `CourseRun`s should be indexed as a list of objects under the key `course_runs`

This may temporarily break the Availability search facet until we update the frontend, but since this course search isn't released yet we're ok with this. This will be fixed in #2123 afterwards.

PR #2130, opened 2019-08-01
Import courses and course runs for MITx

### LINKED ISSUE #2115: mitx course catalog import should write to course runs
How we write course and course run data needs to be updated the data model changes introduced in #2114 when importing from mitx. We should safely be able to make this change without any feature flag branching because it will actually get us to our desired behavior of having one result per logical course immediately, with the problem that the course run metadata isn't searchable. It may break some facets in the short term, but it shouldn't in theory cause any errors in the application.

We should put values available in the top level course object into `Course` and the values from the `course_runs` list into `CourseRun` with the foreign key to `Course` set.

As part of this, we should have a command that repairs the `Course` table by dropping all MITX `Course` records and kicking off the task to reimport them.

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
- [x] Settings

- Closes #2115
- Adds an overlooked setting to api.json and settings.py (`MAX_S3_GET_ITERATIONS`)

#### What's this PR do?
Imports MITx data into both `Course` and `CourseRun` objects.

- Set required .env variables (you can copy from heroku CI):
```
EDX_API_URL
EDX_API_CLIENT_ID
EDX_API_CLIENT_SECRET
OCW_*
```
- Run `docker-compose run web python manage.py backpopulate_edx_courses --delete`
- Verify that `Course` objects with `platform=mitx` were created, and each of these courses have at least 1 `CourseRun` object.
- Verify that the `Course` objects have associated `CourseTopic` objects, and that the `CourseRun` objects have associated `CourseInstructor` and `CoursePrice` objects.
- Go to `http://od.odl.local:8063/courses/search`, there should be results for MITx platform courses, and they should display almost the same as before (except they will always be shown as free because the price info is now in the course run - this will need to be fixed in #2116).  
- Clicking on any MITx course card should still bring up the drawer with more details.  However some of those details will be missing because they'e now stored in the CourseRun model (to be fixed in #2124)
- Try importing some OCW courses too, there shouldn't be any issues with those:
  `docker-compose run web python manage.py backpopulate_ocw_courses --delete`

### DIFF
--- a/course_catalog/api.py
+++ b/course_catalog/api.py
@@ -74,9 +74,7 @@ def parse_mitx_json_data(course_data, force_overwrite=False):
         compare_datetime = datetime.strptime(
             course_modified, "%Y-%m-%dT%H:%M:%S.%fZ"
         ).astimezone(pytz.utc)
-        needs_update = (
-            compare_datetime <= course.last_modified
-        ) or force_overwrite
+        needs_update = (compare_datetime >= course.last_modified) or force_overwrite
         index_func = update_course
     except Course.DoesNotExist:
         index_func = index_new_course
@@ -123,7 +121,7 @@ def parse_mitx_json_data(course_data, force_overwrite=False):
                     # Try and get the CourseRun instance. If it exists check to see if it needs updating
                     try:
                         courserun_instance = CourseRun.objects.get(
-                            course_run_id=course_run.get("key"), course=course
+                            course_run_id=course_run.get("key")
                         )
                         compare_datetime = datetime.strptime(
                             max_modified, "%Y-%m-%dT%H:%M:%S.%fZ"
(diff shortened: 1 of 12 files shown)


course_catalog/api.py
```diff
@@ -52,66 +57,104 @@ def parse_mitx_json_data(course_data, force_overwrite=False):
...
+                        continue
+
+                    # Get the last modified date from the course run
+                    course_run_modified = course_run.get("modified")
+
+                    # Since we use data from both course and course_run and they use different modified timestamps,
+                    # we need to find the newest changes
+                    max_modified = (
+                        course_modified
+                        if course_modified > course_run_modified
```
rhysyngsun: These values are actually strings so we need to parse them to get the correct value.

rhysyngsun: It can also be shortened to `max(course_modified, course_run_modified)`.

mbertrand: 👍

PR #2132, opened 2019-08-02
Updated course_catalog factories to be generate more data out of the box

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Pre-req for #2116 

#### What's this PR do?
Updates the course_catalog factories to generate more thorough data out of the box (e.g. `CourseFactory.create()` will give you a course with runs (if it's not an OCW course).

Tests should still pass and you should be able to create an MITx course and see that it has runs:

```
from course_catalog.factories import *
course = CourseFactory.create(is_mitx=True)
```
(Optional)

(Optional)

(Optional)

(Optional)

### DIFF
--- a/course_catalog/factories.py
+++ b/course_catalog/factories.py
@@ -1,8 +1,12 @@
 """Factories for making test data"""
+import random
+from datetime import timedelta
+
 import factory
 from factory import SubFactory
 from factory.django import DjangoModelFactory
 from factory.fuzzy import FuzzyChoice, FuzzyText
+import pytz
 
 from django.contrib.contenttypes.models import ContentType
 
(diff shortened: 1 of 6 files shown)


course_catalog/factories.py
```diff
@@ -125,31 +196,11 @@ class CourseRunFactory(DjangoModelFactory):
...
-
     class Meta:
         model = CourseRun
 
 
-class BootcampFactory(DjangoModelFactory):
+class BootcampFactory(AbstractCourseFactory):
     """Factory for Bootcamps"""
 
     course_id = factory.Sequence(lambda n: "BOOTCAMP%03d.MIT" % n)
```
mbertrand: The `BootcampSerializer` always sets `offered_by` to `OfferedBy.bootcamps.value`.  Maybe we should do the same for the factory?

```python
offered_by = OfferedBy.bootcamps.value
```

rhysyngsun:
@mbertrand all set

Issue #2101, opened 2019-07-23
Update frontend code to additionally search for `program` and `pathway` record types


pdpinch:
As a user searching for learning resources, I'd like to be able filter on programs 

#### Designs and Mockups

See https://invis.io/XRRX2PH5KZM#/367253746_Open_Courses_with_List_Filter_wAnnotations

#### Acceptance Criteria:

- [ ] Add programs filter to Learning Resources facet
- [ ] Add pathways filter to Learning Resources facet

mbertrand:
In the design for programs/learning paths, [a description is shown](https://projects.invisionapp.com/share/XRRX2PH5KZM#/screens/367253745).  The Program/List model does not currently have a description field. Should I add one?

mbertrand:
@pdpinch @ferdi - the current elasticsearch index document mapping for programs and pathways includes title and short description, but none of the items in each list.  Is that okay?  The consequence is that if you search for a particular course title, the results won't include any program or path that the course is a part of, unless it just happens to be included in the program/pathway title or description.

pdpinch:
@mbertrand can this be closed?

PR #2133, opened 2019-08-02
Add Program and UserList to Django admin

mbertrand:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Indirectly related to #2101, makes it easier to view/edit Program and UserList objects 

#### What's this PR do?
Adds admin pages for Program and UserList objects; fixes the `limit_choices_to` value in those models.

Go to Django admin and you should be able to create/edit/view Programs and UserLists.


aliceriot:
just one thing I noticed, when making a `UserList` the `image` field is required - should we be requiring that? it makes sense that the program require it, since they are more curated data, but for user-entered things it seems a bit weird to require an image.

other than that works great! I can make programs and userlists and so on.

mbertrand:
@aliceriot I added `blank=True` to the `Program` and `UserList` models (they were already `null=True`), so the field doesn't need to be filled out in the admin form anymore.

PR #2137, opened 2019-08-05
Index course runs under courses

### LINKED ISSUE #2116: Index courses and course runs

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #2116 

#### What's this PR do?
Expands course indexing to also include course runs

- Run `./manage.py recreate_index`
- Verify the search page is still as functional as `master`
- You should also see nested `course_runs` data coming back in `_source` on the results
- Try running a query to find courses with specific course runs. An example one is provided in this PR:
```
curl --header "Content-Type: application/json" \
  --request POST -s \
  --data "$(cat test_json/es/2015_course_runs.json)" \
  http://localhost:9101/local_all_default/_search | jq
```

### DIFF
--- a/search/indexing_api.py
+++ b/search/indexing_api.py
@@ -115,6 +115,31 @@ COURSE_OBJECT_TYPE = {
     "published": {"type": "boolean"},
     "availability": {"type": "keyword"},
     "offered_by": {"type": "keyword"},
+    "course_runs": {
+        "type": "nested",
+        "properties": {
+            "id": {"type": "long"},
+            "course_id": {"type": "keyword"},
+            "title": ENGLISH_TEXT_FIELD,
+            "short_description": ENGLISH_TEXT_FIELD,
+            "full_description": ENGLISH_TEXT_FIELD,
+            "language": {"type": "keyword"},
+            "level": {"type": "keyword"},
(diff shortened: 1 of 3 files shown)


mbertrand:
PS The above will also be relevant to price if/when we add a facet for that.

test_json/es/2015_course_runs.json
```diff
@@ -0,0 +1,61 @@
+{
```
mbertrand: I don't think this file is used for anything?

PR #2148, opened 2019-08-09
Added retire_users command and don't email or index inactive users

### LINKED ISSUE #1698: management command for retiring users
Similar in goal to [`retire_users` in MicroMasters](https://github.com/mitodl/micromasters/blob/master/profiles/management/commands/retire_users.py)

### acceptance criteria
- [ ] mark the user as inactive
- [ ] remove any notification objects / email optins 
- [ ] delete socialauth objects 

### out of scope
- we won't remove posts or comments

### questions
**Q: What should we do about profile pages?**
**partial answer** if nothing else, we can't continue to display a profile page. Perhaps we check the `is_active` bit on the account before rendering the profile page, and return a 404 if it's false.

### LINKED ISSUE #2076: disable profile page when user account is inactive
Related to #1698, as a user who wants my account deleted, I would like my profile to be in accessible as well. 

#### Acceptance Criteria:

- [ ] return a 404 for profile pages when the user account as has `is_active` set to False

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #1698
Fixes #2076 
Part of #2118

#### What's this PR do?
- Adds a mangement command to retire users
- Updates email notifications to not send to inactive users
- Disables user profile page if inactive
- Don't index inactive users

- Run `./manage.py retire_user [--email] [--username] [--id]` and verify the user is deactivated
- Verify the user doesn't receive daily or weekly email notifications
- Verify the user profile page results in a 404 page
- Run `./manage.py recreate_index`


pdpinch:
Thank you for adding the --email and --username switches. This will be easier to use than the [micromasters version](https://github.com/mitodl/micromasters/blob/master/profiles/management/commands/retire_users.py)

pdpinch:
What about channel invitations? 

What about socialauth objects? 

What will happen to the user's info in the search index?

rhysyngsun:
I think updating the search index is the only one I didn't account for here yet, so I'll add that. Your last question also flagged a separate issue that inactive users aren't filtered out when we reindex, so I'll address that in this PR too since there's no point in removing them if they'll just be pulled in again on a reindex.

rhysyngsun:
Yeah, those are probably important to fix too.

rhysyngsun:
Removing them from the contributor list is going to be more involved since that's directly hitting reddit at the moment. @pdpinch is this something you consider critical to get this out, or should I just open an issue and move ahead on this as-is?

pdpinch:
> A deleted user still gets notifications if someone responds to that user's post or comment

If we make the users account inactive and stop-sending notifications to inactive accounts, doesn't that stop the notifications?

> A deleted user still shows up in the list of channel members/contributors

I'm not sure that's a problem. Their posts still exist so, on some level, they are still a member. In any case, let's open an issue to revisit that and move ahead with this PR.

PR #2153, opened 2019-08-13
CourseRuns for all courses and bootcamps

mbertrand:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

(Required)

#### What's this PR do?
(Required)

(Required)

(Optional)

(Optional)

(Optional)

(Optional)


search/serializers.py
```diff
@@ -333,39 +325,24 @@ class ESCourseSerializer(ESModelSerializer):
...
+    course_runs = serializers.SerializerMethodField()
 
-    def get_instructors(self, course):
+    def get_course_runs(self, instance):
         """
-        Get a list of instructor names for the course
+        Get the course runs in reverse chronological order
         """
-        return [" ".join([i.first_name, i.last_name]) for i in course.instructors.all()]
+        course_runs = instance.course_runs.all().order_by(
```
rhysyngsun: Defining the `order_by` here is negating the `pretch_related` I think because it's creating a copy of the query for each `Course` instance. So I think what we probably want here to simplify things is to define the original query like this:


```python
from django.db.models import OuterRef, Subquery, Prefetch

Course.objects.prefetch_related(
    Prefetch(
        "course_runs",
        queryset=Subquery(
            CourseRun.objects.filter(
                course_id=OuterRef('pk')
            ).order_by(
                 "-enrollment_start", "-start_date", "-year"
            )
        )
    )
)
```

Then you can change this back to a vanilla `ESCourseRunSerializer(many=True, allow_null=True)` definition.

We'll want to apply this to both the `ViewSets` and the ES indexing code.

Also, `all()` isn't needed if you're calling at least one other `ObjectManager` method.

mbertrand: I'm not sure where that query definition is supposed to go? The model definition for `course_runs` is just `course_runs = GenericRelation(CourseRun)`

rhysyngsun: It would go in these two places:

https://github.com/mitodl/open-discussions/pull/2153/files#diff-44a0e3de85d4ee6ea1e4717ef96cad83L102
https://github.com/mitodl/open-discussions/pull/2153/files#diff-a8d441b8cd2ce3c69a44ce0d1994a23cL601

Similar code would be done for bootcamps in their related locations

mbertrand: I had to change the above a bit to get it working, using a subquery was returning 0 runs:
```python
for course in Course.objects.filter(id__in=ids).prefetch_related(
        "topics",
        Prefetch(
            "course_runs",
            queryset=CourseRun.objects.order_by(
                "-enrollment_start", "-start_date", "-year"
            ),
        ),
    )
```

rhysyngsun:
@mbertrand while testing #2170, I had forgotten to rerun the backpopulate scripts (so by course catalog tables were all empty), and I got this error and the entire site crashed:

```
TypeError: "upcomingCourses is null"
    mapStateToProps NewCoursesWidget.js:88
    Redux 6
    React 14
    onStateChange Redux
    wrappedMethod react-hot-loader.development.js:582
    Redux 3
    wrappedMethod react-hot-loader.development.js:582
    dispatch Redux
    e (index):1
    Redux 7
    callback client.js:611
    Request client.js:436
    emit index.js:133
    onreadystatechange client.js:705
react-hot-loader.development.js:175
```

So I think this requires a fix in `NewCoursesWidget` to tolerate no courses in order to ensure the site doesn't crash between when the migrations run and we have a chance to backpopulate again.

mbertrand:
@rhysyngsun I actually can't reproduce that error locally, it works when I have no courses and no bootcamps.  Are you getting any errors in the API calls that may be trickling down to the frontend? - http://od.odl.local:8063/api/v0/courses/new/, http://od.odl.local:8063/api/v0/courses/upcoming

Screen Shot 2019-08-22 at 2 27 42 PM

mbertrand:
@rhysyngsun I could reproduce that error if I reverted the PR migration but kept the PR code running - because the code doesn't match the schema, the API calls to courses fail and that causes the frontend to blow up as well. So I'll tweak the frontend to handle that situation

PR #2156, opened 2019-08-14
some dependency upgrades

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none, just some dep upgrades

#### What's this PR do?

upgrades a bunch of packages, fixes some tests that were broken (mainly by `sinon.assert` now using a deep equality), and removes Searchkit.

everything should work, tests should pass. note that b/c of a circular dependency issue I had to run a dependency flattener - so the delta in `yarn.lock` is a little bigger than it would have been otherwise. because of this though it's probably a good idea to do some substantial checking to make sure that everything is good.


webpack.config.shared.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
   config: {
     entry: {
       root:  ["@babel/polyfill", "./static/js/entry/root"],
+      root:  ["core-js/stable", "regenerator-runtime/runtime", "./static/js/entry/root"],
```
rhysyngsun: Should the line above be removed or should this be under a different key?

aliceriot: lol :woman_facepalming:

PR #2165, opened 2019-08-20
fix program list favorite

### LINKED ISSUE #2164: favoriting / unfavoriting user lists doesn't work on favorites carousel
### Steps to Reproduce

1. add a user list
2. favorite it on the course search
3. go to the course home page (`/courses`)
4. try to unfavorite it

### Expected Behavior

the user should be able to unfavorite the user list here

### Actual Behavior

instead it issues a request to courses endpoint. there's two issues here:

- we currently merge in an object type of `LR_TYPE_COURSE` in the course carousel, which isn't what we want because we actually do have things other than courses in there. so instead we should at the redux-query level merge in an object type which reflects what that object actually is
- then there's another disconnect where we are looking for user lists to have `LR_TYPE_USERLIST: "user_list"` but the `content_type` that comes back in the favorites API is actually `"user list"` (not `"user_list"`). this value comes from `ContentType.objects.get_for_model(UserList).name`. this can be overridden by setting the `verbose_name` of the model to `user_list`, so it matches the content type we're using elsewhere.

aliceriot:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

closes #2164 

#### What's this PR do?

this fixes an issue with unfavoriting the user lists in the favorites carousel on the course home page. see the issue for a longer description of the issues and repro steps.

checkout master and follow the repro steps on the issue to ensure you can repro the issue. then check this out and confirm it fixes the issue.


static/js/lib/queries/learning_resources.js
```diff
@@ -8,7 +8,7 @@ import {
   LR_TYPE_COURSE,
   LR_TYPE_BOOTCAMP,
   LR_TYPE_PROGRAM,
-  LR_TYPE_USERLIST_FAV
```
mbertrand: Remove this from `lib/constants.js` too

aliceriot: oh yeah, I meant to circle back and do that but forgot

aliceriot:
@mbertrand ok took care of that

PR #2188, opened 2019-08-30
upgrade react-redux, redux-query, react-router

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

this upgrades react-redux, redux-query, and react-router. doing so requires some changes!

the main thing is that react-redux and redux-query are no longer compatible with enzyme shallow rendering, because they use react's hooks APIs internally and hooks aren't supported in React's shallow renderer. The React team has basically said that they do not plan to add shallow rendering support for hooks, so it is up to enzyme or similar to do that if they want to. Enzyme has implemented partial support themselves, but it doesn't seem like full support is very close.

So anyway, I wanted to see how difficult it would be to move away from shallow rendering so we could clear the way for this upgrade - turns out, not so bad really. The main difficulty is that a lot of tests needed to be changed a little bit. I was able to drop `mount` in for `shallow` in the hoc renderer configuration method, wrap it in a `Router` and go from there. Surprisingly, the test suite isn't any slower! I thought it might be but was pleasantly surprised. The one thing I had to add to get things working was a method on `IntegrationTestHelper` for mocking out components - we had some tests which passed in shallow rendering because the child components weren't rendering, and whether those components render or not wasn't important to those tests anyhow, so we might as well mock them out instead of dealing with figuring out all the props they need.

Anyhow, this means we can use hooks anywhere we want now, and, in particular, we can now use the new, hooks-based APIs for redux and redux-query, which are super nice :tada: 

make sure that the app is working everywhere! and read through the code changes and make sure they make sense.


static/js/containers/CourseSearchPage_test.js
```diff
@@ -171,8 +175,9 @@ describe("CourseSearchPage", () => {
         })
       )
     })
+    wrapper.update()
     // from is 5, plus 5 is 10 which is == numHits so no more results
-    assert.isFalse(inner.find("InfiniteScroll").prop("hasMore"))
+    assert.isFalse(wrapper.find("InfiniteScroll").prop("hasMore"))
```
noisecapella: Should this `"InfiniteScroll"` be unquoted? and below as well?

static/js/hoc/withChannelTracker_test.js
```diff
@@ -1,28 +1,28 @@
...
 import { shouldIf, TestPage } from "../lib/test_utils"
 import IntegrationTestHelper from "../util/integration_test_helper"
 
 describe("withTracker", () => {
-  let helper, render, gaGaStub, channel, WrappedPage
+  let helper, gaGaStub, channel
+
+  const WrappedPage = withChannelTracker(TestPage)
+
+  const render = (state, props) => {
```
noisecapella: Should these have default arguments? The `props` one should if it's meant to replace `configureHOCRenderer` since it also has a default argument for `props`.

static/js/util/integration_test_helper.js
```diff
@@ -180,48 +202,28 @@ export default class IntegrationTestHelper {
...
-        }
-      }
-      // one more time to shallow render the InnerComponent
-      inner = await inner.dive()
+      this.wrapper = wrapper
+
+      const inner = wrapper.find(InnerComponent).exists()
+        ? wrapper.find(InnerComponent)
+        : wrapper
+          .findWhere(
```
noisecapella: Elegant!

static/js/hoc/withChannelTracker_test.js
aliceriot: sure, makes sense

static/js/containers/CourseSearchPage_test.js
aliceriot: doesn't seem to matter either way, so I think I'll make them all quoted for consistency

aliceriot:
ok those things should be addressed now

aliceriot:
@noisecapella I also added some `resolutions` to `package.json` to force a few dependency upgrades, I think none are a terribly big deal but might be worth just poking around the app to make sure everything is ok

PR #2190, opened 2019-09-03
Show most relevant availability for search result

### LINKED ISSUE #2162: Course search: start date on results card should reflect current search filter
As a user, I'd like to see course search results that reflect the filter I've applied, with the start date on the card reflecting my selection on the start date facet. 

For example, if I select a start date "in the next 90 days" I should only see start dates that are starting in the next 90 days. 

#### Designs and Mockups

https://invis.io/XRRX2PH5KZM#/378191756_Open_Courses_list_View_3_test

#### Acceptance Criteria:

- [ ] If there are two runs of the same course, one starting in 30 days and one starting 90 days, it should appear in the search results, but the start date should change to fit the currently selected facet.

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes #2162

#### What's this PR do?
- Sends the selected availability facets (if any) to the `LearningResourceCard` card component
- Adds functions for determining the best run/availability to display

- Select 'Available Now' facet only, all the results should show 'Available Now'
- Select 'In the next year' only, results should show a mix from 'In the next week' to 'In the next year'
- Select 'Available Now' and 'In the next year', results should show a mix from 'Available Now' to 'In the next year'.  
- Select 'Programs' and 'Learning Paths' facets, they should all be 'Available Now'

#### Screenshots

Screen Shot 2019-09-03 at 4 36 21 PM

Screen Shot 2019-09-03 at 4 37 10 PM

### DIFF
--- a/static/js/factories/learning_resources.js
+++ b/static/js/factories/learning_resources.js
@@ -24,7 +24,7 @@ import type {
 const incrCourse = incrementer()
 const courseId: any = incrementer()
 
-const dateFormat = "YYYY-MM-DD[T]HH:mm:ss[Z]"
+export const dateFormat = "YYYY-MM-DD[T]HH:mm:ss[Z]"
 
 const incrCourseRun = incrementer()
 
--- a/static/js/factories/learning_resources.js
+++ b/static/js/factories/learning_resources.js
@@ -42,6 +42,8 @@ export const makeCourseRun = (): CourseRun => {
     level:             casual.random_element(["Graduate", "Undergraduate", null]),
     start_date:        casual.date(dateFormat),
     end_date:          casual.date(dateFormat),
+    best_start_date:        casual.date(dateFormat),
+    best_end_date:          casual.date(dateFormat),
     enrollment_start:  casual.date(dateFormat),
     enrollment_end:    casual.date(dateFormat),
     availability:      casual.random_element([
--- a/static/js/factories/learning_resources.js
+++ b/static/js/factories/learning_resources.js
@@ -42,8 +42,8 @@ export const makeCourseRun = (): CourseRun => {
     level:             casual.random_element(["Graduate", "Undergraduate", null]),
     start_date:        casual.date(dateFormat),
     end_date:          casual.date(dateFormat),
-    best_start_date:        casual.date(dateFormat),
-    best_end_date:          casual.date(dateFormat),
+    best_start_date:   casual.date(dateFormat),
+    best_end_date:     casual.date(dateFormat),
     enrollment_start:  casual.date(dateFormat),
     enrollment_end:    casual.date(dateFormat),
     availability:      casual.random_element([
--- a/static/js/factories/learning_resources.js
+++ b/static/js/factories/learning_resources.js
@@ -10,7 +10,8 @@ import {
   LR_TYPE_COURSE,
   LR_TYPE_BOOTCAMP,
   LR_TYPE_PROGRAM,
-  LR_TYPE_USERLIST
+  LR_TYPE_USERLIST,
+  DATE_FORMAT
 } from "../lib/constants"
 
 import type {
@@ -23,9 +24,6 @@ import type {
 
 const incrCourse = incrementer()
 const courseId: any = incrementer()
-
-export const dateFormat = "YYYY-MM-DD[T]HH:mm:ss[Z]"
-
 const incrCourseRun = incrementer()
 
 export const makeCourseRun = (): CourseRun => {
@@ -40,12 +38,12 @@ export const makeCourseRun = (): CourseRun => {
     semester:          casual.random_element(["Fall", "Spring", null]),
     year:              casual.year,
     level:             casual.random_element(["Graduate", "Undergraduate", null]),
-    start_date:        casual.date(dateFormat),
-    end_date:          casual.date(dateFormat),
-    best_start_date:   casual.date(dateFormat),
-    best_end_date:     casual.date(dateFormat),
-    enrollment_start:  casual.date(dateFormat),
-    enrollment_end:    casual.date(dateFormat),
+    start_date:        casual.date(DATE_FORMAT),
+    end_date:          casual.date(DATE_FORMAT),
+    best_start_date:   casual.date(DATE_FORMAT),
+    best_end_date:     casual.date(DATE_FORMAT),
+    enrollment_start:  casual.date(DATE_FORMAT),
+    enrollment_end:    casual.date(DATE_FORMAT),
     availability:      casual.random_element([
       COURSE_ARCHIVED,
       COURSE_CURRENT,
--- a/static/js/lib/learning_resources.js
+++ b/static/js/lib/learning_resources.js
@@ -10,6 +10,7 @@ import {
 } from "./constants"
 import { capitalize, emptyOrNil } from "./util"
 import { AVAILABILITY_MAPPING } from "./search"
+import moment from "moment"
 
 export const availabilityFacetLabel = (availability: ?string) => {
   const facetKey = availability ? AVAILABILITY_MAPPING[availability] : null
@@ -26,6 +27,19 @@ export const availabilityLabel = (availability: ?string) => {
     return availability
   }
 }
+
+export const sortedAvailabilities = (availabilities: Array<string>) => (
+  availabilities.sort((a,b) => AVAILABILITY_MAPPING[a] && AVAILABILITY_MAPPING[b]
+    ? AVAILABILITY_MAPPING[a].order - AVAILABILITY_MAPPING[b].order
+    : 1))
+
+export const inDateRanges = (run: CourseRun, availabilities: Array<string>) => {
+  availabilities.forEach()
+}
+
+export const filterRunsByAvailability = (runs, availabilities) => (
+  runs.filter((run) => inDateRanges(run, availabilities))
+)
 
 export const resourceLabel = (resource: string) => {
   return resource === LR_TYPE_USERLIST
--- a/static/js/lib/learning_resources.js
+++ b/static/js/lib/learning_resources.js
@@ -1,5 +1,5 @@
 //@flow
-import {concat, isNil} from "ramda"
+import { concat, isNil } from "ramda"
 
 import {
   COURSE_ARCHIVED,
@@ -9,10 +9,10 @@ import {
   LR_TYPE_USERLIST
 } from "./constants"
 import { capitalize, emptyOrNil } from "./util"
-import {AVAILABILITY_MAPPING, AVAILABLE_NOW} from "./search"
+import { AVAILABILITY_MAPPING, AVAILABLE_NOW } from "./search"
 import moment from "moment"
-import {dateFormat} from "../factories/learning_resources";
-import type {CourseRun} from "../flow/discussionTypes";
+import { dateFormat } from "../factories/learning_resources"
+import type { CourseRun } from "../flow/discussionTypes"
 
 export const availabilityFacetLabel = (availability: ?string) => {
   const facetKey = availability ? AVAILABILITY_MAPPING[availability] : null
@@ -30,14 +30,14 @@ export const availabilityLabel = (availability: ?string) => {
   }
 }
 
-export const parseDateFilter = (filter) => {
+export const parseDateFilter = filter => {
   const format = /(now)(\+)?(\d+)?([Md])?/
   const match = format.exec(filter)
   if (match) {
-     let dt = moment()
-     if (match[3] && match[4]){
-       dt = dt.add(match[3], match[4] === "d" ? "days" : "months")
-     }
+    let dt = moment()
+    if (match[3] && match[4]) {
+      dt = dt.add(match[3], match[4] === "d" ? "days" : "months")
+    }
     return dt
   }
 }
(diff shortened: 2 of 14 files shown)


mbertrand:
@Ferdi @pdpinch @abdulkdawson screenshots attached

aliceriot:
functionality looks pretty good!

aliceriot:
reading code now :)

static/js/lib/learning_resources.js
```diff
@@ -27,24 +34,131 @@ export const availabilityLabel = (availability: ?string) => {
...
+  for (const range in AVAILABILITY_MAPPING) {
+    if (inDateRanges(run, [range])) {
+      return AVAILABILITY_MAPPING[range].label
+    }
+  }
+}
+
+export const compareRuns = (firstRun: CourseRun, secondRun: CourseRun) =>
+  moment(firstRun.best_start_date || defaultStartDate, dateFormat).diff(
+    moment(secondRun.best_start_date || defaultStartDate, dateFormat),
```
aliceriot: a general suggestion for a lot of this code: this call is repeated a whole bunch of times:

```js
moment(run.best_start_date || defaultStartDate, dateFormat)
```

along with a closely-related call for the end date. The code overall might be easier to read and follow if you write two little helpers, like `runStartDate` and `runEndDate`, that do this for you in one place.

static/js/lib/learning_resources.js
```diff
@@ -27,24 +34,131 @@ export const availabilityLabel = (availability: ?string) => {
...
+  )
+
+export const bestRun = (runs: Array<CourseRun>) => {
+  const now = moment()
+
+  // Runs that are running right now
+  const currentRuns = runs.filter(
+    run =>
+      moment(run.best_start_date || defaultStartDate, dateFormat).diff(now) <=
+        0 &&
```
aliceriot: I think it might be a little clearer for these bits of `.diff` logic to use moments `.isBefore` and `.isAfter` helper methods - you can do `momentObj.isBefore()` and if no argument is passed it defaults to `moment.now()`

static/js/factories/learning_resources.js
```diff
@@ -24,7 +24,7 @@ import type {
 const incrCourse = incrementer()
 const courseId: any = incrementer()
 
-const dateFormat = "YYYY-MM-DD[T]HH:mm:ss[Z]"
+export const dateFormat = "YYYY-MM-DD[T]HH:mm:ss[Z]"
```
aliceriot: I think we should probably move this constant to a constants file and rename it so its `ALL_CAPS` like we typically do for constants in the application code - first because we don't want any imports in the application code from the factory module, and second because when it's used in the `lib` file below `dateFormat` looks like it's a function, not a constant

static/js/lib/learning_resources.js
```diff
@@ -27,24 +34,131 @@ export const availabilityLabel = (availability: ?string) => {
   }
 }
 
+export const parseDateFilter = (filter: string) => {
```
aliceriot: so this function is I think taking the text and returning a moment object for comparing, right? two suggestions:

1. a name like `dateFilterToMoment` or something similar might better capture what the function does
2. we have a fixed set of possible values for the availability filter (in our `AVAILABILITY_MAPPING`) so it might be clearer to have a switch statement on those constant values, like

```js
switch (filter) {
  case AVAILABLE_NOW:
    return moment.now()
  case AVAILABLE_NEXT_WEEK:
    return moment.now().add(7, "days")
```

something along those lines, as-is it's pretty hard to follow what's going on here with the regex and everything.

static/js/lib/learning_resources.js
```diff
@@ -27,24 +34,131 @@ export const availabilityLabel = (availability: ?string) => {
...
+    }
+    return dt
+  }
+}
+
+export const inDateRanges = (run: CourseRun, availabilities: Array<string>) => {
+  let inRange = false
+  availabilities.forEach(availability => {
+    if (AVAILABILITY_MAPPING[availability]) {
+      const from = parseDateFilter(
```
aliceriot: this will always be `moment.now()` I think, so the `parseDateFilter` function doesn't need to handle this case

static/js/lib/learning_resources.js
```diff
@@ -27,24 +34,131 @@ export const availabilityLabel = (availability: ?string) => {
...
+      dt = dt.add(match[3], match[4] === "d" ? "days" : "months")
+    }
+    return dt
+  }
+}
+
+export const inDateRanges = (run: CourseRun, availabilities: Array<string>) => {
+  let inRange = false
+  availabilities.forEach(availability => {
+    if (AVAILABILITY_MAPPING[availability]) {
```
aliceriot: is this a guard in case the user edits the url?

static/js/lib/learning_resources.js
```diff
@@ -27,24 +34,131 @@ export const availabilityLabel = (availability: ?string) => {
...
     ? "Learning Paths"
     : concat(capitalize(resource), "s")
 }
 
-export const maxPrice = (resource: Object) => {
-  const prices = !emptyOrNil(resource.course_runs)
-    ? resource.course_runs[0].prices
-    : []
+export const maxPrice = (courseRun: ?CourseRun) => {
+  // $FlowFixMe: prices property will exist if courseRun not emotyOrNil
```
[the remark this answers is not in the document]
aliceriot: (otherwise we're just turning it off everywhere and defeating the point of using it!)

static/js/lib/learning_resources.js
```diff
@@ -27,24 +34,131 @@ export const availabilityLabel = (availability: ?string) => {
...
+          isNil(to) ||
+          // $FlowFixMe: okay to compare moments
+          startDate <= to)
+      ) {
+        inRange = true
+      }
+    }
+  })
+  return inRange
+}
```
aliceriot: general question about this function, is the idea to return `true` if the `courseRun` is in 'range' for any of the availabilities supplied? because I think, based on the way it is now if there were an availability which wasn't "in range" after one which was we'd return `false` in a case where we want to return `true`.

static/js/lib/learning_resources.js
```diff
@@ -27,24 +34,131 @@ export const availabilityLabel = (availability: ?string) => {
...
+      if (
+        ((isNil(startDate) && availability === AVAILABLE_NOW) ||
+          isNil(from) ||
+          // $FlowFixMe: okay to compare moments
+          startDate >= from) &&
+        ((isNil(startDate) && availability === AVAILABLE_NOW) ||
+          isNil(to) ||
+          // $FlowFixMe: okay to compare moments
+          startDate <= to)
+      ) {
```
aliceriot: for this check here you might be able to use moment's `.isBetween` helper

static/js/lib/learning_resources.js
```diff
@@ -27,24 +34,131 @@ export const availabilityLabel = (availability: ?string) => {
...
+}
+
+export const bestRunLabel = (run: ?CourseRun) => {
+  if (!run) {
+    return AVAILABILITY_MAPPING[AVAILABLE_NOW].label
+  }
+  for (const range in AVAILABILITY_MAPPING) {
+    if (inDateRanges(run, [range])) {
+      return AVAILABILITY_MAPPING[range].label
+    }
```
aliceriot: ok, so if I'm reading this right, this will return the first entry `AVAILABILITY_MAPPING` that matches the dates for the course run, not necessarily one that the user has selected. is that the behavior that we want?

like right now if I selected `in the next year` I see a number of other labels too, like `in the next three months`, etc. this makes sense because obviously anything happening within the next three months is also happening in the next three months. but something happening in the next three months could _also_ be labeled, equally accurately although less specifically, `in the next year`.

I thought I remembered a discussion that the selected filter value should be the one we show, but I don't know. what do you think is correct?

mbertrand: I was thinking that the earliest time frame label that still fits within the user's selected time frame might be better/more informative, ie "in the next 3 months" is better than "in the next year".  @Ferdi or @pdpinch what do you think?

static/js/lib/learning_resources.js
mbertrand: `from` will be `null` for `Available Now`

aliceriot: gotcha, in that case, if it's `moment.now()` for all other cases it might be easier to just put that logic right here, like `from = availability === AVAILABLE_NOW ? null : moment.now()`

static/js/lib/learning_resources.js
mbertrand: The filters defined in `AVAILABILITY_MAPPING` are also used in `lib/search/buildSearchQuery`, so I'd prefer to also use them here rather than ignore them and replace with a longer switch statement that repeats some of the same logic and would risk getting out of sync with the `AVAILABILITY_MAPPING` filters.  Will definitely rename it though and can add an explanatory comment or two.

static/js/lib/learning_resources.js
mbertrand: yes

static/js/lib/learning_resources.js
mbertrand: yes, if the user selects "Available Now" and "In the next week" for example, a run should be returned if it's dates fall within any of those date ranges specified.  `inRange` is set to `false` initially, then all the user-selected availabilities will be iterated over, and if any one of them is a match, it will be set to `true`.  Only if it doesn't find any match for any of the availabilities will it still return `false`.

static/js/lib/learning_resources.js
aliceriot: ah, yeah I didn't think about it getting out of sync with the search logic, that's definitely a possibility. I might suggest then that instead we have a separate function that just takes the availability filter string and returns values that `moment` can use, so that part can be pulled out and be cleanly defined

static/js/lib/learning_resources.js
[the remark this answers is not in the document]
mbertrand: I modified the price functions to make the `FlowFixMe` unnecessary

static/js/lib/learning_resources.js
mbertrand: ✔️

static/js/lib/learning_resources.js
mbertrand: ✔️

static/js/factories/learning_resources.js
mbertrand: ✔️

static/js/lib/learning_resources.js
mbertrand: Ferdi said to keep it as is, working as expected.

static/js/lib/learning_resources.js
mbertrand: Replaced <= and >= with isSameOrBefore, isSameOrAfter. Avoided isBetween because of the possibility of null from/to values.

mbertrand:
@aliceriot ready for review again

mbertrand:
@aliceriot I just noticed a discrepancy for some courses in displayed availability, will work on fixing that

mbertrand:
Made an adjustment to account for ES availability aggregation counts being rounded to the nearest day. There was a course run ending in 7 days plus 1 hour, and ES aggregated this as 'in the next week', but the `inDateRanges` function considered this out of range because of the extra hour.  So I adjusted the function to use the beginning (00:00:00) of the day for `from` and the end of the day (23:59:59) for `to`.

aliceriot:
ok, taking another look now

static/js/lib/learning_resources.js
```diff
@@ -27,24 +32,130 @@ export const availabilityLabel = (availability: ?string) => {
...
+        AVAILABILITY_MAPPING[availability].filter.to,
+        true
+      )
+      const startDate = runStartDate(run)
+      if (
+        (isNil(from) ||
+          // $FlowFixMe: if we get this far, only moments are compared
+          startDate.isSameOrAfter(from)) &&
+        (isNil(to) ||
+          // $FlowFixMe: if we get this far, only moments are compared
```
aliceriot: I think you can remove these flow disables here if you switch `isNil(from)` to `!from`

static/js/lib/learning_resources.js
```diff
@@ -27,24 +32,130 @@ export const availabilityLabel = (availability: ?string) => {
   }
 }
 
+export const availabilityFilterToMoment = (filter: string, ending: boolean) => {
+  // Convert an Elasticsearch date_range filter string to a moment
```
aliceriot: can you add an explanation here of why it returns `undefined` for the availability now filter? it's a confusing detail that would be hard to get from just the code itself so it's good to explain I think

static/js/lib/learning_resources.js
```diff
@@ -27,24 +32,130 @@ export const availabilityLabel = (availability: ?string) => {
...
+export const runEndDate = (courseRun: CourseRun) =>
+  moment(courseRun.best_end_date || DEFAULT_END_DT, DATE_FORMAT)
+
+export const compareRuns = (firstRun: CourseRun, secondRun: CourseRun) =>
+  runStartDate(firstRun).diff(runStartDate(secondRun), "hours")
+
+export const bestRun = (runs: Array<CourseRun>) => {
+  // Runs that are running right now
+  const currentRuns = runs.filter(
+    // $FlowFixMe: runStartDate and runEndDate will always return a Moment
```
aliceriot: you might be able to get rid of this flow disable by adding a return type to those functions

static/js/lib/learning_resources_test.js
```diff
@@ -94,3 +105,245 @@ describe("Course utils", () => {
...
+      AVAILABILITY_MAPPING.availableNow.label
+    ],
+    [
+      "2020-08-02T00:00:00Z",
+      null,
+      ["nextYear"],
+      true,
+      AVAILABILITY_MAPPING.nextYear.label
+    ],
+    ["2030-08-02T00:00:00Z", null, ["nextYear"], false, null]
```
aliceriot: just curious, what does it mean for the start or end date to be null? what circumstances will that be the case?

mbertrand: The quality of the imported course run data is all over the place, some of them are missing start dates and/or end dates.

static/js/lib/learning_resources.js
```diff
@@ -27,24 +32,130 @@ export const availabilityLabel = (availability: ?string) => {
...
+      const startDate = runStartDate(run)
+      if (
+        (isNil(from) ||
+          // $FlowFixMe: if we get this far, only moments are compared
+          startDate.isSameOrAfter(from)) &&
+        (isNil(to) ||
+          // $FlowFixMe: if we get this far, only moments are compared
+          startDate.isSameOrBefore(to))
+      ) {
+        inRange = true
```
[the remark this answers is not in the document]
mbertrand: correct

static/js/lib/learning_resources.js
```diff
@@ -27,24 +32,130 @@ export const availabilityLabel = (availability: ?string) => {
...
+  // Convert an Elasticsearch date_range filter string to a moment
+  const format = /(now)(\+)?(\d+)?([Md])?/
+  const match = format.exec(filter)
+  if (match) {
+    let dt = moment()
+    if (match[3] && match[4]) {
+      dt = dt.add(match[3], match[4] === "d" ? "days" : "months")
+    }
+    if (ending) {
+      // $FlowFixMe: this is fine according to moment docs, and it works
```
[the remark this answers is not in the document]
mbertrand: ✔️

static/js/lib/learning_resources.js
mbertrand: ✔️

static/js/lib/learning_resources.js
mbertrand: ✔️

static/js/lib/learning_resources.js
mbertrand: ✔️

Issue #2163, opened 2019-08-20
Courses should be able to be offered by multiple platforms


rhysyngsun:
As a User I'd like to be able to see that a course is offered by multiple platforms (e.g. mitx and micromasters).

#### Designs and Mockups

List view: https://invis.io/XRRX2PH5KZM#/378191756_Open_Courses_list_View_3_test
Grid view: https://invis.io/XRRX2PH5KZM#/377036772_Open_Courses_with_List_FilterV2

#### Acceptance Criteria:

- [x] Data schema is updated and migrated to reflect this
- [x] Search indexing is updated to support this
- [x] Frontend is updated to support this in search and the drawer

#### Out of Scope

- Not going to do

pdpinch:
In the designs for this, we've decided to go with just text. The only combination of "offered by" brands that we expect in production is `["MicroMasters","MITx"]` which should always be rendered at "MITx MicroMasters" 

I will update the description with links to InVision

PR #2195, opened 2019-09-06
Added course catalog integration with micromasters

### LINKED ISSUE #2181: Search: ingest MM programs via API
Follow-up to https://github.com/mitodl/micromasters/issues/4353

#### Acceptance Criteria:

- [ ] periodically request program metadata from MicroMasters programs API
- [ ] merge MicroMasters metadata with edX course metadata 
  - edX metadata takes precedence, except for program metadata, i.e. program membership, program name, course order in program, etc. 
 - [ ] update search index

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [x] Settings

Fixes #2181 

#### What's this PR do?
Adds operations to pull in micromasters program data, which augments existing course/course run data from mitx.

- In micromasters, configure new `Course` and `CourseRun` records for a program in the RC environment (look at `/api/v0/catalog`) so it's configured against courses mitx is aware of. Double check that the courses are actually being imported. before you run the next step
- Run the `get_micromasters_data` task 
- Verify the `Program` now has `items` populated with `Course` references.

Recent efforts to adapt the DRF serializers currently used in `api.py` proved either impossible or time consuming to get right. This PR introduces a new `etl` module that defines `extract` and `transform` methods that are specific to each provider (only micromasters for now) and a normalized set of loaders to persist the data. This is far simpler and can handle much more complex cases than DRF serializers were designed to address.


course_catalog/etl/loaders.py
```diff
@@ -0,0 +1,45 @@
...
+    courses_data = program_data.pop("courses")
+
+    program, _ = Program.objects.update_or_create(
+        program_id=program_id, defaults=program_data
+    )
+
+    for position, course_data in enumerate(courses_data):
+        # skip courses that don't define a course_id
+        if not course_data.get("course_id", None):
+            continue
```
mbertrand: Might end up with empty programs if none of the courses in the program have course ids. For instance, on my local instance and on RC, the "Analog Learning" and "Digital Learning" programs.  Not sure how likely that is on production.

course_catalog/etl/micromasters.py
```diff
@@ -0,0 +1,37 @@
...
+            "program_id": program["id"],
+            "title": program["title"],
+            "url": program["programpage_url"],
+            "image_src": program["thumbnail_url"],
+            "offered_by": OfferedBy.micromasters.value,
+            # all we need for course data is the relative positioning of courses by course_id
+            "courses": [
+                {
+                    "course_id": course["edx_key"],
+                    # TDB: how to handle merging `offered_by` for courses and course runs
```
mbertrand: small typo: "TBD"

course_catalog/etl/loaders.py
```diff
@@ -0,0 +1,45 @@
...
+
+        course = load_course(course_data)
+
+        # create a program item or update its position
+        ProgramItem.objects.update_or_create(
+            program=program,
+            content_type=ContentType.objects.get(model="course"),
+            object_id=course.id,
+            defaults={"position": position},
+        )
```
mbertrand: Again not sure how likely this is to happen if ever in production, but if I remove a course from a program in MM, then rerun the import task, the course is not removed from the `Program` model here.

course_catalog/tasks.py
```diff
@@ -250,3 +251,9 @@ def get_bootcamp_data(force_overwrite=False):
         bootcamp_json = response.json()
         for bootcamp in bootcamp_json:
             parse_bootcamp_json_data(bootcamp, force_overwrite=force_overwrite)
+
+
+@app.task(acks_late=True)
+def get_micromasters_data():
```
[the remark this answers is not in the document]
rhysyngsun: :woman_facepalming: meant to add it to cron

course_catalog/etl/loaders.py
rhysyngsun: Yeah, that's definitely possible, but that amounts to a configuration mistake on MM. I'm inclined to leave it because it might be more obvious what is wrong if a program is empty rather than the program not showing up at all?

mbertrand: 👍

course_catalog/etl/loaders.py
rhysyngsun: It's not impossible, I could definitely imagine over time perhaps a course is retired, or maybe even renumbered, in which case we'd want to keep the course for historical reasons, but remove the linkage here.

course_catalog/etl/loaders.py
```diff
@@ -0,0 +1,45 @@
+"""Course catalog data loaders"""
+from django.contrib.contenttypes.models import ContentType
+
+from course_catalog.models import Course, Program, ProgramItem
+
+
+def load_course(course_data):
+    """Load the course into the database"""
+    course_id = course_data.pop("course_id")
+    course, _ = Course.objects.update_or_create(
```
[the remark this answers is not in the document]
rhysyngsun: The course data should be coming from mitx, and in the case of MM it _should_ exist there before it does in MM (@pdpinch can you confirm?). My plan is to eventually port all the ingestion for mitx, bootcamps, and ocw over to this more straightforward implementation. In those cases `course_data` will have other fields populated for those integrations, and MM becomes a special case where the integration exists only to join the courses under the program, but doesn't do anything additional.

To give a little more context on the direction this is headed, here's how the integration with xpro is going so far, building on the changes introduced in this PR: https://github.com/mitodl/open-discussions/commit/8f925d1532878348a4cc8ba9392a3545a274fb2d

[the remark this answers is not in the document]
pdpinch: I'm not 100% sure what you're asking me here. It's possible for a course to exist on MM before it does on edx.org, but we should consider edx.org to be canonical. In the case that there is a course in MM that does not exist on edx.org it should probably be ignored. 

I can outline a real-life scenario if that level of detail is helpful.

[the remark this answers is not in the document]
rhysyngsun: Yeah, can you give a concrete example around this? The overlapping integrations here complicate things, so having as much detail on data scenarios is helpful.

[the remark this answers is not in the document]
rhysyngsun: Actually, if a course exists in MM but not in edx, we can simply leave the `edx_key` blank on the `Course` in MM and that will cause that course to get skipped here. If that's good enough for you, it's good enough for me.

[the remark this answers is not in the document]
pdpinch: Earlier this year I had populated MM with a set of course keys with the run `3T2019`. However, due to scheduling changes at edx.org, that run didn't make sense any more and the courses on edX were given the run `2T2019a`. I later updated MM with correct course keys, which includes the correct run. 

I'm fine with the idea that the MM courses would be skipped when their keys didn't match the edx.org feed. I just hope that when they are corrected they can be re-ingested in the future.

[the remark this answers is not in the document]
rhysyngsun: It would be ideal not to have edx_key set on course, you can set course_key on the course run without issue though. If those keys change after being ingested it'll probably require manual cleanup since we're uniquing off the edx ids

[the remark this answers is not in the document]
rhysyngsun: Just to be explicit in case my earlier explanation wasn't direct, the course and course runs will both be skipped if the edx_key on course is not set. If it is set, the we add the course under the program. If not, the course will be imported only from edx. Although as I mentioned in the previous comment, we use the course/ course run keys as the unique identifiers of those records, so once something is synced, if the keys change it would look like a new course/run to us. I think that means we're OK though since discussions is the only thing using edx_key

rhysyngsun:
@mbertrand should be all set

course_catalog/etl/loaders.py
```diff
@@ -0,0 +1,55 @@
...
+
+    courses = []
+    course_content_type = ContentType.objects.get(model="course")
+
+    for position, course_data in enumerate(courses_data):
+        # skip courses that don't define a course_id
+        if not course_data.get("course_id", None):
+            continue
+
+        course = load_course(course_data)
```
mbertrand: if you add this code just before calling `load_course`:
```python
        course_data.update({
            "program_type": "MicroMasters",
            "program_name": program.title,
            "offered_by": OfferedBy.micromasters.value
        })
```
then you could get rid of the following:
- the [micromasters code block](https://github.com/mitodl/open-discussions/blob/28c4e91263e65558848325587e96f581d0feb356/course_catalog/api.py#L440-L447) in `course_catalog.api.tag_edx_program`
- the `course_catalog.api.get_micromasters_data` function
- `MICROMASTERS_COURSE_URL` setting and env variable

rhysyngsun: I think it might be best to leave this as-is for now so this PR stays additive in nature and because we need to be able to support #2163 in order to do that and we aren't tackling that yet. I'm also not sure having those `program_*` fields on the course is right given how our data is structured, but I think that requires further digging. Does that seem right?

mbertrand: ok, it just seems a little weird to update courses with data from 2 different MM URL's using separate functions sharing the same name, but it still works, I'm fine with optimizing this in a future PR.

rhysyngsun: Yeah, I agree, it's just sort of a slippery slope I think in terms of what'd get rolled into this change...

PR #2197, opened 2019-09-10
restyle the course home page

### LINKED ISSUE #2179: restyling for course home page
coursesfront

- [x] fix header text at top
- [x] make big red button text 'View All'
- [x] hide favorites if the user doesn't have any or if the user is anonymous (I think we can just hide whenever the response is empty)
- [x] fix styling for carousel titles
- [x] return to old 'previous' and 'next' button styles

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #2179

#### What's this PR do?

this does some styling tweaks to the course home page. also while I was fiddling with this component I refactored it to use the react-redux and redux-query hook APIs, and rewrote the tests to be tests of the full component.

make sure the styling fixes on the issue have all been satisfactorily implemented, and that it looks good and so on.

make sure all the code makes sense and everything.

updatedcousehome

mobile is pretty sad rn but this is what it looks like:

updatedcousehomemob

### DIFF
--- a/static/js/components/CourseCarousel.js
+++ b/static/js/components/CourseCarousel.js
@@ -1,12 +1,14 @@
 // @flow
 /* global SETTINGS:false */
-import React from "react"
+import React, { useState } from "react"
 import Carousel from "nuka-carousel"
+import R from "ramda"
 
 import LearningResourceCard from "./LearningResourceCard"
 
 import { SEARCH_GRID_UI } from "../lib/search"
-import { CAROUSEL_PAGE_SIZE } from "../lib/constants"
+import { CAROUSEL_PAGE_SIZE, PHONE, TABLET, DESKTOP } from "../lib/constants"
+import { useDeviceCategory } from "../hooks/util"
 
 import type { LearningResourceSummary } from "../flow/discussionTypes"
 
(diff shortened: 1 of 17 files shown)


Ferdi:
If it's not too hard can you display on card in the mobile view ?

aliceriot:
yeah - I haven't done any mobile stuff yet really but I can add that here

static/js/components/CourseCarousel.js
```diff
@@ -16,50 +17,66 @@ type Props = {|
...
-  return disabled ? null : (
-    <div onClick={nextSlide} className="carousel-control next">
-      <i className="material-icons">chevron_right</i>
+  return (
+    <div className="course-carousel">
+      <div className="title-and-controls">
+        <div className="title">{title}</div>
+        <div className="controls">
+          <button
+            className={`dark-outlined compact extra-compact prev ${
```
mbertrand: What do you think of moving these buttons a bit closer to the carousel?  It can be a bit hard to tell which buttons belong to which carousel if you don't already know they should be on top:

Screen Shot 2019-09-11 at 11 31 09 AM

static/js/components/CourseCarousel.js
```diff
@@ -16,50 +17,66 @@ type Props = {|
...
+            <i className="material-icons">arrow_forward</i>
+          </button>
+        </div>
+      </div>
+      <Carousel
+        slideIndex={index}
+        slidesToShow={CAROUSEL_PAGE_SIZE}
+        slidesToScroll="auto"
+        withoutControls={true}
+        afterSlide={slideIndex => setIndex(slideIndex)}
```
mbertrand: Is setting `afterSlide` necessary? The carousel seemed to behave the same with or without this line in some brief testing.

aliceriot:
@ferdi ok I added a responsive thing to do 3 cards wide at desktop, 2 cards on a tablet, and 1 on a phone:

desksksksks
tabeltversion
phoneversion

static/js/factories/learning_resources.js
```diff
@@ -133,3 +134,30 @@ export const makeLearningResource = (objectType: string) => {
...
+  const bootcamps = R.times(makeBootcamp, 3).map(makeFavorite)
+
+  return [
+    ...courses.map(formatFavorite(LR_TYPE_COURSE)),
+    ...programs.map(formatFavorite(LR_TYPE_PROGRAM)),
+    ...bootcamps.map(formatFavorite(LR_TYPE_BOOTCAMP))
+  ]
+}
+
+export const courseListResponse = (list: Array<LearningResource>) => ({
```
mbertrand: This only seems to be used for tests, maybe it should go in `test_utils.js` instead?

static/js/components/CourseCarousel.js
aliceriot: are you sure you pulled down the most recent version? the padding below the buttons should look like it does in the screenshots above

mbertrand: Just pulled again, looks good now 👍

static/js/components/CourseCarousel.js
aliceriot: the `afterSlide` is what keeps the `slideIndex` in sync with the position of the carousel. The carousel has it's own internal state which manages which page it's showing - here basically I'm replicating that state externally (the `index` variable) so that I can control that state with the buttons. Then I pass that state down into the carousel (via the `slideIndex` prop) to control the carousel and sync and carousel-side changes to the position back to the `index` via this `afterSlide` callback.

basically, the case where it can get out of sync is if the user slides the carousel. if we're not syncing that change in the index back (via `afterSlide`) and the user then goes to click the button, we'll change teh `index` based on an incorrect current value for the index (b/c it hasn't been updated to reflect the changed position of the carousel) and we'll get some weird behavior.

does that make sense? I can show you how it can lead to weird behavior if that helps too

mbertrand: sounds good

static/js/factories/learning_resources.js
aliceriot: sure

PR #2230, opened 2019-09-17
Make sure object_type is always merged in when retrieving entities from state

### LINKED ISSUE #2228: Clicking course card to display drawer only works once
### Steps to Reproduce

1. Navigate to /courses/
2 Click on a course card
3. The drawer loads
4. Click anywhere off the drawer to dismiss it. 
5. Click on the same course card again

### Expected Behavior

- Drawer should load again. 

### Actual Behavior

- Nothing happens
- No errors in the console 

### Screenshot or Screencast

https://embed.vidyard.com/share/idwpYJTGUwoFPBZwk6V1v5?

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #2228

#### What's this PR do?
Merges the object_type property with appropriate value into entities retrieved from state.

- Click a new course on the index page to bring up the drawer. Close it and repeat.  The drawer should open every time.
- Add some favorite courses and bootcamps, and click on them too in the Favorites carousel.

### DIFF
--- a/course_catalog/serializers.py
+++ b/course_catalog/serializers.py
@@ -249,6 +249,7 @@ class CourseSerializer(BaseCourseSerializer):
     """
 
     course_runs = CourseRunSerializer(read_only=True, many=True, allow_null=True)
+    object_type = serializers.CharField(read_only=True, default="course")
 
     class Meta:
         model = Course
@@ -336,6 +337,7 @@ class BootcampSerializer(BaseCourseSerializer):
     """
 
     course_runs = CourseRunSerializer(read_only=True, many=True, allow_null=True)
+    object_type = serializers.CharField(read_only=True, default="bootcamp")
 
     def to_internal_value(self, data):
         """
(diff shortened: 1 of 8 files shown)


aliceriot:
I wonder with this, should we go to the step of adding it in at the API level so that we're assured that the value is always set for the frontend?

aliceriot:
the frontend code for a lot of these things could be simplified if we did that, for instance in the favorites API we currently get these `{ content_data, content_type }` records back - if the `content_type` were just set on the content itself then we could just have an `Array<LearningResource>` instead

aliceriot:
I think the problem on the course carousel in particular is that when the drawer is opened we hit the relevant API, and then the response from the course API (the `courseRequest` mutation hits the relevant endpoint) overwrites the value already set in the store but it doesn't set the `object_type` value on it's response

mbertrand:
@aliceriot I changed the approach and added `object_type` to the serializers instead.

Issue #2152, opened 2019-08-12
Update course drawer to open for programs


mbertrand:
When clicking on a search result card for a `Program`, the course drawer should open and show the contents of that program/list.

#### Designs and Mockups

Program:
https://projects.invisionapp.com/d/main#/console/17477251/376322443/preview

mbertrand:
@pdpinch the [design for the Program drawer](https://projects.invisionapp.com/d/main#/console/17477251/376322443/preview) has many fields that still do not exist for the Program model, should they? - length, effort, language, video transcript, instructors.  Nathan is working on adding price.  I think the topics list will always be empty for programs at the moment.  If necessary I could query for all the instructors and topics from all the program's course runs to display on the fly, instead of storing them as part of the Program model.

mbertrand:
@pdpinch I talked with @rhysyngsun about above, the info for the missing fields does not exist, except instructors, which should be imported assuming the CMS is populated correctly.  So the program drawer should include the following fields: title, short description, offered by, instructors, # courses, cost.

mbertrand:
@pdpinch mind if I split out the learning paths into a new issue?

pdpinch:
Yeah, that would be fine. The design for learning paths iis still in progress.

PR #2251, opened 2019-09-26
Open drawer for programs

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [ ] Code is tested

Partially addresses #2152

#### What's this PR do?
Opens the `LearningResourceDrawer` when a program search result/card is clicked and displays the relevant info.

- import some programs if you don't have any yet:
  ```
   get_xpro_data()
   sync_and_upload_edx_data()
   get_micromasters_data()
  ```
  you'll need to set some .env variables for the above including:
  ```
  MICROMASTERS_CATALOG_API_URL=http://docker.for.mac.host.internal:8079/api/v0/catalog/
  XPRO_CATALOG_API_URL=http://docker.for.mac.host.internal:8053/api/programs/
  ```
- Search for programs at `/courses/search`.  Click one of the results.  The drawer should open with the relevant info and it should look roughly like the [design](https://projects.invisionapp.com/d/main#/console/17477251/376322443/preview), but many fields will be missing due to the info not existing for programs (length, effort, video transcript, topics, instructors, etc).
- ~The image will be broken for xpro courses because the xpro API lists them as relative URL's (will be fixed soon)~.
- If there is a link to a Micromasters program, it won't work either because it is imported as a relative URL (https://github.com/mitodl/micromasters/issues/4410)

Screen Shot 2019-09-27 at 10 35 37 AM

Screen Shot 2019-09-27 at 10 33 52 AM


mbertrand:
@ferdi @pdpinch @abdulkdawson screenshots are in the description

Ferdi:
Do we have start_date for xpro/MM ?
Do we have instructors for xpro / MM ?
Why there is not "Take course on" button in the first screenshot
Can you separate "take course on" from the provider like in the design?
replace "Number of courses : 5 courses in program" to "Number of courses : 5" or "Number of courses in program: 5"

mbertrand:
> Do we have start_date for xpro/MM ?
Not yet, it's being worked on for xpro programs, not sure about MM
> Do we have instructors for xpro / MM ?
No, instructors are populated for course runs only.
> Why there is not "Take course on" button in the first screenshot
There is no URL being imported yet for xPro programs.
> Can you separate "take course on" from the provider like in the design?
Sure, but my assumption is that styling changes for the drawer would be in a different PR.  
> replace "Number of courses : 5 courses in program" to "Number of courses : 5" or "Number of courses in program: 5"
OK

mbertrand:
@ferdi I changed the code to display all unique instructors from all course runs included in a program.

mbertrand:
@ferdi regarding "Can you separate "take course on" from the provider like in the design?" - presumably this would apply to courses, bootcamps too?  I would need logos for all the "offered by" options (Micromasters, xPro, etc) - seems like this should be handled in a separate PR.

mbertrand:
@ferdi @pdpinch @abdulkdawson I just updated the screenshots

Ferdi:
@mbertrand 

I suggest we have the providers APIs expose the list of instructors and other metadata. Each of those (MM, xpro etc.) can use their own logic to construct the data, and Open just consumes what comes from these api-s.

mbertrand:
@ferdi ok I will create an issue for that in xpro and micromasters.  Should I remove the new code here that extracts the instructor list from all the course runs, or leave as is until the above is completed?

Ferdi:
@mbertrand up to you.

course_catalog/serializers.py
```diff
@@ -399,6 +399,7 @@ class ProgramSerializer(serializers.ModelSerializer, FavoriteSerializerMixin):
 
     items = ProgramItemSerializer(many=True, allow_null=True)
     topics = CourseTopicSerializer(read_only=True, many=True, allow_null=True)
+    prices = CoursePriceSerializer(read_only=True, many=True, allow_null=True)
```
noisecapella: Do we need to expose all of the prices via REST API? If we only need the most recent one we should expose that specifically.

static/js/lib/learning_resources.js
```diff
@@ -143,16 +147,20 @@ export const resourceLabel = (resource: string) => {
...
 
-export const maxPrice = (courseRun: ?CourseRun) => {
-  const prices = courseRun && courseRun.prices ? courseRun.prices : []
+export const maxPrice = (prices: Array<CoursePrice>) => {
+  if (emptyOrNil(prices)) {
+    return null
+  }
   const price = Math.max(...prices.map(price => price.price))
-  return price > 0 ? `$${price}` : "Free"
+  return price > 0 ? `$${price.toFixed(2)}` : "Free"
```
noisecapella: Can we make a separate function for this? Maybe similar to https://github.com/mitodl/mitxpro/blob/master/static/js/lib/ecommerce.js#L39

course_catalog/serializers.py
mbertrand: Program prices don't have any dates (besides created/modified), and for xpro at least there should be only one anyway.  I don't think the micromasters catalog API returns any prices at all yet.

static/js/lib/learning_resources.js
mbertrand: Done

static/js/components/LearningResourceDrawer.js
```diff
@@ -105,6 +108,10 @@ const getObject = createSelector(
       return querySelectors.isFinished(queries, bootcampRequest(objectId))
         ? bootcamps[objectId]
         : null
+    case LR_TYPE_PROGRAM:
```
noisecapella: It looks like this case isn't covered by any tests. Can you add a test for it?

mbertrand:
@pdpinch do you want to review the screenshots before I merge?

mbertrand:
Oops sorry @pdpinch just noticed you're on vacation, so never mind.  @ferdi do the screenshots look okay to you?

PR #2260, opened 2019-10-01
Remove OCW courses from search if they are unpublished

### LINKED ISSUE #2125: OCW courses that are deleted should be viewable directly, but not searchable
We need to be able to reflect the fact that some OCW courses are deleted, but support the fact that these courses may show up in lists and we don't want to break those.

This means:

- [ ] Don't delete OCW courses from the `Course` table so that it is still available via the `/api/v0/course/:ID:/` API
- [ ] Instead we should flag the course in such a way that we don't index it (maybe set `published = False`? this probably requires some investigation into how OCW reflects a course deletion)
- [ ] Kick off a task to remove it from the index

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #2125 

#### What's this PR do?
Updates indexing code to delete documents from the Elasticsearch index which are unpublished

First, start on master. Make sure you have up to date OCW data, or run `./manage.py backpopulate_ocw_data` if not. Go to `/courses/search?q=linear%20algebra` and there should be some Linear Algebra courses available. Click on each one from the top of the list until you get to one whose instructor is "Gilbert Strang". Click the "As Taught in" dropdown and verify that there are three years: 2002, 2005, and 2010. Select '2002' and click 'Take Course on OCW'. You should get a 404 page.

Run `./manage.py backpopulate_ocw_data --skip-s3 --overwrite`. This will take many hours and you may need to run it overnight. After it finishes, go back to the course search and look for the same course as before. It should now show only two years: 2005 and 2010.

### DIFF
--- a/course_catalog/api.py
+++ b/course_catalog/api.py
@@ -5,11 +5,12 @@ import json
 import logging
 from datetime import datetime
 
-import pytz
-
+import boto3
 from django.db import transaction
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
+from ocw_data_parser import OCWParser
+import pytz
 
 from course_catalog.constants import (
     PlatformType,
@@ -17,14 +18,19 @@ from course_catalog.constants import (
     AvailabilityType,
     OfferedBy,
 )
-from course_catalog.models import Bootcamp, CourseRun
+from course_catalog.models import Bootcamp, Course, CourseRun
 from course_catalog.serializers import (
     BootcampSerializer,
     CourseRunSerializer,
     OCWSerializer,
 )
 from course_catalog.utils import get_course_url
-from search.task_helpers import upsert_course, index_new_bootcamp, update_bootcamp
+from search.task_helpers import (
+    delete_course,
+    upsert_course,
+    index_new_bootcamp,
+    update_bootcamp,
+)
 
 log = logging.getLogger(__name__)
 
@@ -62,12 +68,11 @@ def digest_ocw_course(
         is_published (bool): Flags OCW course as published or not
         course_prefix (str): (Optional) String used to query S3 bucket for course raw JSONs
     """
-
     ocw_serializer = OCWSerializer(
         data={
             **master_json,
             "last_modified": last_modified,
-            "is_published": is_published,
+            "is_published": True,  # This will be updated after all course runs are serialized
             "course_prefix": course_prefix,
             "raw_json": master_json,  # This is slightly cleaner than popping the extra fields inside the serializer
         },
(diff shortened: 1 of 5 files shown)


course_catalog/api.py
```diff
@@ -526,3 +530,192 @@ def parse_bootcamp_json_data(bootcamp_data, force_overwrite=False):
...
+        aws_secret_access_key=settings.OCW_CONTENT_SECRET_ACCESS_KEY,
+    ).Bucket(name=settings.OCW_CONTENT_BUCKET_NAME)
+
+    # get all the courses prefixes we care about
+    ocw_courses = generate_course_prefix_list(raw_data_bucket)
+
+    # loop over each course
+    uids = {}
+    for course_prefix in sorted(ocw_courses):
+        try:
```
mbertrand: just a tip, to make manual testing faster, I temporarily enclosed the following with this `if` statement before running:
`if course_prefix.startswith("PROD/18/18.06"):`

course_catalog/api.py
```diff
@@ -526,3 +530,192 @@ def parse_bootcamp_json_data(bootcamp_data, force_overwrite=False):
...
+    courserun_instance = CourseRun.objects.filter(course_run_id=uid).first()
+
+    # Make sure that the data we are syncing is newer than what we already have
+    if (
+        courserun_instance
+        and last_modified <= courserun_instance.last_modified
+        and not force_overwrite
+    ):
+        log.info("Already synced. No changes found for %s", course_prefix)
+        return uid, is_published
```
mbertrand: If `uid` and `is_published` are returned here, the course will be either reindexed (or deindexed) even if no runs were changed.

course_catalog/api.py
```diff
@@ -526,3 +530,192 @@ def parse_bootcamp_json_data(bootcamp_data, force_overwrite=False):
...
+    Args:
+        force_overwrite (bool): A boolean value to force the incoming course data to overwrite existing data
+        upload_to_s3 (bool): If True, upload course media to S3
+    """
+    uids = sync_ocw_courses(force_overwrite=force_overwrite, upload_to_s3=upload_to_s3)
+    course_ids = update_course_published(
+        CourseRun.objects.filter(course_run_id__in=uids.keys())
+    )
+    for course in Course.objects.filter(id__in=course_ids):
+        # Probably quicker to do this as a bulk operation
```
mbertrand: I think all 3000+ OCW courses will be updated and reindexed (or deindexed) on every run (unless no uid is returned for unchanged runs - see comment above), in which case it might be worth adding bulk ES updates for both of these.

course_catalog/api.py
noisecapella: Ah, I did something similar myself

course_catalog/api.py
noisecapella: Anecdotally it seems like the slow part is `Loading JSON` and the Elasticsearch indexing is a minor piece of it. I would like to leave it as is but I can make an issue to address later.

noisecapella:
Should be ready for review again

course_catalog/api.py
mbertrand: 👍

PR #2270, opened 2019-10-07
refactor the tooltips!

### LINKED ISSUE #1299: Change styling and rendering approach for popups/tooltips

### LINKED ISSUE #2046: "Join MIT Open" pop-ups displays in the wrong place
### Steps to Reproduce

1. Log out
2. Navigate to the home page
3. Click on an upvote button

### Expected Behavior

"Join MIT Open" pop-up appears below the upvote button so that the arrow at the top left of the pop-up points at the upvote button. 

### Actual Behavior

(Optional)
image

### LINKED ISSUE #2209: When anonymous user clicks favorite button, ask them to Join MIT Open
As an anonymous user, I'd like to get a prompt to join when I try to click on a Favorite button

#### Designs and Mockups

Should be similar UX to when an anonymous user clicks on the the vote button:

Untitled

I'm not sure if we need to update the text of this pop-up: 

> As a member, you can share your ideas, subscribe, vote, and comment on educational content that matters to you.

#### Acceptance Criteria:

- [ ] Show standard "Join" pop-up when anonymous user clicks on favorite.

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #1299
closes #2046
closes #2209

#### What's this PR do?

This refactors our LoginPopup and SharePopup to be LoginTooltip and ShareTooltip, and tears out our homemade stuff for showing these popups, to use a library instead (this one: https://react-component.github.io/tooltip/)

Basically, now you can just grab the `<LoginTooltip>` and wrap it around any element you want to trigger a popup when clicked, like this:

```js
<LoginTooltip>
  <div className="authenticated-users-only">
    FUN!
  </div>
</LoginTooltip>
```

and that's it! No messing with the store, no nothing.

Make sure that:

- there are no regressions in the things which currently have a login tooltip on them. Like, clicking on anything that requires the user to be logged in on the PostPage, for instance, should trigger a tooltip if you arent logged in and should do the normal thing if you are logged in
- the "favorite" button on the learning resource card is now something that also triggers a login tooltip
- the "share" popup on the post page (for the post and for the comments) works as it did before

I think that's it. Probably a good idea to read through all the code and so on too :smile: 

newloginsearch
newlogin
newshare


noisecapella:
The triangle on the tooltip is pointing in the wrong direction, and it stays around even though the drawer closes. Feel free to ignore, it's still better than it used to be. Just wanted to point out if it's an easy fix.

Screenshot from 2019-10-08 11-08-51

static/js/components/PostUpvoteButton_test.js
```diff
@@ -1,39 +1,50 @@
...
+    let resolver: Function
+    const prommo = new Promise(resolve => {
+      resolver = resolve
     })
+    upvoteStub.returns(prommo)
+    const wrapper = renderButton()
+    assert.isFalse(wrapper.state().upvoting)
+    wrapper.instance().onToggleUpvote()
+    assert.isTrue(wrapper.state().upvoting)
+    // $FlowFixMe: flow things this isn't a function
```
noisecapella: `thinks`

static/js/lib/comments.js
```diff
@@ -41,3 +41,14 @@ export const findComment = (
 
   return null
 }
+
+const stripReplies = R.evolve({
+  replies: R.always([])
```
noisecapella: Not sure if it's important here but this looks like it would return the same empty array instance for each `replies` instance worked on, which would mean code which calls `.push` would add items to multiple different arrays. I think if we replace with `() => []` it would return a different empty array instance each time.

static/js/components/LoginTooltip.js
```diff
@@ -0,0 +1,41 @@
...
+  children: Array<React$Element<any>> | React$Element<any>
+}
+
+const LoginTooltip = ({ children }: Props) =>
+  userIsAnonymous() ? (
+    <Tooltip placement="top" trigger={["click"]} overlay={LoginTooltipContent}>
+      {children}
+    </Tooltip>
+  ) : (
+    <React.Fragment>{children}</React.Fragment>
```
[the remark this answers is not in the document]
aliceriot: we probably don't!

PR #2272, opened 2019-10-08
Add Open Learning Library Import

### LINKED ISSUE #2227: Courses: OpenLearningLibrary import
As a user, I'd like to have complete, evergreen access to archived courses from MITx and Residential MITx. 

From time to time, these courses are ported to https://openlearninglibrary.mit.edu 
#### Acceptance Criteria:

- [ ] periodically import courses from the API at https://discovery.openlearninglibrary.mit.edu/
- [ ] "offered by" should be set to "OpenLearningLibrary" 
- [ ] ignore any missing subjects and instructors 
- [ ] all courses are free, and the feed should reflect that

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [x] Settings

Fixes #2227 

#### What's this PR do?

This PR adds Open Learning Library as a course catalog import:
  - Uses the existing abstract openedx ETL factory
  - Generates a $0 price for each course run so they appear as "Free" in the course search

It also makes some necessary changes to how we unique courses and course runs:
  - The tuple `(key, platform)` is used for `update_or_create` calls instead of just `key`, this allows us to avoid collisions with MITx courses and runs.
  - Added a `platform` field to `CourseRun`
  - Removed `unique=True` on both `Course.course_id` and `CourseRun.course_run_id`
  - Backpopulates `platform` on `Course` and `CourseRun` just to be safe

There are two things this PR _doesn't_ do, because they both could trigger race conditions that could cause the migrations to fail because we weren't writing `platform` in a few places. A follow-up issues #2275 has been opened to address that later.

- Doesn't add a `unique_together` for `(platform, course_id)` and `(platform, course_run_id)`
- Doesn't remove `null=True` from `CourseRun.course_run_id`

- Configure the environment variables (DM me for the access key / client id)
- Run `./manage.py backpopulate_oll_data`

Known issues:

- A fair amount of data is null in the catalog response because it's not present on OLL, some known examples (this implementation support them, should they ever be configured):
  - Instructors
  - Course images
  - Topics

Screenshot_2019-10-07 MIT Open Learning(1)
Screenshot_2019-10-07 MIT Open Learning


rhysyngsun:
This is causing regressions in the OCW and it's proving difficult to reason about with the state of that code and its tests. Hoping that waiting for #2260 will make the cause of this clearer.

mbertrand:
@rhysyngsun what is the expected behavior when a course is offered on multiple platforms?  For instance there is a course on MITx called "Mechanics: Momentum and Energy".  If I run the OLL import, the course is changed from `offered_by="MITx", platform="mitx"` to `offered_by="Open Learning Library", platform="oll"`.  If I then run `get_mitx_data`, it is switched back again.  

```python
In [17]: Course.objects.filter(title__startswith="Mechanics: Momentum and Energy").values_list("id","offered_by", "platform")
Out[17]: <TimestampedModelQuerySet [(3630, 'Open Learning Library', 'oll')]>

In [18]: get_mitx_data()
[2019-10-10 18:32:03] INFO 1 [root] ocw.py:50 - [4a37616598d4] - Uploading edX courses data to S3

In [19]: Course.objects.filter(title__startswith="Mechanics: Momentum and Energy").values_list("id","offered_by", "platform")
Out[19]: <TimestampedModelQuerySet [(3630, 'MITx', 'mitx')]>
```

mbertrand:
I think `course_catalog.etl.load_course` needs to include `platform` in its `update_or_create` call?

rhysyngsun:
@mbertrand I addressed that issue, plus a few others I noticed:

- Removed `@log_exceptions` from `load_course` and `load_run`, these were silencing important errors. Added one to `load_courses` since that's the only other entry point other than `load_programs` which is already covered
- Updated the generation of ES document ids to account for the use of `("platform", "course_id")` as the unique key for courses (this was causing OLL docs to overwrite MITx docs even after fixing the issue you noted)

I'd suggest exercising the following in order:

- Delete all OLL + MITx `Course`s
- `./manage.py recreate_index
- Verify in the course search page you are now getting no results for either MITx or OLL
- Run `./manage.py backpopulate_edx_data`, you should get 328 courses in the search page/facet for MITx
- Run `./manage.py backpopulate_oll_data`, you should get 19 courses in the search page/facet for OLL
- Run `./manage.py recreate_index` and verify the numbers (328 and 19) remain the same

course_catalog/migrations/0044_uniq_by_platform_id.py
```diff
@@ -0,0 +1,48 @@
...
+
+from django.db import migrations, models
+
+from course_catalog.constants import OFFERED_BY_MAPPINGS
+
+
+def populate_platforms(apps, schema_editor):
+    """
+    Set all avatars to None. Since we don't show avatars yet
+    this avoids the work of populating avatar_small and avatar_medium
```
mbertrand: Wrong text

course_catalog/migrations/0044_uniq_by_platform_id.py
```diff
@@ -0,0 +1,48 @@
...
+    Course = apps.get_model("course_catalog", "Course")
+    LearningResourceRun = apps.get_model("course_catalog", "LearningResourceRun")
+
+    for course in Course.objects.iterator():
+        course.platform = OFFERED_BY_MAPPINGS[course.offered_by]
+        course.save()
+
+    for run in LearningResourceRun.objects.iterator():
+        run.platform = OFFERED_BY_MAPPINGS[course.offered_by]
+        run.save()
```
mbertrand: Assuming all runs have a populated `offered_by` field, I think this should be:
```python
    for run in LearningResourceRun.objects.iterator():
        run.platform = OFFERED_BY_MAPPINGS[run.offered_by]
        run.save()
```

if not, then:

```python
    Course = apps.get_model("course_catalog", "Course")
    Bootcamp = apps.get_model("course_catalog", "Bootcamp")
    Program = apps.get_model("course_catalog", "Program")
    LearningResourceRun = apps.get_model("course_catalog", "LearningResourceRun")

    for run in LearningResourceRun.objects.filter(content_type__model="course").iterator():
        course = Course.objects.get(id=run.object_id)
        course.platform = OFFERED_BY_MAPPINGS[course.offered_by]
        course.save()
        run.platform = OFFERED_BY_MAPPINGS[course.offered_by]
        run.save()

    for run in LearningResourceRun.objects.filter(content_type__model="bootcamp").iterator():
        bootcamp = Bootcamp.objects.get(id=run.object_id)
        bootcamp.platform = OFFERED_BY_MAPPINGS[bootcamp.offered_by]
        bootcamp.save()
        run.platform = OFFERED_BY_MAPPINGS[bootcamp.offered_by]
        run.save()

    for run in LearningResourceRun.objects.filter(content_type__model="program").iterator():
        program = Program.objects.get(id=run.object_id)
        run.platform = OFFERED_BY_MAPPINGS[program.offered_by]
        run.save()
```

I was getting some dupe runs with the code as is, unless I deleted everything and imported from scratch.

rhysyngsun: We want the first one, I also made it a bit safer by switching to `.get()` with the current value as the default in case there's some weird cases where `offered_by` isn't set/

rhysyngsun: I checked, there's no null or empty values in prod, so we shouldn't see an issue, but probably better safe than sorry...

course_catalog/etl/micromasters.py
```diff
@@ -39,6 +40,7 @@ def transform(programs):
             "courses": [
                 {
                     "course_id": course["edx_key"],
+                    "platform": PlatformType.mitx.value,
```
mbertrand: `"platform": PlatformType.micromasters.value,` ?

course_catalog/etl/micromasters_test.py
```diff
@@ -87,12 +96,21 @@ def test_micromasters_transform(mock_micromasters_data):
             "image_src": "http://example.com/program/2/image/url",
             "offered_by": OfferedBy.micromasters.value,
             "courses": [
-                {"course_id": "3", "offered_by": OfferedBy.micromasters.value},
-                {"course_id": "4", "offered_by": OfferedBy.micromasters.value},
+                {
+                    "course_id": "3",
+                    "platform": PlatformType.mitx.value,
```
mbertrand: `"platform": PlatformType.micromasters.value,`?

course_catalog/etl/micromasters_test.py
```diff
@@ -87,12 +96,21 @@ def test_micromasters_transform(mock_micromasters_data):
...
+                    "offered_by": OfferedBy.micromasters.value,
+                },
+                {
+                    "course_id": "4",
+                    "platform": PlatformType.mitx.value,
```
mbertrand: `"platform": PlatformType.micromasters.value,`?

rhysyngsun:
@mbertrand regarding your last comment, that's intentional, there's a few things here at play as to why we want to do this:

- By specifying `"platform": PlatformType.mitx.value` on course and run, that allows the MM courses and runs to merge with the MITx ones. We want it to do that because MITx populate the bulk of the data fields. That's also why only platform/course_id/offered_by are set here. This is important for the next bullet point:
- `offered_by` is more or less a placeholder here until we finish #2163. Currently it ends up being a tug of war and the value is whatever import ran last. For the sake of this PR we can just ignore this.

PR #2286, opened 2019-10-11
fix learning resource details drawer right-to-left behavior

### LINKED ISSUE #2218: Courses: drawer animation should be right to left
The course detail drawer should animate from right to left. 

#### Designs and Mockups

TBD

#### Acceptance Criteria:

- [ ] less jarring animation
- [ ] mobile version

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #2218

#### What's this PR do?

This PR basically fixes the "right to left" issue with the ExpandedLearningResourceDrawer. Basically, on the course search page we want it to come in from the right, but on master currently it sort of ends up on the right but the animation goes left-to right, so it's weird.

Anyway, the way to fix this turns out to be just basically to update to the latest version of the material components package we're using ([rmwc](https://rmwc.io/)), which fixed the issue. So then most of the work I ended up doing was dealing with that upgrade, mainly because:

- the drawer behavior / classnames / etc is different, so I had to do some work to port our current drawer behavior over
- the react-clamp-lines package we were using is, for some strange reason, broken in the new drawer, so I replace that by extending our existing `TruncatedText` component to support the course drawer use-case (showing / hiding the learning resource description text).
- a few miscellaneous other things were necessary because other parts of the rmwc package had changed (like the menu, dialog, I think a few other random things)

I think that's about it

I think this should be tested pretty thoroughly. Make sure that any part of the site where we're using any rmwc components works as it does on master. in particular,

- check anything involving a dialog
- check out all aspects of the current drawer behavior
- check out the course drawer
- check out settings pages and so on

I think those are the main areas, but I would definitely just poke around and see if I broke anything unintentionally.


aliceriot:
ah, forgot about that z-index issue! I'll fix that now

aliceriot:
@mbertrand I actually found an additional issue with the widget dialogs, just went ahead and fixed that. lmk if things look good to you

PR #2295, opened 2019-10-16
Don't show any results if no text matches are found

### LINKED ISSUE #2293: Searching for text with no matches returns all resources
### Steps to Reproduce

Search for text 'flsdfjlksdjflskdjflskdjf' at `../courses/search`

### Expected Behavior

No matching results found.

### Actual Behavior

Returns results for anything matching other filters besides text.

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #2293 

#### What's this PR do?
Modifies ES search query so that it returns no results if a text query like 'sdhfjsdhfksdjfhsk' has no matches.

Make sure search returns expected matches:
- "6.002x" returns "Circuits and Electronics" MITx course as 1st result
- "Aerospace Engineering" returns a course with that in the title as 1st result
- "Capozzola" returns a MITx course with Christopher Capozzola as instructor
- "sjdkfjslfkjsldfkjsdlfks" returns nothing

The text clauses had to be added to the ES query twice: in the original location (which adjusts the result scores based on closeness of match but doesn't exclude non-matching results), and in the "filter -> must" query (which excludes non-matching results but does not score the results returned).
Also boosted some additional fields (title, short description, instructors).

### DIFF
--- a/static/js/lib/search.js
+++ b/static/js/lib/search.js
@@ -151,7 +151,7 @@ export const RESOURCE_QUERY_NESTED_FIELDS = [
   "runs.year",
   "runs.semester",
   "runs.level",
-  "runs.instructors"
+  "runs.instructors^5"
 ]
 
 const LIST_QUERY_FIELDS = [
--- a/static/js/lib/search.js
+++ b/static/js/lib/search.js
@@ -414,34 +414,35 @@ export const buildSearchQuery = ({
 
   const types = getTypes(type)
   for (const type of types) {
-    let textQuery = {}
-    const textClauses = []
-    let textFilter = []
-    if (!emptyOrNil(text)) {
-      textClauses.push({
-        multi_match: {
-          query:     text,
-          fields:    searchFields(type),
-          fuzziness: "AUTO"
-        }
-      })
-      if ([LR_TYPE_BOOTCAMP, LR_TYPE_COURSE, LR_TYPE_PROGRAM].includes(type)) {
-        textClauses.push({
-          nested: {
-            path:  "runs",
-            query: {
-              multi_match: {
-                query:     text,
-                fields:    RESOURCE_QUERY_NESTED_FIELDS,
-                fuzziness: "AUTO"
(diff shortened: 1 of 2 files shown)


static/js/lib/search.js
```diff
@@ -413,37 +414,33 @@ export const buildSearchQuery = ({
...
-            multi_match: {
-              query:     text,
-              fields:    RESOURCE_QUERY_NESTED_FIELDS,
-              fuzziness: "AUTO"
-            }
-          }
-        }
-      })
+      textQuery = { should: textClauses }
+      textFilter = [{ bool: textQuery }]
```
aliceriot: just a little thought I had about how this could maybe be reorganized a little bit for clarity. it looks like we have basically that `textFilter === emptyOrNil(text) ? [] : [{ bool: textQuery }]`, if it was declared with `const` after this block you could declare it that way (I'm just sometimes way of the pattern of declaring a variable with `let` and then assigning it many lines away from it's declaration, I think it's best to avoid `let` wherever we can).

it seems also like the `textClauses` variable doesn't need to be declared outside of the `if () { ... }` block. You could even just do something like:

```js
const textQuery = emptyOrNil(text)
  ? {}
  : {
    should: [
      {
        multi_match: {
          query:     text,
          fields:    searchFields(type),
          fuzziness: "AUTO"
        }
      },
      [LR_TYPE_BOOTCAMP, LR_TYPE_COURSE, LR_TYPE_PROGRAM].includes(type)
        ? {
          nested: {
            path:  "runs",
            query: {
              multi_match: {
                query:     text,
                fields:    RESOURCE_QUERY_NESTED_FIELDS,
                fuzziness: "AUTO"
              }
            }
          }
        }
        : null
    ].filter(clause => clause !== null)
  }

const textFilter = emptyOrNil(text) ? [] : [{ bool: textQuery }]
```

to just declare the variables in one go. Anyway, take this or leave it.

mbertrand: 👍 change made

PR #2316, opened 2019-10-21
Added VideoResource model

### LINKED ISSUE #2306: Video Search: implement VideoResource data model(s)
We need a data model to store and index off of for video resources 
 - Should be a subclass of our abstract LearningResource model
  - Add relevant fields to this new subclass, where fields are common, they should align to our existing search mappings if possible 
  - Transcripts will probably be the only field we don’t already have an equivalent for this

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing

Closes #2306 

#### What's this PR do?
Adds a `VideoResource` model for tracking ingested videos

```python
from course_catalog.factories import VideoResourceFactory
vr = VideoResourceFactory.create()
```


course_catalog/constants.py
```diff
@@ -31,6 +31,7 @@ class PlatformType(Enum):
     bootcamps = "bootcamps"
     xpro = "xpro"
     oll = "oll"
+    youtube = "youtube"
```
mbertrand: Add `youtube = "YouTube"` to `class OfferedBy(Enum)` as well?

course_catalog/factories.py
```diff
@@ -341,3 +342,23 @@ def topics(self, create, extracted, **kwargs):
...
+class VideoResourceFactory(LearningResourceFactory):
+    """Factory for VideoResource"""
+
+    video_id = factory.Sequence(lambda n: "VIDEO-%03d.MIT" % n)
+    platform = FuzzyChoice([PlatformType.youtube.value])
+
+    full_description = factory.Faker("text")
+    image_src = factory.Faker("image_url")
+    image_description = factory.Faker("text", max_nb_chars=300)
+
```
mbertrand: Add a default `offered_by`?
```python
    offered_by = factory.RelatedFactoryList(
        "course_catalog.factories.LearningResourceOfferorFactory",
        size=1,
        name=OfferedBy.youtube.value,
    )
```

course_catalog/factories.py
```diff
@@ -341,3 +342,23 @@ def topics(self, create, extracted, **kwargs):
 
     class Meta:
         model = UserList
+
+
+class VideoResourceFactory(LearningResourceFactory):
+    """Factory for VideoResource"""
+
+    video_id = factory.Sequence(lambda n: "VIDEO-%03d.MIT" % n)
+    platform = FuzzyChoice([PlatformType.youtube.value])
```
mbertrand: Is this a `FuzzyChoice` because other video platforms will be added eventually?

course_catalog/constants.py
rhysyngsun: These videos will actually be offered by existing values (e.g. OCW, MITx)

course_catalog/factories.py
rhysyngsun: Yeah, just figured this will make that change slightly easier

course_catalog/models.py
```diff
@@ -266,3 +266,26 @@ class FavoriteItem(TimestampedModel):
...
+
+class VideoResource(LearningResource):
+    """Data model for video resources"""
+
+    video_id = models.CharField(max_length=80)
+    platform = models.CharField(max_length=128)
+
+    full_description = models.TextField(null=True, blank=True)
+    image_src = models.URLField(max_length=400, null=True, blank=True)
+    image_description = models.CharField(max_length=1024, null=True, blank=True)
```
abeglova: I don't think the video data api returns an image description field.

course_catalog/factories.py
```diff
@@ -341,3 +342,29 @@ def topics(self, create, extracted, **kwargs):
...
+
+    full_description = factory.Faker("text")
+    image_src = factory.Faker("image_url")
+    image_description = factory.Faker("text", max_nb_chars=300)
```
abeglova: You need to remove image_description here as well

rhysyngsun: Do'h, good catch, I'll fix that on the next PR for search. It's unused right now so it doesn't break anything at the moment

course_catalog/migrations/0047_add_video_resource.py
```diff
@@ -0,0 +1,57 @@
...
+                    models.CharField(
+                        blank=True, db_column="offered_by", max_length=128, null=True
+                    ),
+                ),
+                ("video_id", models.CharField(max_length=80)),
+                ("platform", models.CharField(max_length=128)),
+                ("full_description", models.TextField(blank=True, null=True)),
+                ("image_src", models.URLField(blank=True, max_length=400, null=True)),
+                ("last_modified", models.DateTimeField(blank=True, null=True)),
+                ("published", models.BooleanField(default=True)),
```
abeglova: Do we need published? It should always be true since we are not going to get seperate api keys for each channel to get private videos

rhysyngsun: This will cover the case of videos being removed, we have a practice of not deleting the records to maintain the integrity of user lists and learning paths, but `published=False` will mean that the item doesn't get included in the search index. We use this for courses that are no longer published since there's no value in searching for a course you can't take.

course_catalog/migrations/0047_add_video_resource.py
```diff
@@ -0,0 +1,57 @@
...
+                ("transcript", models.TextField(blank=True, default="")),
+                ("raw_data", models.TextField(blank=True, default="")),
+                (
+                    "offered_by",
+                    models.ManyToManyField(
+                        blank=True, to="course_catalog.LearningResourceOfferor"
+                    ),
+                ),
+                (
+                    "topics",
```
abeglova: This doesn't have to change, but the topics returned by the data api won't match exactly to CourseTopic.

course_catalog/models.py
```diff
@@ -266,3 +266,25 @@ class FavoriteItem(TimestampedModel):
...
     class Meta:
         unique_together = ("user", "content_type", "object_id")
+
+
+class VideoResource(LearningResource):
+    """Data model for video resources"""
+
+    video_id = models.CharField(max_length=80)
+    platform = models.CharField(max_length=128)
+
```
abeglova: short_description is missing here

course_catalog/migrations/0047_add_video_resource.py
```diff
@@ -0,0 +1,57 @@
...
+                    ),
+                ),
+                ("video_id", models.CharField(max_length=80)),
+                ("platform", models.CharField(max_length=128)),
+                ("full_description", models.TextField(blank=True, null=True)),
+                ("image_src", models.URLField(blank=True, max_length=400, null=True)),
+                ("last_modified", models.DateTimeField(blank=True, null=True)),
+                ("published", models.BooleanField(default=True)),
+                ("url", models.URLField(max_length=2048, null=True)),
+                ("transcript", models.TextField(blank=True, default="")),
```
abeglova: I'm still figuring out if we will have access to the transcript

course_catalog/migrations/0047_add_video_resource.py
abeglova: Also the api currently returns youtube topic ids but the documentation says that they were only supposed to be support until November 10, 2017. The other way that topics are returned is a list of wikipedia links (!)

course_catalog/migrations/0047_add_video_resource.py
abeglova: Ok! That makes sense

PR #2326, opened 2019-10-22
Allow CRUD operations for UserLists via API

### LINKED ISSUE #2321: Modify UserList API to support creation/deletion/updates
Currently the REST API only supports retrieving a UserList, and adding/removing favorites.  It should be expanded to support creating/deleting UserLists and adding/removing items from a list.

mbertrand:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #2321 

#### What's this PR do?
- Allows UserLists to be created/modified/deleted via REST API calls
- Allows UserListItems to be created/modified/deleted via REST API calls
- Adds permissions for the above

- Go to `http://od.odl.local:8063/api/v0/userlists/` and try to post the following JSON, it should create a new list with correct field values:
```
{
    "title": "MY NEW LIST",
    "short_description": "A description of my new list",
    "privacy_level": "private",
    "list_type": "Learning Path"
}
```

- Go to `http://od.odl.local:8063/api/v0/userlists/{list_id}` and PATCH the following data, the UserList title should be updated:
```
{
"title": "My new title"
}
```
- Try the above again but with a list_id that belongs to another user, you should not be allowed to update it.

- Go to `http://od.odl.local:8063/api/v0/userlists/{id}` and PATCH the following data, two list items should be added to your list:
```
{ "items": 
  [
    {
         "content_type": "course",
        "object_id": <id of existing course>
    },
    {
         "content_type": "program",
        "object_id": <id of existing program>
    }
  ]
}
```
- Try the above again but with a non-existent course id, you should get an error
- PATCH the following data, the existing items' positions should be updated
```
{ "items": [
    {
         "id": <id of item 1>,
         "position": 33
    },
    {
         "id": <id of item 2>,
         "position": 34
    }
]
}
```
- PATCH the following data, the item should be deleted.
```
{ "items": [
    {
         "id": <id of item 1>,
         "delete": true
    }
]
}
```

- PATCH the following data, the 2nd item should be deleted.
```
{ "items": [
    {
         "object_id": <object_id of item 1>,
         "content_type": "course",
         "delete": true
    }
]
}
```


aliceriot:
does this mean that in order to add six items to a new userlist I'd have to issue a POST request for each item, after I've created the list? wondering if it would be possible to have an API where it could be done in one request somehow - just worried about how this could get unwieldy if we're dealing with something like the user re-ordering the items in a list with 50 entries.

mbertrand:
I was assuming the user would be adding/removing things on the list one at a time via the UI, as with Youtube, rather than in batch mode.  Is the user going to be batching multiple items at once instead?  For reordering, I could add logic on the backend that would update all the 'position' values of the items after the item that was changed.

aliceriot:
yeah the main thing I was thinking about was positioning and editing. because if we have a 60-member ordered list, and the user switches the current last item to the first position we'd need to (as far as I can tell looking at the data model) issue a request to edit the 'position' variable for each user list item in the list. this has obvious problems, namely if we do these requests sequentially there is a real possibility on a longer list that the user navigates away or there is a network issue somewhere which leaves the list in an inconsistent state.

the way I would think an API like this should ideally look is one where you just POST/PATCH to a UserList endpoint with an array of object IDs to edit the list, and the backend takes care of the details around creating / updating the UserListItems to save the array that it's been handed. then the frontend doesn't have to know anything about the UserList / UserListItems distinction. it seems to me like UserListItems are more like an implementation detail of how we have to store the data in Postgres.

that way also we just have one API on the frontend. if the API will handle just being handed an array of items and saving that then we can do whatever we want on the frontend (adding an item, removing an item, reordering, etc) with that same interface.

does that make sense?

mbertrand:
I was looking at how Youtube handles reordering, and it does not send the entire array of items in the list, an example request body just sends the id of the video being moved and the id of the preceding video:
```javascript
{"commandMetadata":
    {"webCommandMetadata":
        {"url":"/service_ajax","sendPost":true}},
             "playlistEditEndpoint":
                  {"playlistId":"PL1N347aYQgEiJQ3j-QekBxHg9sTA1-MMl",
                   "actions:
                      [{"action":"ACTION_MOVE_VIDEO_AFTER",
                        "setVideoId":"090796A75D153932",
                        "movedSetVideoIdPredecessor":"56B44F6D10557CC6"}],
                        "params":"CAE%3D"
}}
```

This seems simpler than sending the entire reordered list, and either way the backend will have to update the position value of all the items in the list.  But it can figure out what the positioning of all the other items should be just by the new position of the updated item.

mbertrand:
@aliceriot this is the approach I'm trying out now, what do you think:

- Add a `modified_items` write-only field to the `UserListSerializer`
- Send a PATCH request to the UserList API URL, with the following set for `modified_items`
- To create a new list item:
```
[{
  "content_type": "course",
  "object_id": 2323,
  "action": "create"
}]
```
- To update the position of items:
```
[
  {
   "id": 13,
   "position": 1,
   "action": "update"
  },
  {
   "id": 14,
   "position": 2,
   "action": "update"
  },
  ....etc
]
```
To delete item(s):
```
[
  {
   "id": 13,
   "action": "delete"
  },
  {
   "id": 14,
   "action": "delete"
  },
  ....etc
]
```

Any items not in the `modified_items` list will be left as is.

mbertrand:
@aliceriot I changed the code to allow you to add/update/delete single or multiple items in a UserList.

aliceriot:
awesome, will take a look soon!

aliceriot:
to issue the delete request I have to use the `UserListItem` id of the item, right? I was just wondering, should it work on the basis of the item ID instead? idk if it's possible for their to be collisions there, though...

just wondering about the scenario we discussed of removing the item from lists without necessary having all the information about the lists it's in - maybe we just have the names, but we don't have the whole list and the information about the `UserListItem` ids

aliceriot:
other than that question it works great!

mbertrand:
@aliceriot UserListItem id and item id should be the same thing.  So if a UserList (id of 5) has 3 UserListItems, with id's 101, 102, 103, and you want to remove #102 and #103 from the list, you'd send this request to `http://od.odl.local:8063/api/v0/userlists/5`:
```
{"items": [{"id": 102, "delete": true}, {"id": 103, "delete": true}]
```

Each list item id should be unique.

mbertrand:
Do you mean the id of whatever the UserListItem links to (like a course, program, etc)?

aliceriot:
yeah the first thing I tried was the id of the course - then I thought it wasn't working

aliceriot:
I guess we have to do it this way to deal with the possibility of, say, a program and a course having the same id, was just hoping that the frontend wouldn't have to know anything about `UserListItems`. but that's ok :)

mbertrand:
@aliceriot could send both the item's `object_id` and `course_type`, this would prevent collisions.  I can look into allowing deletes using either way: `{"id": 3, "delete": true}` or `{"object_id": 7544, "content_type": "course", "delete": true}`

aliceriot:
works great! reading through code now

PR #2343, opened 2019-10-29
Added tasks and commands to fetch youtube videos

### LINKED ISSUE #2335: Video Search: cron jobs for syncing data
We need celery tasks hooked into cron to periodically run and pull in data
  - Implement celery tasks to run our ETL pipeline
  - Use APIs integrated with in #2303 
  - Configurable frequency of checks via env var

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #2335 

#### What's this PR do?
- Adds cron task for syncing 
- Adds a management command that supports delete and fetch operations

- Run the task, it should immediately complete
- Run the command, a few examples:
  - `./manage.py backpopulate_youtube_data delete` - removes all `Video` records for youtube
  - `./manage.py backpopulate_youtube_data fetch` - fetch all youtube data (same as the cron task)
  - Run the above fetch commands also specifying `--channel-id ID`, where `ID` is one listed here: https://github.com/mitodl/open-video-data/blob/master/youtube.yaml


rhysyngsun:
@mbertrand this is all set for review, some of the functionality was removed as it was no longer valid due to upstream changes on master

course_catalog/etl/loaders.py
```diff
@@ -203,10 +204,11 @@ def load_video(video_data):
     load_topics(video, topics_data)
     load_offered_bys(video, offered_bys_data)
 
-    if not created and not video.published:
-        search_task_helpers.delete_video(video)
-    elif video.published:
-        search_task_helpers.upsert_video(video)
+    if features.is_enabled(features.VIDEO_INDEXING):
```
mbertrand: Should the youtube task also run only if this is enabled?

rhysyngsun: My thought was that we wanted to pull the data into the database, but not have them show up on the search page (so we could validate the data), but I think this PR has sat long enough we may not need this at all, let me check on that.

rhysyngsun: Removed this flag after verifying with peter it's not needed

rhysyngsun:
@mbertrand feedback addressed, I also reverted the changes to `requirements.[in|txt]` since that dependency is no longer used and should've been in `test_requirements.txt` anyway

PR #2368, opened 2019-11-06
add transcripts

abeglova:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [ ] Testing
  - [x] Code is tested
- [ ] Settings

https://github.com/mitodl/open-discussions/issues/2375

#### What's this PR do?
This pr adds functions to download and parse the transcript for a video.

```
  from course_catalog.etl import youtube
  v = Video.objects.all()[0]
  c = youtube.get_captions_for_video(v)
```
Should generate the captions in xml format

```
youtube.parse_video_captions(c)
```

Should parse the transcript from the captions
(Optional)

Generating transcripts is fairly slow. On my machine, I was able to process the transcripts for only 30 videos per minute. Transcripts will need to be generated and updated in a separate cron job from the rest of the youtube video etl

(Optional)

(Optional)


course_catalog/etl/youtube.py
```diff
@@ -82,6 +84,67 @@ def get_videos_for_playlist(youtube_client, playlist_id, offered_by):
...
 
+def get_captions_for_video(video):
+    """
+    Function which fetches and returns xml captions for a video object
+
+    Args:
+        video (object) : Video Object
+
+    Returns:
+        a string with the transcript in xml format
```
rhysyngsun: This needs a type for the return value - `str:`

course_catalog/etl/youtube.py
```diff
@@ -82,6 +84,67 @@ def get_videos_for_playlist(youtube_client, playlist_id, offered_by):
             break
 
 
+def get_captions_for_video(video):
+    """
+    Function which fetches and returns xml captions for a video object
+
+    Args:
+        video (object) : Video Object
```
rhysyngsun: I think this is actually a `course_catalog.models.Video` type instead of `object`?

course_catalog/etl/youtube.py
```diff
@@ -82,6 +84,67 @@ def get_videos_for_playlist(youtube_client, playlist_id, offered_by):
...
+        None,
+    )
+
+    if not selected_capiton:
+        selected_capiton = next(
+            (caption for caption in all_captions if caption.code == "en"), None
+        )
+
+    if selected_capiton:
+        return selected_capiton.xml_captions
```
rhysyngsun: Could you add an explicit `return None` if there's no selected caption

course_catalog/etl/youtube.py
```diff
@@ -82,6 +84,67 @@ def get_videos_for_playlist(youtube_client, playlist_id, offered_by):
...
+
+    Returns:
+        a string with the transcript in xml format
+
+    """
+
+    pytube_client = pytube.YouTube(video.url)
+    all_captions = pytube_client.captions.all()
+
+    selected_capiton = next(
```
rhysyngsun: I'd suggest using this as an alternative to using `next()`, `next` is sort-of a lower-level operation, it's something you'll probably rarely use and typically if you find yourself using it there's probably a more Pythonic:

```python
english_captions = [caption for caption in all_captions if caption.code == "en"]
# sort auto-generated captions to the bottom of the list
sorted_captions = sorted(english_captions, key=lambda caption: "auto-generated" in caption.name)

return sorted_captions[0] if sorted_captions else None
```

A few key differences here:

- Only filters out non-english captions once
- The sort uses the property of `bool` values evalating as the integer values 0 and 1 so that means it sorts with auto-generated ones at the bottom
- Since the list is now sorted, we can just grab the first item in the list, if any and return that or `None` if not

course_catalog/etl/youtube.py
```diff
@@ -82,6 +84,67 @@ def get_videos_for_playlist(youtube_client, playlist_id, offered_by):
...
+
+    """
+
+    if not xml_captions:
+        return ""
+
+    transcript = ""
+    root = ElementTree.fromstring(xml_captions)
+
+    for child in root.iter():
```
rhysyngsun: Could you update this to separate items by a newline?

course_catalog/etl/youtube.py
```diff
@@ -82,6 +84,67 @@ def get_videos_for_playlist(youtube_client, playlist_id, offered_by):
...
+        text = child.text or ""
+        text = unescape(text.replace("\n", " ").replace("  ", " "))
+
+        if text:
+            transcript += " " + text
```
rhysyngsun: It's best to avoid string concatenation when building up a string - every time you have a `+` here it adds the overhead of creating a new string, so in this case the code is potentially generating 2 new strings per item. It's probably not much in the small scale but it can add up.

A more performant way to approach this is typically to append each piece of text to a list and then use [`str.join`](https://docs.python.org/2/library/string.html#string.join) to combine them in a single operation.

course_catalog/etl/youtube.py
```diff
@@ -82,6 +84,67 @@ def get_videos_for_playlist(youtube_client, playlist_id, offered_by):
...
+    if selected_capiton:
+        return selected_capiton.xml_captions
+
+
+def parse_video_captions(xml_captions):
+    """
+    Function which parses xml captions and returns a string with just the transcript
+
+    Args:
+        xml_captions (string) : Caption with xml with text and timestamps
```
rhysyngsun: `string` -> `str`

course_catalog/etl/youtube.py
```diff
@@ -82,6 +84,67 @@ def get_videos_for_playlist(youtube_client, playlist_id, offered_by):
...
+
+    Returns:
+        a string with the transcript with timestamps removed
```
rhysyngsun: Needs a return type

rhysyngsun:
Functionality looks good

course_catalog/etl/youtube.py
```diff
@@ -89,60 +89,50 @@ def get_captions_for_video(video):
     Function which fetches and returns xml captions for a video object
 
     Args:
-        video (object) : Video Object
+        video (course_catalog.models.Video)
 
     Returns:
-        a string with the transcript in xml format
+        transcript in xml format (str)
```
rhysyngsun: The format for return types is slightly different: `str: transcript in xml format`

course_catalog/etl/youtube_test.py
```diff
@@ -151,44 +151,44 @@ def mock_pytube_caption_object(code, name, xml_caption):
...
     ]
 
-    with patch("pytube.YouTube") as mock_pytube_client:
-        mock_pytube_client.return_value.captions.all.return_value = captions_list
+    mock_pytube_client = mocker.patch("pytube.YouTube")
+    mock_pytube_client.return_value.captions.all.return_value = captions_list
 
-        video = MagicMock()
-        assert youtube.get_captions_for_video(video) == "English: Not Auto-generated"
+    video = Mock()
```
rhysyngsun: `mocker.Mock()` here so you don't need the import at the top of the file

rhysyngsun:
:+1: when travis passes

PR #2379, opened 2019-11-13
Show a filled-in star when a resource is in a user's list

### LINKED ISSUE #2358: Fill in "favorited" star on resource card if it is included in the user's lists
As a user, I'd like the star in each learning resource card to be filled in if either of the following is True:
- the resource has been favorited
- the resource is in one of my lists (top level, not nested in a list inside a list)

#### Designs and Mockups
- Create a `InListSerializerMixin` similar to `FavoriteSerializerMixin` and add it to the relevant learning resource serializers.
- Change the front end code to fill in the star if either `is_favorite` or `in_list` is true.

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #2358

#### What's this PR do?
- Fills in the 'favorited' star on a `LearningResourceCard` if the resource is in one of the user's lists. 

- Click the 'star' icon on a resource (both on the search results page and the learning home page).  Add it to a list but don't favorite it.  The star should get filled in.  Remove it from the list and the star should revert back.

### DIFF
--- a/course_catalog/serializers.py
+++ b/course_catalog/serializers.py
@@ -79,6 +79,28 @@ class FavoriteSerializerMixin(serializers.Serializer):
             return False
 
 
+class ListsSerializerMixin(serializers.Serializer):
+    """
+    Mixin to serialize lists for various models
+    """
+
+    lists = serializers.SerializerMethodField()
+
+    def get_lists(self, obj):
+        """
+        Return a list of user's lists/path id's that a resource is in.
+        """
+        request = self.context.get("request")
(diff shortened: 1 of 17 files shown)


rhysyngsun:
Code looks good, but I noticed one place where the star button isn't behaving as described: if you go to `/courses/lists/ID` none of the stars are filled in, however, at `/courses/lists/favorites` they are.

mbertrand:
@rhysyngsun good catch, the request context wasn't being passed down to the nested item serializers so on the item detail page the star would never be filled in (whether a favorite or in a list).  Should work now with the latest commit.

PR #2383, opened 2019-11-14
Filter out unauthored lists on UserListsPage and AddToListDialog

### LINKED ISSUE #2382: Favorited user lists leak into UI that shows my own lists
### Steps to Reproduce

- login as user A, confirm you have no lists at /courses/lists/
- as user A, favorite some of user B's lists
- as user A navigate directly to /courses/search (so your react state starts from zero)
- click on the star next to a search result and observe that these lists from the other user that you favorited appear in your lists
- click the "My Lists" link on the header and observe that the lists you favorited show up as yours

The issue seems to be that this code transforms those objects onto `entities.userLists`, but the UI pieces don't filter those down to lists authored by the current user:
https://github.com/mitodl/open-discussions/blob/master/static/js/lib/queries/learning_resources.js#L36-L47

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #2382 

#### What's this PR do?
Filters out any lists in `state.entities.userLists` that are not authored by the current user

- As user A, search for user lists or learning paths and favorite 1+ that are authored by user B
- Go to the "My Lists" page, the recently favorited lists by user B should not show up here
- Click the star on any resource card to open the 'Add to List' dialog, only your own lists should show up here as well.

### DIFF
--- a/static/js/factories/learning_resources.js
+++ b/static/js/factories/learning_resources.js
@@ -67,7 +67,7 @@ export const makeRun = (): LearningResourceRun => {
         full_name:  casual.name
       }
     ],
-    prices: [{ mode: "audit", price: casual.number }]
+    prices: [{ mode: "audit", price: casual.integer(1, 1000) }]
   }
 }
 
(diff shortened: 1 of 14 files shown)


aliceriot:
works great! looking at code now.

static/js/factories/learning_resources.js
```diff
@@ -163,7 +163,8 @@ export const makeUserList = (): UserList => ({
   list_type:     "userlist",
   profile_img:   casual.url,
   profile_name:  casual.name,
-  privacy_level: casual.random_element(["public", "private"])
+  privacy_level: casual.random_element(["public", "private"]),
+  author:        casual.number
```
aliceriot: I think here for `author` you need to use `casual.integer(low, high)`, where `low` and `high` are integer values for a range from which the function should select an integer (e.g. 1 and 1000 or something)

aliceriot: just realized we are using `casual.number` in a number of places - but if you check in the repl `casual.number` isn't a function that `casual` provides

mbertrand: I replaced all the `casual.number`s with `casual.integer`s

Issue #2378, opened 2019-11-13
API response for list of user lists gets too large


mbertrand:
### Steps to Reproduce

- Make some user lists with many courses
- Add those user lists to other user lists

### Expected Behavior

View 'My Lists', it should load quickly

### Actual Behavior

It takes awhile to load because the API response is so large.  

The API response should be simplified to exclude most of the list item data when retrieving all lists.  That information is only required when viewing the details of a specific list.

aliceriot:
this is a good idea to address now I think, users could have lists with hundreds of items in them so it could get pretty bad.

for the UI on the list index page (`/courses/lists/`) the only piece of data we'll need about the items _in_ the list to display the UI we're currently displaying is how many items and the cover image for the first one

aliceriot:
@mbertrand is there more to do in order to simplify / smallify the response?

mbertrand:
If/when we figure out a way to get the star button to work without requiring all favorites/lists at once, then reducing the page size, but maybe that should be a separate issue and I can close this one.

PR #2385, opened 2019-11-15
Simplified serializers for UserListView list response

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

- Partially address #2378

#### What's this PR do?
- When requesting a list of UserLists, returns a simplified response via a UserListItem serializer that contains only the essential properties needed to maintain existing functionality (favoriting, adding/remove items to list, etc).

- The 'My Lists' page should still work as before but load faster
- You should still be able to add/remove items in a list, un/favorite items, create/delete lists, view list details

### DIFF
--- a/course_catalog/serializers.py
+++ b/course_catalog/serializers.py
@@ -79,6 +79,28 @@ class FavoriteSerializerMixin(serializers.Serializer):
             return False
 
 
+class ListsSerializerMixin(serializers.Serializer):
+    """
+    Mixin to serialize lists for various models
+    """
+
+    lists = serializers.SerializerMethodField()
+
+    def get_lists(self, obj):
+        """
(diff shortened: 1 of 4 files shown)


aliceriot:
wow, the `/courses/lists/` page is really speedy now!

I think though that we need to tweak something, because as-is this breaks the display of a thumbnail for the lists:

coverimagesnotthere

for reference, on `master` currently this cover image is taken from the first item in the list. maybe we can modify the serializer to grab that value and return a `thumbnail` or `image_src` attribute on the userlist itself?

mbertrand:
oops, sorry for missing that! Will fix.

mbertrand:
@aliceriot should be fixed now

Issue #2395, opened 2019-11-19
Add and remove subjects (topics) for a list


mbertrand:
As a user, I'd like to add/remove subjects for a list.

#### Designs and Mockups

The list form should include a multiselect dropdown of available topics, and buttons with x's for selected topics (clicking the x will remove the topic).  Users should not be able to create their own topics.

@aliceriot do you know of an npm module that provides a component like the above?

#### Acceptance Criteria:

- [x] API supports viewing, adding, removing list categories
- [ ] UI for viewing, adding, removing list categories

mbertrand:
@pdpinch @abdulkdawson @Ferdi @aliceriot latest screenshots:

Initial view:
![Screen Shot 2019-11-21 at 12 33 54 PM](https://user-images.githubusercontent.com/187676/69361942-8324b680-0c5b-11ea-96ac-2319e0e47942.png)

No dropdown will be shown until the user either clicks the arrow button, or starts typing, at which point options with matching text are shown:
![Screen Shot 2019-11-21 at 12 34 35 PM](https://user-images.githubusercontent.com/187676/69361911-6d16f600-0c5b-11ea-9689-f521c85ad89c.png)

Once a subject is selected, it is added to the display as a button with an x to remove it, and all dropdown options are shown until/unless user starts typing again:
![Screen Shot 2019-11-21 at 12 35 06 PM](https://user-images.githubusercontent.com/187676/69362123-e878a780-0c5b-11ea-965b-e38aa4af2ded.png)

PR #2397, opened 2019-11-19
Support topics CRUD in UserList API

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Partially addresses #2395 

#### What's this PR do?
Supports changing the topics in a list on creation/update via API

- Go to `http://od.odl.local:8063/api/v0/lists`
- Enter the following data and post it to create a new list with topics:
  ```
  {
      "topics": [{"id": <valid_topic_id>}, {"id": <valid_topic_id>}],
      "title": "A new list with topics",
      "short_description": "short description",
      "privacy_level": "private",
      "list_type": "userlist"
  }
  ```
- a new list should be created with the topics you selected
- Go to `http://od.odl.local:8063/api/v0/userlists/<new_userlist_id>/`
- Set the patch data to include a different topic or no topic at all, for example:
  ```
  {"topics": [{"id": <different_topic_id>}]}
  ```
  or
  ```
  {"topics": []}
  ```
- The userlist's topics should be updated when the data is patched.
- If you don't include "topics" in the request data at all, the current list topic(s) if any should remain unchanged.

### DIFF
--- a/course_catalog/serializers.py
+++ b/course_catalog/serializers.py
@@ -32,10 +32,6 @@ from course_catalog.utils import (
 from open_discussions.serializers import WriteableSerializerMethodField
 from search.task_helpers import upsert_user_list
 
-import logging
-
-log = logging.getLogger()
-
 
 class GenericForeignKeyFieldSerializer(serializers.ModelSerializer):
     """
@@ -480,7 +476,6 @@ class SimpleUserListSerializer(
 
     def validate_topics(self, topics):
         """Validator for topics"""
-        log.error(f"VALIDATING {topics}")
         if CourseTopic.objects.filter(
             id__in=[topic["id"] for topic in topics]
         ).count() != len(topics):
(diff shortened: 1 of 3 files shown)


course_catalog/serializers.py
```diff
@@ -445,9 +445,24 @@ class SimpleUserListSerializer(serializers.ModelSerializer, FavoriteSerializerMi
     """
 
     items = SimpleUserListItemSerializer(many=True, allow_null=True, read_only=True)
-    topics = CourseTopicSerializer(read_only=True, many=True, allow_null=True)
+    topics = WriteableSerializerMethodField()
     object_type = serializers.CharField(read_only=True, default="userlist")
 
+    def validate_topics(self, topics):
+        """Validator for topics"""
+        topic_ids = {topic["id"] for topic in topics}
```
rhysyngsun: This would be prone to a `KeyError` if `"topics": [{}]` was passed

PR #2416, opened 2019-11-25
Prevent users from adding lists to each other

### LINKED ISSUE #2414: Recursion error when adding lists to each other
### Steps to Reproduce

Add List 1 to List 2, then add List 2 to List 1 . --> Recursion error

Need to prevent users from doing do.

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes #2414

#### What's this PR do?
Prevents users from adding lists to each other (List 1 in List 2 and List 2 in List 1)

- Create List 1
- Create List 2
- Click the List 2 'favorite' star and add it to List 1
- Click the List 1 star, List 2 should be excluded from the checkbox list

### DIFF
--- a/static/js/components/AddToListDialog.js
+++ b/static/js/components/AddToListDialog.js
@@ -122,7 +122,16 @@ export default function AddToListDialog() {
           (![LR_TYPE_LEARNINGPATH, LR_TYPE_USERLIST].includes(
             resource.object_type
           ) ||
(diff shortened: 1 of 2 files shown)


noisecapella:
This PR fixes the A -> B -> A case but not the A -> B -> C -> A case. But @mbertrand said future work with the serializer may fix the latter issue.

PR #2423, opened 2019-11-27
Calculate and return 'is_favorite' and 'lists' fields in ES search results

### LINKED ISSUE #2422: Return user-specific "is_favorite" and "lists" attributes for each search result
As a user, I'd like to get search results indicating which ones are favorited or in one of my lists, without having to query the API for the full list of my favorites and lists first.

#### Designs and Mockups
Use the approach described in #2142: implement Elasticsearch script fields to return user-specific info on favorites/lists for each search result.

#### Acceptance Criteria:

- [ ] Include 'is_favorite' and 'lists' (array of list ids) in each search result
- [ ] Remove API calls on the `CourseSearchPage` to retrieve all favorites and user lists

#### Out of Scope

- Shrink the page size of the favorites and userlists API calls from 1000 to 10 (this will require a paging mechanism on the UI too).

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #2422 

#### What's this PR do?
- Uses scripted fields to include user-specific "is_favorite" and "lists" fields in each ES search result
- Removes API calls to retrieve all favorites and lists on the `CourseSearchPage`

- Run a search on the course search page.  The stars should be filled on any items that you've already favorited or added to lists.
- Add more search results as favorites or to lists, and unfavorite/unlist others.  The stars for the results should change appropriately.


search/api.py
```diff
@@ -133,6 +141,55 @@ def gen_video_id(video_obj):
...
+    Args:
+        user(User): a user to retrieve UserListItems for
+
+    Returns:
+        dict: a mapping of list ids (value) per resource item ES id (key)
+
+    """
+    lists_dict = collections.defaultdict(list)
+    items = (
+        UserListItem.objects.filter(user_list__author=user)
```
rhysyngsun: Suggest we add a `.only("id", "user_list_id")` here so we're only fetching the fields we absolutely need to return the result.

search/api.py
```diff
@@ -133,6 +141,55 @@ def gen_video_id(video_obj):
...
+        .select_related("user_list")
+        .iterator()
+    )
+    for item in items:
+        lists_dict[gen_doc_ids([item])[0]].append(item.user_list.id)
```
rhysyngsun: If we alter this to `item.user_list_id` then we don't need to do the join/`select_related` on line 185 and it should be a bit more performant.

search/api.py
```diff
@@ -215,18 +287,39 @@ def execute_search(*, user, query):
     search = Search(index=index, using=get_conn())
     search.update_from_dict(query)
     search = _apply_general_query_filters(search, user)
-    return transform_aggregates(search.execute().to_dict())
+    if not user.is_anonymous and is_learning_query(query):
```
rhysyngsun: This `is_learning_query` is ok for now, but I think this kind of introspection of the incoming query in order to specialize the query further is a sign we need to look into providing an alias for the current search endpoint that applies different query logic for different logical (learning resources vs. posts/comments) searches. It's also impossible to detect when this logic breaks outside of a full integration test between the frontend and backend.

That said, I think this approach is a reasonable tradeoff vs. spending time on a separate endpoint for the time being since things shouldn't change that much here, but we should keep an eye on it and any further changes of a similar nature should probably be the trigger point for such a refactor.

PR #2462, opened 2019-12-05
Don't allow userlists to be added to userlists

### LINKED ISSUE #2461: Prevent users from adding lists to lists
Prevent users from adding a user list/path to another userlist/path.

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #2461 

#### What's this PR do?
Prevents users from adding lists to lists

On database migration, any existing userlist items in lists should be removed
Clicking the star on any resource besides a list/path should work the same as before.
Clicking the star on a list/path should only give you "Favorites" as an option


mbertrand:
@Ferdi @abdulkdawson here's what the 'Add To List' dialog looks like for a list/path: I removed all checkboxes except 'Favorites" as well as the 'Create New List' button - is that okay?

Screen Shot 2019-12-05 at 3 25 41 PM

Ferdi:
@mbertrand I think that's good. Does adding to favorites affects performance ?

mbertrand:
@ferdi yes, a large userlist favorite will have the same problem

static/js/components/AddToListDialog.js
```diff
@@ -123,22 +123,12 @@ export default function AddToListDialog() {
...
-                    item.object_id === userList.id &&
-                    [LR_TYPE_LEARNINGPATH, LR_TYPE_USERLIST].includes(
-                      item.content_type
-                    )
-                )
-              ))) ? (
+      {resource &&
+      ![LR_TYPE_LEARNINGPATH, LR_TYPE_USERLIST].includes(
+        resource.object_type
+      ) ? (
```
aliceriot: we end up making this check in a bunch of places - what do you think about adding a `isUserList` or similar function in `lib/learning_resources`?

mbertrand: Good idea, just pushed that change

PR #2463, opened 2019-12-05
add history navigation to drawer

### LINKED ISSUE #2440: add history and 'back' button to user list browsing
when you have a drawer open for looking at search results you can click on like a course in say a user list, and the drawer will change to show details of the course. theoretically you could have a user list in another user list, and so on, so it could be possible to have a number of levels of navigation here. We want to store those changes in a history, so that the user can back out to where they started without just closing the drawer.

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #2440

#### What's this PR do?

This refactors how we control what is shown in the `LearningResourceDrawer`. Previously we basically set an object ID and object type, and then we'd use those to fetch the right object, and then we'd show it. This worked for opening and looking at one object, but it broke down a little when it came to navigating to a new object from the initial one. The specific problem is that if you click in you can't get back to the object you we're looking at before without closing the whole drawer and opening and navigating again.

We need to support this use-case well because the user can do this from a userlist or a learning path.

Anyhow, to fix this I changed it so it works on the basis of a history object, which we store in the `ui` reducer. It's just an array, and this is how it works:

- if the array is empty then nothing is on the history, so don't show the drawer
- if there is any item on the history, then show the drawer
- take the last item on the history and use that item ID and object type to fetch an object to display in the `ExpandedLearningResourceDisplay`

so I implemented 3 actions: push, pop, and clear. they do what you'd expect. overall it seems to work pretty well!

Make sure you can open the drawer for any resource. Initially, you should just see a 'close' button in the upper right of the drawer.

If you click on a user list, you should then be able to click on an item in that list, and it should then open in the drawer.

Then you should additionally see a 'back' button, which will take you back to the initial item.

initial item:

initialffff

second item:

initialfffffffdasdfasdf

### DIFF
--- a/static/js/actions/ui.js
+++ b/static/js/actions/ui.js
@@ -22,10 +22,14 @@ export const setShowDrawerMobile = createAction<string, *>(
 export const SET_SHOW_DRAWER_HOVER = "SET_SHOW_DRAWER_HOVER"
 export const setShowDrawerHover = createAction<string, *>(SET_SHOW_DRAWER_HOVER)
 
-export const SET_SHOW_RESOURCE_DRAWER = "SET_SHOW_RESOURCE_DRAWER"
-export const setShowResourceDrawer = createAction<string, *>(
-  SET_SHOW_RESOURCE_DRAWER
-)
+export const PUSH_LR_HISTORY = "PUSH_LR_HISTORY"
+export const pushLRHistory = createAction<string, *>(PUSH_LR_HISTORY)
+
+export const POP_LR_HISTORY = "POP_LR_HISTORY"
+export const popLRHistory = createAction<string, *>(POP_LR_HISTORY)
+
+export const CLEAR_LR_HISTORY = "CLEAR_LR_HISTORY"
+export const clearLRHistory = createAction<string, *>(CLEAR_LR_HISTORY)
 
 export const SET_SNACKBAR_MESSAGE = "SET_SNACKBAR_MESSAGE"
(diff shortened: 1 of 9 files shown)


noisecapella:
Is it possible to do the history navigation in the react-router browser history rather than in the redux store?

aliceriot:
yeah I think that should be possible, I'll give it a go and see if I run into any difficulties

aliceriot:
alright, so I gave it a go and go things partially working, but I ran into some difficulties which make me think driving this feature purely off the browser history isn't the best way to go.

there's a few things that are difficult to coordinate properly, mainly the fact that I need to show a back button in the drawer whenever the user has gone more than one "level" deep in the drawer (i.e. clicked from a user list to a course which is listed therein). it's difficult to tell whether this is the case from the history and URL state alone, since on `window.history` there is no access to history entries for security reasons.

this means that even if we're driving things off the URL I have to store somewhere the state of whether the user has opened the drawer, how many items we've already looked at, etc - so I have to keep more or less the existing code even if I'm now using `history.goBack()` to navigate instead of dispatching a redux action.

now, I do need to drive a feature we want to have off of the URL, namely, being able to link to a state where the drawer opens. But I'm thinking the best way to accomplish that is to keep this redux-based history setup and have a `useEffect` call that runs only at component mount, looks for URL parameters and, if they are present, dispatches an action to open the drawer. 

what do you think?

noisecapella:
Sure, makes sense to me

Issue #2464, opened 2019-12-06
Enable paging for userlists, userlist items, and favorites


mbertrand:
As a user, I'd like to be able to page through userlists, userlist items, and favorites instead of retrieving all of them at once.

#### Designs and Mockups
The UI needs to be changed to support this, @abdulkdawson should it be like search in that it automatically retrieves more as you scroll down the page?
- The following still needs work with paged results: 
  - 'Add to List Dialog' needs to show all the user's lists, or have a paging UI tool
  - 'Favorite' star needs to be correctly filled in (can be done via #2423)

PR #2470, opened 2019-12-08
List Items API

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

- Partially addresses #2464

#### What's this PR do?
- Adds a nested UserListItems API

- All the following should work the same as before from the UI:
  - Adding/removing items from lists
  - Creating/updating/deleting lists

New functionality:
- Go to `http://od.odl.local:8063/api/v0/userlists/<user_list_id>/items/` for one of your own lists or another user's public list, the response should include the items and their details in that list.
- Try the above for another user's private list, you should get a 403
- Try POSTing a new item, it should be created:
  ```
  { "content_type": "course", "object_id": <course_id>}
  ```
- Go to `http://od.odl.local:8063/api/v0/userlists/<user_list_id>/items/<item_id>/` for one of the items in your own list or another user's public list, the response should include the item details.
- Try the above for another user's private list, you should get a 403
- Try PATCHing the item with a new position, the position should be updated.
  ```
  { "position": 8675309}
  ```
- Try DELETEing the item, it should be deleted.


course_catalog/permissions.py
```diff
@@ -22,3 +23,26 @@ def has_object_permission(self, request, view, obj):
                 or request.user == obj.author
             )
         return request.user == obj.author
+
+
+class HasUserListItemPermissions(BasePermission):
+    """Permission to view/modify UserListItems"""
+
+    def has_permission(self, request, view):
+        user_list = UserList.objects.get(
```
noisecapella: Might not be important but this will cause a 500 error when a user views this API with the wrong id. For a 404 we would need to use `get_object_or_404` or some similar logic.

noisecapella:
Just the one take-or-leave comment. Everything looks good. If you're ready to merge let me know and I'll approve.

course_catalog/permissions.py
mbertrand: Fixed this and added a test for it

PR #2480, opened 2019-12-11
Display similar learning resources in drawer

### LINKED ISSUE #2436: Display similar learning resources in drawer
As a search user, I'd like to see a list of similar resources when I open the drawer to view the details of al learning resource.

#### Designs and Mockups

https://projects.invisionapp.com/d/main#/console/18704964/392640678/preview

#### Acceptance Criteria:

- [ ] Add "Similar Learning Resources" header below learning resource details in drawer
- [ ] Display list of 3 similar resources when expanded
- [ ] Clicking a resources replaces current drawer contents with resource details
  - [ ] display back button so user can return to prior item (similar to user list UX)

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #2436 

#### What's this PR do?
- Adds a backend API to search for similar resources and adjusts the UI to display them.

- Open a drawer for any resource, from either the course search page or home page.  At the bottom there should be a 'Similar Learning Resources' section with a '+' next to it.
- Click the '+', it should change to a '-' and the similar resources should be displayed.
- Click the '-' it should change to a '+' and the similar resources should be hidden again.

Initially I tried sending `object_id` and `object_type` to the API as parameters, then using `like={"_id": gen_doc_id(object_id, object_type)}` for the ES 'More Like This' query, but this only returned resources of the same type.  

So I switched to sending a document body in the request containing the following fields -  `id`, `object_type`, `title`, `short_description` - and using `like={"doc": value_doc}` instead.  Another option would be to continue sending object id and type, retrieve and serialize the correct object from that, and use the serializer data as the `value_doc`.  But that's an additional database call so I decided to not do it this way.

(Optional)


mbertrand:
@abdulkdawson @pdpinch @Ferdi screenshots of 'similar resources':

Screen Shot 2019-12-12 at 9 45 58 AM

Screen Shot 2019-12-12 at 9 46 18 AM

aliceriot:
works great! the resources seem generally pretty similar, and it works well with the drawer history I just merged. reading code now

search/api.py
```diff
@@ -316,6 +317,42 @@ def find_related_documents(*, user, post_id):
...
+    return [
+        hit.to_dict()
+        for hit in response.hits
+        if (
+            hit["id"] != value_doc.get("id", None)
+            or hit["object_type"] != value_doc.get("object_type", None)
+        )
+    ][0 : settings.OPEN_DISCUSSIONS_SIMILAR_RESOURCES_COUNT]
+
+
```
aliceriot: just wondering, we have a similar thing above for getting related posts. in this method we limit the returned 'object_type' to return only learning resources on line 335 - it looks like we don't have a similar limitation on the post code above. do you think we should add one while we're in here? was just thinking that maybe that could end up returning learning resources on the 'related posts' thing, which the UI is currently not set up to handle (and I'm not sure we would want that besides).

static/scss/button.scss
```diff
@@ -87,7 +87,8 @@ a.blue-btn {
   font-size: 12px;
 }
 
-.remove {
+button.remove,
+a.remove {
```
aliceriot: :+1:

static/js/components/ExpandedLearningResourceDisplay.js
```diff
@@ -246,6 +251,28 @@ const ExpandedLearningResourceDisplay = (props: Props) => {
...
+            ? similarItems.map((resource, i) => (
+              <LearningResourceRow
+                key={i}
+                object={searchResultToLearningResource(resource)}
+                searchResultLayout={SEARCH_LIST_UI}
+              />
+            ))
+            : null}
+        </div>
+      ) : null}
```
aliceriot: two questions:

- I seem to remember some discussion that the list of items, in the case of a program, userlist, or learningpath, should collapse similarly to the similar items list. would it make sense to add this behavior in while we're in here?
- also since we have some duplicate code here between the items list and the similar items list, it might be helpful to create a separate component in this same file which takes like a title and list of items and does the rest. what do you think?

search/api.py
mbertrand: For some reason that remains a mystery to me, a 'More Like This' query to ES using `like={"_id:" <doc_id>}`, as done in `related_posts`, only returns docs of the same type.  But `like={"doc": <values_doc>}` as used for similar resources, returns multiple types.

static/js/components/ExpandedLearningResourceDisplay.js
mbertrand: Sure, I'll make those changes

search/api.py
aliceriot: oh haha, sounds like some obscure elasticsearch quirk

static/js/components/ExpandedLearningResourceDisplay.js
mbertrand: Changes made.  @abdulkdawson here's a screenshot of both userlist/program items and similar resources, in their default collapsed state - I put 5px of padding between them, does that look okay?

Screen Shot 2019-12-13 at 2 05 30 PM

mbertrand: @aliceriot ready for review again

PR #2482, opened 2019-12-12
Fix bug unchecking lists

### LINKED ISSUE #2481: Unchecking a checked list on AddToListDialog sometimes unchecks wrong box
### Steps to Reproduce

- Make a few lists
- Check a few lists on the AddToListDialog then close
- Reopen the dialog and uncheck some of the boxes

### Expected Behavior

The resource is removed from that list and the box you clicked on is unchecked

### Actual Behavior

The resource is removed from that list and sometimes another box you didn't click on is unchecked

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #2481

#### What's this PR do?
Fixes #2481

- Make a few lists
- Check a few lists on the AddToListDialog then close
- Reopen the dialog and uncheck some of the boxes

### DIFF
--- a/static/js/components/AddToListDialog.js
+++ b/static/js/components/AddToListDialog.js
@@ -100,8 +100,8 @@ export default function AddToListDialog() {
     const matchingList = find(propEq("list_id", userListId), resource.lists)
     if (checked && !matchingList) {
       resource.lists.push({ list_id: userListId })
-    } else {
-      resource.lists.pop(matchingList)
+    } else if (matchingList && !checked) {
+      resource.lists.splice(resource.lists.indexOf(matchingList), 1)
     }
   }
 
--- a/static/js/components/AddToListDialog_test.js
+++ b/static/js/components/AddToListDialog_test.js
@@ -123,6 +123,35 @@ describe("AddToListDialog", () => {
     })
   })
 
+  it(`should uncheck the correct userlist`, async () => {
+    course.lists.push({ list_id: userLists[0].id, item_id: 1 })
+    course.lists.push({ list_id: userLists[1].id, item_id: 1 })
+    course.lists.push({ list_id: userLists[2].id, item_id: 1 })
+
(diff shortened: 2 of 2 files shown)


aliceriot:
ah, I ran into this issue too but thought it was just because of my slow computer!

static/js/components/AddToListDialog.js
```diff
@@ -100,8 +100,8 @@ export default function AddToListDialog() {
     const matchingList = find(propEq("list_id", userListId), resource.lists)
     if (checked && !matchingList) {
       resource.lists.push({ list_id: userListId })
-    } else {
-      resource.lists.pop(matchingList)
+    } else if (matchingList && !checked) {
+      resource.lists.splice(resource.lists.indexOf(matchingList), 1)
     }
   }
```
aliceriot: is this `updateResource` basically just to modify the object on the frontend in order to render the right UI?

aliceriot: just wondering if there might be a better way to do this overall - i.e. if we're going to be editing the object it is probably better to that in the redux store itself rather than indirectly here

mbertrand: Yes (1st comment), maybe? (2nd) - not sure what the best way of doing that is, the userlist entitities are being modified/returned in the API calls rather than the resource being added to those lists.  If I modify `updateResource` to be something like:
```
  const [, updateResource] = useRequest(
    resource ? courseRequest(resource.id) : null
  )
```
then the dialog is closed as soon as I check or uncheck a box.

aliceriot: yeah that would fetch a new one, I don't think we need to do that but if we wanted to we could try to do a modification in the `update` method for the `userListMutation`. it's just probably not the best thing to modify the object here, but there's not an easy way to do that so I think it's probably ok

Issue #2466, opened 2019-12-06
Learn: update drawer styling


pdpinch:
#### Designs and Mockups

[InVision](https://projects.invisionapp.com/d/main#/console/17477251/392253808/preview)

#### Acceptance Criteria:

- [x] #2253 change course button and logo
- [x] Update font and size of title
- [x] Update text styling of description
- [x] Hide "Read More" link when there is nothing more to read
- [x] remove "Info" header"
- [x] Change styling of metadata section
  - [x] text color
  - [x] two-column layout
  - [x] change "topics" label to "subjects" 
  - [x] remove button styling for topics/subjects

aliceriot:
@pdpinch @Ferdi @abdulkdawson how's this looking? still need to add the platform logo next to 'take course'

![ffnfnfnfnf2](https://user-images.githubusercontent.com/6207644/70645669-67d21900-1c13-11ea-8852-6444ce0de5ba.png)
![ffnfnfnfnf](https://user-images.githubusercontent.com/6207644/70645670-67d21900-1c13-11ea-820e-1f474e9004f8.png)

Ferdi:
In the metadata section the labels are to far from the values. It makes it hard to read imo. I'd bring them as close as possible

Ferdi:
@aliceriot I like better the treatment of "subjects" in the current design /cc @pdpinch @abdulkdawson

aliceriot:
sure, I can do either way as soon as we decide how we want to go

abdulkdawson:
@aliceriot @pdpinch This is looking good so far but I do agree that there is too much spacing between the metadata labels and the values.

aliceriot:
yeah, I was going based on the designs which have a two-column layout for the metadata. I could try out a 1/3 - 2/3 split on the column widths instead of 50 / 50 or something, whatever we want to go with.

do we have a decision on what we want to do with the subjects?

abdulkdawson:
@aliceriot both the metadata labels and values should be left-justified and lined up accordingly as its shown in the design we can just try to reduce the white space between the labels and values. As for the topics or "Subjects", I agree with @Ferdi that the current treatment feels better to me.

aliceriot:
how's this look?

![fjfjfndfasdfasdf](https://user-images.githubusercontent.com/6207644/70828275-12d00780-1db9-11ea-80ed-5f4d3230245a.png)

abdulkdawson:
[...]

PR #2484, opened 2019-12-13
fix display of read more / read less on TruncatedText component

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

one part of #2466 

#### What's this PR do?

Currently when the `TruncatedText` component is used with the `showExpansionControls` prop we always show the 'read more' / 'read less' controls, regardless of whether any content has been hidden by the `Dotdotdot` component. This changes things so that we only show the expansion controls in the case where they can, you know, do something ^_^

In order to do this I had to figure out how to tell when the text was being truncated. Basically, this is being done with a div with `overflow: hidden` on it, and the `Dotdotdot` component doesn't have a way to pass in a callback or something to get out whether or not it's done the truncated.

So what I had to do was get a ref to the div used by `Dotdotdot` and then look at some of it's properties. Basically if the `scrollHeight`, which measures the height of the content, is higher than the `offsetHeight`, which represents the high of the element on screen, then there is some content which has been hidden. So in this case we show the expansion controls, and otherwise we can safely hide them. I had to do a very slight hack to make this work, basically to force our `TruncatedText` component to always do two render calls so that the `Dotdotdot` component has completed it's render. But things seem ok!

make sure that the behavior described above has been implemented and nothing works weirdly!

we use this in the compact post display and in the learning resource description.

basically, if the text is short you shouldn't see any controls in the learning resource description, and if it's long enough to overflow and get truncated you should see a 'read more / read less' control.

showmorefffshort
showmorefff
showmore

### DIFF
--- a/static/js/components/TruncatedText.js
+++ b/static/js/components/TruncatedText.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react"
+import React, { useState, useRef, useLayoutEffect } from "react"
 
 import Dotdotdot from "react-dotdotdot"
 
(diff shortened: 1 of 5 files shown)


static/js/components/TruncatedText.js
```diff
@@ -29,6 +29,31 @@ export default function TruncatedText(props: Props) {
...
+      // the .container on the Dotdotdot holds the ref for the div it uses to
+      // wrap the text
+      const { container } = dotRef.current
+
+      if (container.scrollHeight === 0 && container.offsetHeight === 0) {
+        // the div which holds the text has been added to the DOM but doesn't have
+        // a calculated height yet. in this case we need to use a setState call
+        // to force the component to re-render so we can get the values we need
+        // to figure out when the div has overflow
+        forceReRender(val => val + 1)
```
mbertrand: I think this loop is being called endlessly if the resource's short description is "", for example course MITx+12.340x ("Outdated - Global Warming Science") causes the following exception and the page goes blank:

Screen Shot 2019-12-16 at 12 01 42 PM

aliceriot: ah, that is not good! will fix haha

aliceriot:
@mbertrand ok that should be fixed now

PR #2520, opened 2020-01-02
Update indexing methods to reduce data sent to and from redis via celery 

### LINKED ISSUE #2518: Refactor indexing updates away from sending the entire document through celery
We have a few indexing tasks that have become problematic as we've started to index larger documents: `create_document`, `upsert_document`, and `update_document_with_partial`. These tasks expect a document in the task payload and that document can be memory-heavy, which causes high redis utilization. 

Long term we should fix this across all our indexed document types, but the immediate concern is around video documents.

For each document type we need an indexing task that accepts only an object id and then fetches, serializes, and updates the index with the updated document. We should then replace the current calls in `search/task_helpers.py` with these new tasks so that we stop sending large celery tasks over the wire.

Implement for:

- [ ] Videos
- [ ] Courses
- [ ] Programs
- [ ] User Lists
- [ ] Bootcamps
- [ ] Profiles
- [ ] Posts
- [ ] Comments

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #2518 

#### What's this PR do?
Converts create and update indexing methods to take a primary key value and pass it to the task to index, instead of serializing the data then sending it to the task. This should cut down on the amount of data in each task redis is keeping track of. 

- For videos, run youtube ETL task (backpopulate youtube data)
 - User list: create a user list and verify that 
 - Courses/programs: xpro integration (backpopulate xpro data)

### DIFF
--- a/search/tasks.py
+++ b/search/tasks.py
@@ -17,9 +17,22 @@ from embedly.api import get_embedly_content
 from open_discussions.celery import app
 from open_discussions.utils import merge_strings, chunks, html_to_plain_text
 from search import indexing_api as api
-from search.constants import VALID_OBJECT_TYPES, VIDEO_TYPE
+from search.constants import (
+    BOOTCAMP_TYPE,
+    COURSE_TYPE,
+    PROGRAM_TYPE,
+    VALID_OBJECT_TYPES,
+    VIDEO_TYPE,
+    USER_LIST_TYPE,
+)
 from search.exceptions import RetryException, ReindexException
-from search.serializers import ESVideoSerializer
+from search.serializers import (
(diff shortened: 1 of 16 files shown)


search/tasks.py
```diff
@@ -64,6 +81,74 @@ def create_document(doc_id, data):
     return api.create_document(doc_id, data)
 
 
+@app.task
+def index_new_bootcamp(bootcamp_id):
```
mbertrand: Maybe add a unit test for this one?  Otherwise everything looks great.

noisecapella:
Should be all set now

PR #2526, opened 2020-01-06
share learning resources

### LINKED ISSUE #2435: Share learning resources
As a user, I would like to be able to share learning resources (courses, programs, learning paths, videos, etc.) from search results 

#### Designs and Mockups

TBD

#### Acceptance Criteria:

- [ ] sharing button in resource detail view (i.e. the drawer)
- [ ] unique URL for each detail view
- [x] URL opens the "learn" home page with the drawer open, populated with the correct resource details

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested
- [x] Settings

closes #2435

#### What's this PR do?

this PR sets us up for allowing the user to share learning resources via a link. There are two components:

- on the learning home page we now look for URL params on startup which, if present, are used to open the learning resource drawer
- in the learning resource drawer there is now a "share" button, which opens a little popup (the same one we use for posts and comments) which has the URL for the course home page with the right URL parameters in it.

make sure that different learning resources work right! all learning resources should be supported.

Screenshot from 2020-01-06 13-56-26
Screenshot from 2020-01-06 13-59-56

### DIFF
--- a/static/js/components/ShareTooltip.js
+++ b/static/js/components/ShareTooltip.js
@@ -15,7 +15,8 @@ import { setSnackbarMessage } from "../actions/ui"
 type HelperProps = {
   url: string,
   hideSocialButtons?: boolean,
-  setSnackbarMessage: Function
+  setSnackbarMessage: Function,
+  objectType?: string
 }
 
 export class ShareTooltipHelper extends React.Component<HelperProps> {
@@ -37,11 +38,13 @@ export class ShareTooltipHelper extends React.Component<HelperProps> {
   }
 
   render() {
-    const { url, hideSocialButtons } = this.props
+    const { url, hideSocialButtons, objectType } = this.props
 
     return (
       <div className="tooltip">
-        <div className="tooltip-text">Share a link to this post:</div>
+        <div className="tooltip-text">
+          {`Share a link to this ${objectType || "post"}:`}
+        </div>
         <input ref={this.input} readOnly value={url} />
         <div className="bottom-row">
           {hideSocialButtons ? null : (
@@ -74,13 +77,14 @@ const ConnectedShareTooltipHelper = connect(
 type Props = {
   url: string,
   hideSocialButtons?: boolean,
-  children: Array<React$Element<any>> | React$Element<any>
+  children: Array<React$Element<any>> | React$Element<any>,
+  placement?: string
 }
 
-export default function ShareTooltip({ children, ...props }: Props) {
+export default function ShareTooltip({ children, placement, ...props }: Props) {
   return (
     <Tooltip
-      placement="top"
+      placement={placement || "top"}
       trigger={["click"]}
       overlay={() => <ConnectedShareTooltipHelper {...props} />}
     >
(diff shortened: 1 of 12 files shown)


Ferdi:
@aliceriot can the URL be a bit more user friendly ? Something like  open.mit.edu/learn/lr/1234

aliceriot:
It's a bit more work to get that to work with our existing setup, but if we want to I can do that

static/js/components/ShareTooltip.js
```diff
@@ -37,11 +38,13 @@ export class ShareTooltipHelper extends React.Component<HelperProps> {
...
 
     return (
       <div className="tooltip">
-        <div className="tooltip-text">Share a link to this post:</div>
+        <div className="tooltip-text">
+          {`Share a link to this ${objectType || "post"}:`}
+        </div>
         <input ref={this.input} readOnly value={url} />
         <div className="bottom-row">
           {hideSocialButtons ? null : (
```
mbertrand: `hideSocialButtons` is returning `False` for private user lists / learning paths

static/scss/learning-resource-drawer.scss
```diff
@@ -92,6 +92,23 @@
   }
 }
 
+.elr-share {
+  display: flex;
+  justify-content: flex-end;
```
mbertrand: Add `  cursor: pointer;` ?

static/js/components/ShareTooltip.js
aliceriot: ah, forgot about that option, I'll add a check for it

aliceriot: ah, forgot about that option, I'll add a check for it

aliceriot:
@mbertrand ok I made those changes

aliceriot:
ah good point - @abdulkdawson what do you think, should we show the "share" button at all for private lists?

abdulkdawson:
@aliceriot @pdpinch I think it can be useful for private lists. If a user creates a list but wants to keep it private from the general public can choose who they want to share it without making the list public. Just because the user decides to keep the list private doesn't necessarily mean that they won't share it.

Ferdi:
@aliceriot , sorry, I thought I reply back to you but don't see it here. Anyway. Which part is difficult to do ?  how about open.mit.edu/learn/?lr=1234  ?

aliceriot:
we could do something like `?lr=1234&lrtype=course` if we'd prefer that - because of the way the different LR APIs are set up we need to include a parameter for what type of LR we're dealing with so we know which API to hit

@abdulkdawson as far as sharing goes currently we don't have support (afaik) for granularity in sharing. sharing is either public and anyone has permission to access it or private and it is only accessible to the user. I don't think it would be terribly hard to change that, but if we wanted to allow users to share private lists by URL or by selecting other users to give view permissions to we'd need to make some backend changes to support that. My feeling is that for now we should probably hide the sharing on private lists and add it back in the future if / when we add backend support.

abdulkdawson:
@aliceriot That's fine. If that functionality is not currently supported than our only choice is like you said to hide the share in private and add it back later. I'm totally ok with that.

aliceriot:
@mbertrand I added a check to hide the share button for private userlists

PR #2537, opened 2020-01-13
Improve ES performance by not validating connection for read operations

### LINKED ISSUE #2534: Search: Performance on first request
### Steps to Reproduce

Go to /learn/search and observe that most of the time this first request to the search api is slow. this is more noticeable on RC than local environments

The first request should be as performant as subsequent loads, even if it means we have to back out other recent changes.

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #2534 

#### What's this PR do?
This improves performance by removing extraneous requests to validate a connection. We end up making requests to validate that all of the indexes exist, but actually only assert that at least one exists, which isn't a particularly useful property because it doesn't really guarantee us anything.

I also noticed that `elasticsearch_dsl.connections` internally caches the connections which is the exact same thing `get_conn` is doing. So I moved configuration code into a `configure_connections` which is now called from the `ready()` function of the app. I modified `get_conn` to return the default connection and removed `get_conn` calls where the default behavior would've been to use the default connection anyway. That means only code dealing with the indicies APIs uses `get_conn` now.

- Go to the various search pages and verify they all still work
- You may or may not notice a performance gain locally between this branch and master 
- Verify reindexing works

### DIFF
--- a/search/connection.py
+++ b/search/connection.py
@@ -9,55 +9,36 @@ from elasticsearch_dsl.connections import connections
 
 from search.constants import VALID_OBJECT_TYPES
 
-_CONN = None
-# When we create the connection, check to make sure all appropriate mappings exist
-_CONN_VERIFIED = False
 
-
-def get_conn(*, verify=True):
+def configure_connections():
     """
-    Lazily create the connection.
+    Create connections for the application
 
-    Args:
-        verify (bool): If true, check the presence of indices and mappings
+    This should only be called once
+    """
+    # this is the default connection
+    http_auth = settings.ELASTICSEARCH_HTTP_AUTH
+    use_ssl = http_auth is not None
+    # configure() lazily creates connections when get_connection() is called
+    connections.configure(
+        default={
+            "hosts": [settings.ELASTICSEARCH_URL],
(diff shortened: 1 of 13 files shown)


search/connection.py
```diff
@@ -2,62 +2,50 @@
...
+                "http_auth": http_auth,
+                "use_ssl": use_ssl,
+                # make sure we verify SSL certificates (off by default)
+                "verify_certs": use_ssl,
+            }
         )
-        # Verify connection on first connect if verify=True.
-        do_verify = verify
+    else:
+        log.error("ELASTICSEARCH_URL is not set")
```
noisecapella: Should we raise an exception here? This setting should always be true.

noisecapella:
Functionality looks good, just the one question

search/connection.py
rhysyngsun: I removed the condition here and instead put a check for the setting in `settings.py` so the app will fail to deploy and heroku will roll the app back.

rhysyngsun:
@noisecapella all set

Issue #2545, opened 2020-01-16
Bug: search "MIT Sports Entrepreneurship Bootcamp" without quotes


abeglova:
### Steps to Reproduce
If you search for "MIT Sports Entrepreneurship Bootcamp" without quotes 
the MIT Sports Entrepreneurship Bootcamp does not come up on the first page. 
https://discussions-rc.odl.mit.edu/learn/search?q=MIT%20Sports%20Entrepreneurship%20Bootcamp. And possibly does not show up at all

### Expected Behavior

Searching for "MIT Sports Entrepreneurship Bootcamp" without quotes should return the MIT Sports Entrepreneurship Bootcamp as the first or second result 

### Issue
Bootcamps don't show up in the search because elasticsearch normalizes document scores individually for each index by default and the results for bootcamps are off because there are not a lot of items in the index.

Setting the search type to "dfs_query_then_fetch" fixes the issue
Elasticsearch documentation: https://www.elastic.co/blog/understanding-query-then-fetch-vs-dfs-query-then-fetch

odlbot:
➤ Peter Pinch commented:

duplicate of https://trello.com/c/iR3IbFwz/31-if-you-search-for-mit-sports-entrepreneurship-bootcamp-as-text-the-mit-sports-entrepreneurship-bootcamp-does-not-come-up-on-the

PR #2558, opened 2020-01-22
improve elasticsearch scoring between indexes

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/open-discussions/issues/2545

#### What's this PR do?
Currently bootcamps don't show up in the search because elasticsearch normalizes document scores individually for each index by default and the results for bootcamps are off because there are not a lot of items in the index.

This PR sets a setting for elasticsearch to normalize document scores accross indexes.
Elasticsearch documentation: https://www.elastic.co/blog/understanding-query-then-fetch-vs-dfs-query-then-fetch

Search for "MIT Sports Entrepreneurship Bootcamp" without quotes. Ensure that the Sports Entrepreneurship Bootcamp is the first result. Also this and other searches should not be slower than usual.

### DIFF
--- a/search/api.py
+++ b/search/api.py
@@ -31,6 +31,7 @@ from search.constants import (
 
 RELATED_POST_RELEVANT_FIELDS = ["plain_text", "post_title", "author_id", "channel_name"]
 SIMILAR_RESOURCE_RELEVANT_FIELDS = ["title", "short_description"]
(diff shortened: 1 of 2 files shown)


pdpinch:
@abeglova Do you think we should close this PR?

Issue #2494, opened 2019-12-18
Learn: set the default sort when there is no search text


pdpinch:
As a user, I'd like to see the most representative learning objects at the top of the page, before I enter a search term. In particular, I'd like to see free or no cost items that are available now.

#### Acceptance Criteria:

- [ ] sort first by price ascending, then availability. The first page of results should be all free courses available now. 

#### Out of Scope

- we should consider allowing the user to change the sort
- it would be nice if the sort could reflect the users profile and/or search history

abdulkdawson:
@pdpinch Here is my recommendation for the default sort in Open. Take a look and let me know what you think.


New Users and Users not signed in:

The default sort should have the most popular free courses and programs displayed as soon as the user lands on the Open search page. These results will reflect the Learning Resources that have been clicked, searched and saved by site-wide users.


New Signed in: 

The default sort should have the most popular free courses and programs displayed as soon as the user gets to the Open search page. Since this is the first time the user is visiting the site they will begin to create a history that will be reflected in the default results upon their return visit to Open. 


Returning Users Signed in: 

By default, the user will see courses related to their previous search history along with other popular courses and programs. 
As the user continues to search and click on various learning resources they will add to their user history and altering the sorting of Learning Resources. 


All users will also have the ability to sort those results by using the “Sort By” dropdown. The “Sort By” dropdown will allow the user to sort the results by popularity, alphabetically A-Z, alphabetically Z-A and Newest.

pdpinch:
After discussing in Slack, we decided to opt for defaulting to sorting by price ascending and then availability (now first). 

I've updated the description.

abeglova:
After a discussion with Peter, we decide to opt for sorting by price ascending then the date the resource was added to the search

PR #2570, opened 2020-01-30
Additional index updates for default search

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes https://github.com/mitodl/open-discussions/issues/2494

#### What's this PR do?
I am breaking up the PR for the sorting order in the default search page into two PRs so there is no downtime when it is released. This PR adds `minimum_price` and `default_search_priority` to the elasticsearch index. https://github.com/mitodl/open-discussions/pull/2571
updates the search query.

Recreate your elasticsearch index. The script should not break and the course search UI should still look and work the same as before after it runs

### DIFF
--- a/search/indexing_api.py
+++ b/search/indexing_api.py
@@ -110,6 +110,9 @@ LEARNING_RESOURCE_TYPE = {
     "image_src": {"type": "keyword"},
     "topics": {"type": "keyword"},
     "offered_by": {"type": "keyword"},
+    "created": {"type": "date"},
+    "default_search_priority": {"type": "integer"},
+    "minimum_price": {"type": "scaled_float", "scaling_factor": 100},
     "runs": {
         "type": "nested",
         "properties": {
@@ -130,12 +133,16 @@ LEARNING_RESOURCE_TYPE = {
             "instructors": {"type": "text"},
             "prices": {
                 "type": "nested",
-                "properties": {"mode": {"type": "text"}, "price": {"type": "float"}},
+                "properties": {
+                    "mode": {"type": "text"},
+                    "price": {"type": "scaled_float", "scaling_factor": 100},
+                },
             },
             "image_src": {"type": "keyword"},
             "published": {"type": "boolean"},
             "availability": {"type": "keyword"},
             "offered_by": {"type": "keyword"},
+            "created": {"type": "date"},
         },
     },
 }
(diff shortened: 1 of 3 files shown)


noisecapella:
Can you edit the PR description to say `Fixes #2494` instead of just `#2494`, just so github knows to close the related issue?

search/indexing_api.py
```diff
@@ -109,6 +109,9 @@
     "image_src": {"type": "keyword"},
     "topics": {"type": "keyword"},
     "offered_by": {"type": "keyword"},
+    "created_on": {"type": "date"},
+    "default_search_priority": {"type": "integer"},
+    "minimum_price": {"type": "float"},
```
noisecapella: Can we use a `scaled_float` here instead? It would help avoid rounding errors. https://stackoverflow.com/questions/30767089/how-to-store-money-in-elasticsearch

search/indexing_api.py
```diff
@@ -109,6 +109,9 @@
     "image_src": {"type": "keyword"},
     "topics": {"type": "keyword"},
     "offered_by": {"type": "keyword"},
+    "created_on": {"type": "date"},
```
noisecapella: It looks like we are naming a similar field `created` on `CONTENT_OBJECT_TYPE` above. Can we rename this to match that field for consistency?

search/serializers.py
```diff
@@ -300,6 +302,27 @@ class LearningResourceSerializer(serializers.ModelSerializer):
...
+        if hasattr(instance, "runs") and instance.runs:
+            prices = [
+                run.prices.values_list("price", flat=True)
+                for run in instance.runs.all()
+            ]
+
+            if prices:
+                prices = reduce(lambda x, y: x | y, prices)
+
+                return float(min(prices, default=0))
```
noisecapella: Is it possible to avoid using `float` for prices here? If we return a `Decimal` here I think it will be passed along as a string, which would avoid floating point precision problems. But I'm not sure how that interacts with the `scaled_float` type I suggested above.

search/indexing_api.py
abeglova: Done. I updated the price in the runs serializer to use `scaled_float` as well so the two have the same format

search/serializers.py
abeglova: I was getting `TypeError: Object of type 'Decimal' is not JSON serializable` with an unconverted decimal. I updated the code to explicitly convert to a string here

abeglova:
Thanks @noisecapella! Back to you

search/indexing_api.py
```diff
@@ -109,6 +109,9 @@
     "image_src": {"type": "keyword"},
     "topics": {"type": "keyword"},
     "offered_by": {"type": "keyword"},
+    "created": {"type": "date"},
+    "default_search_priority": {"type": "integer"},
+    "minimum_price": {"type": "scaled_float", "scaling_factor": 10},
```
noisecapella: I think the `scaling_factor` needs to be `100` for prices based on [this page](https://www.elastic.co/guide/en/elasticsearch/reference/master/number.html) in the docs. Can you update here and below?

abeglova: Done!

PR #2571, opened 2020-01-30
Sort default search results

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes https://github.com/mitodl/open-discussions/issues/2494

#### What's this PR do?
Sorts results in the search page by price, then resource type, then date added when there is not text search set. 

Courses and programs should come before videos and user lists in the search when they have the same cost

Recreate your elasticsearch index
Verify that search results with no text are sorted by price, then resource type , then created on
Verify that search results with text filters are not changed

### DIFF
--- a/static/js/lib/search.js
+++ b/static/js/lib/search.js
@@ -505,6 +505,14 @@ export const buildSuggestQuery = (text: string) => {
   return suggest
 }
 
+export const buildDefaultSort = () => {
+  return [
+    { minimum_price: { order: "asc" } },
+    { default_search_priority: { order: "desc" } },
+    { created: { order: "desc" } }
+  ]
+}
+
 export const buildOrQuery = (
   builder: any,
   searchType: string,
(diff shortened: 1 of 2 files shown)


abeglova:
@rhysyngsun I discussed this with Peter and Ferdi and we decided to not include the sort if there is any filter

static/js/lib/search.js
```diff
@@ -620,8 +620,8 @@ export const buildLearnQuery = (
     if (!emptyOrNil(text)) {
       // $FlowFixMe: if we get this far, text is not null
       builder = builder.rawOption("suggest", buildSuggestQuery(text))
-    } else {
-      builder.rawOption("sort", buildDefaultSort())
+    } else if (facetClauses.length === 0 && types === LR_TYPE_ALL) {
```
rhysyngsun: Strictly equals in JS can fail for lists even if the elements match if the two references are different objects, `R.equals` is safer: https://ramdajs.com/docs/#equals

PR #2576, opened 2020-02-05
Refactor index_items

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
Refactors indexing code to allow calling code to construct the iterable of documents instead of passing in a serializer function

Run recreate_index and verify that no errors happened and the results look fine

### DIFF
--- a/search/indexing_api.py
+++ b/search/indexing_api.py
@@ -432,20 +432,19 @@ def update_post(doc_id, post):
     )
 
 
-def index_items(serialize_bulk_items, object_type, ids):
+def index_items(documents, object_type):
     """
     Index items based on list of item ids
 
     Args:
-        serialize_bulk_items (callable): the function to serializer a list of objects by id
+        documents (iterable of dict): An iterable with ElasticSearch documents to index
         object_type (str): the ES object type
-        ids(list of int): List of item id's
     """
     conn = get_conn()
     for alias in get_active_aliases(conn, [object_type]):
         _, errors = bulk(
             conn,
-            serialize_bulk_items(ids),
+            documents,
             index=alias,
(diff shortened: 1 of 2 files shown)


search/indexing_api.py
```diff
@@ -432,20 +432,19 @@ def update_post(doc_id, post):
...
+        documents (iterable of dict): An iterable with ElasticSearch documents to index
         object_type (str): the ES object type
-        ids(list of int): List of item id's
     """
     conn = get_conn()
     for alias in get_active_aliases(conn, [object_type]):
         _, errors = bulk(
             conn,
-            serialize_bulk_items(ids),
+            documents,
```
noisecapella: Because this is a generator we won't be able to read from it twice, and there could be two or more `alias`es in the for loop... I'll see what we can do to work around that

search/indexing_api.py
```diff
@@ -432,27 +433,32 @@ def update_post(doc_id, post):
...
-            index=alias,
-            doc_type=GLOBAL_DOC_TYPE,
-            # Adjust chunk size from 500 depending on environment variable
-            chunk_size=settings.ELASTICSEARCH_INDEXING_CHUNK_SIZE,
-        )
-        if len(errors) > 0:
-            raise ReindexException(f"Error during bulk {object_type} insert: {errors}")
+    # bulk will also break an iterable into chunks. However we should do this here so that
+    # we can use the same documents when indexing to multiple aliases.
+    for chunk in chunks(
```
noisecapella: @mbertrand I added this for loop to break the stream into chunks so we can use the same chunk for each alias in the inner for loop below.

mbertrand:
Just fix tests

Issue #2219, opened 2019-09-16
Courses: investigate duplicate MITx courses


pdpinch:
### Steps to Reproduce

1. Navigate to https://discussions-rc.odl.mit.edu/courses/search?q=%22just%20money%22

### Expected Behavior

- There should be only one version of "Just Money: Banking as if Society Mattered" that is available now. 
- The llnk to "Take Course on edX" should take the user to a marketing page

### Actual Behavior

- There are two courses
- The link on the second one goes directly to a course page

rhysyngsun:
Regarding this,  the duplicate "Just Money" course, one has the course key MITx+11.405x_2, the other is MITx+11.405x, the former appears to only have 1 run from Summer 2017, the latter has summer runs from 2016-2019, excluding 2017.

This means it's actually two different logical courses on the mitx side. We could merge this, but we'd need to make a decision on which course is the authoritative one.

Ferdi:
I don't think we should show/index courses that have no runs in the future.

rhysyngsun:
@Ferdi last time we discussed this I think the consensus was that we didn't want to do that because it can lead to other undesirable behavior (e.g. courses that haven't had a future run setup disappearing as soon as the current run ends). That's reflected in the course run selection logic we specified in #2124.

I think our best options here are to either build some mechanism where we can logically merge these two runs on our end, or a mechanism to hide courses and/or runs that are historical anomalies.

Ferdi:
@rhysyngsun , yes you are right, that's what we said.

In the "just money" course example above, one of them links to the marketing page of the course and the other one to the courseware url.  Do you know why ?

rhysyngsun:
@Ferdi import code prefers marketing urls over courseware urls if they're present, but the course that does not link to the marketing url has its marketing url set to `https://courses.edx.org//courses/course-v1:MITx+11.405x_2+1T2017/course/` in the data coming back from edx.

pdpinch:
That marketing URL is definitely a sign of a broken course. I can try to get MITx/edX to clean it up, or maybe we could set up a regex pattern for marketing URLs and we ignore anything that doesn't match it?

rhysyngsun:
I'm not sure that cleanup is practical, the only course run in `MITx+11.405x` that has an actual marketing url is `2T2019`. That still wouldn't solve the problem for `11.405x_2` since that course would end up getting the courseware url anyway since that's our fallback.

odlbot:
[...]

PR #2577, opened 2020-02-06
dedupe mitx courses

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
- [x] Settings

https://github.com/mitodl/open-discussions/issues/2219

#### What's this PR do?
This pr merges course runs from duplicate courses. The course with the most relevant run is displayed. The most relevant run is
1) A current run, if available
2) If there is no current run, the upcoming run that starts the soonest
3) If there is no current or upcoming  run, the archived run that started the most recently

Set 
DUPLICATE_COURSES_URL=https://gist.githubusercontent.com/abeglova/5dfb0717c6c54edc6e0f5e0d5287e620/raw/bc77456780bd66ee98bcf4fa1653e6ef03bcdccc/gistfile1.yml
in your .env file

run 
```
get_mitx_data()
```
There should be only one MITx course with the following title and the search page should link to a run with the following start date:
"Discrete-Time Signal Processing" - Aug 20, 2016
"Foundations of Development Policy" - Feb 4, 2020
"Global Warming Science" - Feb 19, 2020 - also there should be no course with the title "Outdated - Global Warming Science"
"A Global History of Architecture"  - Dec 19, 2019
"The Challenges of Global Poverty"  -  Feb 4, 2020
"Just Money: Banking as if Society Mattered" - Nov 26, 2019

### DIFF
--- a/course_catalog/etl/deduplication.py
+++ b/course_catalog/etl/deduplication.py
@@ -1,0 +1,41 @@
+"""Functions to combine duplicate courses"""
+from course_catalog.constants import AvailabilityType
+
+
+def get_most_relevant_run(runs):
+    """
+    Helper function to determine the most relevant course run.
+
+    Args:
+        runs (QuerySet): a set of LearningResourseRun objects
+    Returns:
+        A LearningResourseRun object
+    """
+
+    # if there is a current run in the set pick it
+    most_relevant_run = next(
+        (run for run in runs if run.availability == AvailabilityType.current.value),
+        None,
+    )
+
+    if not most_relevant_run:
+        # if there a future runs in the set, pick the one with earliest start date
+        runs = runs.order_by("best_start_date")
+        most_relevant_run = next(
+            (
+                run
+                for run in runs
+                if run.availability
+                in [
+                    AvailabilityType.upcoming.value,
+                    AvailabilityType.starting_soon.value,
+                ]
+            ),
+            None,
+        )
+
+        if not most_relevant_run:
+            # get latest past run by start date
+            most_relevant_run = next((run for run in runs.reverse()))
+
+    return most_relevant_run
--- a/course_catalog/etl/loaders.py
+++ b/course_catalog/etl/loaders.py
@@ -27,7 +27,8 @@ from course_catalog.models import (
     UserListItem,
     ContentFile,
 )
-from course_catalog.utils import load_course_blacklist
+from course_catalog.utils import load_course_blacklist, load_course_duplicates
+from course_catalog.etl.deduplication import get_most_relevant_run
 from search import task_helpers as search_task_helpers
 from search.constants import COURSE_TYPE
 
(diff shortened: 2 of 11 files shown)


course_catalog/tasks.py
```diff
@@ -28,6 +28,7 @@
 def get_mitx_data():
     """Task to sync mitx data with the database"""
     pipelines.mitx_etl()
+    mitx.dedupe_courses()
```
rhysyngsun: I think deduplicating the courses at the end of everything could be problematic, a few issues I could see arising from this:

- Courses will reappear in search results for a few minutes until `mitx_etl()` returns and we start deduplicating
- Any failure along the way could cause this task to exit and leave us in a partial state (e.g. multiple courses showing up again)

I think we need to move the deduplication to an earlier stage

course_catalog/etl/mitx.py
```diff
@@ -25,6 +34,72 @@ def _is_mit_course(course):
...
+        )
+
+        if not most_relavant_run:
+            # get latest past run by start date
+            most_relavant_run = next((run for run in runs.reverse()))
+
+    return most_relavant_run
+
+
+def dedupe_courses():
```
rhysyngsun: Is this MITx-specific logic or can we generalize it to other platforms? I'm specifically thinking about how we occasionally archive MITx courses off to OLL, so we could start seeing them show up there as well.

course_catalog/utils.py
```diff
@@ -176,3 +176,17 @@ def load_course_blacklist():
...
+    """
+    Get a list of blacklisted course ids
+    Returns:
+        list of lists of courses which are duplicates of eachother
+    """
+    duplicates_url = settings.DUPLICATE_COURSES_URL
+    if duplicates_url is not None:
+        response = requests.get(duplicates_url)
+        response.raise_for_status()
+        return [str(line, "utf-8").split(",") for line in response.iter_lines()]
```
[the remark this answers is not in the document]
abeglova: I removed "utf-8"

course_catalog/tasks.py
abeglova: I updated the pr to not update any of the duplicate courses in elasticsearch until the deduplication step happens. This prevents duplicate courses from appearing while the tasks runs. If the task fails  before the dedupe step, courses with duplicates will be not updated in the search but no duplicates will appear

abeglova:
Back to you @noisecapella. You should set 
DUPLICATE_COURSES_URL=https://gist.githubusercontent.com/abeglova/5dfb0717c6c54edc6e0f5e0d5287e620/raw/bc77456780bd66ee98bcf4fa1653e6ef03bcdccc/gistfile1.yml
to get the new duplicate list format

course_catalog/etl/deduplication.py
```diff
@@ -0,0 +1,72 @@
...
+    Helper function to determine the most relavant course run.
+
+    Args:
+        runs (QuerySet): a set of LearningResourseRun objects
+    Returns:
+        A LearningResourseRun object
+    """
+
+    # if there is a current run in the set pick it
+    most_relavant_run = next(
```
rhysyngsun: Spelling: `most_relevant_run`

course_catalog/utils.py
```diff
@@ -176,3 +177,23 @@ def load_course_blacklist():
...
     return []
+
+
+def load_course_duplicates(platform):
+    """
+    Get a list of blacklisted course ids for a platform
+    Args:
+        platform (string): the platform for which course duplicates are needed
+    Returns:
+        list of lists of courses which are duplicates of eachother
```
rhysyngsun:
```suggestion
        list of lists of courses which are duplicates of each other
```

course_catalog/etl/loaders.py
```diff
@@ -209,7 +223,26 @@ def load_program(program_data, blacklist):
...
     blacklist = load_course_blacklist()
-    return [load_program(program_data, blacklist) for program_data in programs_data]
+
+    if (
+        programs_data
+        and len(programs_data) > 0
+        and programs_data[0].get("courses")
+        and len(programs_data[0].get("courses")) > 0
+    ):
+        platform = programs_data[0].get("courses")[0].get("platform")
```
rhysyngsun: I think you can make this more straightforward by having `platform` be passed as a keyword argument to a [curried version](https://toolz.readthedocs.io/en/latest/api.html#toolz.functoolz.curry) of `load_programs`, which would be paramterized in `pipelines.py`.

course_catalog/etl/deduplication.py
```diff
@@ -0,0 +1,113 @@
...
+        runs = runs.order_by("created_on")
+        first_run = next(run for run in runs)
+        runs.update(object_id=first_run.object_id)
+        most_relevant_run = get_most_relevant_run(runs)
+
+        copy_course_data(
+            courses.get(pk=first_run.object_id),
+            courses.get(pk=most_relevant_run.object_id),
+            courses,
+        )
```
rhysyngsun: It may have slipped my mind since it's been almost a week, but could you give some context around why this copying of fields was added in your latest changes? My gut says if we're having to do something like this we maybe need to back up and look at the design here again.

course_catalog/utils.py
```diff
@@ -176,3 +177,23 @@ def load_course_blacklist():
         response.raise_for_status()
         return [str(line, "utf-8") for line in response.iter_lines()]
     return []
+
+
+def load_course_duplicates(platform):
```
rhysyngsun: This should probably be in `deduplication.py` instead since it doesn't load anything into the database.

course_catalog/etl/deduplication.py
abeglova: We were concerned that sharable links would no longer work if the visible copy for a duplicate course switched over to a newer copy. This keeps the original course but overwrites course content with the data corresponding to the most relevant run.

As we discussed last week, we don't know which copy of a course has the latest data until all the copies are loaded

course_catalog/utils.py
abeglova: I think it makes sense for this to live in the same place as `load_course_blacklist` but moving it to `deduplication.py` would be fine too

PR #2597, opened 2020-02-26
Fix blank thumbnails for OCW courses

mbertrand:
#### Pre-Flight checklist
- [ ] Testing
  - [ ] Code is tested

N/A?

#### What's this PR do?
Runs `OCWParser.upload_all_media_to_s3` before `digest_ocw_course`, because the parser updates the course JSON while doing so (including thumbnail URL).

Run `manage.py backpopulate_ocw_data --overwrite` and let it run for a little bit until at least some OCW courses are (re-)imported.  Check that the imported courses have a non-blank thumbnail.

### DIFF
--- a/course_catalog/api.py
+++ b/course_catalog/api.py
@@ -407,15 +407,6 @@ def sync_ocw_course(
     if course_json["course_id"] in blacklist:
         is_published = False
 
-    log.info("Digesting %s...", course_prefix)
-    try:
-        course, run = digest_ocw_course(
-            course_json, last_modified, is_published, course_prefix
-        )
-    except TypeError:
-        log.info("Course and run not returned, skipping")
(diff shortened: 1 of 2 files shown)


course_catalog/api.py
```diff
@@ -432,13 +423,19 @@ def sync_ocw_course(
                 parser.upload_all_media_to_s3(upload_master_json=True)
         except:  # pylint: disable=bare-except
             log.exception(
-                (
-                    "Error encountered uploading OCW files for %s, deleting run",
-                    course_prefix,
-                )
+                ("Error encountered uploading OCW files for %s", course_prefix)
             )
-            run.delete()
```
abeglova: Is this change deliberate?

PR #2601, opened 2020-02-28
Refresh requirements.txt

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
Fixes a rebase screwup (ocw-data-parser version should be 0.3.0)

All tests should pass


noisecapella:
It looks like defusedxml is downgraded, which is probably my fault from updating requirements.txt in #2594. If the tests pass it's not an issue, but I think it will break the tests because of a deprecation warning. I'll see if this can be easily fixable in requirements.in

mbertrand:
@noisecapella using `python3-saml==1.9.0` prevents the downgrade, I'll see if the tests pass with that.

noisecapella:
Sounds good. I looked at the changelog for python3-saml and nothing jumped out at me as something which would break with an upgrade, though I'm not familiar with saml.

PR #2603, opened 2020-03-02
Initial proposal for reddit migration work

rhysyngsun:
[Rendered](https://github.com/mitodl/open-discussions/blob/nl/reddit-migration-docs/docs/reddit-migration/readme.md)

N/A

#### What's this PR do?
This PR is a set of markdown docs of my recommendation for how to migrate away from reddit

### DIFF
--- a/docs/reddit-migration/considerations/sorting.md
+++ b/docs/reddit-migration/considerations/sorting.md
@@ -1,0 +1,34 @@
+### Sorting
+
+There are several areas where reddit sorts objects coming back from their APIs in a way that isn't necessarily trivial via standard django ORM methods. This list is meant to be comprehensive so we know the problem space:
+
+- `Hot` (named "Active" in our UI)
+  - A logarithmic sort, where as an item ages, it requires an order of magnitude more `upvotes - downvotes` to be sorted above a newer item, the interval for the order of magnitude increase is 12.5 hours
+  - Implementation: https://github.com/mitodl/reddit/blob/stable/r2/r2/lib/db/_sorts.pyx#L47
+- `Confidence` (named "Best" in our UI)
+  - Items are sorted with a [weighted algorithm](https://www.evanmiller.org/how-not-to-sort-by-average-rating.html)
+  - See the corresponding code in the reddit source:
+    - https://github.com/mitodl/reddit/blob/stable/r2/r2/lib/db/_sorts.pyx#L70
+    - https://github.com/mitodl/reddit/blob/stable/install/setup_postgres.sh#L42
+- `New` - posts are sorted in reverse chronological order
+- `old` - posts are sorted in chronological order
+
+
+#### Posts
+
+Posts are sorted in two places: channels and the frontpage. These both support seemingly similar sorts but the implementations actually differ.
+
+##### Posts in Channels
+
+Post sorting in channels is the more straightforward version of post sorting, there are 3 kinds we use: `Hot`, `New`, `Top`
+
+##### Posts in Frontpage
+
+ - `Hot` - posts are sorted using a normalized version of the hot algorithm. This is done to ensure that content from less active channels isn't overshadowed by content from more popular channels.
+   - See the corresponding code in the reddit source:
+     - https://github.com/mitodl/reddit/blob/stable/r2/r2/lib/normalized_hot.py#L38-L79
+ - `New` and `Top` - same method as channels
+
+ #### Comments
+
+ Comments only have a few sorting options we're currently using: `New`, `Old`, and `Confidence`
--- a/docs/reddit-migration/strategy/reads-and-writes.md
+++ b/docs/reddit-migration/strategy/reads-and-writes.md
@@ -1,0 +1,36 @@
+# Migrate Reads and Writes
+
+We need to gracefully migrate our APIs over to the new schema. We want to avoid or at the very least minimize any sort of downtime.
+
+In general, migrating reads and writes can happen in a siloed manner, in order of highest to lowest level in the relationship hierarchies:
+
+- Channels
+- Posts
+- Comments
+- etc...
+
+### Add writes to the new tables
+
+This should be fairly straightforward, we should add another write to where we are currently writing (calling APIs) to reddit. Most/all of these should be in `channels/api.py` and we should include the writes in any of the existing atomic transactions in that file. There are some cases where we haven't added transactions because they provide no benefit. Those tend to be when we're creating a piece of data (post or comment) and the source of truth is reddit. In those cases a transaction rollback is of no value because we can't rollback a reddit api call.
+
+### Backpopulate historical data into new tables
+
+This can be done with backpopulate commands and celery tasks. We've done this plenty of times so it shouldn't be too difficult.
+
+### Verify new data
+
+We want some kind of verification of the new data vs the old data. Probably a command that runs through all of the data and verifies records exist and are accurate.
+
+### Move reads to the new tables
+
+- Move APIs, celery tasks, and any other reads to the new data models. After this step there should be no references in application code to the old models.
+- Migrate any BI queries to the new tables
+  - If we silo our changes on object type, this might have to wait until everything is moved over
+
+### Backup reddit data
+
+It can't hurt and is probably prudent to take a backup of our reddit data prior to cutting off writes.
+
+### Remove writes from the old schema
+
+This means removing writes to the models defined in the `channels` app and the writes to reddit itself. After this step we should see no traffic to reddit and no further writes to the tables. This can be tested by temporarily stopping the reddit API instances and verifying there are no errors when exercising the app.
--- a/docs/reddit-migration/strategy/schema.md
+++ b/docs/reddit-migration/strategy/schema.md
@@ -1,0 +1,211 @@
+# Establish a New Schema
+
+We want to establish a new schema so that we have a clean slate to work with and because there will be some incompatible changes with the current schema. An example of this is the `Post.post_id` column is a reference to the reddit id, but in the future needs to be the primary key of the row (see [Reddit Ids](reddit-ids.md)). We can't do that in-place with the current table.
+### Channels
+
+Channels should have the fields carried over with the following adjustments:
+
+ - `channel_id` -> `id`
+
+### Posts
+
+Posts should have the fields carried over with the following adjustments:
+
+ - `post_id` -> `id`
+ - `reports_ignored = BooleanField(default=False)`
+   - Flag to indicate whether we're ignoring subsequent reports of the post
+   - See "Moderation" for more details
+ - `text` / `preview_text` / `plain_text`
+   - We currently have several representations of textual post content, the 3 fields/properties listed above:
+     - `text` is the markdown text for text posts
+     - `preview_text` is precomputed plain text representation of link posts, sourced from and Embedly API request for that link
+     `plain_text` is a `@property`, not a db column, but its logic is as follows:
+       - If the post is an article, render the article to plain text
+       - If the post is a text post, render the markdown to plain text
+       - Otherwise use `preview_text`
+    - See also: https://github.com/mitodl/open-discussions/blob/9a96f09ab4c8c617da3ba8842adb9c419c5d321d/docs/rfcs/0002-indexing-embedly-link-preview.md#4-related-but-separate-issue-getting-rid-of-either-the-plain_text-or-text-field-in-es-documents
+    - We should to normalize `preview_text` / `plain_text` down to a single stored field so we're not recomputing it every time a read occurs.
+    - We still need to keep the `text` field and the `article` relation
+
+
+### Comments
+
+Comments should have the fields carried over with the following adjustments:
+
+ - `channel_id` -> `id`
+ - `reports_ignored = BooleanField(default=False)`
+   - Flag to indicate whether we're ignoring subsequent reports of the comment
+   - See "Moderation" for more details
+
+#### Comment Tree
+
+Prior art:
+
+ - SQL Antipatterns ("Naive Trees"):http://www.r-5.org/files/books/computers/languages/sql/style/Bill_Karwin-SQL_Antipatterns-EN.pdf
+ - Models for Hierarchical Data: https://www.slideshare.net/billkarwin/models-for-hierarchical-data
+ - Django treebeard: https://django-treebeard.readthedocs.io/en/latest/
+
+Reddit stores comment trees in Cassandra, which is overkill for our current use cases. That said, we need to be able to migrate the comments in such a manner that we achieve the following:
+
+- Maintain current API result structures
+  - Specifically, we need to support some form of reddit's "more comments" pagination. This **does not** need to partition the comments exactly as reddit would, the interface just needs to be compatible with how the frontend traverses it.
+- Support currently used sorting methods (best, old, new)
+- Support rendering an arbitrary subtree of the comments for the comment detail view
+
+From the Prior Art listed above, Materialized Paths are probably our best bet and we're already familiar to some degree with them because they're what `wagtail` uses under the hood for page hierarchies.
+
+The big thing here is that in order to support the sorts we need one table/model for *each type of sort*. That precludes using a single table for everything but it simplifies reads because we can pick our sorting data model and then `prefetch_related` on the `Comment` table so that code can work generically over similarly structured tables. This approach means these sorting hierarchy tables end up being a sorted lookup table for extracting a portion of the comment tree.
+
+An example of what the data model would look like (working branch [here](https://github.com/mitodl/open-discussions/tree/nl/reddit_schema_migration)):
+
+```python
+from django.db import models
+from treebeard.mp_tree import MP_Node
+
+class Post(models.Model):
+  pass
+
+
+class Comment(models.Model):
+    post = models.ForeignKey(Post, on_delete=models.CASCADE)
+
+
+class CommentTreeNode(MP_Node):
+    # specialize the alphabet for postgres
+    # see: https://django-treebeard.readthedocs.io/en/latest/mp_tree.html#treebeard.mp_tree.MP_Node.alphabet
+    alphabet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+
+    # a post can have many comments, but a reverse lookup into this tree isn't pragmatic
+    post = models.ForeignKey(
+        Post, on_delete=models.CASCADE, related_name="+"
+    )
+
+    class Meta:
+        abstract = True
+
+
+class NewCommentTreeNode(CommentTreeNode):
+    comment = models.OneToOneField(
+        Comment,
+        on_delete=models.CASCADE,
+        unique=True,
+        null=True,
+        related_name="new_comment_node",
+    )
+
+    created = models.DateTimeField()
+
+    node_order_by = ["-created"]
+
+
+class OldCommentTreeNode(CommentTreeNode):
+    comment = models.OneToOneField(
+        Comment,
+        on_delete=models.CASCADE,
+        unique=True,
+        null=True,
+        related_name="old_comment_node",
+    )
+
+    created = models.DateTimeField()
+
+    node_order_by = ["created"]
+
+
+class BestCommentTreeNode(CommentTreeNode):
+    comment = models.OneToOneField(
+        Comment,
+        on_delete=models.CASCADE,
+        unique=True,
+        null=True,
+        related_name="best_comment_node",
+    )
+
+    score = models.IntegerField()
+
+    node_order_by = ["-score"]
+```
+
+When a sorted column changes, it requires a move of the node. For the Old/New sorts those will never change. For the Best sort, it will change whenever a vote is made or updated. Fortunately, Materialized Paths are pretty efficient at this. An example with the above data models is:
+
+```python
+from django.db import transaction
+
+post = Post.objects.create()
+comment1 = Comment.objects.create(post=post)
+comment2 = Comment.objects.create(post=post)
+comment3 = Comment.objects.create(post=post)
+
+@transaction.atomic
+def rescore(node, new_score):
+    node.score = score
+    node.save()
+    node.move(node)
+
+# comment trees start with a root node w/ `comment=None`
+# this would get created when the post is created
+root = BestCommentTreeNode.add_root(post=post, score=0)
+# top level comment
+c1n = root.add_child(comment=comment1, score=1)
+# replies to comment1
+c2n = c1n.add_child(comment=comment1, score=4)
+c3n = c1n.add_child(comment=comment1, score=5)
+
+# c3n should list above c2n
+print(root.dump_bulk())
+
+rescore(c2n, 8)
+# c2n is now sorted above c3n and has the updated score
+print(root.dump_bulk())
+```
+
+This results 4 queries, two `SELECT ... LIMIT 1` and two `UPDATE ...` queries:
+
+```sql
+UPDATE "discussions_bestcommenttreenode" SET "path" = '000100010002', "depth" = 3, "numchild" = 0, "post_id" = 2, "comment_id" = 3, "score" = 8 WHERE "discussions_bestcommenttreenode"."id" = 5;
+SELECT "discussions_bestcommenttreenode"."id", "discussions_bestcommenttreenode"."path", "discussions_bestcommenttreenode"."depth", "discussions_bestcommenttreenode"."numchild", "discussions_bestcommenttreenode"."post_id", "discussions_bestcommenttreenode"."comment_id", "discussions_bestcommenttreenode"."score" FROM "discussions_bestcommenttreenode" WHERE ("discussions_bestcommenttreenode"."depth" = 3 AND "discussions_bestcommenttreenode"."path" BETWEEN '000100010000' AND '00010001zzzz' AND "discussions_bestcommenttreenode"."score" > 8) ORDER BY "discussions_bestcommenttreenode"."path" ASC  LIMIT 1;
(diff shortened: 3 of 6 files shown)


docs/reddit-migration/strategy/reads-and-writes.md
```diff
@@ -0,0 +1,36 @@
...
+In general, migrating reads and writes can happen in a siloed manner, in order of highest to lowest level in the relationship hierarchies:
+
+- Channels
+- Posts
+- Comments
+- etc...
+
+### Add writes to the new tables
+
+This should be fairly straightforward, we should add another write to where we are currently writing (calling APIs) to reddit.
```
pdpinch: Curious if this will be part of a transaction, with the call to reddit?

rhysyngsun: Where we have explicit transactions already these should be added as part of those yes, there are some places where that hasn't been possible or valuable so it's ok in those places not to have a transaction. I believe they're documented as such in the code.

docs/reddit-migration/strategy/schema.md
```diff
@@ -0,0 +1,61 @@
...
+
+Votes are probably the trickiest moving piece in this reddit migration because it's impractical to get the data directly from the reddit backend for a backpopulate. This probably means we're going to need either some kind of ETL process. This is only a 1-time event, so an approach as basic as an export/import of a CSV would be adequate.
+
+
+### Reported Content
+
+We need a data model for reported content (e.g. report post/comment to a moderator). We don't need much data here, the following is sufficient:
+
+- `GenericForeignKey` to `Post` / `Comment` - we should unique on this FK so that we only get one report object per content object. We should handle new reports as a `get_or_create`
+- `resolved_on = DateTimeField(null=True)` - when a report is resolved, this is set the the current datetime, nullable datetimes have useful properties that we might want to take advantage of down the road (e.g. if a post is edited, delete any reports prior to the edit time)
```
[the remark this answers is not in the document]
rhysyngsun: Ah, i forgot about the ignore report action, I think that still fits into this data model though, because my expectation is that resolved reports don't show up to moderators and so subsequent reports just end up being a no-op on the already resolved report. Unless you feel like there's value in knowing _how_ a report was resolved (approved, removed, or ignored). That would just be an extra column.

docs/reddit-migration/strategy/schema.md
```diff
@@ -0,0 +1,61 @@
...
+
+- Maintain current API result structures
+  - Specifically, we need to support some form of reddit's "more comments" pagination. This **does not** need to partition the comments exactly as reddit would, the interface just needs to be compatible with how the frontend traverses it.
+- Support currently used sorting methods (best, old, new)
+- Support rendering an arbitrary subtree of the comments for the comment detail view
+
+From the Prior Art listed above, Materialized Paths are probably our best bet and we're already familiar to some degree with them because they're what `wagtail` uses under the hood for page hierarchies.
+
+The big thing here is that in order to support the sorts we need one for *each type of sort*. That precludes using a single table for everything so we'll want to have a hierarchical table for each sort and then `prefetch_related` on the `Comment` table so that code can work generically over similarly structured tables. This approach means these sorting hierarchy tables end up being an acceleration data structure for extracting a portion of the comment tree.
```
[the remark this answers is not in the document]
rhysyngsun: Sorry, I'll drop that terminology, I didn't realize it's not commonly used outside of graphical programming.

It generally refers to a data representation of precomputed lookups for optimizations.

docs/reddit-migration/strategy/schema.md
```diff
@@ -0,0 +1,61 @@
...
+Channels should have the fields carried over with the following adjustments:
+
+ - `channel_id` -> `id`
+
+### Posts
+
+Posts should have the fields carried over with the following adjustments:
+
+ - `post_id` -> `id`
+ - `preview_text` / `plain_text` - we should to normalize this property / attribute down to a single field so we're not recomputing it every time a read occurs
```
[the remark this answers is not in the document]
rhysyngsun: I'll clarify that,  we currently have `text` and `preview_text` columns on the data model. `text` is markdown and used for text posts. `preview_text` comes from Embedly-extracted plain text for link posts. We _also_ have an `@property` that at runtime renders either the article text or the markdown to plain text **OR** if it's a link post uses `preview_text`. My proposal is to not compute all that every time we serialize a post to a response and instead put all pre-rendered plain text representations of the post into a new field called `plain_text`. We would then only have `text` and `plain_text`.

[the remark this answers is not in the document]
rhysyngsun: See also Gavin's RFC for the embedly link preview: https://github.com/mitodl/open-discussions/blob/master/docs/rfcs/0002-indexing-embedly-link-preview.md#4-related-but-separate-issue-getting-rid-of-either-the-plain_text-or-text-field-in-es-documents

docs/reddit-migration/strategy/schema.md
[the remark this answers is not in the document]
pdpinch: I'm not looking to add new functionality. I was just trying to understand how the data model represents the current functions.

docs/reddit-migration/strategy/reads-and-writes.md
pdpinch: I think it would be worth saying that explicitly here, because I can see this question coming up again.

docs/reddit-migration/strategy/schema.md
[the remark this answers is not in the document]
rhysyngsun: Actually, you're right, we need some kind of value to represent the action that was taken on the report, otherwise `ignored` vs `approved` is ambiguous since both cause no change to the post or comment, but `ignored means further reports don't show up.

pdpinch:
How is this modeled on Reddit?

rhysyngsun:
> How is this modeled on Reddit?

Reddit uses a [Entity Attribute Value](https://tapoueh.org/blog/2018/03/database-modelization-anti-patterns/#entity-attribute-values) schema, which "works" for large scale data but is generally regarded as a sql antipattern. It's more or less an implementation of a document store in sql.

As a bit of an oversimplification, the data models we care about (subreddits, submissions, comments) and generally stored on 2 tables:

- A "Thing" table - a record of `{id, type}` (e.g. `{id: 123, type: "submission"}`)
- A data table - a key/value store of all the properties on the thing (e.g. `{thing_id: 123, key: "title", value: "My new post"}`

rhysyngsun:
> I have a couple of small questions, plus some fixes for typos.
> 
> Big question: it's not obvious to mean how you intend to chunk this. Is this it?
> 
>     1. Establish New Schema
> 
>     2. Migrate reads and writes
>        (possible subtasks here; would like to do this without delay)
> 
>     3. Cleanup
>        (possible subtasks here; no urgency)

Yes, that's correct. I've clarified it in my next commit.

docs/reddit-migration/strategy/schema.md
```diff
@@ -0,0 +1,61 @@
...
+Prior art:
+
+ - SQL Antipatterns ("Naive Trees"):http://www.r-5.org/files/books/computers/languages/sql/style/Bill_Karwin-SQL_Antipatterns-EN.pdf
+ - Models for Hierarchical Data: https://www.slideshare.net/billkarwin/models-for-hierarchical-data
+ - Django treebeard: https://django-treebeard.readthedocs.io/en/latest/
+
+Reddit stores comment trees in Cassandra, which is overkill for our current use cases. That said, we need to be able to migrate the comments in such a manner that we achieve the following:
+
+- Maintain current API result structures
+  - Specifically, we need to support some form of reddit's "more comments" pagination. This **does not** need to partition the comments exactly as reddit would, the interface just needs to be compatible with how the frontend traverses it.
```
aliceriot: is there a way we can determine how many posts currently on the production instance actually pass the "more comments" barrier? just wondering if we may be able to get away with leaving out this detail from an MVP for migrating away from reddit.

meaning, if we do (as I suspect) have very very few posts with more than, say, 20 comments we may be falling into a bit of a YAGNI trap if we add comment pagination to the tree initially. Of course we do want to make sure we use a solution that is going to be compatible with it when we do implement it.

docs/reddit-migration/considerations/sorting.md
```diff
@@ -0,0 +1,32 @@
...
+   - See the corresponding code in the reddit source:
+     - https://github.com/mitodl/reddit/blob/stable/r2/r2/lib/normalized_hot.py#L38-L79
+ - `New` and `Top` - same method as channels
+
+ #### Comments
+
+ Comments only have a few sorting options we're currently using:
+
+ - `New` and `Old` - chronological sorts, exactly when they say
+ - `Best` - sorted by `upvotes / downvotes`
```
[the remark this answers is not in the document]
rhysyngsun: I've clarified and reorganized this whole document and added more links to the source code, hopefully it makes more sense now.

docs/reddit-migration/strategy/schema.md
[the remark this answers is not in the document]
rhysyngsun: Ok, decided to go with a `reports_ignored` prop on `Post`/`Comment` and keeping the current data model as-is.

docs/reddit-migration/strategy/schema.md
```diff
@@ -0,0 +1,99 @@
...
+- Support currently used sorting methods (best, old, new)
+- Support rendering an arbitrary subtree of the comments for the comment detail view
+
+From the Prior Art listed above, Materialized Paths are probably our best bet and we're already familiar to some degree with them because they're what `wagtail` uses under the hood for page hierarchies.
+
+The big thing here is that in order to support the sorts we need one for *each type of sort*. That precludes using a single table for everything so we'll want to have a hierarchical table for each sort and then `prefetch_related` on the `Comment` table so that code can work generically over similarly structured tables. This approach means these sorting hierarchy tables end up being a sorted lookup table for extracting a portion of the comment tree.
```
aliceriot: does "one" in the first sentence in this paragraph mean that we need a separate materialized path for each sort? that's what I take this to mean but just wanted to confirm.

that seems like a decent way to do it to me. Django-treebeard supports it as well. 

By having a hierarchical table for each sort, do you mean (to use some terminology from the treebeard project) we would have something like

```py
class BestComment(MP_Node):
    comment = models.ForeignKey(Comment)
    post = models.ForeignKey(post)

class OldComment(MP_Node):
    comment = models.ForeignKey(Comment)
    post = models.ForeignKey(post)

class NewComment(MP_Node):
    comment = models.ForeignKey(Comment)
    post = models.ForeignKey(post)
```

or something along those lines? it seems like it could get tricky to update all of those and keep them in sync. I guess in my pseudocode I assumed that the comment text, votes, etc would be stored in a separate table so it's not duplicated. I guess this reason is why reddit comments still show up in the tree after they have been deleted?

My question here would be to ask whether we're over complicating by assuming we want to support all of these sorts. Are we sure this is something we want to emulate, or would we be satisfied with only one of the three?

It seems to me that the 'old' and 'new' ones are the simplest, since they are simply adding new entries to the tree. but the 'best' one would have to be recalculated on every vote, and potentially a whole series of adjustments made. I guess we could do this intelligently - like if there's a vote on a top-level comment we're only going to be making adjustments to the top-level, not to the whole tree, and the only possible adjustment would be if the score of one comment at a given level in the hierarchy is now higher or lower than another they have to be reordered. But idk, I'm wondering about the performance / reliability aspects of trying to do something like that efficiently and reliably when a lot of upvotes are coming in (not that we see much traffic on production now but still).

docs/reddit-migration/strategy/schema.md
rhysyngsun: I think the problem is that if we want to limit the api response to a reasonable size we need to pagination it somehow. I can see 2 ways of doing that:

- Select the first N items from the tree and return "more" links as appropriate
  - This is easy to do with the way `treebeard` structures the table
  - This is reasonably close, but not exactly how reddit does it. I think reddit does some kind of branch clipping based on the relevance/popularity of tree nodes, but we don't need anything nearly that fancy
  - This would mean that we would have 0-2 "more" links in the response:
    - 0 if we loaded the entire tree in 1 response
    - 1 if we cleanly paginated between 2 top-level comments
    - 2 if we paginated in the middle of a branch - 1 in the middle of the branch and another at the top level
- Select *at least* the first N comments, but only paginate between top-level comments
  - The query here might be a little tricky, because you don't really know how many top-level comments you need to select. Might require 2 queries - first one to get N number of comments, second one to get the rest of the branch of the last comment in that query response. Haven't thought too much about this kind of approach though.
  - There's no upper bound on how many items this response could return since the branches are unbounded, although in practice it might not be too bad.
  - Although if conversational engagement on the app gets high enough, we could see a performance hit here

So we could simplify it, but at the risk of having to reintroduce this mechanism when engagement goes up.

docs/reddit-migration/considerations/sorting.md
```diff
@@ -0,0 +1,34 @@
...
+##### Posts in Frontpage
+
+ - `Hot` - posts are sorted using a normalized version of the hot algorithm. This is done to ensure that content from less active channels isn't overshadowed by content from more popular channels.
+   - See the corresponding code in the reddit source:
+     - https://github.com/mitodl/reddit/blob/stable/r2/r2/lib/normalized_hot.py#L38-L79
+ - `New` and `Top` - same method as channels
+
+ #### Comments
+
+ Comments only have a few sorting options we're currently using: `New`, `Old`, and `Confidence`
```
aliceriot: does it make sense to think about storing these sorts in a materialized view or something so we're not always recomputing? I guess the frontpage (which is the thing I was thinking of, bc of the email send) is different for every users anyhow so we wouldn't get much mileage out of something like that there. Maybe we could go ahead and try to implement our own sorting and then see how it shakes out performance wise and see if we have any need of something like that.

docs/reddit-migration/considerations/sorting.md
```diff
@@ -0,0 +1,32 @@
+### Sorting
+
+There are several areas where reddit sorts objects coming back from their APIs in a way that isn't necessarily trivial via standard django ORM methods. This list is meant to be comprehensive so we know the problem space:
```
mbertrand: Are we generally happy with reddit's sorting or could this be an opportunity to come up with something different instead of trying to match its algorithms?

docs/reddit-migration/strategy/schema.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Votes
+
+We should have a table for `PostVote` and `CommentVote`, these should have the following columns:
+
+- `comment = ForeignKey(Comment)` / `post = ForeignKey(Post)` - item that was voted on
+- `voter = ForeignKey(User)` - the user who made the vote
+- `direction = SmallIntegerField()` - we store either a `1` or `-1` to indicate which direction the user voted. In practice, `PostVote` will always have a `direction == 1`, but it's useful to carve this out now in case that changes at a later date. It's also useful to be able to run a `SUM()` over that column.
+
+Votes are probably the trickiest moving piece in this reddit migration because it's impractical to get the data directly from the reddit backend for a backpopulate. This probably means we're going to need either some kind of ETL process. This is only a 1-time event, so an approach as basic as an export/import of a CSV would be adequate.
```
aliceriot: sincere question: is it worth it to recover all of this data?

When I look at the production instance of open there is really not a lot of engagement going on right now. Many channels have a few posts at most, most with no comments and only one or two upvotes. some are more active, but from the channels I have access to it looks like the only ones that have semi-active engagement on them are the ones affiliated with micromasters programs (dedp and so on). even there, we have a post every couple of days. It doesn't seem like we have a lot of active users who are regularly posting, upvoting, etc, other than people who work in OL.

What if instead of spending a lot of time and effort getting every single vote from reddit we did the following:

- implement immediately a new vote table in postgres, and save all votes going forward in that table as well as in reddit
- at the time of migration, when we're ready to start calculating our own scores, do a backfill where we look at all the votes we have stored vs the scores for posts, comments, etc and create dummy votes for any differences. so like if there's a post which has a score of 10 but we have 6 upvotes in postgres then just create 4 upvotes with no user linked to them. that will keep the score the same for the post when we switch over to our new system.

My assumption, which could be wrong, is that switching off reddit completely is, even if we start working immediately, something which is going to take a pretty long time. So I was thinking that if we took the approach above, and starting saving votes ourselves immediately, by the time we're actually in a position to switch reddit off we'll probably already have accurate saved votes for the past four-six months of content. This means that as we switch to our own way to calculate score and show whether the user has upvoted content or not we'll have that data to use for recent content which the user is likely to see.

my thought is that we should at least ask whether showing a personalized upvote button to a user on old content is something that is worth spending a lot of engineering time on.

Anyhow - I put this out as food for thought, I'm not convinced one way or the other but I felt it would be worthwhile to raise the question.

I do think that probably in any case we may want to look into getting our own table going for recording votes soon, since it may make our job easier going forward.

docs/reddit-migration/strategy/schema.md
```diff
@@ -0,0 +1,61 @@
...
+- Support currently used sorting methods (best, old, new)
+- Support rendering an arbitrary subtree of the comments for the comment detail view
+
+From the Prior Art listed above, Materialized Paths are probably our best bet and we're already familiar to some degree with them because they're what `wagtail` uses under the hood for page hierarchies.
+
+The big thing here is that in order to support the sorts we need one for *each type of sort*. That precludes using a single table for everything so we'll want to have a hierarchical table for each sort and then `prefetch_related` on the `Comment` table so that code can work generically over similarly structured tables. This approach means these sorting hierarchy tables end up being an acceleration data structure for extracting a portion of the comment tree.
+
+### Votes
+
+We should have a table for `PostVote` and `CommentVote`, these should have the following columns:
```
mbertrand: Would a `GenericForeignKey` to `Post`/`Comment` make sense here?

docs/reddit-migration/strategy/schema.md
rhysyngsun: Yeah, that's generally the data model we'd end up using, I have a functional branch here: https://github.com/mitodl/open-discussions/tree/nl/reddit_schema_migration

Materialized paths are pretty efficient with moves/reordering. Treebeard abstracts that all away from us so it ends up only being a few lines of code. Here's how you'd do that with my branch:

```python
from django.db import transaction
from discussions.models import *
channel = Channel.objects.create()
post = Post.objects.create(channel=channel)
comment1 = Comment.objects.create(post=post)
comment2 = Comment.objects.create(post=post)
comment3 = Comment.objects.create(post=post)

@transaction.atomic
def rescore(node, new_score):
    node.score = score
    node.save()
    node.move(node)

# comment trees start with a root node w/ `comment=None`
# this would get created when the post is created
root = BestCommentTreeNode.add_root(post=post, score=0) 
# top level comment
c1n = root.add_child(comment=comment1, score=1)
# replies to comment1
c2n = c1n.add_child(comment=comment1, score=4)
c3n = c1n.add_child(comment=comment1, score=5)

# c3n should list above c2n
print(root.dump_bulk())

rescore(c2n, 8)
# c2n is now sorted above c3n and has the updated score
print(root.dump_bulk())
```

The 4 queries the move generates are (fair warning it's a bit noisy, but it's 2 `UPDATE`s and 2 `SELECT LIMIT 1`s):

```
UPDATE "discussions_bestcommenttreenode" SET "path" = '000100010002', "depth" = 3, "numchild" = 0, "post_id" = 2, "comment_id" = 3, "score" = 8 WHERE "discussions_bestcommenttreenode"."id" = 5;
SELECT "discussions_bestcommenttreenode"."id", "discussions_bestcommenttreenode"."path", "discussions_bestcommenttreenode"."depth", "discussions_bestcommenttreenode"."numchild", "discussions_bestcommenttreenode"."post_id", "discussions_bestcommenttreenode"."comment_id", "discussions_bestcommenttreenode"."score" FROM "discussions_bestcommenttreenode" WHERE ("discussions_bestcommenttreenode"."depth" = 3 AND "discussions_bestcommenttreenode"."path" BETWEEN '000100010000' AND '00010001zzzz' AND "discussions_bestcommenttreenode"."score" > 8) ORDER BY "discussions_bestcommenttreenode"."path" ASC  LIMIT 1; 
SELECT "discussions_bestcommenttreenode"."id", "discussions_bestcommenttreenode"."path", "discussions_bestcommenttreenode"."depth", "discussions_bestcommenttreenode"."numchild", "discussions_bestcommenttreenode"."post_id", "discussions_bestcommenttreenode"."comment_id", "discussions_bestcommenttreenode"."score" FROM "discussions_bestcommenttreenode" WHERE ("discussions_bestcommenttreenode"."depth" = 3 AND "discussions_bestcommenttreenode"."path" BETWEEN '000100010000' AND '00010001zzzz') ORDER BY "discussions_bestcommenttreenode"."path" DESC  LIMIT 1;
UPDATE "discussions_bestcommenttreenode" SET path='000100010005'||SUBSTR(path, 13) WHERE path LIKE '000100010002%';
```

docs/reddit-migration/considerations/sorting.md
rhysyngsun: I don't think a materialized view on top of treebeard has much benefit in most cases, but we can always look into it as an option if we're not liking the out-of-the-box performance.

docs/reddit-migration/strategy/schema.md
rhysyngsun: That's a fair question to ask, I don't think they're mutually exclusive either (we could take your suggestion initially and then remove and backpopulate later on as long as we have a backup of the original data). I defer to @pdpinch on what makes sense product/user-wise

docs/reddit-migration/considerations/sorting.md
rhysyngsun: I'm not opposed to it, but I think we should be careful about what changes could impact the UI and for me that would be the line of whether we do it as part of the migration or as a subsequent step.

That said, keeping behavior as similar as possible is probably going to make verifying the switchover easier on ourselves, so there's that to consider.

docs/reddit-migration/strategy/schema.md
aliceriot: that doesn't look too bad at all. has your experience been that the library is pretty nice to work with? I am no data structures expert but based on a little reading it does seem like a materialized path is the most straightforward option, and this package seems to have a good implementation of it.

does it handle removal / deletion of nodes that have children? it seems that the leaf node case is easy to handle.

docs/reddit-migration/considerations/sorting.md
aliceriot: oh sorry! meant for that comment to be w/r/t to the whole file, and I was mainly thinking about posts. but yeah we probably don't need to worry about that too much.

docs/reddit-migration/strategy/schema.md
rhysyngsun: I debated on that, I think the reason I settled for separate models is because I was thinking it'd be more efficient to perform aggregates over the table. It'd also be better for referential integrity.

docs/reddit-migration/strategy/schema.md
rhysyngsun: Yeah, it handles all of that pretty easily, I'm not sure in practice if we'll use the entirety of the possible operations, but it's very easy to work with for the use cases we have.

docs/reddit-migration/strategy/reads-and-writes.md
```diff
@@ -0,0 +1,36 @@
...
+This can be done with backpopulate commands and celery tasks. We've done this plenty of times so it shouldn't be too difficult.
+
+### Verify new data
+
+We want some kind of verification of the new data vs the old data. Probably a command that runs through all of the data and verifies records exist and are accurate.
+
+### Move reads to the new tables
+
+- Move APIs, celery tasks, and any other reads to the new data models. After this step there should be no references in application code to the old models.
+- Migrate any BI queries to the new tables
```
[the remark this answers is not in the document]
rhysyngsun: Yeah, I've noted that in `cleanup.md` since it's not critical to moving the app off its reddit dependency.

docs/reddit-migration/strategy/schema.md
aliceriot: that makes sense to me, especially given that treebeard has some support for doing this

docs/reddit-migration/strategy/schema.md
```diff
@@ -0,0 +1,211 @@
...
+The big thing here is that in order to support the sorts we need one table/model for *each type of sort*. That precludes using a single table for everything but it simplifies reads because we can pick our sorting data model and then `prefetch_related` on the `Comment` table so that code can work generically over similarly structured tables. This approach means these sorting hierarchy tables end up being a sorted lookup table for extracting a portion of the comment tree.
+
+An example of what the data model would look like (working branch [here](https://github.com/mitodl/open-discussions/tree/nl/reddit_schema_migration)):
+
+```python
+from django.db import models
+from treebeard.mp_tree import MP_Node
+
+class Post(models.Model):
+  pass
```
aliceriot:
```suggestion
    pass
```

docs/reddit-migration/strategy/schema.md
```diff
@@ -0,0 +1,211 @@
...
+    node.save()
+    node.move(node)
+
+# comment trees start with a root node w/ `comment=None`
+# this would get created when the post is created
+root = BestCommentTreeNode.add_root(post=post, score=0)
+# top level comment
+c1n = root.add_child(comment=comment1, score=1)
+# replies to comment1
+c2n = c1n.add_child(comment=comment1, score=4)
```
aliceriot:
```suggestion
c2n = c1n.add_child(comment=comment2, score=4)
```

docs/reddit-migration/strategy/schema.md
```diff
@@ -0,0 +1,211 @@
...
+c3n = c1n.add_child(comment=comment1, score=5)
```
aliceriot:
```suggestion
c3n = c1n.add_child(comment=comment3, score=5)
```

PR #2624, opened 2020-03-11
Support adding users to MIT-only channels

### LINKED ISSUE #2630: Support MIT-only channels
We want to be able to support MIT-only channels. Some criteria we want to be able to filter on:

- @mit.edu and friends email suffixes
- Touchstone authenticated users

rhysyngsun:
Closes #2630 

#### What's this PR do?
This adds support for generating channel contributors/subscribers via configured filters. It uses the `django-filter` library in combination with a new `ChannelMembershipConfig` data model. This allows us to define subsequent filter capabilities simply by extending `UserFilter`

- Create 3 new channels:

```python
from channels.api import get_admin_api
api = get_admin_api()
api.create_channel("mmc1", title="managed membership channel 1", channel_type="private")
api.create_channel("mmc2", title="managed membership channel 2", channel_type="private")
api.create_channel("mmc3", title="managed membership channel 3", channel_type="private")
```
- Ensure you have at least 2 users with different domains
- Ensure one of your users has a saml or micromasters social auth and the other one doesn't
- Go to django-admin and create 2 new membership configs:
  - One should use the email suffix and configure with the value "@domain.com" (NOTE: it is best to have the `@` symbol in there if you want `@domain.com` to match but not `@sub.domain.com`)
    - Select the first and second channels for this config
 - The other config should use an auth provider configuration set to the same provider that your user has
    - Select the second and third channels for this config
- Login with each of these users, they should be added as contributors/subscribers of the correct channels

Configuration templates ("Email suffix" and "Social Auth"):
Screenshot from 2020-03-13 09-45-48

Django admin config validation errors
Screenshot_2020-03-13 Change channel membership config Django site admin(1)

Available options:
Screenshot_2020-03-13 Change channel membership config Django site admin


rhysyngsun:
Hmm, that is an issue, i'll look into it.

rhysyngsun:
@mbertrand fixed that, it was a typo in the django admin field config. Also added a test for multiple filters cause it seems prudent as well.

PR #2626, opened 2020-03-12
Pin redis and nginx versions

noisecapella:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

https://trello.com/c/1KnhFWrP/64-anonymous-access-to-user-profiles

#### What's this PR do?
Pin a couple docker container versions so that everybody is on the same version

Site should load

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,7 +34,7 @@ services:
       - "5432"
 
   redis:
-    image: redis
+    image: redis:5.0.8
     ports:
       - "6379"
 
@@ -50,7 +50,7 @@ services:
       - "9101:9200"
 
   nginx:
-    image: nginx
+    image: nginx:1.9.5
     ports:
       - "8063:8063"
     links:


docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
       - "5432"
 
   redis:
-    image: redis
+    image: redis:3.0.3
```
aliceriot: is this the redis version we're using on prod? I just checked and I have 5.0.7 installed locally, and this version doesn't seem to work properly on my laptop

aliceriot: 3.0.3 is also [almost five years old](https://github.com/antirez/redis/releases/tag/3.0.3) as far as I can tell, so we may want to shoot for a newer version

noisecapella: Yes, I'll use 5.0.7. I think I had used the version that micromasters was pinned to

noisecapella:
@aliceriot Ready for review again

PR #2631, opened 2020-03-13
More intelligent OLX ingestion, and ingest static files for xPRO

noisecapella:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

https://trello.com/c/BzzuItbZ/62-index-xpro-static-files-in-addition-to-olx

#### What's this PR do?
 - Uses the `display_name` on the vertical for the title of the `ContentFile`
 - Ingest static file content after passing through tika, using the file extension to indicate file type to tika

 - Run `./manage.py backpopulate_xpro_data` and `./manage.py backpopulate_xpro_files`.
 - In a Django shell query for `ContentFile.objects.filter(content_type='file', run__platform='xpro')`. You should see many objects which represent static file content.

Note that none of the `ContentFile`s will have a `vertical` `content_type` because it looks like none of the OLX data for courses which contain verticals are also in the courses API for xPRO


pdpinch:
@noisecapella can you note somewhere what "more intelligent ingestion" means?

noisecapella:
@pdpinch I think "intelligent" here will mean using only child text and stripping out other pieces of XML from the vertical, stripping script and style tags altogether, and using mime types to tell tika how to parse documents

abeglova:
👍 This change looks good to me but I'm confused about why the xpro api and the courses in the tarfile aren't matching.

PR #2687, opened 2020-03-20
OCW webhook

### LINKED ISSUE #2612: S3-triggered webhook to import/update OCW courses
Currently a daily celery task is run to crawl through all OCW S3 subfolders looking for updated content.  Replace with an S3 event on the OCW source bucket that sends an SNS message to `../api/v0/ocw_webhook` whenever an S3 object is added, updated, or deleted.  This webhook endpoint should then trigger a task to process that particular OCW course/run.

- The event should be triggered by any S3 object creation/deletion event having a suffix of `0/1.json`
- There should be a small delay between reception of the webhook call and import of the affected OCW course/run, to make sure all relevant JSON files have been created/updated on S3.

mbertrand:
#### Pre-Flight checklist
* [x] Testing
  * [x] Code is tested
  * [x] Changes have been manually tested
* [x] Settings
  * [x] New settings are documented and present in `app.json`
  * [x] New settings have reasonable development defaults, if applicable

Closes #2612 
closes https://trello.com/c/AhXkk9oe/63-s3-triggered-webhook-to-import-update-ocw-courses

#### What's this PR do?
* Adds a new feature flag (`WEBHOOK_OCW`)
* Creates a webhook endpoint that will trigger an OCW import for a specific course based on the S3 key contained in the AWS notification JSON (if the new flag is `True`)
* The task to import the course should be delayed by `OCW_WEBHOOK_DELAY` seconds (as a precaution just in case it takes a little extra time for all relevant json files for a course to be written to the ocw bucket).
* The current daily task to iterate over all OCW courses won't do anything if the new flag is `True`
* The manual command `backpopulate_ocw_data` should still work because it sends an optional kwarg to override the feature flag.

* Add `FEATURE_WEBHOOK_OCW=True` to your .env file
* Via Postman or similar tool, post this data to the new endpoint `../api/v0/ocw_webhook/`:

```
{
    "Records": [
        {
            "s3": {
                "object": {
                    "key": "PROD/6/6.003/Fall_2011/6-003-signals-and-systems-fall-2011/0/1.json",
                }
            }
        }
    ]
}
```

* In the celery log you should soon see `Received task: course_catalog.tasks.get_ocw_courses`
* After a delay of at least 120 seconds, you should see `Task course_catalog.tasks.get_ocw_courses[<uid>] succeeded`


course_catalog/views.py
```diff
@@ -328,3 +339,40 @@ class TopicViewSet(viewsets.ReadOnlyModelViewSet):
...
+    def handle_exception(self, exc):
+        """Raise any exception with request info instead of returning response with error status/message"""
+        raise WebhookException(
+            "REST Error (%s). BODY: %s, META: %s"
+            % (exc, self.request.body, self.request.META)
+        ) from exc
+
+    def post(self, request):
+        """Process webhook request"""
+        content = rapidjson.loads(request.body.decode())
```
noisecapella: It looks like you can do `request.data` to avoid parsing at this point

course_catalog/views.py
```diff
@@ -328,3 +339,40 @@ class TopicViewSet(viewsets.ReadOnlyModelViewSet):
...
+        records = content.get("Records")
+        if features.is_enabled(features.WEBHOOK_OCW) and records is not None:
+            for record in content.get("Records"):
+                s3_key = record.get("s3", {}).get("object", {}).get("key")
+                prefix = s3_key.split("0/1.json")[0]
+                get_ocw_courses.apply_async(
+                    countdown=settings.OCW_WEBHOOK_DELAY,
+                    kwargs={
+                        "course_prefixes": [prefix],
+                        "blacklist": load_course_blacklist(),
```
noisecapella: We should move this out of the for loop since it initiates a request each time

course_catalog/views.py
```diff
@@ -328,3 +339,40 @@ class TopicViewSet(viewsets.ReadOnlyModelViewSet):
...
+            % (exc, self.request.body, self.request.META)
+        ) from exc
+
+    def post(self, request):
+        """Process webhook request"""
+        content = rapidjson.loads(request.body.decode())
+        records = content.get("Records")
+        if features.is_enabled(features.WEBHOOK_OCW) and records is not None:
+            for record in content.get("Records"):
+                s3_key = record.get("s3", {}).get("object", {}).get("key")
```
noisecapella: This should end with `.get("key", "")` so that the `split` on the next line will work

course_catalog/views.py
```diff
@@ -328,3 +339,40 @@ class TopicViewSet(viewsets.ReadOnlyModelViewSet):
...
     permission_classes = (AnonymousAccessReadonlyPermission,)
+
+
+class WebhookOCWView(APIView):
+    """
+    Handle webhooks coming from the OCW bucket
+    """
+
+    permission_classes = ()
+    authentication_classes = ()
```
noisecapella: Is there any way to authenticate that the webhook came from S3?

course_catalog/views.py
```diff
@@ -328,3 +339,40 @@ class TopicViewSet(viewsets.ReadOnlyModelViewSet):
...
+
+    def handle_exception(self, exc):
+        """Raise any exception with request info instead of returning response with error status/message"""
+        raise WebhookException(
+            "REST Error (%s). BODY: %s, META: %s"
```
noisecapella: Can we use an f-string or `.format` here?

course_catalog/views.py
mbertrand: I did it this to way purposely so an exception would be raised on the next line if it's not there - if there is no S3 key, it's not a valid request (tested via sending `{"Records": [{"foo": "bar"}`)

course_catalog/views.py
mbertrand: @markbreedlove is there some kind of extra authentication key parameter that could always be sent with an SNS notification?  The code could check that the origin is amazon.com but that can be spoofed.  For now I'll add a check that the bucket in the request is the same as `OCW_CONTENT_BUCKET_NAME`

course_catalog/views.py
mbertrand: I think the standard for logging exceptions is to use `%s` - something to do with how sentry handles them, it prevents duplicate issues somehow?

course_catalog/views.py
markbreedlove: It's not clear from the documentation, but I will assume that I can add a querystring to the URL of the request that SNS will send to your endpoint.

mbertrand: @markbreedlove can you give that a try, maybe something like `?webhook_key=<somerandomstring>`?  If it works I'll add a new .env `WEBHOOK_KEY` set to that string

markbreedlove: OK. I've just added an SNS subscription that posts to `discussions-ci`. I can discuss the details with you over Slack. The subscription is currently "pending confirmation."

markbreedlove:
I don't have anything more to add. Ready to try it out when it's on CI.

course_catalog/views.py
mbertrand: @markbreedlove is there a way to retrieve the confirmation URL?  Otherwise we can try sending another confirmation message once it's on CI, and it should log an exception containing the message body in sentry.

course_catalog/views.py
noisecapella: Right but this is an exception being raised, not one being logged.

course_catalog/views.py
markbreedlove: We'll have to do the latter, because I can't retrieve the confirmation URL.
I should have looked at the documentation earlier.
I'll need to take the information that's logged and paste it into a form in the control panel to confirm the subscription.

An additional feature, which I'm not inclined to request right now, would be for `WebhookOCWView.post()` to log a specific message or trigger an alert of some kind if the `x-amz-sns-message-type` header is "SubscriptionConfirmation"

course_catalog/views.py
```diff
@@ -328,3 +339,42 @@ class TopicViewSet(viewsets.ReadOnlyModelViewSet):
...
+
+    def handle_exception(self, exc):
+        """Raise any exception with request info instead of returning response with error status/message"""
+        raise WebhookException(
+            f"REST Error ({exc}). BODY: {rapidjson.dumps(self.request.data or {})}, META: {self.request.META}"
+        ) from exc
+
+    def post(self, request):
+        """Process webhook request"""
+        if request.GET.get("webhook_key", None) != settings.OCW_WEBHOOK_KEY:
```
noisecapella: Can we use [compare_digest](https://docs.python.org/3/library/hmac.html#hmac.compare_digest) here to avoid [this vulnerability](https://codahale.com/a-lesson-in-timing-attacks/) with short-circuit comparison?

noisecapella:
@mbertrand You should be able to do `from hmac import compare_digest` instead of importing from within `websocket`

PR #2701, opened 2020-03-25
Remove tox, use pytest directly instead

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to https://github.com/mitodl/mitxpro/pull/44

#### What's this PR do?
Removes tox. Instead people will use pytest directly to run tests. Instead of running `black` in tox, it will be run separately in Travis. This is intended to match the current behavior in xPRO

Tests should pass


README.md
```diff
@@ -168,6 +168,11 @@ docker-compose run --rm web black .
 docker-compose run -p 9001:9001 watch npm run storybook
 ```
 
+### Running tests
```
gsidebo, line 171: I just updated the handbook to account for projects moving away from `tox` so you can remove this section. I also added a note about using the `black` formatter, so you can also remove that if you want to

docker-compose.travis.yml
```diff
@@ -1 +1,27 @@
...
+      CELERY_TASK_ALWAYS_EAGER: 'True'
+      SECRET_KEY: 'not_very_secret_key'
+      OPEN_DISCUSSIONS_BASE_URL: 'http://localhost:8063/'
+      OPEN_DISCUSSIONS_COOKIE_NAME: 'cookie_monster'
+      OPEN_DISCUSSIONS_COOKIE_DOMAIN: 'localhost'
+      OPEN_DISCUSSIONS_DEFAULT_SITE_KEY: 'mm_test'
+      MAILGUN_SENDER_DOMAIN: 'other.fake.site'
+      MAILGUN_KEY: 'fake_mailgun_key'
+      ELASTICSEARCH_INDEX: 'testindex'
+      INDEXING_API_USERNAME: 'mitodl'
```
gsidebo, line 27: Can we not use `x-environment` to specify all of these and reuse them like `docker-compose.yml` does?

open_discussions/tests/app_tests.py
```diff
@@ -0,0 +1,13 @@
+"""Application level tests"""
```
gsidebo, line 1: A couple things: 
-  `open_discussions/app_tests.py` still exists, so this seems redundant
- Adding this `tests/` directory is confusing. The pattern in this codebase and all of our other codebases is have test files as siblings to the files they test, and to avoid grouping them into a `tests/` directory. I think it would make more sense to get rid of this directory and to rename `render_bundle_test.py` to `templatetags_test.py` or something like that

travis/python_tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+status=0
+
+function run_test {
```
gsidebo, line 4: Personally I've found it helpful in xpro to have the command printed before we run it: https://github.com/mitodl/mitxpro/blob/master/travis/python_tests.sh#L4-L9

noisecapella:
Closing this PR since it's getting outdated

Issue #2714, opened 2020-03-30
Bug: searching for "Curate" in the teachremote channel returns people named Curt


odlbot:
## Steps to Reproduce
Go to https://open.mit.edu/c/teachremote/search/?q=curated

## Expected Results
* Posts that explicitly mention "curated" should be top of the results
* Names like "Murat" and "Curt" shouldn't be returned at all, or should be very low in the results

## Actual results 
See attached

┆Attachments: <a href="https://trello-attachments.s3.amazonaws.com/5dfd295df7c63e06791c4cae/5e8210e9b48ad912d5ef939c/b1a10a73caa9498bf8ca7ec2f79e135a/image.png">image.png</a>

odlbot:
➤ Peter Pinch commented:

Based on the conversation in Slack, it seems the thing to do is to reduce or eliminate the fuzzy searching, similar to what we did for learning resources.

PR #2720, opened 2020-03-30
remove fuzzy search

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/open-discussions/issues/2714

#### What's this PR do?
This PR removes the fuzzy search parameter from channel search queries. Fuzzy search produces results that are too broad. For example profiles matching "Curt" and "Murat" were returned in a search for "curated", which was de-stemmed to "curat" by elasticsearch

Create a new post titled "Outside the box"
Create a new post titled "I like foxes" 

Searching for "Fox" in http://localhost:8063/search/ should not return the second post but not the first.

### DIFF
--- a/static/js/lib/search.js
+++ b/static/js/lib/search.js
@@ -554,9 +554,8 @@ export const buildChannelQuery = (
         should: [
           {
             multi_match: {
-              query:     text,
-              fields:    searchFields(type),
-              fuzziness: "AUTO"
+              query:  text,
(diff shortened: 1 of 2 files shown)


abeglova:
If this is RTM I think suggestions can be a follow up pr

pdpinch:
Suggestions should be a follow-up PR.

PR #2764, opened 2020-04-10
Podcasts: add podcasts, podcast episodes to the admin

aliceriot:

no ticket

#### What's this PR do?

^^ what it says

should be sections in the admin now

### DIFF
--- a/course_catalog/models.py
+++ b/course_catalog/models.py
@@ -110,6 +110,9 @@ class LearningResourceOfferor(TimestampedModel):
 
     name = models.CharField(max_length=256, unique=True)
 
+    def __str__(self):
+        return self.name
+
 
 class LearningResource(TimestampedModel):
     """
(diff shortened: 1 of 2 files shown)


course_catalog/models.py
```diff
@@ -504,6 +504,9 @@ class Podcast(LearningResource, LearningResourceGenericRelationsMixin):
     url = models.URLField(null=True, max_length=2048)
     searchable = models.BooleanField(default=True)
 
+    def __str__(self):
```
noisecapella, line 507: Can we do this or something similar for `PodcastEpisode` too/

aliceriot: I was going to do so, but then I though that what I would want to do is something like

```
Episode {self.episode_number} of {self.podcast.title}
```

which would require pulling in the podcast object - what would you suggest for a good `__str__` for this object?

noisecapella: I'm fine either way. I don't think the performance hit will be significant since `__str__` is only used in the console for this

PR #2765, opened 2020-04-10
Add recent podcasts API

### LINKED ISSUE #2763: Podcasts recent episodes API
As a podcast UI frontend, I'd like to have a list of the 10 most recent episodes to show to the end user

#### Acceptance Criteria:

- [ ] Create `/api/v0/podcasts/recent/` with a serialized `PodcastEpisode`, also including the id to reference a `Podcast` from `/api/v0/podcasts/`
- [ ] This should use the same feature flag as `/api/v0/podcasts/`

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #2763 

#### What's this PR do?
Adds `/api/v0/podcasts/recent/` to list recent podcast episodes

Populate some podcast episodes then go to `/api/v0/podcasts/recent/` and verify that it looks right. The `last_modified` value should start with the most recent and go backward in time.

### DIFF
--- a/course_catalog/serializers.py
+++ b/course_catalog/serializers.py
@@ -689,6 +689,7 @@ class PodcastEpisodeSerializer(serializers.ModelSerializer):
             "image_src",
             "url",
             "offered_by",
+            "podcast",
         ]
 
 
--- a/course_catalog/serializers.py
+++ b/course_catalog/serializers.py
@@ -690,6 +690,7 @@ class PodcastEpisodeSerializer(serializers.ModelSerializer):
             "url",
             "offered_by",
             "podcast",
+            "last_modified",
         ]
 
 
--- a/course_catalog/views.py
+++ b/course_catalog/views.py
@@ -436,6 +436,7 @@ class PodcastViewSet(viewsets.ReadOnlyModelViewSet):
     """
 
     serializer_class = PodcastSerializer
+    pagination_class = DefaultPagination
     permission_classes = (ReadOnly & PodcastFeatureFlag,)
 
     queryset = Podcast.objects.filter(published=True).prefetch_related(
@@ -457,6 +458,7 @@ class RecentPodcastEpisodesViewSet(viewsets.ReadOnlyModelViewSet):
     """
 
     serializer_class = PodcastEpisodeSerializer
+    pagination_class = DefaultPagination
     permission_classes = (ReadOnly & PodcastFeatureFlag,)
 
     queryset = (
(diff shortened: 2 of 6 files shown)


course_catalog/serializers.py
```diff
@@ -689,6 +689,7 @@ class Meta:
             "image_src",
             "url",
             "offered_by",
+            "podcast",
```
aliceriot, line 692: could we also add the `last_modified` field here, which is used for the published date? I suspect we'll want that in the UI

course_catalog/views.py
```diff
@@ -448,3 +449,21 @@ class PodcastViewSet(viewsets.ReadOnlyModelViewSet):
...
+    permission_classes = (ReadOnly & PodcastFeatureFlag,)
+
+    queryset = (
+        PodcastEpisode.objects.filter(published=True)
+        .order_by("-last_modified", "-id")
+        .prefetch_related(
+            Prefetch("offered_by", queryset=LearningResourceOfferor.objects.all()),
+            Prefetch("topics", queryset=CourseTopic.objects.all()),
+        )
+    )
```
aliceriot, line 469: does this by default return a list of all the podcast episodes? just curious if we should limit it to 10 or something (obviously ignore this comment if its already in there haha)

noisecapella: Hmm, I thought `DefaultPagination` was set by default but apparently it's not for these ViewSets. It sets a limit of 10 per page with a 100 items max. I'll add it to the PR but it will change the JSON returned. Alternatively, it can put the pagination links in the header but it doesn't look like we do that for other APIs.

Do we have any code in the frontend that currently handles pagination? It looks like `/api/v0/videos/` and other APIs already use the paginated format so I assume we are doing something in the frontend already to handle that.

course_catalog/serializers.py
noisecapella: Sounds good

course_catalog/views.py
aliceriot: we have pagination I think for the contents of user lists and a few other things

noisecapella: I'll return the paginated output here. It looks like the `api/v0/videos/new` API does not handle pagination, it only works with whatever is in the first page. I guess we can handle pagination later on if necessary.

odlbot: Not sure you are are trying to make the api flexible, but for the purpose of this particular feature (show latest episodes) we don't need any pagination. A fixed number would be fine. 
When we index podcasts in "Open-learn" , that's where user can see list of all podcast/episodes.

noisecapella:
@aliceriot Comments addressed

course_catalog/views.py
aliceriot: that makes sense to me - I think in the designs right now it's only showing 6 recent episodes, so just even limiting it to that would be fine

PR #2782, opened 2020-04-15
Audio player

gumaerc:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

https://trello.com/c/C0KaEV0e/268-podcasts-build-a-simple-podcast-player

#### What's this PR do?
Adds an audio player component to open-discussions based on amplitudejs, mainly for podcasts but it could be used with any audio.

Run open-discussions and visit `/podcasts`.  Click play on any episode and it should open the player.  Try different resolutions, podcasts and run the tests.  Try adjusting the playback speed as well as seeking by buttons and the seek bar.

Screen Shot 2020-04-22 at 5 29 06 PM
Screen Shot 2020-04-22 at 5 29 32 PM

swanson


static/js/components/AudioPlayer.js
```diff
@@ -0,0 +1,116 @@
...
+            </div>
+            <div className="audio-player-playback-speed-container">
+              <button className="light-outlined amplitude-playback-speed" />
+            </div>
+          </div>
+        </div>
+      </div>
+    )
+  }
+}
```
aliceriot, line 116: one general comment: we're trying to mainly use functional components and react's [hooks](https://reactjs.org/docs/hooks-intro.html) API for new features. I think if you want to stick with a class-based component for this that's fine - but you may find the hooks APIs nicer to work with (I do!).

static/scss/audio-player.scss
```diff
@@ -0,0 +1,135 @@
...
+  content: "1.0x";
+}
+
+.amplitude-playback-speed-15:before {
+  content: "1.5x";
+}
+
+.amplitude-playback-speed-20:before {
+  content: "2.0x";
+}
```
aliceriot, line 135: small scss comment - when I'm styling something like this (something we're building as a real self-contained UI component) I like to scope all the styles specific to it within an outer selector (like `.audio-player-container-outer`) which will just make sure that all these rules don't accidentally affect anything outside of it.

of course I think we're unlikely to have elements elsewhere on the site that will be selected by `.amplitude-playback-speed-15:before` but still a good general rule to follow I think :smile:

gumaerc: The latest commits to the branch should show everything nested nicely.

aliceriot:
not sure why, but I can't seem to pause it on mobile? I started the podcast playing on desktop, switched to the chrome mobile simulator, and then pressing the 'pause' button doesn't do anything. the +/- 30s buttons seem to work well, as does the seek bar

gumaerc:
> not sure why, but I can't seem to pause it on mobile? I started the podcast playing on desktop, switched to the chrome mobile simulator, and then pressing the 'pause' button doesn't do anything. the +/- 30s buttons seem to work well, as does the seek bar

Hmm, that's odd.  I just did that and it seemed to work for me.

static/js/components/AudioPlayer.js
```diff
@@ -0,0 +1,122 @@
+// @flow
+import _ from "lodash"
+import React, { useEffect, useRef, useCallback, useState } from "react"
+import { useSelector } from "react-redux"
+import { currentlyPlayingAudioSelector } from "../lib/redux_selectors"
+import Amplitude from "amplitudejs"
```
aliceriot, line 6: this is a fiddly thing, but we usually put a blank line to separate outside dependency imports from imports from our own modules

static/js/components/AudioPlayer_test.js
```diff
@@ -0,0 +1,65 @@
...
+      "https://chtbl.com/track/F9DD6B/cdn.simplecast.com/audio/2c64ac/2c64ace6-baf4-4e86-b527-445e611e6a31/aa0ca88f-3c4c-4d33-9897-36e45c43e012/film-is-for-everyone-with-prof-david-thorburn_tc.mp3"
+  }
+
+  beforeEach(() => {
+    helper = new IntegrationTestHelper()
+    wrapper = renderAudioPlayer({}, helper.store)
+    dispatchThen = helper.store.createDispatchThen(state => state.audio)
+    dispatchThen(setCurrentlyPlayingAudio(exampleAudio), [
+      SET_CURRENTLY_PLAYING_AUDIO
+    ])
```
aliceriot, line 36: I think you may be able to simplify some of this code if you use the `reduxQueryRenderer` in `IntegrationTestHelper`, you won't have to set up the `dispatchThen` thing for instance

static/js/components/PodcastEpisodeCard.js
```diff
@@ -20,14 +23,30 @@ type Props = {
...
+
   return (
     <Card className="podcast-episode-card low-padding">
       <div className="left-col">
         <div className="episode-title">
           <Dotdotdot clamp={2}>{episode.title}</Dotdotdot>
         </div>
         <div className="podcast-name">{podcast.title}</div>
-        <div className="play-placeholder black-surround">Play</div>
+        <div className="play-placeholder black-surround" onClick={playClick}>
```
aliceriot, line 47: we should probably rename `play-placeholder` to `play-button` while we're here :)

static/js/components/PodcastEpisodeCard_test.js
```diff
@@ -1,31 +1,34 @@
...
 describe("PodcastEpisodeCard", () => {
-  let podcast, episode
+  let podcast, episode, render, helper
   beforeEach(() => {
     podcast = makePodcast()
     episode = makePodcastEpisode(podcast)
+    helper = new IntegrationTestHelper()
+    render = helper.configureReduxQueryRenderer(PodcastEpisodeCard, {
+      podcast: podcast,
+      episode: episode
```
aliceriot, line 22: formatting nit: you can declare this object like `{ podcast, episode }` because those variables are already bound to the name you want them to have in the object

static/js/reducers/audio.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+import { SET_CURRENTLY_PLAYING_AUDIO } from "../actions/audio"
+import type { Action } from "../flow/reduxTypes"
```
aliceriot, line 3: can you add a blank line between the JS and `type` import here?

static/scss/audio-player.scss
```diff
@@ -0,0 +1,269 @@
...
+    @include breakpoint(desktop) {
+      padding: 8px 12px;
+      height: calc(100% - 16px);
+      width: calc(100% - 24px);
+    }
+    @include breakpoint(mobile) {
+      flex-direction: column;
+      padding: 8px 12px;
+      height: calc(100% - 16px);
+      width: calc(100% - 24px);
```
aliceriot, line 38: is the `24px` here because it's 2x the `12px` in the padding? if so might be good to make these things variables so it's clear how they relate

static/scss/audio-player.scss
```diff
@@ -0,0 +1,269 @@
...
+            font-size: 48px;
+          }
+
+          .amplitude-play-pause.amplitude-playing span:before {
+            content: "pause_circle_outline";
+          }
+
+          .amplitude-play-pause.amplitude-paused span:before {
+            content: "play_circle_outline";
+          }
```
aliceriot, line 167: is it necessary to change the display of the icon this way because it's laggy otherwise or something? just wondering if it would be easier to grab the boolean for whether it's playing or not out of `Amplitude` and use that to switch between these two icons in the react component itself

aliceriot:
yeah, on that play / pause issue I can't get it to do it now, so :woman_shrugging: let's not worry about it :)

static/scss/audio-player.scss
gumaerc: I switched this up so that AudioPlayer.js uses Amplitude's `initialized` callback to set event handlers on the `Audio` element's `play` and `pause` events.  I liked the CSS solution because it was technically less code and doesn't require any JS.  However I like the JS solution more now because I realized that instead of programmatically clicking the button, we can use the `Amplitude.play()` function and simply run `setAudioPlaying(true)` to set the state.

static/js/components/AudioPlayer_test.js
gumaerc: So I did this in the PodcastEpisodeCard tests and tried the same here, but I couldn't get the last test to work right with reduxQueryRenderer for some reason.  Amplitude is never initialized when the test is run in this way, but it is if I create the dispatch manually.

aliceriot: how were you setting up the redux state when you were using it? after you create the renderer there is a second argument you can pass to `render` called `beforeRenderActions`, you can pass a list of actions that will be dispatched to the store before the component is mounted. you should be able to use it like

```js
const { wrapper } = await render({}, [setCurrentlyPlayingAudio(exampleAudio)])
```

I think you can look at `components/AddToListDialog_test.js` for an example of this

gumaerc: Ah yea that worked.  Thanks!

static/js/components/AudioPlayer.js
```diff
@@ -0,0 +1,140 @@
...
+import { currentlyPlayingAudioSelector } from "../lib/redux_selectors"
+
+export default function AudioPlayer() {
+  const seekBar = useRef()
+  const playPauseButton = useRef()
+  const currentlyPlayingAudio = useSelector(currentlyPlayingAudioSelector)
+  const [audioLoaded, setAudioLoaded] = useState(false)
+  const [audioPlaying, setAudioPlaying] = useState(false)
+
+  useEffect(
```
aliceriot, line 16: small readability suggestion: it might be good to move the binding for the `amplitudeInitialized` up above this `useEffect` call

aliceriot: same for the `playbackStarted` and `playbackPaused` callbacks

PR #2799, opened 2020-04-17
Remove nested episodes from podcasts API to improve performance

noisecapella:
#### Pre-Flight checklist
- [x] Testing

None

#### What's this PR do?
Fixes slow performance in podcasts API due to a large amount of data being provided. The nested episodes are the largest piece of the REST API and they're not needed so they are removed from the results.

Populate some podcast data. Then go to `/api/v0/podcasts/` and make sure it feels reasonably snappy.


aliceriot:
locally this takes it from ~1s to ~80ms so that looks like a win to me :smiley:

PR #2815, opened 2020-04-21
Add episodes per podcast view

### LINKED ISSUE #2813: Add nested episodes list within podcasts API
Add `/api/v0/podcasts/:number/episodes` to list all serialized podcast episodes for a given podcast

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #2813

#### What's this PR do?
Adds `/api/v0/podcasts/:number/episodes/` to list all serialized episodes for a given podcast id

Go to the endpoint and verify that the episodes match what you expect to see. Unpublished episodes should not be included in the list.

### DIFF
--- a/course_catalog/views.py
+++ b/course_catalog/views.py
@@ -478,7 +478,9 @@ class EpisodesInPodcast(viewsets.ReadOnlyModelViewSet):
 
     def get_queryset(self):
         return (
-            PodcastEpisode.objects.filter(published=True, podcast=self.kwargs["pk"])
+            PodcastEpisode.objects.filter(
+                published=True, podcast__published=True, podcast=self.kwargs["pk"]
+            )
             .order_by("-last_modified", "-id")
             .prefetch_related(
                 Prefetch("offered_by", queryset=LearningResourceOfferor.objects.all()),
--- a/course_catalog/views.py
+++ b/course_catalog/views.py
@@ -466,3 +466,22 @@ class PodcastEpisodesViewSet(viewsets.ReadOnlyModelViewSet):
             Prefetch("topics", queryset=CourseTopic.objects.all()),
         )
     )
+
+
+class EpisodesInPodcast(viewsets.ReadOnlyModelViewSet):
+    """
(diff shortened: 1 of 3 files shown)


course_catalog/views.py
```diff
@@ -466,3 +466,22 @@ class PodcastEpisodesViewSet(viewsets.ReadOnlyModelViewSet):
...
+    """
+    Viewset for listing PodcastEpisodes for a given Podcast primary key
+    """
+
+    serializer_class = PodcastEpisodeSerializer
+    permission_classes = (ReadOnly & PodcastFeatureFlag,)
+
+    def get_queryset(self):
+        return (
+            PodcastEpisode.objects.filter(published=True, podcast=self.kwargs["pk"])
```
mbertrand, line 481: This may be an unrealistic edge case, but if I have published episodes in an unpublished podcast, I get the list of the published episodes returned, whereas the API URL for the podcast itself returns a 404.   Maybe that's more of a bad data situation?

course_catalog/views.py
```diff
@@ -466,3 +466,22 @@ class PodcastEpisodesViewSet(viewsets.ReadOnlyModelViewSet):
...
     )
+
+
+class EpisodesInPodcast(viewsets.ReadOnlyModelViewSet):
+    """
+    Viewset for listing PodcastEpisodes for a given Podcast primary key
+    """
+
+    serializer_class = PodcastEpisodeSerializer
+    permission_classes = (ReadOnly & PodcastFeatureFlag,)
```
mbertrand, line 477: do we need paging at all for the episodes? For instance if a podcast has 100's/1000's of episodes?

course_catalog/views.py
noisecapella: I added a filter so that an unpublished podcast will return an empty list

course_catalog/views.py
noisecapella: @aliceriot looked at this and I think we're good for now, performance-wise

PR #2833, opened 2020-04-23
Change queryset to show empty podcasts

### LINKED ISSUE #2830: Podcast episode filters out podcasts without any published episodes
We should show all podcasts which are published regardless of whether episodes are published

┆Attachments: <a href="https://github.com/mitodl/open-discussions/pull/2833">open-discussions #2833 Change queryset to show empty podcasts</a>

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #2830 

#### What's this PR do?
Fixes a bug where published podcasts which have no episodes were being omitted from the list. Instead we want to filter podcast episodes from `episode_count` while leaving the podcast in the list.

Mark all `PodcastEpisode` as unpublished for a particular published `Podcast`. You should not see it at `/api/v0/podcasts/` on master but you should see it on this PR, with `"episode_count": 0`

### DIFF
--- a/course_catalog/views.py
+++ b/course_catalog/views.py
@@ -7,7 +7,7 @@ from hmac import compare_digest
 import rapidjson
 from django.contrib.contenttypes.models import ContentType
 from django.db import IntegrityError
-from django.db.models import Prefetch, Count
+from django.db.models import Prefetch, Count, Q
 from django.utils import timezone
 from rest_framework import viewsets, status
 from rest_framework.decorators import action, api_view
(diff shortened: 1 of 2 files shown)


course_catalog/views.py
```diff
@@ -439,12 +440,26 @@ class PodcastViewSet(viewsets.ReadOnlyModelViewSet):
...
 
     queryset = (
-        Podcast.objects.filter(published=True, episodes__published=True)
+        Podcast.objects.filter(published=True)
         .prefetch_related(
             Prefetch("offered_by", queryset=LearningResourceOfferor.objects.all()),
             Prefetch("topics", queryset=CourseTopic.objects.all()),
         )
-        .annotate(episode_count=Count("episodes"))
+        .annotate(
```
gsidebo, line 448: Can you include a comment about why this somewhat complex query is needed here instead of just `Count("episodes")`?

noisecapella: Sure, just added one

noisecapella:
Any other comments?

course_catalog/views.py
gsidebo: Sorry I forgot to get to this earlier? Can we just use conditional aggregation here if the only change is that we want to count published episodes only? https://docs.djangoproject.com/en/2.2/ref/models/conditional-expressions/#conditional-aggregation

I think it would look something like this:
```python
Podcast.objects.filter(published=True)
        .prefetch_related(	        .prefetch_related(
            Prefetch("offered_by", queryset=LearningResourceOfferor.objects.all()),	            Prefetch("offered_by", queryset=LearningResourceOfferor.objects.all()),
            Prefetch("topics", queryset=CourseTopic.objects.all()),	            Prefetch("topics", queryset=CourseTopic.objects.all()),
        )	        )
        .annotate(
            episode_count=Count("episodes", filter=Q(is_published=True))
        )
```

noisecapella:
@gsidebo good catch. I needed to tweak your code a bit but it appears to work fine and it's much more concise

PR #2872, opened 2020-04-28
Audio player padding adjustments

gumaerc:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

None at the moment

#### What's this PR do?
Adds redux selectors to check if the audio player is currently loaded, then implements them on the classes that have outermost container UI elements; `App.js`, `Drawer.js` and `LearningResourceDrawer.js`.  When the audio player is active, a class is added to the outermost elements that adds bottom padding based on the value of the audio player height sass variables.  This would all be much simpler if CSS4 was released, because we could just use the `:has` selector to only apply a rule to an outer element that contains a `:not(.hidden)` `.audio-player-container-outer`.  This is the best way to do it right now, from what I can tell.

Go to /podcasts and click play on one of the episodes.  When the audio player loads, you should be able to scroll to the bottom of the page and the end of the content should line up with the top of the audio player.  You should be able to do the same in the drawer.

### DIFF
--- a/static/js/pages/App.js
+++ b/static/js/pages/App.js
@@ -54,7 +54,7 @@ import { setChannelData } from "../actions/channel"
 import { AUTH_REQUIRED_URL, SETTINGS_URL } from "../lib/url"
 import { isAnonAccessiblePath, needsAuthedSite } from "../lib/auth"
 import { isMobileWidth, preventDefaultAndInvoke } from "../lib/util"
-import { getOwnProfile } from "../lib/redux_selectors"
+import { getOwnProfile, isAudioPlayerLoaded } from "../lib/redux_selectors"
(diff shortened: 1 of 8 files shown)


static/js/pages/App.js
```diff
@@ -355,7 +361,8 @@ const mapStateToProps = state => {
     snackbar,
     banner,
     showUserMenu,
-    profile: getOwnProfile(state)
+    audioPlayerLoaded: isAudioPlayerLoaded(state),
```
aliceriot, line 364: you should be able to just use your `isAudioPlayerLoadedSelector` here (and in the `Drawer.js` too)

Issue #2845, opened 2020-04-24
Podcasts: +30 button causes playback to go to 0, then to +30 seconds


noisecapella:
### Steps to Reproduce

Click the +30 or -30 seconds button.

### Expected Behavior

It should jump to that time in the podcast

### Actual Behavior

It appears to flicker, showing it at 0:00 before starting at the expected time

### Screenshots
![plus30podcast](https://user-images.githubusercontent.com/863262/80236108-aac3e480-8628-11ea-8465-8e94670c6f07.gif)

PR #2891, opened 2020-04-30
Audio player skip forward / backward progress reset bug fix

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/open-discussions/issues/2845

#### What's this PR do?
Currently, the audio player is using Amplitude's `skipTo` function to skip to a specific point in the audio file.  Apparently, this causes Amplitude to re-load the audio file, causing it to temporarily reset its progress, detailed here:

https://github.com/521dimensions/amplitudejs/issues/384

Instead, I set these buttons to simply calculate the percentage played to set the audio to and use `Amplitude.setSongPlayedPercentage` like the seek bar is using because it doesn't cause the audio to reload.

Play a podcast, press the forward 30 / back 10 buttons and ensure that the progress does not blank out / reset to 0:00.

### DIFF
--- a/static/js/babelhook.js
+++ b/static/js/babelhook.js
@@ -53,12 +53,14 @@ global.Audio = class Audio {
     this._pauseStub()
   }
 
-  addEventListener() {
+  addEventListener(name, fn) {
     this._addEventListenerStub()
+    window.addEventListener(name, fn)
   }
 
-  removeEventListener() {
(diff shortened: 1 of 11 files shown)


gumaerc:
Had to do a stupid merge commit because I accidentally created this branch based off of your play/pause branch and not master...

PR #2982, opened 2020-06-04
fix favorite / not favorite display for 'similar items' in LR drawer

### LINKED ISSUE #2871: Bug: Favorite icon in similar items is not marked when it should be
### Steps to Reproduce

Assuming you have similar data in your database:
 - Go to `/learn/search`
 - Filter on courses and type "educational" in the box.
- The second item should be titled "Producing Educational Videos". Click the bookmark icon and add it to the favorites list. The bookmark should turn from a blue outlined white icon to a fully blue icon.
 - Clear the courses filter and type "physics demostration videos" in the search box.
 - Click the first item which should be "Physics Demostration Videos". A panel should open.
 - Look in the similar items in the panel. The first item should be "Producing Educational Videos" but it does not have a selected bookmark icon. If you click the icon you can see that it is marked as a favorite, but the icon does not show that.

### Expected Behavior

Item should be marked on both learning resources

### Actual Behavior

Item is marked on the search page for that item but not in the similar items of another item

Screenshot from 2020-04-28 16-20-42

Screenshot from 2020-04-28 16-21-10

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #2871

#### What's this PR do?

this PR fixes the issue with items in the 'similar items' panel in the LR drawer not showing up as favorited / added to a list when they have been.

I found an issue that was unexpected: although the individual APIs for all of our learning resources have an `is_favorite` boolean field on them, for some reason in the favorites API this field is not present on the objects in the response. So to fix it I made a change to the favorite item serializer to insert the field manually.

go to the course search and open up an item in the drawer. favorite one of the items in the 'similar items' panel. if you reload the page and open the drawer for that same item you should see the 'favorite' icon filled-in for the item you favorited (or added to a user list).

Another way to test is to look at <http://localhost:8063/api/v0/favorites/> and ensure that you see the `is_favorite` field in the responses.

Screenshot from 2020-06-04 12-35-43

### DIFF
--- a/course_catalog/serializers.py
+++ b/course_catalog/serializers.py
@@ -675,6 +675,12 @@ class FavoriteItemSerializer(serializers.ModelSerializer):
     content_data = GenericForeignKeyFieldSerializer(source="item")
     content_type = serializers.CharField(source="content_type.name")
 
+    def to_representation(self, instance):
+        """put `is_favorite` in to the content data"""
+        data = super().to_representation(instance)
+        data["content_data"]["is_favorite"] = True
+        return data
+
     class Meta:
         model = FavoriteItem
         fields = "__all__"
--- a/course_catalog/serializers_test.py
+++ b/course_catalog/serializers_test.py
@@ -17,6 +17,7 @@ from course_catalog.factories import (
     PodcastFactory,
     PodcastEpisodeFactory,
     LearningResourceOfferorFactory,
+    VideoFactory,
 )
 from course_catalog.models import FavoriteItem, UserListItem
 from course_catalog.serializers import (
@@ -29,6 +30,7 @@ from course_catalog.serializers import (
     CourseTopicSerializer,
     PodcastSerializer,
     PodcastEpisodeSerializer,
+    VideoSerializer,
 )
 from open_discussions.factories import UserFactory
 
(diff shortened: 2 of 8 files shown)


aliceriot:
ahh I forgot that this icon is supposed to be filled in if the thing is in a list too. I'll fix that

aliceriot:
yeah, after thinking about that a little more I'm not sure there's a good solution to that issue, other than trying to figure out how to represent this data in elasticsearch.

this is because the similar items comes from elasticsearch. right now to show the favorited items as filled-in here we are using the favorites API to fetch the list of all items that hte user has favorited. then it's pretty easy to check, for each item as we're putting it into the list of similar items, whether we already have an item saved in the redux store or not. this works well enough for the favorites list.

but we can't apply the same principle to the user list, because in order to make that work we would have to not only fetch data _about_ the lists a user has created but also fetch all of their contents. I can't remember where but iirc @noisecapella made a change to have the user lists API not return the items in a list to avoid some performance problems that were cropping up when it worked that way.

so I'm not sure how to proceed here. it seems like the best solution is to find some way to mark things as favorited or not in the similar items thing from elastic search, but I don't know enough about elasticsearch to know how difficult that is.

mbertrand:
I think you might be able to modify `search.api.find_similar_resources` to include some of the same code in `search.api.transform_results` that adds in the favorite/list info to each search result returned by ES: https://github.com/mitodl/open-discussions/blob/3005018737fda124f306d8668633cbc4743ee6d3/search/api.py#L397-L412

aliceriot:
@mbertrand ok I have made those changes in the search API methods and fixed tests, so I think this should be ready for review again (re-review?)

mbertrand:
One of the travis tests failed but I think it was just a timeout fluke

PR #2988, opened 2020-06-10
Restrict creation of public lists

### LINKED ISSUE #2824: Update Search - Restrict Access to Creating Learning Lists
As a user of Open Learning, I want to ensure Learning Lists are high quality.

**Designs and Mockups**
N/A

**Acceptance Criteria:**

* [ ] Turn off the ability for anyone to create a Learning List
* [ ] Create / Update functionality for creating Learning Lists and ensure it's only available to 'Trusted People'. 

**Out of Scope**
NA

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
- [x] Settings

Closes #2824 

#### What's this PR do?
- Adds `STAFF_MOIRA_LISTS` .env setting (moira lists of users who can create public learning paths/lists).
- Adds a validation function to `UserListSerializer` to ensure that only users in the above moira lists (or django staff/super users) are allowed to create public learning lists.
- On the front end UserListFormDialog, hides the privacy options and sets the default to 'private' if the user is not on the moira lists.

- Set `STAFF_MOIRA_LISTS` to include at least one moira list and create a user whose MIT email is on that list (and not a staff or superuser).
- Log in as that user and verify that you can create public lists and see the privacy options on the form.
- Log in as a staff or superuser and you should still be able to create public lists.
- Log in as a user who is not on the specified moira lists, not staff, and not a superuser.  You should still be able to create learning lists/paths, but the privacy options should be hidden and the lists you create should be private.

### DIFF
--- a/course_catalog/serializers.py
+++ b/course_catalog/serializers.py
@@ -8,7 +8,7 @@ from django.db.models import Max, F, prefetch_related_objects
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
 
-from course_catalog.constants import PlatformType, ResourceType, ListType
+from course_catalog.constants import PlatformType, ResourceType, ListType, PrivacyLevel
 from course_catalog.models import (
     Course,
     CourseInstructor,
@@ -30,6 +30,7 @@ from course_catalog.utils import (
     get_course_url,
     semester_year_to_date,
 )
+from moira_lists.moira_api import is_list_staff
 from open_discussions.serializers import WriteableSerializerMethodField
 from search.task_helpers import upsert_user_list, delete_user_list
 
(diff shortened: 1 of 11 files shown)


course_catalog/views_test.py
```diff
@@ -215,21 +216,41 @@ def test_user_list_endpoint_get(client, is_public, is_author, user):
         assert resp.data.get("title") == another_user_list.title
 
 
+@pytest.mark.parametrize("is_public", [True, False])
+@pytest.mark.parametrize("is_staff", [True, False])
+@pytest.mark.parametrize("is_super", [True, False])
+@pytest.mark.parametrize("on_moira", [True, False])
 @pytest.mark.parametrize("is_anonymous", [True, False])
-def test_user_list_endpoint_create(client, is_anonymous, mock_user_list_index):
+def test_user_list_endpoint_create(  # pylint: disable=too-many-arguments
```
rhysyngsun, line 224: Should expand the test for `PATCH` to also cover these conditions.

course_catalog/serializers.py
```diff
@@ -568,6 +569,18 @@ def validate_list_type(self, list_type):
...
+    def validate_privacy_level(self, privacy_level):
+        """
+        Validator for privacy_level
+        """
+        request = self.context.get("request")
+        if request and hasattr(request, "user") and isinstance(request.user, User):
+            if privacy_level == PrivacyLevel.public.value and not is_list_staff(
+                request.user
+            ):
+                raise ValidationError("Invalid permissions for creating public lists")
```
rhysyngsun, line 581: This error could be hit when updating a user list, could you make the error message more general to both cases?

course_catalog/views_test.py
mbertrand: Just added an `is_staff` condition in the following test.

mbertrand:
@rhysyngsun all set

Issue #2984, opened 2020-06-05
Search: Invalid Dates on Micromasters and MITx Courses


odlbot:
As a potential Micromasters or MITx enroller, I want to know when courses begin. Many of the Micromasters and MITx cards in Open Search are showing Invalid Date for the start dates.

**Designs and Mockups**
Attached

**Acceptance Criteria:**

* [ ] Review the comments
* [ ] Implement the best solution

**Out of Scope**

**Dependencies**

┆Attachments: <a href="https://trello-attachments.s3.amazonaws.com/5dfd295df7c63e06791c4cae/5eda4e1871daf308c3783c31/b5db112966ecb80e74acba64ece1fbe3/Micromasters.png">Micromasters.png</a> | <a href="https://trello-attachments.s3.amazonaws.com/5dfd295df7c63e06791c4cae/5eda4e1871daf308c3783c31/1e0e79d1b4ad45c62020efa9811313fb/JSON.png">JSON.png</a>

odlbot:
➤ Brian Grossman commented:

From Slack: https://mitodl.slack.com/archives/G5RHT8GDD/p1591298625048900

Brian Grossman  3:23 PM
There’s a number of ‘Invalid dates’ coming up on MicroMasters courses in Open Search. Anyone know why?
(Attached screenshot)

nlevesque  4:05 PM
I'm seeing a bunch of future runs for those courses with almost all null values in their fields

Anastasia Beglova  4:07 PM
There should probably be a ui change not to display that even if the date is missing

nlevesque  4:07 PM
(Attached JSON image)

yeah, I think you're right, it shouldn't be selecting that
although really, we shouldn't even be indexing a course run like that
a lot of those fields should be a bare minimum to consider the run for ingestion (title, most of the date fields)
it's just a stub course atm

mbertrand:
Works great, just 1 question. When there is only one run and it has no dates, no date will appear in the search result card.  At the top of the course's drawer the start date will be "Ongoing".  Should they be consistent with each other (both show either nothing or "Ongoing") or is it okay as is?

<img width="811" alt="Screen Shot 2020-06-19 at 1 51 05 PM" src="https://user-images.githubusercontent.com/187676/85165562-04651b80-b234-11ea-87b4-b9fd4152b1ea.png">

[...]

PR #2997, opened 2020-06-18
do not select run with missing dates as best run

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/open-discussions/issues/2984

#### What's this PR do?
There is currently a bug in the code that preferentially selects runs with no start or end date as the best run (due to how we set default dates). This fixes it.

If there is currently a course with "Invalid date" in your local app, this should fix it. Otherwise create one by picking a course with a certificate then running

```
from search import task_helpers

run = course.runs.last()
run.best_start_date = None
run.best_end_date = None
run.save()
task_helpers.upsert_course(course.id)
```

The course should have "Invalid date" in the card in the master branch but not this branch

### DIFF
--- a/static/js/lib/constants.js
+++ b/static/js/lib/constants.js
@@ -119,8 +119,6 @@ export const readableLearningResources = {
 }
 
 export const DATE_FORMAT = "YYYY-MM-DD[T]HH:mm:ss[Z]"
-export const DEFAULT_START_DT = "1970-01-01T00:00:00Z"
-export const DEFAULT_END_DT = "2500-01-01T23:59:59Z"
 export const DISPLAY_DATE_FORMAT = "MMMM D, YYYY"
 
 export const PHONE = "PHONE"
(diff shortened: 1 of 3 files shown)


mbertrand:
Works great, just 1 question. When there is only one run and it has no dates, no date will appear in the search result card.  At the top of the course's drawer the start date will be "Ongoing".  Should they be consistent with each other (both show either nothing or "Ongoing") or is it okay as is?

Screen Shot 2020-06-19 at 1 51 05 PM

Screen Shot 2020-06-19 at 1 51 24 PM

abeglova:
@Ferdi @pdpinch @briangrossman Do we want to list "Ongoing" if the only run has no start or end dates? Are there runs with no start and end date that are not just bad data? I think we should remove "Ongoing" from the run selection dropdown (and probably remove that dropdown entirely, at least for MITx)

abeglova:
Just talked to Ferdi on slack and he said this is ok to merge as is now while we decide what we want to do about runs with no start dates

PR #3005, opened 2020-06-19
limit to one recent episode per podcast

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3004

#### What's this PR do?
This limits the podcast episodes in the recent list on the /podcasts page to one episode per podcast

Check if you have multiple episodes from the same podcast in the recent episodes list on the /podcasts page when you run the app with the master branch.  If not, update the `last_modified` date on a podcast episode so that you do have multiple episodes from the same podcast in the list. 

Run the code in this branch. You should no longer have more than one episode per podcast in that list.

### DIFF
--- a/course_catalog/views.py
+++ b/course_catalog/views.py
@@ -2,6 +2,8 @@
 course_catalog views
 """
 import logging
+import operator
+
 from hmac import compare_digest
 
 import rapidjson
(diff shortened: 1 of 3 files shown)


course_catalog/views.py
```diff
@@ -441,16 +460,31 @@ def get_queryset(self):
...
-                Prefetch("topics", queryset=CourseTopic.objects.all()),
-            )
-            .annotate_is_favorite_for_user(user)
-            .prefetch_list_items_for_user(user)
-            .select_related("podcast")
+        return shared_podcast_episode_query(user).order_by("-last_modified", "-id")
+
+
+class RecentPodcastEpisodesViewSet(viewsets.ReadOnlyModelViewSet, FavoriteViewMixin):
+    """
```
mbertrand, line 467: This view could inherit from PodcastEpisodesViewSet, which would slightly reduce the amount of code:

```python
class RecentPodcastEpisodesViewSet(PodcastEpisodesViewSet):
    """
    Viewset for most recent PodcastEpisode for each Podcast
    """

    def get_queryset(self):
        user = self.request.user

        latest_episode_by_podcast = (
            shared_podcast_episode_query(user)
            .order_by("podcast_id", "-last_modified", "-id")
            .distinct("podcast_id")
        )

        return sorted(
            latest_episode_by_podcast,
            key=operator.attrgetter("last_modified"),
            reverse=True,
        )
```

course_catalog/views_test.py
```diff
@@ -725,31 +726,50 @@ def test_recent_podcast_episodes_no_feature_flag(settings, client):
...
+        )
+
         resp = client.get(reverse(basename))
         assert resp.status_code == status.HTTP_200_OK
-        assert resp.json()["count"] == 5
+        assert resp.json()["count"] == count
 
-        assert resp.json()["results"] == [
-            {"is_favorite": False, **episode}
+        expected_results = [
```
mbertrand, line 766: Once this test failed on me because the topics were in different orders.  Could modify the serializer to always return topics in a sorted order, or just pop out the topics from the JSON before making the comparison.

abeglova: I think I just need to add sort_keys=True to `dumps(`. That sorts nested keys as well

PR #3015, opened 2020-06-26
ensure audio player is present on all `/learn` pages

### LINKED ISSUE #3014: no audio player UI on `/learn`
you can play the audio from the drawer if you're looking at a podcast, but the audio player component isn't present

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #3014

#### What's this PR do?

makes sure that on all `/learn` pages we have the audio player component, so that you can play podcast episodes from all the places they are accessible.

Go to the search page and look for podcasts, ensuring you can play an episde.
do the same on the course home page (`/learn`) and on the user list pages too.

Screenshot from 2020-06-26 13-58-24


aliceriot:
@abeglova there was some conversation on the issue (#3014) about exactly what we want to do - I ended up making those changes here.

basically,

- the audio player is now present site-wide, so if you start listening to an episode on the podcasts page and then navigate to the home page the player UI will stay present
- the audio player also now has a little 'x' button that lets you close it and stop the audio.

aliceriot:
which is to say, I think it needs another look now :)

PR #3021, opened 2020-06-29
fix podcast etl

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/open-discussions/issues/3019

#### What's this PR do?
There is a bug in the loader so elasticsearch is not updated when an episode is marked as no longer published.  Also, we generate episode ids based on the url, which can change instead of the guid which is not supposed to change. This fixes both those issues.

Run `docker-compose run web ./manage.py backpopulate_podcast_data`. Verify that there are not duplicate episodes in /learn/search

### DIFF
--- a/course_catalog/etl/loaders.py
+++ b/course_catalog/etl/loaders.py
@@ -749,9 +749,14 @@ def load_podcast(podcast_data, *, config=PodcastLoaderConfig()):
         episode = load_podcast_episode(episode_data, podcast, config=config.episodes)
         episode_ids.append(episode.id)
 
-    PodcastEpisode.objects.filter(podcast=podcast).exclude(id__in=episode_ids).update(
-        published=False
+    unpublished_episodes = PodcastEpisode.objects.filter(podcast=podcast).exclude(
+        id__in=episode_ids
     )
+
+    for episode in unpublished_episodes:
+        search_task_helpers.delete_podcast_episode(episode)
+
+    unpublished_episodes.update(published=False)
 
     if not created and not podcast.published:
         search_task_helpers.delete_podcast(podcast)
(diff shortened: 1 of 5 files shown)


course_catalog/etl/loaders.py
```diff
@@ -749,10 +749,16 @@ def load_podcast(podcast_data, *, config=PodcastLoaderConfig()):
         episode = load_podcast_episode(episode_data, podcast, config=config.episodes)
         episode_ids.append(episode.id)
 
-    PodcastEpisode.objects.filter(podcast=podcast).exclude(id__in=episode_ids).update(
-        published=False
+    unpublished_episodes = PodcastEpisode.objects.filter(podcast=podcast).exclude(
+        id__in=episode_ids
     )
 
+    if unpublished_episodes:
```
mbertrand, line 756: I think you can get rid of this `if` statement

course_catalog/etl/podcast.py
```diff
@@ -127,7 +127,7 @@ def transform_episode(rss_data, offered_by, topics, parent_image):
         dict:
             normalized podcast episode data
     """
-    episode_id = generate_unique_id(rss_data.enclosure["url"])
+    episode_id = generate_unique_id(rss_data.guid.text)
```
mbertrand, line 130: This change seems to create many duplicate episodes in the database, one with the old id and one with the new id (episode count after running backpopulate on master: 1297; after running again on this branch: 2413).  But the older ones get unpublished so it's not apparent to the end user.  My initial feeling was that the episodes with old id's should be deleted before importing again, but on second thought they don't seem to be doing any obvious harm.  What do you think?

abeglova: As you said, those unpublished episodes will be ignored. I think it makes sense to manually delete all unpublished episodes from the database after this is merged just to avoid clutter in the database

abeglova:
Thanks @mbertrand! Back to you

PR #3051, opened 2020-07-14
Block certain emails during registration

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
- Adds a `BlockedEmailRegex` model
- Does not send a registration confirmation if the email address matches any `BlockedEmailRegex` objects.

- From django admin, create a new `BlockedEmailRegex` object with a `match` value that will match the email that you intend to register as.  You can look at the new test parameters in this PR for examples.
- Register with that email.  You should not get a confirmation email.
- Register with a different email that does not match.  You should get a confirmation email.

Screen Shot 2020-07-14 at 4 22 07 PM


mail/admin.py
```diff
@@ -0,0 +1,18 @@
...
+
+from .models import BlockedEmailRegex
+
+
+class BlockedEmailRegexAdmin(admin.ModelAdmin):
+    """Admin for BlockedEmailRegex"""
+
+    model = BlockedEmailRegex
+    list_display = ["match", "created_on", "updated_on"]
+    search_fields = ["match"]
```
pdpinch, line 15: Can we add hint text in the admin with example values? I'm not sure whether I would use "*@example.com" or "\w+@example.com" to ban a whole domain.

mail/models.py
```diff
@@ -0,0 +1,12 @@
+""" Models for mail app """
+from django.db import models
+
+from open_discussions.models import TimestampedModel
+
+
+class BlockedEmailRegex(TimestampedModel):
```
rhysyngsun, line 7: I'm in the process of extracting the `authentication` app out into a reusable django app, and I think this model should live in the `authentication` app instead so I can include it in that effort and all our apps can benefit. I'd probably also have the check for the blocked email happen in a pipeline action, but I'm ok with it living in ` mail/verification_api.py` for the sake of getting this out sooner. Moving the model later will be more difficult though.

mail/admin.py
mbertrand: How's this for a few examples?

Screen Shot 2020-07-14 at 4 45 41 PM

Issue #748, opened 2018-05-31
Can't comment on posts older than 6 months


rhysyngsun:
We just saw this for the first time: https://sentry.io/mit-office-of-digital-learning/discussions/issues/568931822/?environment=production

See: https://www.reddit.com/r/changelog/comments/25kvjo/reddit_change_the_logic_for_archiving_posts_has/

It looks like we might be able to increase that value with a simple configuration tweak, but we should verify that: https://github.com/mitodl/reddit/blob/3b01331785c846e6c4a24ba2a8b70321b803310d/r2/r2/lib/app_globals.py#L332

At a minimum we should fix this so it doesn't raise a 500.

rhysyngsun:
Sentry issue: [DISCUSSIONS-6P](https://sentry.io/mit-office-of-digital-learning/discussions/issues/568931822/)

aliceriot:
the current user experience here is this when trying to make a comment:

![Screenshot from 2020-07-16 10-00-03](https://user-images.githubusercontent.com/6207644/87680226-25456180-c74b-11ea-9053-04b6db2a53bb.png)

but this is what happens when you try to *upvote* a comment:

![Screenshot from 2020-07-16 10-00-49](https://user-images.githubusercontent.com/6207644/87680296-3db57c00-c74b-11ea-8219-1016b549d32b.png)

we should probably at the least have the same 'sorry we couldn't do that' behavior for voting that we do for commenting on these old posts

pdpinch:
👍

> we should probably at the least have the same 'sorry we couldn't do that' behavior for voting that we do for commenting on these old posts

alicewriteswrongs:
unassigning myself - this probably still theoretically needs some attention though!

PR #3057, opened 2020-07-16
fix a few issues with comment voting

### LINKED ISSUE #221: show spinner when updating score
We've noticed that changing a vote can be slow to update. 

As a stopgap measure we can show a spinner while the score is being updated.

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #221 (from 2017!)
partially addresses #748

#### What's this PR do?

this PR fixes a few issues with comment voting. namely:

- now instead of failing and blowing up the UI (see #748) when a comment vote fails 
- I also fixed the optimistic update behavior for the comment vote to immediately reflect what the score on the comment will be *after* the user's vote is applied.

if you have a comment on your local reddit that is older than 6 months you can use that to test this, else you can apply the following patch:

```patch
diff --git a/channels/views/comments.py b/channels/views/comments.py
index be8e5af0..d1fbfe43 100644
--- a/channels/views/comments.py
+++ b/channels/views/comments.py
@@ -209,6 +209,7 @@ class CommentDetailView(APIView):
 
     def patch(self, request, *args, **kwargs):  # pylint: disable=unused-argument
         """Update a comment"""
+        return Response(status=status.HTTP_500_INTERNAL_SERVER_ERROR)
         with translate_praw_exceptions(request.user):
             comment = self.get_object()
             subscriptions = lookup_subscriptions_for_comments(
```

then try upvoting and downvoting comments to try out the new user experience.

also play around with upvoting and downvoting various comments *without* the 500 error to make sure the optimistic score update works as it should.

also try upvoting and downvoting comments on the profile page, that should work as expected.

this is what used to happen when a comment vote failed:

Screenshot from 2020-07-16 12-32-49

this is what happens now:

Screenshot from 2020-07-16 12-33-47

so the user at least doesn't see a total error state.

### DIFF
--- a/static/js/components/CommentVoteForm.js
+++ b/static/js/components/CommentVoteForm.js
@@ -1,4 +1,5 @@
 // @flow
+/* global SETTINGS:false */
 import React, { useState, useCallback } from "react"
 import { useDispatch } from "react-redux"
 
@@ -6,6 +7,7 @@ import LoginTooltip from "./LoginTooltip"
 
 import { actions } from "../actions"
 import { userIsAnonymous } from "../lib/util"
+import { setBannerMessage } from "../actions/ui"
 
 import type { CommentInTree } from "../flow/discussionTypes"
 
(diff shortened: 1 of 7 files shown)


static/js/components/CommentVoteForm.js
```diff
@@ -55,9 +80,38 @@ export default function CommentVoteForm(props: Props) {
   const downvoted = comment.downvoted !== downvoting && !upvoting
   const upvoted = comment.upvoted !== upvoting && !downvoting
 
+  // here we create an optimistic score based on assuming that the action in
+  // progress (if any) will succeed
+  let { score } = comment
```
abeglova, line 85: Can you add a check for whether the comment author is the user? Upvoting/downvoting your own comment does not affect the score. When the user votes on their own comment this causes an incorrect score to flash before reverting to the previous score

static/js/components/CommentVoteForm.js
```diff
@@ -20,32 +22,55 @@ export default function CommentVoteForm(props: Props) {
   const [upvoting, setUpvoting] = useState(false)
   const [downvoting, setDownvoting] = useState(false)
 
+  const showErrorBanner = useCallback(
```
abeglova, line 25: Should we display an error banner when upvoting posts fails as well?

aliceriot: probably - but I think I'm going to tackle that in a separate PR. I'm sort of just going through all this older code in open when I have time and trying to fix it up :)

static/js/components/CommentVoteForm.js
aliceriot: yeah, that's a good idea, I'll add that in

aliceriot:
@abeglova I think everything should be updated now

Issue #3046, opened 2020-07-10
Podcasts: Episode Read More/Less Link Appears When Not Needed (and Remembers State)


briangrossman:
### Steps to Reproduce
Note: This is a sneaky one...
1. Go to the Podcasts home page: [`https://open.mit.edu/podcasts/`](https://open.mit.edu/podcasts/)
2. Click on the `MIT Bootcamps` podcast (you'll need to scroll a little)
3. Click on the oldest episode in the drawer (`Deep Dive: Q&A Session 7 Working with Universities`)
4. ***Note*** that there is (appropriately) no `Read More` link because the description is short enough to be displayed
5. Click the arrow on the top-left of the drawer to return to the list of episodes for the drawer
6. Click on the 2nd oldest episode in the drawer (`Introduction to Deep Technology`)
7. ***Note*** that this episode requires the `Read More` link. Click it to view the full episode description.
8. Click the arrow on the top-left of the drawer to return to the list of episodes
9. Click on the oldest episode in the drawer (`Deep Dive: Q&A Session 7 Working with Universities`) again
10. Two things to ***note***:
   - The `Read Less` link is now present. It doesn't need to be because the description is so short. 
   - The link remembered its state. It says `Read Less`. In fact, if you go to another episode that actually requires the link, it will be in the fully expanded, `Read Less` state. It should default to `Read More` when you view new episodes.

### Screenshot or Screencast

![Screen Shot 2020-07-10 at 1 43 47 PM](https://user-images.githubusercontent.com/21342568/87184245-ab7a2780-c2b5-11ea-8d76-4ec0beb6c1a4.png)

aliceriot:
that is a weird bug - I think I have an idea of what's going on, will try to dig into it

PR #3069, opened 2020-07-23
fix read more

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3046

#### What's this PR do?
Fixes the bug with the read more/less control.

Follow the instructions in the ticket to verify that there is a bug with the read more control in the master branch. Follow the instructions again in this branch to verify that the bug has been fixed

### DIFF
--- a/static/js/components/TruncatedText.js
+++ b/static/js/components/TruncatedText.js
@@ -31,9 +31,16 @@ export default function TruncatedText(props: Props) {
   const [expanded, setExpanded] = useState(false)
   const [hasOverflow, setHasOverflow] = useState(false)
   const [, forceReRender] = useState(0)
+  const [currentText, setCurrentText] = useState(text)
   const dotRef = useRef(null)
 
   useLayoutEffect(() => {
+    if (text !== currentText) {
+      setCurrentText(text)
+      setExpanded(false)
+      setHasOverflow(false)
+    }
+
     if (dotRef.current) {
       // the .container on the Dotdotdot holds the ref for the div it uses to
       // wrap the text
--- a/static/js/components/TruncatedText_test.js
+++ b/static/js/components/TruncatedText_test.js
@@ -96,4 +96,21 @@ describe("TruncatedText", () => {
     })
     assert.isNotOk(wrapper.find(".tt-expansion-control").exists())
   })
+
+  it("should reset if the text changes", () => {
+    const wrapper = render({
+      text:                  "text",
+      lines:                 1,
+      showExpansionControls: true
+    })
+    wrapper.find(".tt-expansion-control").simulate("click")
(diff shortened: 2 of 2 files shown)


static/js/components/TruncatedText.js
```diff
@@ -31,9 +31,16 @@ export default function TruncatedText(props: Props) {
...
   const [, forceReRender] = useState(0)
+  const [currentText, setCurrentText] = useState(text)
   const dotRef = useRef(null)
 
   useLayoutEffect(() => {
+    if (text !== currentText) {
+      setCurrentText(text)
+      setExpanded(false)
+      setHasOverflow(false)
+    }
```
aliceriot, line 42: I think here if you wanted to you could just do this:

```js
useEffect(() => {
  setExpanded(false)
} [ text ])
```

which will undo the expansion whenever the `text` prop changes. then below in the `useLayoutEffect` we could just always set `hasOverflow` to the value of `container.scrollHeight > container.offsetHeight`

abeglova: I change the code to use useEffect() as you suggested.  The reset for hasOverflow is needed since dotdotdot is replaced by a plain div when the text is expanded

aliceriot: right - I was just suggesting that in the existing `useLayoutEffect` callback we could change this line:

```js
if (container.scrollHeight > container.offsetHeight) {
    // scrollHeight is the height of the rendered text, while offsetHeight
    // is the height of the portion which is shown
    setHasOverflow(true)
}
```

to simply

```js
setHasOverflow(container.scrollHeight > container.offsetHeight)
```

aliceriot: sorry my suggestion wasn't clear earlier :)

aliceriot: although just resetting `hasOverflow(false)` whenever the `text` prop changes accomplishes basically the same thing - so if you want to stick with that I think that's fine

abeglova: That does not work because when the text is expanded `dotRef.current` is null (since there is no dotdotdot component` so there is no container

abeglova: Also I reverted the last commit. We don't want a separate useEffect() hook for the text because we want the code in useLayoutEffect() to run too when the text changes

aliceriot: it does - when you call `useLayoutEffect` or `useEffect` and only pass the first argument (a function) it runs on every re-render

aliceriot: if you want to merge this as-is it's fine with me - I just thought that if we could avoid the need for the currentText state we might want to, but it's nbd

PR #3081, opened 2020-07-28
Added server-side 404 page for posts

### LINKED ISSUE #2785: Removed Posts should not be visible for non-moderator users
### Steps to Reproduce
Ideally, when items are removed, they should no longer be visible for anonymous and non-moderator users. Currently, while no longer in the channel and front pages, removed items can still be accessed directly from the URL. 

Steps:

1. Go to URL of item to remove. Copy the URL.
1. Remove the item.
1. Log out of Open or open another browser that isn't logged in.
1. Paste URL of removed item and access that page.
1. You can view the removed post.

### Expected Behavior
Ideally there'd be a 404 for non-moderator and anonymous users when they try to view removed items.

### Actual Behavior
Everyone can still view the removed items directly with the URL, even after removed.

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #2785 

#### What's this PR do?
This adds a server-side 404 page for removed posts for non-moderator users, to ensure that for search engines and anyone else don't see this content perpetually. Also covered 400/403 errors just in case since they were a quick win.

- Remove a post in a public and/or private channel
- You should still be able to view the post as a moderator
- As a logged in user you should get a 404 page
- As an anonymous user you should get a 404 page for the public post

Intentionally not handled here:

- 500 error pages: not in scope and a bit of effort
- Anonymous users trying to view a removed private post

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -5,7 +5,7 @@ services:
       - .:/src
       - django_media:/var/media
     environment:
-      DEBUG: 'True'
+      DEBUG: ${DEBUG:-True}
       NODE_ENV: 'development'
       OPEN_DISCUSSIONS_USE_WEBPACK_DEV_SERVER: 'True'
 
@@ -14,7 +14,7 @@ services:
       - .:/src
       - django_media:/var/media
     environment:
-      DEBUG: 'True'
+      DEBUG: ${DEBUG:-True}
       NODE_ENV: 'development'
       OPEN_DISCUSSIONS_USE_WEBPACK_DEV_SERVER: 'True'
 
--- a/open_discussions/templates/base.html
+++ b/open_discussions/templates/base.html
@@ -10,12 +10,7 @@
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons" />
     <link rel="stylesheet" type="text/css" href="{% static 'hijack/hijack-styles.css' %}" />
     <link rel="icon" href="{% static 'images/favicon.ico' %}" />
-    {% if js_settings_json %}
-    <script type="text/javascript">
-    var SETTINGS = {{ js_settings_json|safe }};
-    </script>
-    {% render_bundle 'style' %}
-    {% endif %}
+
     <title>{% block title %}{% trans "Open Discussions" %}{% endblock %}</title>
     <meta name="description" content="{% block description %}{% endblock %}">
     <meta name="keywords" content="{% block keywords %}{% endblock %}">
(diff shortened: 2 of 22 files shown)


pdpinch:
What effect will this have on the profile page's list of posts?

rhysyngsun:
@pdpinch that wasn't addressed here, but the links to the posts will resolve to the 404

open_discussions/views.py
```diff
@@ -102,8 +108,50 @@ def index(request, **kwargs):  # pylint: disable=unused-argument
...
+    return _render_app(request)
+
+
+def channel_post(request, **kwargs):
+    """Render a channel post's page as long as it isn't removed"""
+    post_id = kwargs.get("post_id", None)
+    if not post_id:
+        raise Http404("No post specified")
+
+    post = get_object_or_404(Post, post_id=post_id)
```
mbertrand, line 125: I noticed that the author of a removed post now gets a 404, but a removed comment is still reachable to the author via direct link.  Not sure if that really matters or not, all other non-admin users get a 404.

open_discussions/templates/base.html
```diff
@@ -10,15 +10,11 @@
...
-    <script type="text/javascript">
-    var SETTINGS = {{ js_settings_json|safe }};
-    </script>
-    {% render_bundle 'style' %}
-    {% endif %}
+
     <title>{% block title %}{% trans "Open Discussions" %}{% endblock %}</title>
     <meta name="description" content="{% block description %}{% endblock %}">
     <meta name="keywords" content="{% block keywords %}{% endblock %}">
+    {% render_bundle 'style' %}
```
mbertrand, line 17: On this branch but not master I'm seeing this error in the console:

Screen Shot 2020-07-30 at 10 13 45 AM

open_discussions/views.py
mbertrand: To clarify, a removed comment on a non-removed post

docker-compose.override.yml
```diff
@@ -5,7 +5,6 @@ services:
       - .:/src
       - django_media:/var/media
     environment:
-      DEBUG: 'True'
```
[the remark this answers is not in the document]
rhysyngsun: good catch, i made that change to test locally, but i like your fix

open_discussions/views.py
rhysyngsun: Hm, are you ok with me opening a separate issue for that since it'll involve a bunch more test reworking?

mbertrand: Fine with me!

PR #3082, opened 2020-07-28
Blocked IP model and middleware

### LINKED ISSUE #3070: Add ip blocklist for incoming requests
This should be configurable via a data model, my initial idea is to have a starting ip and an ending ip so we can block a whole IP range if necessary (blocking a single ip would in theory involve just using the same ip for both values).

- [ ] Data model for block IP ranges
- [ ] Middleware to only allow readonly actions for blocked IPs
- [ ] Avoid adding non-routable IPs into the configuration (e.g. 168.0.0.0/24 or 10.0.0.0/24 blocks)

See also: we currently use https://github.com/un33k/django-ipware for ip detection, that is the preferred library as opposed to inspecting headers manually.

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #3070 

#### What's this PR do?
- Adds a `BlockedIPRange` model
- Adds middleware to block POST/PUT/PATCH requests if coming from a user IP in a blocked range (if the IP is routable and the user is not a superuser).
- Adds a `blocked_ip_exempt` decorator so elasticsearch requests are not affected.

- As a superuser, add a new `BlockedIPRange` object that includes your current IP (likely 10.something-something). You can also try entering some invalid values (null, non-ip values, end < start) and verify that the validation works.
- You should still be able to do everything (including create posts/comments).
- In an incognito browser, try logging in as a non-superuser and registering as a new user.  You should not be able to.
- Try searching for courses, podcasts, etc.  You should still be able to do that as an anonymous user.

Screen Shot 2020-07-28 at 4 36 29 PM

Screen Shot 2020-07-28 at 4 36 13 PM


authentication/middleware.py
```diff
@@ -33,3 +40,35 @@ def process_exception(self, request, exception):
...
+
+            if user_ip is None or (
+                is_routable
+                and request.method not in SAFE_METHODS
+                and BlockedIPRange.objects.filter(
+                    Q(ip_start__lte=user_ip) & Q(ip_end__gte=user_ip)
+                ).count()
+                > 0
+            ):
+                raise PermissionDenied
```
rhysyngsun, line 74: This should `return django.http.HttpResponseForbidden()` instead, because django isn't aware of `rest_framework` exceptions so this turns into a server 500 instead.

PR #3122, opened 2020-08-24
Fix flaky test

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
Fixes a flaky test due to list items which were out of order

### DIFF
--- a/open_discussions/test_utils.py
+++ b/open_discussions/test_utils.py
@@ -77,7 +77,23 @@ def drf_datetime(dt):
     return dt.isoformat().replace("+00:00", "Z")
 
 
-def assert_json_equal(obj1, obj2):
+def _sort_values(dict_like):
+    """
+    Sort each value of a dictionary if applicable to the value
+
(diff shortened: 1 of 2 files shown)


aliceriot:
looking at the test output, it looks like some tests can fail if the values within a list in a dict cannot be sorted with `sorted` - in particular it looks like it doesn't work when we have a dict with a list of dicts. I confirmed in a python shell, if I call

```py
sorted([ { "foo": "bar" } , { "biz": "baz" } ])
```

I get the same error. Is there a way we can provide an implementation of the `<` operator for dicts? tbh I forget how this kind of thing works in Python - in Rust or something we could just implement some sort of 'comparable' trait. Maybe it could sort based on the first value in a sorted list of the keys? like

```
sorted(dict.keys())[0]
```

idk

noisecapella:
I defined a key function for sorting dicts. What do you think?

PR #3125, opened 2020-08-26
upgrade eslint config, fix errors

aliceriot:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

this just updates to the latest version of our eslint configuration, which adds the react-hooks preset. this allows the linter to find a lot of issues where the logic of hooks is being violated - it found a bunch of callbacks which did not have their dependencies properly specified, for instance, which could lead to bugs related to stale data inside the closure of the callback.

there aren't really any code changes - the only things are renaming some things so that the linter knows they are hooks and turning off rules in a few, specific places. so I think just read through it and make sure things look good, and possibly do a quick smoke test of the app to make sure I didn't break anything.

### DIFF
--- a/static/js/hooks/learning_resources.js
+++ b/static/js/hooks/learning_resources.js
@@ -35,6 +35,9 @@ export function useLearningResourcePermalink() {
         })
       )
     }
+    // dependencies intentionall blank here
+    // this effect should only run on component mount
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 }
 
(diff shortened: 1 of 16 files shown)


static/js/hooks/learning_resources.js
```diff
@@ -35,6 +35,9 @@ export function useLearningResourcePermalink() {
         })
       )
     }
+    // dependencies intentionall blank here
```
noisecapella, line 38: intentionally

aliceriot: lol

PR #3128, opened 2020-08-27
Fixed password reset

### LINKED ISSUE #3095: Password reset flow works but appear to fail to the user
- On the password reset form, when you submit it display an error due to a 204 response, even though you still get the reset email.
- On clicking the link, the form submits and changes the password but nothing happens form the user perspective

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #3095 

#### What's this PR do?
Fixes the password reset views so the password reset flow works again

Run through a full password reset workflow (fill out the form, click the link in the resulting email, and use the password change form)

Sooo due to a naming issue, our two Djoser test cases were not being run. It looks like something changed in the Djoser internals which broke our functionality that overrides it, and this wasn't caught because of said test cases being unrecognized 😬

### DIFF
--- a/open_discussions/test_utils.py
+++ b/open_discussions/test_utils.py
@@ -5,6 +5,8 @@ from contextlib import contextmanager
 import traceback
 from unittest.mock import Mock
 
+from django.http.response import HttpResponse
+
 import pytest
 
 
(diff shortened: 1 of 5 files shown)


open_discussions/test_utils.py
```diff
@@ -50,18 +52,20 @@ def assert_not_raises():
...
 
-class MockResponse:
+class MockResponse(HttpResponse):
     """
-    Mock requests.Response
+    Mocked HTTP response object that can be used as a stand-in for request.Response and
+    django.http.response.HttpResponse objects
     """
 
     def __init__(self, content, status_code):
```
noisecapella, line 61: Can we add a docstring and mark `content` as `str`, so that pycharm can show a warning if bytes are passed in?

noisecapella:
What was the naming issue for djoser?

gsidebo:
@noisecapella The test class was named `DjoserViewTests`. By default pytest settings, it needs to be prepended with `Test` for pytest to collect its methods and run it

PR #3135, opened 2020-09-02
fix course search textbox behavior

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings

no ticket, but basically to accomplish the same thing I did here: https://github.com/mitodl/hugo-course-publisher/pull/190 for the OCW search.

#### What's this PR do?

this removes the automatic search behavior, so that the user has to manually trigger the search instead via the enter key.

I also found a few performance issues on the page and diagnosed those, adding `React.memo` to fix.

also refactored `CourseSearchPage` from class-based to functional component.

This touched a lot of parts of the search, so please test it exhaustively. Facets should all work, should be able to change text search, clear facet values, etc etc.

### DIFF
--- a/static/js/components/search/CourseFilterDrawer.js
+++ b/static/js/components/search/CourseFilterDrawer.js
@@ -28,76 +28,89 @@ type FilterDrawerProps = {
   onUpdateFacets: Function
 }
 
-export function FilterDisplay(props: FilterDrawerProps) {
-  const {
-    activeFacets,
-    clearAllFilters,
-    toggleFacet,
-    facetOptions,
-    onUpdateFacets
-  } = props
+// I had to add a `React.memo` call here because this component
+// was slowing down rendering for the whole page.
+export const FilterDisplay = React.memo<FilterDrawerProps>(
+  function FilterDisplay(props: FilterDrawerProps) {
+    const {
+      activeFacets,
+      clearAllFilters,
+      toggleFacet,
+      facetOptions,
+      onUpdateFacets
+    } = props
 
-  const anyFiltersActive =
-    _.flatten(_.toArray(Object.values(activeFacets))).length > 0
+    const anyFiltersActive =
+      _.flatten(_.toArray(Object.values(activeFacets))).length > 0
 
-  return (
-    <>
-      {anyFiltersActive ? (
-        <div className="active-search-filters">
-          <div className="filter-section-title">
-            Filters
-            <span
-              className="clear-all-filters"
-              onClick={clearAllFilters}
-              onKeyPress={e => {
-                if (e.key === "Enter") {
-                  clearAllFilters()
-                }
-              }}
-              tabIndex="0"
-            >
-              Clear All
-            </span>
+    return (
+      <>
+        {anyFiltersActive ? (
+          <div className="active-search-filters">
+            <div className="filter-section-title">
+              Filters
+              <span
+                className="clear-all-filters"
+                onClick={clearAllFilters}
+                onKeyPress={e => {
+                  if (e.key === "Enter") {
+                    clearAllFilters()
+                  }
+                }}
+                tabIndex="0"
+              >
+                Clear All
+              </span>
+            </div>
+            {facetDisplayMap.map(([name, , labelFunction]) =>
+              (activeFacets[name] || []).map((facet, i) => (
+                <SearchFilter
+                  key={i}
+                  value={facet}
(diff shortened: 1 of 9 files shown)


static/js/components/search/CourseFilterDrawer.js
```diff
@@ -28,76 +28,89 @@ type FilterDrawerProps = {
...
 
-export function FilterDisplay(props: FilterDrawerProps) {
-  const {
-    activeFacets,
-    clearAllFilters,
-    toggleFacet,
-    facetOptions,
-    onUpdateFacets
-  } = props
+// I had to add a `React.memo` call here because this component
```
noisecapella, line 31: Just curious, did you figure out what the slow part was? Is there just a lot of content here?

static/js/pages/CourseSearchPage.js
```diff
@@ -112,415 +63,442 @@ export type CourseSearchParams = {
...
+      if (!loaded) {
+        // this function will be triggered repeatedly by <InfiniteScroll />, filter it to just once at a time
+        return
+      }
 
-    this.updateSearchState({
-      activeFacets
-    })
-  }
+      runSearch(text, activeFacets, true)
```
noisecapella, line 267: Should there be an `await` here?

static/js/pages/CourseSearchPage.js
```diff
@@ -112,415 +63,442 @@ export type CourseSearchParams = {
...
-    const text = event ? event.target.value : ""
-    await this.updateSearchState({ text }, true)
-  }
+  // this effect here basically listens to activeFacets for changes and re-runs
+  // the search whenever it changes. we always want the facet changes to take
+  // effect immediately, so we need to either do this or call runSearch from
+  // our facet-related callbacks. this approach lets us avoid having the
+  // facet-related callbacks (toggleFacet, etc) be dependent on then value of
+  // the runSearch function, which leads to too much needless churn in the
+  // facet callbacks and then causesexcessive re-rendering of the facet UI
```
noisecapella, line 286: `causes excessive`

static/js/pages/CourseSearchPage.js
```diff
@@ -112,415 +63,442 @@ export type CourseSearchParams = {
...
-    case LR_TYPE_PODCAST_EPISODE:
-      return podcastEpisodes ? podcastEpisodes[result.id] || null : null
-    case LR_TYPE_LEARNINGPATH:
-      return userLists ? userLists[result.id] || null : null
-    }
-  }
+      if (isEnabled) {
+        newFacets[name] = _.union(newFacets[name] || [], [value])
+      } else {
+        newFacets[name] = _.without(newFacets[name] || [], value)
```
noisecapella, line 302: Is it convenient to write a test covering this line? It's missing from code coverage

noisecapella (changes requested):

noisecapella:
Functionality works great, just a few minor comments

static/js/components/search/CourseFilterDrawer.js
aliceriot: I think that's basically it - when we're dealing with the full set of topics and so on it's a fair amount of rendering that's happening. And I think unfortunately it was just sort of all being rendered...

I used the react dev tools and tracked down the component that was taking longest to the `CourseFilterDrawer` iirc

aliceriot: but the `React.memo` is pretty smart about how it works, so it will skip unnecessary renders but if there's any call to set state on a subcomponent it will re-render that part of the tree so I think even though it's not clear exactly where the issue is we should be ok. It also wasn't like a terrible, awful slowdown, it just got a bit sluggish - although then again this is on my work laptop which is pretty beefy, so it might have been terrible on a phone.

static/js/pages/CourseSearchPage.js
aliceriot: you know, I just copied over the `clearSearchCB` from the class component, idk why that was implemented as an async function because there is nothing async about it.

static/js/pages/CourseSearchPage.js
aliceriot: yeah, I can add a test tehre

aliceriot:
@noisecapella ok I think those comments should be addressed now

noisecapella (approved):

PR #3154, opened 2020-09-15
Youtube video resource RFC

mbertrand:
[Rendered](https://github.com/mitodl/open-discussions/blob/mb/rfcresourcevideo/docs/rfcs/0005-video-resources.md)

#3155

#### What's this PR do?
RFC for completing #3156

### DIFF
--- a/docs/rfcs/0005-video-resources.md
+++ b/docs/rfcs/0005-video-resources.md
@@ -35,7 +35,8 @@ The current ETL process for OCW resources will need to be modified to process th
   - If a matching video is found, copy that video's `image_src` and `transcript` values to the resource's `image_src` and `content` fields.
   - If a matching `Video` object is NOT found:
     - Find the `course_embedded_media` object with `id` of "Thumbnail-YouTube-JPG" and use its `media_location` as the `image_src` value.
-    - Find the `course_embedded_media` object with `id` of "<Youtube_video_id>.pdf". Use tika to retrieve (via `technical_location` url) and parse the contents and save to the `content` attribute.
+    - If the `transcript` attribute of the object exists and use that (stripped of HTML via BeautifulSoup) for the `content` value.
+    - If the `content` is still empty/null, find the `course_embedded_media` object with `id` of "<Youtube_video_id>.pdf". Use tika to retrieve (via `technical_location` url) and parse the contents and save to the `content` attribute.
   - The `ContentFile` will have a `content_type` of "video" and a `file_type` of "youtube"
   - If no Youtube video object is found, ignore the section and move on.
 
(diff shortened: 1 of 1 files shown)


docs/rfcs/0005-video-resources.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Table of Contents
+{: .no_toc .text-delta }
+
+- Table of Contents
+{:toc}
+
+### Abstract
+Parse the `course_embedded_media` section of OCW master.json files.  Search for Youtube videos within that section.
+If found, add the video as a resource, including the the thumbnail image and transcript text.
```
pdpinch, line 19:
```suggestion
If found, add the video as a resource, including the thumbnail image and transcript text.
```

mbertrand: 👍

docs/rfcs/0005-video-resources.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+### Architecture Changes
+
+The `ContentFile` model will need an additional nullable field, `image_src`, to store the value of the video thumbnail if it exists.  The existing `content` attribute can be used to hold the transcript text.
+The current ETL process for OCW resources will need to be modified to process the `course_embedded_media` section:
+  - Youtube videos in that section will have a `title` & `id` of "Video-YouTube-Stream", with a `media_location` equal to the Youtube video id.
+  - Once a Youtube video is identified, check for the existence of a `Video` object with the same `video_id`
+  - If a matching video is found, copy that video's `image_src` and `transcript` values to the resource's `image_src` and `content` fields.
+  - If a matching `Video` object is NOT found:
```
rhysyngsun, line 36: So this sounds like we don't intend for these files to be searchable individually on Open, is that correct?

rhysyngsun (commented):
I think this is looking good, I just have one question

docs/rfcs/0005-video-resources.md
Ferdi: Good point. @mbertrand how about we index all and filter out the rest for ocw ?

mbertrand: That is correct, only the video would show up in search results, not the video's thumbnail and transcript file (though the content of the transcript file would be used to find relevant results since it's saved as part of the video resource).  @ferdi I'm not sure if there's a point to indexing the supporting files unless you think we want users to get these returned in results at some point in the future.

Ferdi: @mbertrand sure, got it.

mbertrand: @rhysyngsun should I merge this yet or would you like to take another look?

rhysyngsun: @mbertrand lgtm!

rhysyngsun (approved):

PR #3185, opened 2020-10-16
ab/remove-course-catalog-acks-late

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?
Recently we had an issue on ci with a ocw update job never  exiting successfully and continuously rerunning due to acks_late setting. This pr sets acks_late=False (the default) from most course catalog update tasks. It keeps keep  acks_late=True for ocw updates generated by the webhook since those jobs won't be re-run by themselves.

Test that the ocw webhook still works. Instructions here:
https://github.com/mitodl/open-discussions/pull/2687

Test that backpopulate commands (backpopulate_podcast_data for example) still work

### DIFF
--- a/course_catalog/tasks.py
+++ b/course_catalog/tasks.py
@@ -34,8 +34,27 @@ def get_mitx_data():
 
 
 @app.task(acks_late=True)
+def get_ocw_course_with_retry(
+    *, course_prefix, blocklist, force_overwrite, upload_to_s3
+):
+    """
+    Task to sync a batch of OCW courses. We want acks_late=True for tasks launched
+    with the webhook so that there will be a retry if the task fails or the worker is killed
+    """
+    sync_ocw_courses(
+        course_prefixes=[course_prefix],
+        blocklist=blocklist,
+        force_overwrite=force_overwrite,
+        upload_to_s3=upload_to_s3,
+    )
+
+
+@app.task
 def get_ocw_courses(*, course_prefixes, blocklist, force_overwrite, upload_to_s3):
-    """Task to sync a batch of OCW courses"""
+    """
+    Task to sync a batch of OCW courses. We want to set acks_late=False for refreshes
+    launched by the backpopulate command or the scheduled task
+    """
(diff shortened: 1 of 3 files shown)


course_catalog/tasks.py
```diff
@@ -34,8 +34,27 @@ def get_mitx_data():
 
 
 @app.task(acks_late=True)
+def get_ocw_courses_with_retry(
```
abeglova, line 37: I don't love this solution, but I couldn't find a way to set acks_late conditionally.

course_catalog/tasks.py
```diff
@@ -34,8 +34,27 @@ def get_mitx_data():
...
+    *, course_prefixes, blocklist, force_overwrite, upload_to_s3
+):
+    """
+    Task to sync a batch of OCW courses. We want acks_late=True for tasks launched
+    with the webhook so that there will be a retry if the task fails or the worker is killed
+    """
+    sync_ocw_courses(
+        course_prefixes=course_prefixes,
```
mbertrand, line 45: Since the webhook is always for just one course_prefix, you could change this task name to `get_ocw_course_with_retry`, the input parameter name to `course_prefix`, and the code to:
```python
sync_ocw_courses(
    course_prefixes=[course_prefix]
    ...
```
and also in `course_catalog.views.py`, change the input parameter there from a list to a single prefix:
```
                get_ocw_course_with_retry.apply_async(
                    countdown=settings.OCW_WEBHOOK_DELAY,
                    kwargs={
                        "course_prefix": prefix,
```

mbertrand (changes requested):

course_catalog/tasks.py
mbertrand, line 37: minor nitpick: this could be renamed to `get_ocw_course_with_retry` since it only affects 1 course at a time.

mbertrand (approved):
Works great, one minor nitpick, feel free to change or ignore.

Issue #3087, opened 2020-07-30
Removed comments should not be visible to non-moderator users


rhysyngsun:
Removed comments are still visible at a detail url, this should return a 404 similar to the work done to do this for the post page in #2785

PR #3202, opened 2020-10-30
404 for removed comments

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/open-discussions/issues/3087

#### What's this PR do?
This adds a server-side 404 page for removed comments for non-moderator users

Make a comment as a non-moderator user
Get the share url for the comment it should be something like `http://localhost:8063/c/public_channel/5/i-like-robotics/comment/3/`
Mark the comment as spam using 
 `docker-compose run web ./manage.py reclassify_spam --spam --comment-id <object id of comment>`
Go to the comment url. You should be able to see the comment as a moderator but get a 404 as a non-mderator or anonymous user

### DIFF
--- a/cassettes/channels.views.comments_test.test_get_removed_comment[None-404].json
+++ b/cassettes/channels.views.comments_test.test_get_removed_comment[None-404].json
@@ -1,0 +1,922 @@
+{
+  "http_interactions": [
+    {
+      "recorded_at": "2020-11-05T14:43:31",
+      "request": {
+        "body": {
+          "encoding": "utf-8",
+          "string": ""
+        },
+        "headers": {
+          "Accept": [
+            "*/*"
+          ],
+          "Accept-Encoding": [
+            "gzip, deflate"
+          ],
+          "Connection": [
+            "keep-alive"
+          ],
+          "User-Agent": [
+            "python-requests/2.24.0"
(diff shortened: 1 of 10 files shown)


mbertrand (commented):
When logged in as the spam comment writer I can still see the comment under the post, but the direct link to the comment returns a 404.  Should that be a separate issue/PR?

abeglova:
Making the removed comment look normal to the comment writer is default reddit behavior - to discourage the comment writer from just re-posting their spam. I think as long as the comment is not visible to other  non-moderator users and anonymous users, we don't care if the spam comment is visible to the spam commenter.

mbertrand:
Should the direct link to the comment also be visible to the comment writer too?  Currently the writer can see the comment under the post url but the direct comment url returns a 404.

mbertrand:
Seems like there's some inconsistency in behavior with a spam comment author being able to see the comment under the post URL but getting a 404 for the direct comment URL.  Maybe the comment URL should still work for the author?

abeglova:
Ok! I'll make an update to make the URL still work for the spam writer

mbertrand (approved):
Works great 👍

PR #3204, opened 2020-11-02
Migrate to Github Actions

rhysyngsun:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
Removes Travis in favor of github actions

- Tests should pass
- Codecov still works

Everybody is replaceable


.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+      - name: Code formatting
+        run: black --check .
+
+      - name: Lint
+        run: pylint ./**/*.py
+
+      - name: Tests
+        run: |
+          export MEDIA_ROOT="$(mktemp -d)"
+          tox
```
aliceriot, line 64: am I right in reading this that the tests are now being run outside of docker? just want to make sure I'm grokking this correctly

.github/workflows/ci.yml
```diff
@@ -0,0 +1,121 @@
...
+      - name: Set up NodeJS
+        uses: actions/setup-node@v2-beta
+        with:
+          node-version: 10.15.1
+
+      - name: Setup environment
+        run: sudo apt-get install libelf1
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile --ignore-engines
```
aliceriot, line 103: do you know if running yarn will use yarn's cache between builds? I looked here and didn't see a mention of the yarn cache: https://github.com/actions/setup-node

It looks like the yarn install only takes around 40 seconds anyhow, so probably not the biggest deal either way

aliceriot: I think the total JS build time is roughly the same as travis as well

.travis.yml
```diff
@@ -1,26 +0,0 @@
...
-    - language: node_js
-      before_install:
-        - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.13.0
-        - export PATH="$HOME/.yarn/bin:$PATH"
-      node_js:
-        - "10.15.1"
-      script: bash ./travis/js_tests.sh
-      cache: yarn
-      env:
-        name: JavaScript
```
aliceriot, line 26: RIP

aliceriot (changes requested):
just a few questions and things

.github/workflows/ci.yml
rhysyngsun: Yes, I was hoping it would run a bit faster not having to build the containers. Seeing a 2-3 minute speedup but i'm not sure if that's caused by this or different hardware.

.github/workflows/ci.yml
rhysyngsun: I don't think there's an implicit cache in how `setup-node` works, let me take a quick look at what's possible here.

.github/workflows/ci.yml
aliceriot: it's probably related to ditching the containers - I saw a similar speedup on travis from moving from docker to "bare metal" (ha) for the JavaScript tests

aliceriot (approved):
LGTM! :+1:

Issue #3188, opened 2020-10-21
Implement celery priorities so spam tasks run faster


abeglova:
We recently had problems with spam check update commands taking 40min-1 hr to run because they were queued behind slow jobs in celery. We want both the manual spam reclassification job for when akismet gets it wrong and the initial spam classification to run immediately.

We should implement separate queues for different types of tasks

PR #3206, opened 2020-11-02
speparate celery queue for spam checks

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3188

#### What's this PR do?
Adds a separate queue for spam tasks. Since the spam queue should have fewer items, spam tasks will run sooner.

add
```
import time
@app.task
def sleep(seconds):
    time.sleep(seconds)
```

to `course_catalog/tasks.py`

Fill the celery queues with default priority tasks by running

```

for i in range(0,1000):
    sleep.delay(15)
```

from the shell

`docker-compose run web ./manage.py reclassify_spam --spam --post-id <object id of post>`

should still run fairly quickly and if you create a post or comment with a non-moderator account a `check_post_for_spam` job should run to create a SpamCheckResult record

All other celery tasks will be queued behind the sleep jobs and will not run until those finish

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -106,7 +106,7 @@ services:
     command: >
       /bin/bash -c '
       sleep 3;
-      celery -A open_discussions.celery:app worker -B -l ${OPEN_DISCUSSIONS_LOG_LEVEL:-INFO}'
+      celery -A open_discussions.celery:app worker -Q spam,default -B -l ${OPEN_DISCUSSIONS_LOG_LEVEL:-INFO}'
     links:
       - db
       - redis
--- a/open_discussions/celery.py
+++ b/open_discussions/celery.py
@@ -12,4 +12,12 @@ from django.conf import settings  # noqa pylint: disable=wrong-import-position
 
 app = Celery("open_discussions")
 app.config_from_object("django.conf:settings", namespace="CELERY")
+app.conf.task_default_queue = "default"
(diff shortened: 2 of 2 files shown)


channels/management/commands/reclassify_spam.py
```diff
@@ -65,12 +66,16 @@ def handle(self, *args, **options):
             self.stdout.write("Either --spam or --ham flag is required")
             return
 
-        task = tasks.update_spam.delay(
-            spam=is_spam,
-            comment_ids=options["comment_ids"],
-            post_ids=options["post_ids"],
-            retire_users=options["retire_users"],
-            skip_akismet=options["skip_akismet"],
+        task = tasks.update_spam.apply_async(
```
rhysyngsun, line 69: I think it'd be better to use [automatic routing](https://docs.celeryproject.org/en/stable/userguide/routing.html#automatic-routing) since you don't have to worry about covering all of the task call sites now or in the future. It's probably not much of a gain right now for this work, but later on if we start splitting other tasks into their own queues we'll want one centralized place to designate the routing.

rhysyngsun (changes requested):

abeglova:
Back to you @rhysyngsun

rhysyngsun (approved):

PR #3213, opened 2020-11-05
add fmt:check to CI

aliceriot:

none

#### What's this PR do?

this just adds back a call to `npm run fmt:check` to our JS CI check, which was I think unintentionally omitted from https://github.com/mitodl/open-discussions/pull/3204

The build should pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,6 +125,15 @@ jobs:
       - name: Lints
         run: npm run lint
 
+      - name: Code formatting
+        run: npm run fmt:check
+
+      - name: Scss lint
+        run: npm run scss_lint
+
+      - name: Flow
+        run: npm run flow
+
       - name: Tests
         run: npm run test
         env:


.github/workflows/ci.yml
```diff
@@ -125,6 +125,15 @@ jobs:
       - name: Lints
         run: npm run lint
 
+      - name: Code formatting
+        run: npm run fmt:check
+
+      - name: Scss lint
+        run: npm run fmt:check
```
rhysyngsun, line 132:
```suggestion
        run: npm run scss_lint
```

rhysyngsun (changes requested):
Typo

.github/workflows/ci.yml
aliceriot: oh lol, that's what copy-paste gets you

rhysyngsun (approved):
:+1: when tests pass

PR #3225, opened 2020-11-16
fix styling of long search filters in course search

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

none, but see https://github.com/mitodl/hugo-course-publisher/issues/337

(we have the same issue over here)

#### What's this PR do?

this just makes a small styling change to deal with a text overflow issue on longer filter names.

Normally the filters were display here aren't quite long enough to trigger the issue, but we're doing department name over in OCW and that is long enough. But I figured we should fix the styling here too so we're "future proofed"

To repro the issue you need to just edit the text node inside of the `.active-search-filter` to be long enough to trigger the problem. Then confirm checking out this branch fixes it (see screenshots below for reference).

before:

Screenshot from 2020-11-16 10-29-31

after:

Screenshot from 2020-11-16 10-29-12

### DIFF
--- a/static/scss/search.scss
+++ b/static/scss/search.scss
@@ -100,7 +100,6 @@
       display: inline-flex;
       align-items: center;
       flex-wrap: nowrap;
-      white-space: nowrap;
       border: 1px solid $font-grey-mid;
       border-radius: 14px;
 


abeglova (approved):
👍

PR #3228, opened 2020-11-18
Add Elasticsearch shard count variable

### LINKED ISSUE #3140: Add configurable shard count to search.indexing_api.clear_and_create_index
`search.indexing_api.clear_and_create_index` should specify a configurable shard count when creating a new index, instead of settling for Elasticsearch's default. Elasticsearch's default of five shards per index (with version 6.x) is too high for our production cluster, which has only three nodes. The high shard count can be detrimental to server performance, especially during shard reallocation if a node leaves and rejoins the cluster; and can also affect indexing performance.

The `clear_and_create_index` function:
https://github.com/mitodl/open-discussions/blob/3005018737fda124f306d8668633cbc4743ee6d3/search/indexing_api.py#L267

(The index setting to assign is `settings.index.number_of_shards`.)

markbreedlove:
Add an environment variable that specifies the number of shards to allocate when creating a new index.

Closes #3140


open_discussions/settings.py
```diff
@@ -669,6 +669,7 @@
 ELASTICSEARCH_MIN_QUERY_SIZE = get_int("ELASTICSEARCH_MIN_QUERY_SIZE", 2)
 ELASTICSEARCH_MAX_SUGGEST_HITS = get_int("ELASTICSEARCH_MAX_SUGGEST_HITS", 1)
 ELASTICSEARCH_MAX_SUGGEST_RESULTS = get_int("ELASTICSEARCH_MAX_SUGGEST_RESULTS", 1)
+ELASTICSEARCH_SHARD_COUNT = get_int("ELASTICSEARCH_SHARD_COUNT", 5)
```
rhysyngsun, line 672: I found a default of 5 to fail locally when reindexing, if a default of 5 makes sense for deployment scenarios, could you add `ELASTICSEARCH_SHARD_COUNT: 1` to the `docker-compose` environment variables so local development works out of the box? 

https://github.com/mitodl/open-discussions/blob/master/docker-compose.yml#L3

rhysyngsun (changes requested):

open_discussions/settings.py
markbreedlove: 5 shards was the default in Elasticsearch for a long time until version 6 or 7, and my recollection is that an earlier version is being used here. I'm happy to set the default lower, though. There's no reason for it to be higher than 1 or 2 in a development context.

rhysyngsun: either way is fine by me, locally for whatever reason this new setting causes ES to hard throttle the reindexing so it ends up failing. that probably needs to be fixed too...

markbreedlove: I've just changed the default to 2. Does it work better for you?

rhysyngsun (approved):

Issue #3244, opened 2020-11-30
Allow moderators to individually opt out of email notification


abeglova:
We added notifications of new posts for channel moderators in https://github.com/mitodl/open-discussions/issues/3165

Add an option in /settings/notifications that is only visible to moderators to allow moderators to individually opt in or out of notifications

PR #3257, opened 2020-12-11
moderator notification setting ui

abeglova:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested
Screen Shot 2020-12-14 at 7 59 09 AM

closes https://github.com/mitodl/open-discussions/issues/3244

#### What's this PR do?
This PR adds an option in /settings/notifications that is only visible to moderators to allow moderators to individually opt in or out of notifications

Run `docker-compose run web ./manage.py populate_notification_settings`

Log in as a non-moderator user and go to `http://localhost:8063/settings/`. There should be no changes

Log in as a moderator user and go to `http://localhost:8063/settings/`. There should be a new checkbox to allow moderator notifications.

Keep moderator notifications enabled in that ui and enable notifications in `http://localhost:8063/manage/c/edit/<channel>/basic/`

Make a post in the channel. Then run

```
```
You should get a notification on the MAILGUN_RECIPIENT_OVERRIDE email

If moderator notifications are turned off for either the user or the channel, you should not receive a notification email

### Updated you based on @Ferdi's comment:
Screen Shot 2020-12-22 at 2 11 34 PM

### Updated UI based on additional comment from @Ferdi:
Screen Shot 2021-01-11 at 9 02 17 PM


notifications/views.py
```diff
@@ -27,5 +29,14 @@ class NotificationSettingsViewSet(
...
+
         user = self.request.user
-        return user.notification_settings
+
+        if admin_api.is_moderator_for_any_channel(user.username):
+            return user.notification_settings
+        else:
+            return user.notification_settings.exclude(
+                notification_type=NOTIFICATION_TYPE_MODERATOR
+            )
```
mbertrand, line 42: Is this part necessary, would a non-moderator not have that notificiation type?

Ferdi:
looking at the screenshot , it's not clear what this does .  
If i'm guessing right what the intent is , this should work something like : 

When do you want to receive notification ?
[ ] when someone responds to my post / comment 
[ ] when someone post to a channel I moderate

mbertrand (changes requested):
Functionality works great! 

One thing I noticed not related to this PR is that `docker-compose run web ./manage.py populate_notification_settings` didn't do anything on my dev instance without the `--inactive` flag because `Profile.last_active_on` was null for all my users.  The same seems to be true on RC and production, including my own account which I just logged in today.  I think it's probably a bug?

notifications/views.py
abeglova: Yes, we do not delete the notification setting if a user is removed as a moderator

Ferdi:
@abeglova even better would be something like this : 
When do you want to receive notification ?
[ ] when someone responds to my post / comment
[ ] when someone post to a channel I moderate
       [ ] channel 1
       [ ] channel 2 
       [ ] etc
       
The "when someone post to a channel I moderate" shows up only when user is moderating 1 ore more channels  AND those channels have the settings turned on in the setting page

notifications/views.py
abeglova: I updated the code to remove the moderator setting when a user is removed as a moderator

abeglova:
Back to you @mbertrand. To test, delete all your local NotificationSetting objects with notification_type="moderator_posts" and run `populate_notification_settings` again

mbertrand (approved):
Unfortunately I'm not able to test the functionality right now due to ongoing laptop issues, but the code looks good so approving based on that.

mbertrand (approved):
Functionality works great too, though the filtering by inactive users is still a problem because of https://github.com/mitodl/open-discussions/issues/3266

PR #3278, opened 2021-01-13
upgrade to the latest version of redux-hammock

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

I recently upgraded some dependencies in redux-hammock (https://github.com/mitodl/redux-hammock/pull/31). This just bumps us up to that version here.

We're using redux-hammock for a bunch of rest APIs still, mainly in the code for creating posts, channels, etc. So I think exercising various features with posts like voting, commenting, editing, etc are a good idea. There aren't any code changes really in this new version of redux-hammock, basically just dependency upgrades, but still we should check thoroughly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "redux-actions": "^2.6.5",
     "redux-asserts": "^0.0.12",
     "redux-debounced": "^0.5.0",
-    "redux-hammock": "^0.3.1",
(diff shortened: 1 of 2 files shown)


abeglova (approved):
LGTM

PR #3287, opened 2021-01-21
Don't show suggestion if it is effectively the same as search text

### LINKED ISSUE #3286: Search suggestion is same as search term
### Steps to Reproduce

Same issue as https://github.com/mitodl/ocw-www/issues/34 - the suggestion is shown if it differs from the search term in case, or if the search term includes quotes or non-alphanumeric characters like ":" or "!".  

Search for "radiative transfer" with a topic of "Particle Physics", i.e. https://ocwnext.odl.mit.edu/search/?q=%22radiative%20transfer%22&t=Particle%20Physics

### Expected Behavior

No suggestion is shown because the suggestion is the same as the search text.

### Actual Behavior

I get the suggested search message: Did you mean "radiative transfer"?

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #3286 

#### What's this PR do?
Ignores quotes, case, and special characters when comparing a search suggestion to the search text.

Perform a search included a search term in quotes with a colon in between words, that only returns 1 result, like "Fiction: Great Books!" On https://discussions-rc.odl.mit.edu/learn/search?q=%22Fiction%3A%20Great%20Books%21%22 the suggested search term "fiction great books" will display.  On this branch, the suggested search term should not display.  Change the search term to "Fiction: Great Boos!" and it should display.

### DIFF
--- a/static/js/pages/CourseSearchPage.js
+++ b/static/js/pages/CourseSearchPage.js
@@ -236,6 +236,7 @@ export default function CourseSearchPage(props: Props) {
         [
           text
             .toLowerCase()
+            .trim()
             .replace(/^"(.*)"$/, "$1")
             .replace(/[\W]+/g, " ")
             .trim()
--- a/static/js/pages/CourseSearchPage.js
+++ b/static/js/pages/CourseSearchPage.js
@@ -232,7 +232,16 @@ export default function CourseSearchPage(props: Props) {
   const resultsColumnWidth = searchResultLayout === SEARCH_GRID_UI ? 9 : 8
   const suggestions =
     !emptyOrNil(suggest) && !emptyOrNil(text)
-      ? R.without([text], suggest).map(
+      ? R.without(
+        [
+          text
+            .toLowerCase()
+            .replace(/^"(.*)"$/, "$1")
+            .replace(/[\W]+/g, " ")
+            .trim()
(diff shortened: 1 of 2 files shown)


static/js/pages/CourseSearchPage.js
```diff
@@ -232,7 +232,16 @@ export default function CourseSearchPage(props: Props) {
...
   const suggestions =
     !emptyOrNil(suggest) && !emptyOrNil(text)
-      ? R.without([text], suggest).map(
+      ? R.without(
+        [
+          text
+            .toLowerCase()
+            .replace(/^"(.*)"$/, "$1")
+            .replace(/[\W]+/g, " ")
+            .trim()
```
noisecapella, line 241: Can we move this before the double quote regex so we can handle ` "search phrase"` for example?

noisecapella (changes requested):

static/js/pages/CourseSearchPage.js
mbertrand: Thanks, looks like I need it in both places (for example, ` "aerospace: biomedical!" `

mbertrand:
@noisecapella ready for another look

noisecapella (approved):

PR #3291, opened 2021-01-25
fix inactive setting

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
The populate_notification_settings job currently ignores the --inactive flag

Update a user to be inactive and delete notifications for that user

```
from notifications.models import NotificationSettings

user = User.objects.last()
profile = User.profile
profile.last_active_on = None
profile.save()

NotificationSettings.objects.filter(user=user).delete()
```

Run
`docker-compose run web ./manage.py populate_notification_settings`

Notification settings should not be created for the user

Run
`docker-compose run web ./manage.py populate_notification_settings --inactive`

Now notification settings should not be created for the user

### DIFF
--- a/notifications/management/commands/populate_notification_settings.py
+++ b/notifications/management/commands/populate_notification_settings.py
@@ -12,7 +12,7 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument("--username", nargs="?", action="append")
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
👍

Issue #3266, opened 2020-12-22
last_active_on is not being updated in profiles


abeglova:
last_active_on is null for users with recent logins

pdpinch:
Is that field visible to users anywhere?

abeglova:
No it shouldn't be. I noticed that the field isn't being populated because  populate_notification_settings ignores inactive profiles based on that field

abeglova:
Looks like that field is only used in `populate_notification_settings` . The functionality is also duplicated with `last_login` in the  Django user model. I made a pr to delete the field entirely.

pdpinch:
Well, I found another dependency on this field. It was being used in a [BI report](https://bi.odl.mit.edu/queries/360/source#571). I'll have to see if the report can be changed to `auth_user.last_login`

PR #3293, opened 2021-01-26
Ab/remove profile updated on

abeglova:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3266

#### What's this PR do?
`Profile.last_active_on` does not get updated. The field is also only used in the `populate_notification_settings` command and the functionality of tracking user activity is duplicated with `last_login` in the django user model, which does get updated. 

Tests should pass. You should still be able to login in the app.


notifications/management/commands/populate_notification_settings.py
```diff
@@ -12,15 +12,10 @@ class Command(BaseCommand):
...
     def add_arguments(self, parser):
         parser.add_argument("--username", nargs="?", action="append")
-        parser.add_argument("--inactive", action="store_true", default=False)
 
     def handle(self, *args, **options):
         users = get_user_model().objects.all()
 
-        if not options["inactive"]:
-            # unless the --inactive flag is specified, default to the safer subset of active users
-            users = users.filter(profile__last_active_on__isnull=False)
```
mbertrand, line 22: Instead of deleting this option, maybe change to this?:

```
users = users.filter(last_login__isnull=False)
```

open_discussions/settings.py
```diff
@@ -123,7 +123,6 @@
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "corsheaders.middleware.CorsMiddleware",
     "authentication.middleware.BlockedIPMiddleware",
-    "open_discussions.middleware.user_activity.UserActivityMiddleware",
```
mbertrand, line 126: This was added by @rhysyngsun for https://github.com/mitodl/open-discussions/issues/473 - is it not needed anymore?

rhysyngsun: Yeah, we can just remove that entirely.

notifications/management/commands/populate_notification_settings.py
abeglova: I'm not sure why this option is needed. @rhysyngsun you added it a long time ago, do you remember why it exists and is it ok to get rid of it?

The Notification Settings page assumes that all users have comment and front page notification settings and breaks if they don't have them, so we don't want users without notification settings. Also the job skips over users with existing notification settings and makes no changes.

rhysyngsun: That can be removed too, it's a holdover from the massive micromasters backpopulate

mbertrand (approved):

Issue #3297, opened 2021-01-28
Create models for enrollment data


abeglova:
Enrollments table should have user_id, course_run_id and enrollment_date as columns

PR #3299, opened 2021-01-29
add enrollment models

abeglova:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3297

#### What's this PR do?
Adds a table to store enrollments in open

Make sure the migration runs


pdpinch:
👍 Schema looks OK to me.

mbertrand (changes requested):
Getting an error on startup: "module 'enrollments' has no attribute 'default_app_config" - adding an `__init__.py` file to the app should fix this.

Might be useful to add the Enrollment model to the Django Admin page.

I probably would have placed this model in the existing `course_catalog` app, but a separate app works too and maybe makes more sense if there's lots more functionality coming (tasks, views?) that's centered around these enrollments.

abeglova:
I moved the model to course_catalog and added an admin view

course_catalog/admin.py
```diff
@@ -145,6 +146,15 @@ class PodcastEpisodeAdmin(admin.ModelAdmin):
     search_fields = ("full_description",)
 
 
+class EnrollmentAdmin(admin.ModelAdmin):
+    """Enrollment Admin"""
+
+    model = Enrollment
+    search_fields = ("run_id", "user_id")
+    list_display = ("run_id", "user_id", "enrollment_timestamp")
+    search_fields = ("run_id", "user_id")
```
mbertrand, line 155: `search_fields` occurs twice.  Depending on whether or not Peter/Brian think they might need to use this, might be worth making it more user-friendly by doing something like this:

```python
    def run_slug(self, obj):
        return f"{obj.run.slug}"

    run_slug.admin_order_field = 'run__slug'

    model = Enrollment
    search_fields = ("run__run_id", "run__slug", "run__title", "user__email", "user__username")
    list_display = ("run_slug", "user", "enrollment_timestamp")
    autocomplete_fields = ("run", "user")
```

mbertrand (changes requested):

course_catalog/admin.py
abeglova: Thanks! I updated the code to use your suggestion

mbertrand (approved):
👍

Issue #3298, opened 2021-01-28
Create update job to pull enrollment data for a single email


abeglova:
Create a celery job in the open codebase to query the Athena cluster for enrollments for a specific user. The job will run whenever a new user signs up for open. The results will be appended to the enrollments table in the open database.

PR #3312, opened 2021-02-19
enrollments for single user

abeglova:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [x] Settings

closes https://github.com/mitodl/open-discussions/issues/3298

#### What's this PR do?
This pr queries for enrollments from the Athena enrollment table for a single user and creates Enrollment records in the open database. It attempts to match the course code from the Athena database (which comes from the mitx bigquery database) to a course run or, failing that, a course.

Ask me for Athena credentials. Set 
ATHENA_ACCESS_KEY_ID
ATHENA_SECRET_ACCESS_KEY
ATHENA_WORK_GROUP
and
ATHENA_REGION_NAME
in your .env file

1) Verify that you can still create a new user through the UI and that you see `Task course_catalog.tasks.update_enrollments_for_email` run on celery after you set the password.

2) Change the email of the user you just created to `pdpinch@mit.edu` Run

```
update_enrollments_for_email.delay('pdpinch@mit.edu')
```
Verify that 154 enrollments are created


course_catalog/admin.py
```diff
@@ -149,21 +149,38 @@ class PodcastEpisodeAdmin(admin.ModelAdmin):
...
+        if obj.course:
+            return f"{obj.course.course_id}"
+        else:
+            return ""
 
     model = Enrollment
     search_fields = (
+        "course__course_id",
+        "course__title",
+        "enrollments_table_run_id",
```
pdpinch, line 170: What changes are you making the admin here?

requirements.txt
```diff
@@ -4,418 +4,147 @@
...
+uritemplate==3.0.0        # via google-api-python-client
+urllib3==1.25.10          # via -r requirements.in, botocore, elasticsearch, requests, sentry-sdk
+uwsgi==2.0.18             # via -r requirements.in
+vine==1.3.0               # via amqp, celery
+wcwidth==0.1.8            # via prompt-toolkit
+webencodings==0.5.1       # via html5lib, tinycss2
+wrapt==1.10.11            # via -r requirements.in, deprecated
+xbundle==0.3.1            # via -r requirements.in
+xmlsec==1.3.3             # via python3-saml
+zeep==3.4.0               # via mit-moira
```
pdpinch, line 147: small thing: is there anyway to get pip-compile to maintain the line breaks, so this diff isn't so huge?

course_catalog/admin.py
abeglova: I'm making changes to the Enrollments admin page (currently no data).

There aren't matches for runs for all enrollments. 
This updates the admin page to display just the course match for courses where there is a match on the course id but not the run id. It also shows the run id from the enrollments table

app.json
```diff
@@ -39,6 +39,22 @@
       "description": "Algolia Places API key",
       "required": false
     },
+    "ATHENA_ACCESS_KEY_ID": {
```
blarghmatey, line 42: We shouldn't be adding separate access keys for Athena, we should just add the appropriate permissions to the AWS credentials that the application is already using.

course_catalog/etl/enrollments.py
```diff
@@ -0,0 +1,78 @@
...
+        SELECT 
+            enrollment_course_id,
+            enrollment_created
+        FROM ol_warehouse.mitx_enrollments 
+        WHERE email = '{email}'
+        """.format(
+        email=user.email
+    )
+
+    cursor = athena_cursor()
```
blarghmatey, line 47: I think it would make sense to pass the cursor as a parameter to this function so that it is easier to swap to Trino from Athena once we have that deployed. It would also simplify mocking the cursor object for tests.

course_catalog/etl/enrollments.py
```diff
@@ -0,0 +1,78 @@
...
+    Args:
+        user (User): user to fetch enrollments for
+    """
+
+    if not (
+        settings.ATHENA_ACCESS_KEY_ID
+        and settings.ATHENA_SECRET_ACCESS_KEY
+        and settings.ATHENA_WORK_GROUP
+        and settings.ATHENA_REGION_NAME
+    ):
```
blarghmatey, line 34: It seems like this function shouldn't have any awareness of the settings requirements for the cursor object and should be external to the logic to process the query results.

app.json
abeglova: Agreed! Can you give the rc credentials for open access to Athena?

course_catalog/etl/enrollments.py
abeglova: I was using the credentials as a feature flag so we don't get noisy errors if this is merged before the credentials are set. If we are going to just use the existing AWS credentials, this isn't necessary. The only one of these credentials that we will keep is ATHENA_WORK_GROUP, and 'primary' will be the default so it will work without any credentials set

course_catalog/etl/enrollments.py
abeglova: Once we remove the check for credentials above, swapping the cursor will just involve changing athena_cursor()
 to trino_curosr() here. Passing the cursor to this functions works too though

course_catalog/etl/enrollments.py
```diff
@@ -0,0 +1,78 @@
...
+            course = run.content_object
+        else:
+            course = Course.objects.filter(
+                course_id__contains=run_id_parts[1], published=True
+            ).last()
+
+        timezone = pytz.timezone("UTC")
+        enrollment_timestamp = timezone.localize(result["enrollment_created"])
+
+        Enrollment.objects.update_or_create(
```
mbertrand, line 73: Is it okay to create an enrollment if both `run` and `course` are `None`?  Should there be an error logged if `course` is also `None` to notify someone that the Athena `enrollment_course_id` doesn't have any match in open?

course_catalog/tasks_test.py
```diff
@@ -479,3 +481,13 @@ def test_get_podcast_data(mocker):
...
+
+
+def test_update_enrollments_for_email(mocker):
+    """Verify that update_enrollments_for_email call update_enrollments_for_user job"""
+    user = UserFactory.create()
+    mock_task = mocker.patch(
+        "course_catalog.etl.enrollments.update_enrollments_for_user"
+    )
+    update_enrollments_for_email.delay(user.email)
+    mock_task.assert_called_once_with(user)
```
mbertrand, line 493: Maybe also add a test to verify that the task is not called if no user is found with a matching email address

```
    update_enrollments_for_email.delay(fake_email)
    mock_task.assert_not_called()
```

mbertrand (changes requested):

requirements.txt
abeglova: fixed

course_catalog/etl/enrollments.py
abeglova: There are enough courses without matches that raising an error does not make sense. Once this is merged and the data is updated, the plan is to review the enrollment_course_ids with no matches to see if there are any we care about then add a look up table to match them to the correct course id

app.json
blarghmatey: Please open a ticket for that, preferably at https://trello.com/b/jLF7h6ZW/platform-engineering-open-learning-engineering

mbertrand (commented):
Code looks great, but I'm now getting an AWS AccessDenied error probably because the CI AWS key I'm using doesn't have the correct permissions yet.  Probably should hold off on merging until those permissions are in place?


```

File: app.json (hunk 3 above)
Comment (abeglova) [reply app.json]: I don't have access to that trello board

Comment (abeglova) [reply app.json]: Here is an open issue https://github.com/mitodl/open-discussions/issues/3318

File: app.json [hunk 8]
```diff
@@ -39,6 +39,14 @@
...
       "required": false
     },
+    "ATHENA_WORK_GROUP": {
+      "description": "work group for Athena queries",
+      "required": false
+    },
+    "ATHENA_REGION_NAME": {
+      "description": "Region name for Athena queries",
+      "required": false
+    },
```
blarghmatey, line 49: As part of setting up the permissions for Open to access Athena I have set up per-environment workgroups and database names, so there should be an additional parameter of `"ATHENA_DATABASE_NAME"` that is used to determine which database to execute queries against.

course_catalog/etl/enrollments.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Args:
+        user (User): user to fetch enrollments for
+    """
+
+    query = """
+        SELECT 
+            enrollment_course_id,
+            enrollment_created
+        FROM ol_warehouse.mitx_enrollments
```
blarghmatey, line 33: This line needs to be configurable to the database name that will be queried.

course_catalog/etl/enrollments.py
```diff
@@ -30,9 +30,10 @@ def update_enrollments_for_user(user):
         SELECT 
             enrollment_course_id,
             enrollment_created
-        FROM ol_warehouse.mitx_enrollments 
+        FROM {database}.mitx_enrollments 
         WHERE email = '{email}'
         """.format(
+        database=settings.ATHENA_WORK_GROUP,
```
blarghmatey: This should be `settings.ATHENA_MITX_DATABASE_NAME`. We might also want to make the table name configurable as it will likely evolve/change as we add more enrollments data sources.

blarghmatey (approved):

blarghmatey:
:+1:

Issue #3316, opened 2021-03-01
Channel-specific GA Tracking not working with G4 Measurement ID 


pdpinch:
### Steps to Reproduce

- Add a G4 style Measurment ID to a channel, e.g. G-312MTJSH61 using the django admin /admin/channels/channel

### Expected Behavior

- Google Analytics should gather web usage data when users visit the channel. 

### Actual Behavior

- No data flows into the channel-specific Google Analytics property, but the site-wide property still gets data. 

### Related Issues

In PR #1019 we added per-channel Google Analytics support, but we used the now-replaced UTM- style codes. 

In PR #248 we set up site-wide Google Analytics. We want to maintain this feature. If necessary, we can upgrade the existing property to G4.

PR #3319, opened 2021-03-05
make tracking with new google analytics g-tags possible

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
- [x] Settings

closes https://github.com/mitodl/open-discussions/issues/3316

#### What's this PR do?
Google analytics no longer supports `analytics.js`, which the `react-ga` module uses. This upgrades our code to use `gtag.js` so that the code works with newly created google tags that begin with G-.  The old universal analytics tags which start with U- are also supported with gtag.js 

To test out the new system and ensure we get the same metrics for both, this pr adds a new GA_G_TRACKING_ID variable so we can run both at the same time initially. GA_TRACKING_ID and `react-ga` will be removed in a followup pr once we confirm that everything is working properly on production. 

The only channel on production that currently has a tracking tag hasn't had posts in two years, so we can upgrade channel-specific tracking in this pr.

Set GA_G_TRACKING_ID in your .env file.
Go to the network tab in developer tools and search for the tag.
You should see calls to https://www.google-analytics.com with the tag when you click around the app.

Set the GA tracking id for a channel from the admin (http://localhost:8063/admin/channels/)
Go to the network tab in developer tools and search for the tag.
You should see calls to https://www.google-analytics.com with the tag when you click around the app in that channel but not when you are in a different channel or on other pages.


static/js/util/withTracker.js
```diff
@@ -14,9 +14,30 @@ const withTracker = (WrappedComponent: Class<React.Component<*, *>>) => {
     ReactGA.initialize(SETTINGS.gaTrackingID, { debug: debug, testMode: test })
   }
 
+  if (SETTINGS.gaGTrackingID) {
+    const url = `https://www.googletagmanager.com/gtag/js?id=${
+      SETTINGS.gaGTrackingID
+    }`
+    const gaScript = document.createElement("script")
```
abeglova, line 21: This appends the gtag.js snippet to the header https://developers.google.com/analytics/devguides/collection/gtagjs#install_the_global_site_tag

rhysyngsun (commented):
I'm not sure if this was an existing bug or not, but i'm not seeing page view events getting sent when i navigate around, only on the first load of the application.

Issue #3345, opened 2021-03-24
The first page view after a user clicks away from a channel gets recorded by the channel tracker


abeglova:
Brian noticed some page views that don't belong in the channel when testing the channel tracker.

It looks like  the first page view after a user clicks away from a channel gets recorded by the channel tracker

PR #3339, opened 2021-03-19
manually send gtag events

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
https://github.com/mitodl/open-discussions/issues/3335

#### What's this PR do?
The automatic pageview metric in gtag.js (implemented in https://github.com/mitodl/open-discussions/pull/3319) doesn't pick up everything we consider a page view. This pr turns off automatic pageviews and sends manual pageviews instead.

#### How should this be manually tested?
Set GA_G_TRACKING_ID in your .env file. You can use G-B0EKT59MQF for testing
Go to the network tab in developer tools and search for the tag.
You should see calls to https://www.google-analytics.com with the tag when you click around the app.

PR #3357, opened 2021-03-31
Ab/fix channel tracking

abeglova:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  
closes https://github.com/mitodl/open-discussions/issues/3345

#### What's this PR do?
This PR fixes the channel tracker to ensure that events are not sent after the user moves away from the channel.
It also reverses the changes from https://github.com/mitodl/open-discussions/pull/3339 since the extra page views turned out to be automated traffic

Set GA_G_TRACKING_ID in your .env file. You can use G-B0EKT59MQF for testing.
Set the GA tracking id for a channel from the admin (http://localhost:8063/admin/channels/). You can use G-BEM7JCXH4D for testing.

Go to the networks tab and filter for G-BEM7JCXH4D. You should see calls to
`https://www.google-analytics.com/g/collect` when you are in the channel main page or viewing channel posts. You should not see calls when you are viewing a different channel or on other open pages.

If you filter to  G-B0EKT59MQF you should see calls to `https://www.google-analytics.com/g/collect` on all pages

### DIFF
--- a/static/js/hoc/withChannelTracker.js
+++ b/static/js/hoc/withChannelTracker.js
@@ -2,14 +2,7 @@
 import React from "react"
 import R from "ramda"
 
-const shouldLoadData = R.complement(
-  R.allPass([
-    // if channel name doesn't match
-    R.eqBy(R.path(["channel", "name"])),
-    // if URL path doesn't match
-    R.eqBy(R.path(["location", "search"]))
-  ])
-)
+const channelChanged = R.complement(R.eqBy(R.path(["channel", "name"])))
 
 export const withChannelTracker = (
   WrappedComponent: Class<React.Component<*, *>>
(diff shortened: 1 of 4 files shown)


mbertrand (commented):
I also happen to have `GA_TRACKING_ID` set on my local instance, and looks like there are 2 pageview requests sent to google, one for `GA_TRACKING_ID`, another for `GA_G_TRACKING_ID`.  Should only one of these be enabled?  

Screen Shot 2021-03-31 at 10 13 34 AM

abeglova:
That's intentional. Once we are confident that we are getting the same metrics, I will remove GA_TRACKING_ID and the analytics.js/reactGA code. We are migrating away from reactGA because it only works for older google tags that start with UA-. The new tracking code works both with the UA- tags and the new G- tags

mbertrand (approved):
👍

Issue #3361, opened 2021-04-02
Add a update_index command to search


abeglova:
The only command we currently have to make changes to the elasticsearch index is the `recreate_index` command which creates new indexes for every content type.

 Sometimes we need to reindex because we had a change in the data and not in the index mapping. For those situations we will create a new `update_index` command which will loop through the database and re-upsert objects to the existing indexes instead of creating new indexes.

This command should also have an optional flag to filter by content type

PR #3370, opened 2021-04-14
update-index-command

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3361

#### What's this PR do?
This pr adds a `update_index` command that updates indexes in place instead of recreating them

Allowed options are
  --all                 Update all indexes
  --comments            Update the comment index
  --courses             Update the course index
  --podcasts            Update the podcast index
  --podcastepisodes     Update the podcastepisode index
  --posts               Update the post index
  --profiles            Update the profile index
  --programs            Update the program index
  --userlists           Update the userlist index
  --resourcefiles   Update course resource files
There is also a `--course_platform` option that filters course and resource file updates to a specific platform

Test this by running `manage.py update_index` with different resource options

### DIFF
--- a/search/connection.py
+++ b/search/connection.py
@@ -80,19 +80,39 @@ get_default_alias_name = partial(make_alias_name, False)
 get_reindexing_alias_name = partial(make_alias_name, True)
 
 
-def get_active_aliases(conn, object_types):
+def get_active_aliases(conn, *, object_types=None, include_reindexing=True):
     """
     Returns aliases which exist for specified object types
 
     Args:
         conn(elasticsearch.client.Elasticsearch): An Elasticsearch client
         object_types(list of str): list of object types (post, comment, etc)
+        include_reindexing(boolean): include reindexing indexes if true
 
     Returns:
         list of str: Aliases which exist
     """
     if not object_types:
         object_types = VALID_OBJECT_TYPES
+    if include_reindexing:
+        return active_aliases_with_reindexing(conn, object_types)
+    else:
+        return [
+            alias
+            for alias in [get_default_alias_name(obj) for obj in object_types]
+            if conn.indices.exists(alias)
+        ]
+
+
+def active_aliases_with_reindexing(conn, object_types):
+    """
+    Returns aliases which exist for specified object types including reindexing aliases
+
+    Args:
+        conn(elasticsearch.client.Elasticsearch): An Elasticsearch client
+        object_types(list of str): list of object types (post, comment, etc)
+    """
+
     return [
         alias
         for alias_tuple in [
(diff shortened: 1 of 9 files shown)


search/connection.py
```diff
@@ -93,15 +93,22 @@ def get_active_aliases(conn, object_types):
...
+    else:
+        return [
+            alias
+            for alias_tuple in [
+                (get_default_alias_name(obj), get_reindexing_alias_name(obj))
+                for obj in object_types
+            ]
+            for alias in alias_tuple
+            if conn.indices.exists(alias)
         ]
```
rhysyngsun, line 111: This function is getting kind of complex now so I think it'd be worth pulling apart into some separate functions and composing them together:
```suggestion
    aliases = []
    for obj_type in object_types:
        aliases.append(get_default_alias_name)
        if include_reindexing:
            aliases.append(get_reindexing_alias_name(obj_type))
    return [alias for alias in alias if conn.alias.exists(alias)]
```

rhysyngsun (changes requested):

search/management/commands/update_index.py
```diff
@@ -0,0 +1,107 @@
...
+            PlatformType.ocw.value,
+            PlatformType.mitx.value,
+            PlatformType.micromasters.value,
+            PlatformType.bootcamps.value,
+            PlatformType.xpro.value,
+            PlatformType.oll.value,
+            PlatformType.see.value,
+            PlatformType.mitpe.value,
+            PlatformType.csail.value,
+        ]
```
rhysyngsun, line 29:
```suggestion
        allowed_course_platforms = [pt.value for pt in PlatformType]
```

search/tasks.py
```diff
@@ -689,6 +841,237 @@ def start_recreate_index(self, indexes):
     )
 
 
+@app.task(bind=True)
+def start_update_index(self, indexes, platform):
```
rhysyngsun, line 845: This task is fairly large, I think it would be cleaner to instead fan out into a set of per-object-type tasks and then have each of those fan out further using the per-type code below.

rhysyngsun (changes requested):

search/management/commands/update_index.py
abeglova: The array only includes course platform types: `youtube` and `podcast` are not included. But starting with all PlatformTypes then excluding is cleaner

rhysyngsun: Oh, ok :+1:

rhysyngsun (approved):
LGTM

PR #3398, opened 2021-05-12
set ocw-data-parser to 0.28.0 in requirements.in and run pip-compile

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

None

#### What's this PR do?
This PR updates `ocw-data-parser` to 0.28.0.  It is the result of changing that setting in `requirements.in` and running `pip-compile`.

Run OCW import on some courses and ensure that it does not fail.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -40,7 +40,7 @@ ipython
 lxml==4.6.3
 markdown2==2.3.9
 newrelic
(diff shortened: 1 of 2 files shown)


abeglova:
👍

Issue #3395, opened 2021-05-10
Don't include posts and comments from first-time and new users in mailings


briangrossman:
We should make an effort to prevent spam posts and comments from appearing in the digest emails sent to users.

- [ ] If the post or comment is the first one from a user, don't include it in the digest emails
- [ ] If the user who created the post has just joined MIT Open within the last day, do not include their posts or comments in the digest emails

This should prevent most of the instances of spam from ending up in the digests and give us time to remove it.

PR #3403, opened 2021-05-17
avoid new user posts in notification

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3395

#### What's this PR do?
This pr only includes posts in weekly and daily digests if it has been 24 hours since the post authors first post

Create a new user. 
Create a post by the new user. 
Create a post by an old user.
Run
```
send_daily_frontpage_digests() (or send_weekly_frontpage_digests depending on your user's setting)
```
Verify that you don't see the new user's post

You may need to delete some EmailNotification objects if you run this multiple times


notifications/notifiers/frontpage.py
```diff
@@ -88,7 +88,11 @@ def _posts_since_notification(notification_settings, notification):
         if not post.stickied and _is_post_after_notification(post, notification)
```
mbertrand, line 88: Could this be added here?

```python
and not post._self_post.exclude_from_frontpage_emails
```

instead of filtering them out later below?

mbertrand (commented):
Works great, just have one question

notifications/notifiers/frontpage.py
abeglova: No proxy_posts() replaces the reddit Submission object with an object that has ._self_post

mbertrand (approved):
👍

PR #3404, opened 2021-05-18
fix to digest email task queue

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
The current code queuing frontpage notifications in a separate celery queue was broken. This fixes the issue.

add
```
import time
@app.task
def sleep(seconds):
    time.sleep(seconds)
```

to `course_catalog/tasks.py`

Fill the celery queues with default priority tasks by running

```

for i in range(0,1000):
    sleep.delay(15)
```

from the shell run
```
send_daily_frontpage_digests.delay()
```

You should receive a notification email before the sleep tasks finish running

You may need to delete some EmailNotifiation objects if you run this multiple times

### DIFF
--- a/open_discussions/celery.py
+++ b/open_discussions/celery.py
@@ -23,8 +23,7 @@ app.conf.task_routes = {
     "notifications.tasks.send_frontpage_email_notification_batch": {
         "queue": "digest_emails"
     },
-    "notifications.tasks.test_send_daily_frontpage_digests": {"queue": "digest_emails"},
-    "notifications.tasks.test_send_weekly_frontpage_digests": {
-        "queue": "digest_emails"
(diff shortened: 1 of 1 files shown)


abeglova:
I think it's ok for `send_unsent_email_notifications` to be in the general queue. It's not a time consuming task and we want notifications letting moderators know about new tasks to be sent quickly. `send_unsent_email_notifications` spawns separate `send_email_notification_batch` and `send_frontpage_email_notification_batch` tasks and  `send_frontpage_email_notification_batch` gets properly queued in the `digest_emails` queue

mbertrand (approved):
👍

PR #3412, opened 2021-05-25
remove node-sass, use sass instead

aliceriot:

none

#### What's this PR do?

This removes the `node-sass` package, which depends on a native (C) module, and replaces it with `sass`, which is a pure-JS implementation of Sass (actually compiled from a Dart implementation). This means our frontend build has one less native code dependency, which is 💯  in my book! (libsass is also [deprecated](https://sass-lang.com/blog/libsass-is-deprecated), so there's that as well)

Everything should look normal across the site. I believe that the libsass implementation was actually slightly non-standard in a few ways, so it is _possible_ that some of our `.scss` code may be interpreted slightly differently now. So it's really important to check pretty thoroughly!


abeglova:
Tested thoroughly. Everything looks good

abeglova (approved):

Issue #3408, opened 2021-05-24
Add extra departments to search


abeglova:
When a course is cross listed in multiple departments, the course numbers from extra departments are saved as “linked_course_number”  in the raw json and "extra_course_number" in the parsed json.

We should store a list in the department field and return all the departments in the search

PR #3419, opened 2021-06-02
support multiple departments

abeglova:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3408

#### What's this PR do?
This pr indexes all the course departments to elasticsearch when a course is crosslisted to multiple departments

Make sure the migration works backwards and forwards

Run
```
from course_catalog.api import *

sync_ocw_courses(course_prefixes=["PROD/6/6.901/Fall_2005/6-901-inventions-and-patents-fall-2005/"], blocklist=[], force_overwrite=True, upload_to_s3=False)
```

Run ocw-hugo-themes with this branch for the index
Search for "Inventions and Patents"  
Verify that you see one  course and three departments

I wasn't able to add the data migration in the migration because django tries to save the array field to postgres with `[` brackets instead of `{` when the `Course` model is accessed through `apps.get_model()` which results in a database error. Updating the data works fine when `Course` is accessed normally. I plan to run the following code from the command line to backfill the data manually:

```
        from course_catalog.models  import Course 
        from course_catalog.constants import PlatformType

        for course in Course.objects.filter(platform=PlatformType.ocw.value).iterator():
            course_runs = course.runs.filter(
                raw_json__isnull=False,
                published=True
            )

            for course_run in course_runs:
                for extra_course_number_json in (course_run.raw_json.get("extra_course_number") or []):
                    department_number=extra_course_number_json.get("linked_course_number_col").split('.')[0]
                    if department_number not in course.department:
                        course.department.append(department_number)
                        course.save()
```


mbertrand (commented):
LGTM.  For the code to run afterward, unless you plan to reindex all courses, you could add a line to run `search.task_helpers.upsert_course(course.id)`.  Do you think it might be worth packaging that into a management command?

abeglova:
There is already a command tp reupsert courses without reindexing them
https://github.com/mitodl/open-discussions/blob/master/search/management/commands/update_index.py

abeglova:
I'm trying to run a reindex now because we don't currently have a way to remove course documents without reindexing. That won't be an issue for this PR

mbertrand:
Sorry, I meant a management command to run the code block in the PR description (for updating the departments).  But I guess it shouldn't have to be done more than once so that might be overkill.

mbertrand (approved):
👍

PR #3429, opened 2021-06-08
add CELERY_CELERYD_CONCURRENCY

abeglova:
#### Pre-Flight checklist
- [x] Settings

None

#### What's this PR do?
This adds a setting variable to set the max concurrency for celery workers

Celery goes by the number of CPUs on the machine to determine the default max concurrency.  Since the 2x workers are on a shared machine with 8 CPUs celery tries to run eight jobs at at time on the 2x workers even thought we only have access to a fourth of that machine. This causes the memory issues we have been observing.

The PM machines have two CPUs and have default max concurrency 2. 

The tests should pass


blarghmatey:
:+1: looks good

PR #3430, opened 2021-06-08
upgrade jsdom

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none, upgrades the dependency in #3425 

#### What's this PR do?

upgrades jsdom so as to pull in the latest version of [ws](https://www.npmjs.com/package/ws).

I had to make some modifications to one test file because of a change in how jsdom handles `document.activeElement`. Previously, jsdom's implemention would allow for elements which are not children of `document.body` to be focused, which is not standard behavior. In jsdom v15.2.1 this was changed to be in-line with standard behavior in browsers. Our test was relying on this nonstandard behavior, so I just had to make a small adjustment to ensure that in a few tests that enzyme's `mount` is mounting to a DOM node attached to `document.body`.

See this issue for details: https://github.com/jsdom/jsdom/issues/2723

tests should pass, that's really it.


abeglova (approved):
LGTM

Issue #3439, opened 2021-06-11
Use yield to stream course documents during ocw imports


abeglova:
Currently we download and serialize all ocw course documents for a course at once during imports. This is a problem for courses with very may documents.

Issue #3436, opened 2021-06-10
Avoid serializing all course_file objects for a course at once


abeglova:
We currently serialize all the document data for a course run before loading to elasticearch
https://github.com/mitodl/open-discussions/blob/master/search/indexing_api.py#L672

We should load and serialize only ELASTICSEARCH_INDEXING_CHUNK_SIZE documents at a time and probably also add a separate ELASTICSEARCH__DOCUMENT_INDEXING_CHUNK_SIZE env variable

PR #3442, opened 2021-06-15
avoid parsing all documents at once

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
- [x] Settings

closes https://github.com/mitodl/open-discussions/issues/3436
closes https://github.com/mitodl/open-discussions/issues/3439

#### What's this PR do?
Some of our courses have thousands of course documents, for example PROD/8/8.01/Fall_2016/8-01sc-classical-mechanics-fall-2016/

This PR updates the ocw update and reindex code to avoid loading all the course document data into memory at the same time.

Verify that the backpopulate_ocw_data, backpopulate_ocw_files, backpopulate_xpro_data and recreate_index --course commands run without errors


app.json
```diff
@@ -130,6 +130,10 @@
       "description": "Chunk size to use for Elasticsearch indexing tasks",
       "required": false
     },
+    "ELASTICSEARCH_DOCUMENT_INDEXING_CHUNK_SIZE": {
```
noisecapella, line 133: Any reason not to reuse `ELASTICSEARCH_INDEXING_CHUNK_SIZE` here? Is the intention to have a smaller chunk size for courses than for other indexed items?

search/indexing_api.py
```diff
@@ -669,21 +669,30 @@ def index_run_content_files(run_id, upate_only=False):
...
+            for content_file in ContentFile.objects.filter(pk__in=ids_chunk)
+            .select_related("run")
+            .prefetch_related("run__content_object")
+            .defer("run__raw_json")
+        )
+
+        index_items(
+            documents,
+            COURSE_TYPE,
+            upate_only,
```
noisecapella, line 691: While we're here can we rename this to `update_only`? It doesn't look like there are any calling functions which use this argument as a keyword argument so it should be a simple change hopefully

course_catalog/etl/loaders.py
```diff
@@ -666,28 +666,28 @@ def load_content_file(course_run, content_file_data):
...
             load_content_file(course_run, content_file)
-            for content_file in content_files_json
+            for content_file in content_files_data
         ]
         if course_run.published:
             search_task_helpers.index_run_content_files(course_run.id)
         else:
             search_task_helpers.delete_run_content_files(course_run.id)
-        return content_files
+        return content_files_ids
```
noisecapella, line 690: The return value should be updated in the docstring

course_catalog/etl/loaders.py
```diff
@@ -657,7 +657,7 @@ def load_content_file(course_run, content_file_data):
         content_file, _ = ContentFile.objects.update_or_create(
             run=course_run, key=content_file_data.get("key"), defaults=content_file_data
         )
-        return content_file
+        return content_file.id
```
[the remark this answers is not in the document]
noisecapella: Actually, should we return anything here? It looks like we only use the return value in `load_content_files`, and that function's return value isn't used anywhere, unless I'm missing something

search/indexing_api.py
```diff
@@ -669,21 +669,30 @@ def index_run_content_files(run_id, upate_only=False):
...
-        routing=gen_course_id(
-            run.content_object.platform, run.content_object.course_id
-        ),
-    )
+    run = LearningResourceRun.objects.get(pk=run_id)
+    content_file_ids = run.content_files.values_list("id", flat=True)
+
+    chunk_size = settings.ELASTICSEARCH_DOCUMENT_INDEXING_CHUNK_SIZE
+
+    for index in range(0, len(content_file_ids), chunk_size):
```
noisecapella, line 677: There is a function `open_discussion.utils.chunks` which may be useful here

noisecapella (changes requested):

noisecapella (approved):

PR #3450, opened 2021-07-07
Add course argument to filter backpopulate_ocw_data

### LINKED ISSUE #3449: Add an argument to run backpopulate_ocw_data on individual courses
As a developer, I'd like to have the option to backpopulate a single course so that I can update course state locally for testing purposes. This would save time because I would not need to run `backpopulate_ocw_data` on thousands of courses which would take hours.

#### Acceptance Criteria:

- [ ] Add an optional argument for course id to the `backpopulate_ocw_data` management command which causes only the course matching that string to be backpopulated

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Fixes #3449 

#### What's this PR do?
Adds a `--course` argument to backpopulate only courses which match the prefix string. This is meant to make it easier for developers to backpopulate only specific courses that they are working on, to avoid the hours-long process of backpopulating the whole catalog.

 - Run `./manage.py backpopulate_ocw_data --course 6-890-algorithmic --overwrite --skip-s3`. It should take about 5 minutes.
 - In a shell, find the `LearningResourceRun` for that course via `Course.objects.get(url__contains='6-890-algorithmic').runs.first()`. The `updated_on` value should be close to the current date and time.

### DIFF
--- a/course_catalog/management/commands/backpopulate_ocw_data.py
+++ b/course_catalog/management/commands/backpopulate_ocw_data.py
@@ -32,8 +32,8 @@ class Command(BaseCommand):
             help="skip uploading course files to s3",
         )
         parser.add_argument(
-            "--course",
-            dest="course",
+            "--course-url-substring",
+            dest="course_url_substring",
             required=False,
             help="If set, backpopulate only this course",
         )
@@ -41,7 +41,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         """Run Populate ocw courses"""
-        course_url_substring = options.get("course")
+        course_url_substring = options.get("course_url_substring")
         if options["delete"]:
             self.stdout.write("Deleting all existing OCW courses")
             for course in Course.objects.filter(platform="ocw"):
--- a/course_catalog/management/commands/backpopulate_ocw_data.py
+++ b/course_catalog/management/commands/backpopulate_ocw_data.py
@@ -31,10 +31,17 @@ class Command(BaseCommand):
             action="store_false",
             help="skip uploading course files to s3",
         )
+        parser.add_argument(
+            "--course",
+            dest="course",
+            required=False,
+            help="If set, backpopulate only this course",
+        )
         super().add_arguments(parser)
 
     def handle(self, *args, **options):
         """Run Populate ocw courses"""
+        course_id = options.get("course")
         if options["delete"]:
             self.stdout.write("Deleting all existing OCW courses")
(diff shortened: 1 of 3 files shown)


course_catalog/management/commands/backpopulate_ocw_data.py
```diff
@@ -45,12 +52,15 @@ def handle(self, *args, **options):
...
+                match_courses=course_id,
             )
             self.stdout.write(
-                "Started task {task} to get ocw course data w/force_overwrite={overwrite}, upload_to_s3={s3}".format(
+                "Started task {task} to get ocw course data "
+                "w/force_overwrite={overwrite}, upload_to_s3={s3}, course_id={course}".format(
                     task=task,
                     overwrite=options["force_overwrite"],
                     s3=options["upload_to_s3"],
+                    course=course_id,
```
abeglova, line 63: This should this something like `url_substring` since you are not matching on course id

abeglova: Or maybe `url_contains`

course_catalog/tasks.py
```diff
@@ -91,6 +95,20 @@ def get_ocw_data(
     ).Bucket(name=settings.OCW_CONTENT_BUCKET_NAME)
     ocw_courses = generate_course_prefix_list(raw_data_bucket)
 
+    total_course_count = len(ocw_courses)
+    if match_courses is not None:
```
abeglova, line 99: Does it make sense to pass this to `generate_course_prefix_list` and filter there?

course_catalog/management/commands/backpopulate_ocw_data.py
```diff
@@ -31,10 +31,17 @@ def add_arguments(self, parser):
             action="store_false",
             help="skip uploading course files to s3",
         )
+        parser.add_argument(
+            "--course",
```
abeglova, line 35: This should be `course_url_substring` as well

course_catalog/management/commands/backpopulate_ocw_data.py
```diff
@@ -31,10 +31,17 @@ def add_arguments(self, parser):
...
+            dest="course",
+            required=False,
+            help="If set, backpopulate only this course",
```
abeglova, line 38: Also the description needs to be updated

noisecapella:
@abeglova any other comments?

abeglova (approved):
LGTM

PR #3465, opened 2021-08-02
make ocw backpopulate restartable

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
https://github.com/mitodl/open-discussions/issues/3460

#### What's this PR do?
This PR makes ocw backpopulate jobs resilient to worker shut downs by setting `acks_late=True` and passing a `start_timestamp` to the task avoid re-importing the same courses after a restart

#### How should this be manually tested?
Run

```
docker-compose run web ./manage.py backpopulate_ocw_data --course 6-890-algorithmic --overwrite
```
The job should succeed and
```
from course_catalog.models import *
Course.objects.get(url__contains='6-890-algorithmic').runs.first().updated_on
```
should be updated to today

Run
```
from datetime import datetime, timedelta
run = Course.objects.get(url__contains='6-890-algorithmic').runs.first()
new_updated_on = run.updated_on + timedelta(days=10)
LearningResourceRun.objects.filter(pk=run.pk).update(updated_on=new_updated_on)
```

Run 
```
docker-compose run web ./manage.py backpopulate_ocw_data --course 6-890-algorithmic --overwrite
```

The job should succeed but you should see

```
Already synced. No changes found for PROD/6/6.890/Fall_2014/6-890-algorithmic-lower-bounds-fun-with-hardness-proofs-fall-2014/
```
in the logs.

 ```
Course.objects.get(url__contains='6-890-algorithmic').runs.first().updated_on
```
Should still be 10 days from now

PR #3469, opened 2021-08-09
fix ocw webhook

abeglova:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?
https://github.com/mitodl/open-discussions/pull/3465/files introduced a bug in the ocw webhook. This fixes it

Set OCW_WEBHOOK_KEY to something in your .env file

Using postman or similar post to 
`localhost:8063/api/v0/ocw_webhook/?webhook_key=key`

with body

```
{
    "Records": [
        {
            "s3": {
                "object": {
                    "key": "PROD/2/2.611/Fall_2006/2-611-marine-power-and-propulsion-fall-2006/0/1.json"
                }
            }
        }
    ]
}
```

You should get a 200 response.

Since `force_overwrite=False` on the webhook task, it won't actually update the course since the course didn't change on s3. 

You can temporarily set `force_overwrite=True` in line 405 in `course_catalog/views.py` and push to the webhook again. Then updated_on for the course should be updated.

### DIFF
--- a/course_catalog/views.py
+++ b/course_catalog/views.py
@@ -400,7 +400,7 @@ class WebhookOCWView(APIView):
                 get_ocw_courses.apply_async(
                     countdown=settings.OCW_WEBHOOK_DELAY,
                     kwargs={
-                        "course_prefix": prefix,
+                        "course_prefixes": [prefix],
                         "blocklist": blocklist,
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
:+1:

PR #3480, opened 2021-12-07
change video duration model field, make the migration for it

### LINKED ISSUE #3478: DataError: value too long for type character varying(10)
Sentry Issue: [OPEN-3RV](https://sentry.io/organizations/mit-office-of-digital-learning/issues/2712758661/?referrer=github_integration)

```
StringDataRightTruncation: value too long for type character varying(10)

  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "newrelic/hooks/database_psycopg2.py", line 35, in execute
    **kwargs)
  File "newrelic/hooks/database_dbapi2.py", line 25, in execute
    *args, **kwargs)

DataError: value too long for type character varying(10)

(20 additional frame(s) were not displayed)
...
  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "django/db/utils.py", line 89, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "newrelic/hooks/database_psycopg2.py", line 35, in execute
    **kwargs)
  File "newrelic/hooks/database_dbapi2.py", line 25, in execute
    *args, **kwargs)
```

alicewriteswrongs:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested

closes #3478

#### What's this PR do?

fixes a schema issue that was causing errors when importing youtube videos. In particular, the duration field we had set as `VARCHAR(10)`, but the [ISO 8601 duration string](https://en.wikipedia.org/wiki/ISO_8601#Durations) returned by the YouTube API can be up to 11 characters long.

These strings look like this:

```ts
`PT${numHours}H${numMinutes}M${numSeconds}S`
```

There are some videos on our list to import which are more than 10 hours long, so we'll have a duration string like `"PT17H28M57S"`, which is 11 characters long. I don't think we'll need more than 11 characters because 1) after setting it to 11 I am able to import all the videos we currently import and 2) youtube currently [doesn't allow uploading videos longer than 12 hours](https://support.google.com/youtube/answer/71673?hl=en&co=GENIE.Platform%3DDesktop) (I believe there are longer videos on the platform which were previously uploaded though).

Make sure you can reproduce the issue:

- On `master`, run the `backpopulate_youtube_videos` management command. It should exit at some point with an error similar to the one in the linked Sentry issue.

then check out this branch, run the migration, and confirm that it fixes the issue and the import can exit without erroring.

When I ran it locally, on `master` I was able to get through 2227 videos before it errored, and then after fixing the issue it pulled down 9001 before doing so. I don't know if the order is deterministic, so your results may be different.


alicewriteswrongs:
There are test failures on the python side, but I'm seeing that on other PRs too (e.g. https://github.com/mitodl/open-discussions/runs/4096300503?check_suite_focus=true), not sure whats up

abeglova (approved):
LGTM

PR #3593, opened 2022-06-22
Reorganize frontend(s) to use yarn workspaces for separation

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Related to https://github.com/mitodl/hq/discussions/95

#### What's this PR do?
A few related things:

1. Reorganizes the frontend code to use yarn workspaces, similar to https://github.com/mitodl/mitxonline/pull/514. This should let develop independent frontends (with shared dependencies, if needed).
2. Update django-webpack-loader to 1.4, and remove our customized stuff. The custom stuff was, I believe, for overwriting publicPath. We can do this with django-webpack-loader, and the upgraded version of django-webpack-loader supports multiple webpack configs, whcih we needed.
3. Updates to Node 16
4. Updates several npm dependencies. Mostly these were minor version bumps. A noteable exception is webpack, see comments in PR.
5. Adds a demo frontend "infinite-corridor".
There are almost no changes to application code.

#### How should this be manually tested?

Test in dev mode:

1. Ensure your .env file includes
    ```
    DEBUG=True
    NODE_ENV=development
	OPEN_DISCUSSIONS_HOSTNAME // hostname, e.g., yourname.od.odl.local ... defaults to od.odl.local
    ```
2. Rebuild docker to get new python dependency (updated version of django-webpack-loader): `docker compose up --build`
3. Visit http://yourname.od.odl.local:8063 ... you should see the MIT Open app. Try making a change in react code (e.g., frontends/open-discussions/src/components/IntroCard.js). The app should reload.
4. Test anything else you want in the frontend.
5. Visit http://yourname.od.odl.local:8063/infinite ... you should see a placeholder infinite corridor frontend. Try making changes in frontends/infinite-corridor/src/entry/root.tsx ... The page should reload.

Test in prod mode:

1. Ensure your .env file includes
    ```
    DEBUG=False
    NODE_ENV=production
	OPEN_DISCUSSIONS_HOSTNAME // hostname, e.g., yourname.od.odl.local ... defaults to od.odl.local
    ```
2. Restart docker.
3. View the json files in `webpack-stats/` and ensure the publicpaths point to `static/`.
4. View http://yourname.od.odl.local:8063 and http://yourname.od.odl.local:8063/infinite ... The apps should work but will not live reload.

#### Where should the reviewer start?
[...]

PR #3597, opened 2022-06-28
fix yarn postinstall script

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Followup to #3593 

#### What's this PR do?
Goal is to fix the deploy error here: https://github.com/mitodl/open-discussions/runs/7094209264?check_suite_focus=true#step:3:35012

This fixes the package.json postinstall script for open-discussions. Previously,
- the script only ran if node_env was not dev
- it referenced an incorrect yarn workspace name

Now it runs in prod and dev and does not reference the workspace at all. Running in dev seems OK since it only runs on install anyway.

The reason this slipped through review / my testing is: the build error was only visible if you were installing with node_env was set to something non-empty and not development. I believe in my testing I probably installed in dev mode, then ran the build script directly.

Delete node modules and run `docker compose run --rm watch yarn install`.  Install should succeed.

### DIFF
--- a/scripts/run-django-dev.sh
+++ b/scripts/run-django-dev.sh
@@ -5,7 +5,7 @@
 health_urls=(
     "http://watch:8052/health"
 )
-wait_time=300
+wait_time=600
 
 if [[ $NODE_ENV == "production" ]] ; then
     # kick off healthchecks as background tasks so they can check concurrently
--- a/frontends/infinite-corridor/package.json
+++ b/frontends/infinite-corridor/package.json
@@ -26,7 +26,7 @@
     "node": "16"
   },
   "scripts": {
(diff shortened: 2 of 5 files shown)


scripts/run-django-dev.sh
```diff
@@ -5,7 +5,7 @@
 health_urls=(
     "http://watch:8052/health"
 )
-wait_time=300
+wait_time=600
```
ChristopherChudzicki, line 8: If you're running in prod mode and node_modules is empty, 5 minutes was not enough time for me to install and build.

I'm not sure what a good timeout would be. If this does timeout, you can always just restart the `web` container.

abeglova (approved):
LGTM

Issue #3608, opened 2022-07-07
Update prettier


ChristopherChudzicki:
We should update Prettier from v1 to v2. Prettier v2 has better typescript support.

For example, `import type` and `export type` cause Prettier v1's parser to error. This was fixed in prettier v2 (https://prettier.io/blog/2020/03/21/2.0.0.html#typescript)

ChristopherChudzicki:
Closed by #3609

PR #3609, opened 2022-07-07
Fix CI commands, update prettier

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/open-discussions/issues/3608

#### What's this PR do?
This PR:
- fixes `yarn-run-with-fallback` ... previously its exit code was always 0 😱 
- updates prettier to v2 so we can use `import type` and `export type`.

Cause some formatting or linting error... e.g., add a semicolon or an unused import. And run `docker compose run --rm watch yarn run fmt-check` (or `lint-check`). The process should exit with code 1.

I have an incoming PR for the front page that uses `import type` and `export type`, having forgotten that Prettier v1 doesn't support it. Initially I tried to use v2 in the "new" workspaces and keep the old `open-discussions` on v1 so the diff wouldn't be so big, but this is a little complicated and I think not worth the effort. 

The reason that the diff is so big is that Prettier changed how it handles chained method calls from v1 to v2. See https://github.com/prettier/prettier/issues/7884 for details. _IMO the v1 way looked better for fluent APIs like enzyme has, but...whatever._

### DIFF
--- a/frontends/infinite-corridor/tsconfig.json
+++ b/frontends/infinite-corridor/tsconfig.json
@@ -1,6 +1,11 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "outDir": "./build"
+    "outDir": "./build",
+    "rootDir": "./src"
   },
+  "include": [
+    "./src/**/*.ts",
+    "./src/**/*.tsx"
+  ]
 }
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki:
Closing this. I'm going to do the CI fix and the prettier stuff separately.

Issue #3613, opened 2022-07-07
Infinite Corridor front page


ChristopherChudzicki:
#### Designs and Mockups
https://github.com/mitodl/hq/discussions/98

#### Acceptance Criteria:

- [ ] A search box . On enter, it sends users to the search page
- [ ] A list of fields, each a link to the field page

#### Out of Scope

- Field pages themselves

pdpinch:
This can be closed, right?

PR #3614, opened 2022-07-07
Extract CourseSearchbox (as Searchbox)

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Toward #3613 

#### What's this PR do?
This PR extracts CourseSearchbox to a new local package, `ol-search-ui`. It does not add any new functionality to infinite corridor.

Additionally, I set up themes for Styled Components... Good for colors, breakpoints, etc.

1. Visit http://yourname.od.odl.local:8063/learn/ ... make sure the search box still works. In particular, it should be able to be submitted by
    - clicking the magnifier icon
    - tabbing to the magnifier and hitting enter
    - hitting enter on the search input
    - hitting the "Explore" button below search input *In this case, there seems to be an existing bug on rc that the text is not preserved when redirected to search page*
    - **New:** non-empty text should be clearable. *Previously this only worked on the search page*
2. On the search page, the input at top should also work... Check similar things.

    This does fix a small bug where tabbing to the clear icon on RC and hitting enter throws a console error and does not clear the text; now it works.


package.json
```diff
@@ -7,7 +7,7 @@
     "scripts": {
         "foreach-with-fallback": "yarn workspaces foreach --verbose --topological exec node ${PROJECT_CWD}/scripts/yarn-run-with-fallback.js $0 $1",
         "postinstall": "yarn run build",
-        "start-dev": "yarn workspaces foreach --verbose --parallel --interlaced run start-dev",
+        "start-dev": "yarn workspaces foreach --verbose --parallel --jobs unlimited --interlaced exec node ${PROJECT_CWD}/scripts/yarn-run-with-fallback.js start-dev global:start-dev",
```
ChristopherChudzicki, line 10: By default, yarn limits the number of parallel jobs, which is a problem in dev mode since the processes never exit.

I'm not sure if this "run `tsc --watch` for each package in parallel" approach is great... I suspect Typescript has a better way to help manage this. Something to look into, but not right now, I think.

frontends/open-discussions/src/pages/CourseIndexPage.js
```diff
@@ -103,13 +103,13 @@ export default function CourseIndexPage({ history }: Props) {
           }}
         >
           <ResponsiveWrapper onlyOn={[TABLET, DESKTOP]}>
-            <div className="link-wrapper">
-              <Link className="link-button" to={COURSE_SEARCH_URL}>
+            <div className="link-wrapper d-flex justify-content-end mt-3">
+              <Link className="link-button me-0" to={COURSE_SEARCH_URL}>
```
ChristopherChudzicki, line 107: Why these new classes?

Previously, CourseSearchbox applied `course-searchbox` class, and that class affected the style of the the link wrappers and link buttons inside of it. 

The new styling for Searchbox does not affect the children in this way. Instead, children need to be responsible for styling themselves.

This was the only case that is affected, and the child styling is easy to get with a few of bootstrap's utility classes.

frontends/open-discussions/src/Router.js
```diff
@@ -27,7 +29,9 @@ export default class Router extends React.Component<Props> {
         <Provider store={store}>
           <ReduxQueryProvider queriesSelector={getQueries}>
             <ReactRouter history={history}>
-              <ScrollToTop>{children}</ScrollToTop>
+              <ThemeProvider theme={combinedTheme}>
+                <ScrollToTop>{children}</ScrollToTop>
+              </ThemeProvider>
```
ChristopherChudzicki, line 34: When I moved CourseSearchbox to typescript, I got rid of the scss styling and used Styled Components instead. Some colors / media queries are part of a theme now, so we need to provide the theme.

frontends/ol-search-ui/src/components/Searchbox.tsx
```diff
@@ -0,0 +1,191 @@
...
+  onClear: React.MouseEventHandler
+  value: string
+} & SearchboxCommon
+export type SearchboxProps = SearchboxUncontrolled | SearchboxControlled
+
+/**
+ * Renders a searchbox with Search and Clear icons, and optional child content.
+ * Can be used as a controlled or uncontrolled input.
+ */
+const Searchbox = (props: SearchboxProps): JSX.Element => {
```
ChristopherChudzicki, line 106: This was previously [CourseSearchbox](https://github.com/mitodl/open-discussions/blob/7973c4375bd7bc794da924aa4a0ddb1866e3d0a1/frontends/open-discussions/src/components/CourseSearchbox.js#L16)

I did make some logical changes within this component as I was trying to add types, but nothing major and it's all pretty thoroughly tested.

For example, on of the changes I made was using real buttons for the search and clear buttons. Before they were just icons. (Now icons wrapped in a button.)

This should be more accessible and easier for us... E.g., before the clear icon had:
```tsx
          <i
            className="material-icons clear-icon"
            onClick={onClear}
            onKeyPress={e => {
              if (e.key === "Enter" && onClear) {
                onClear()
              }
            }}
            tabIndex="0"
          >
            clear
          </i>
```
But with a real button you don't need separate `onKeyDown` and `onClick` handlers. Buttons can be clicked with the Enter key. (Or spacebar). So now it's just:
```tsx
          <PositionedButton onClick={onClear} rightAlign>
            <SearchboxIcon className="material-icons clear-icon">
              clear
            </SearchboxIcon>
          </PositionedButton>
```

frontends/ol-search-ui/src/components/Searchbox.tsx
```diff
@@ -0,0 +1,191 @@
+import { ValidationError } from "ol-forms"
+import React, { useCallback, useState } from "react"
+import styled, { css } from "styled-components"
+
+const SearchboxContainer = styled.div`
+  display: flex;
```
abeglova, line 6: I don't think we want the styles in the component  code. Can this be moved to an scss file? Mixing css in javascript files isn't great

abeglova (changes requested):
Back to you.

frontends/ol-search-ui/src/components/Searchbox.tsx
ChristopherChudzicki: Re ... Some of these comments from slack (Though I reordered some of these.)

> 1. Is the first place we’re using Styled Components right?
> 2. Mixing css in javascript files isn't great
> 3. I don’t think it’s a good idea generally, since it makes it harder to re-use css code among components
> 4. It's very likely that the styling we eventually have for the infinite corridor site will be different than open discussions
> 5. And also having styles in open-discussions work differently from all our other codebases is undesirable

**Re 1:** This isn't the first place we are using styled-components, though looking back, our usage is much more limited than I realized. We were originally using it in PageBanner.js (which I converted to tsx) and AuthEmailForm.js. We also use it (very limitedly) in mitxonline.

**Re 2-4:** Re mixing css and javascript... One thing that I've heard people say about this in the past is that this is mixing concerns, and we want separations of concerns.

In fact, I think this sort of pattern is a much better separation of concerns: We co-locate the styling specific to Searchbox with the javascript code specific to Searchbox. (Whereas previously the styling and layout/js were tightly coupled but in very different areas of our codebase.) Before, we were separating technologies. Now, we are isolating (and separating) concerns: All the searchbox code is together. 

I don't mean to suggest I've done this perfectly, by the way. E.g., the `height` properties in `SearchboxContainer` is a little awkward and that's probably better handled some other way, but I'd rather cross that bridge when we get to it. To a large extent, this the css is searchbox specific and makes sense to have with the searchbox component. (Example: the alignment of the icons within the textbox.¹) 

Re sharing styles and different styles for open-discussions / infinite-corridor: We can definitely do this with styled components via themes. I set it up in this PR... e.g., see [color based on theme here](https://github.com/mitodl/open-discussions/pull/3614/files#diff-4fd28178e751ae5b887c461130e5b555a8654a9b22d29e84e13104fb92148697R39)

**Re 5**: Yeah, that's definitely a concern. As mentioned, I just saw we had styled-components install previously, I did not realize our usage was so limited.

**All in all**: I definitely think we should be isolating component-specific and co-locatting all the relevant code for a single component. Styled components is a pretty easy way to do this. There are other solutions, e.g., css / scss modules.

------

¹I mean, the alignment of icons in the searchbox could certainly be abstracted to a different component... InputWithIcons. But that's more refactoring than I think we should do at the moment. Or better yet we could find a library for that.

ChristopherChudzicki: @abeglova I just put up one more commit that moves some of the course searchbox styling back to open-discussions. 

I've tried to keep the stuff that really is searchbox specific (example: css related to positioning the icons) isolated / co-located with the component. But now you can apply a className the to component to set stuff like font size, border radius. Some things maybe could still be moved out (e.g., the 500px max-width) but imo we should defer that movement till its needed.

abeglova (approved):
👍

PR #3619, opened 2022-07-11
Infinite Corridor front page

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#3613 

#### What's this PR do?
This PR sets up a front page, router, and some jest-related testing-utils.

1. Create a few fields if you don't have any. Do this via the form at http://yourname.od.odl.local:8063/api/v0/fields as a superuser. 
2. Visit http://yourname.od.odl.local:8063/infinite and you should see a list of the fields you created.
3. Each field should link to a placeholder page.
4. Back at the front page, submitting text in the searchbox should send you to the search page with the correct query param.

⚠️ This PR does add a new dependency, [react-query](https://www.npmjs.com/package/react-query). 

I have heard us (Alice before she left; @rhysyngsun via https://github.com/mitodl/ol-npm-libraries#best-practices, possibly via discussions, too) say that this is the direction we should move, though I'm not sure to what extent we've actually used this library before in our apps.

See https://github.com/mitodl/hq/discussions/153 for more, but its usage is basically:
```ts
// status is one of "loading", "error", or "success"
const { data, status, ...lotsOfOtherProperties } = useQuery(queryKey, queryFn, options)
```
where `queryFn` is a function that returns a promise resolving to `data`. We could use any data-fetching tool (native fetch; axios, superagent, etc) to write the `queryFn` component. 

**Update:** Now using axios.


frontends/ol-util/src/components/PageBanner.tsx
```diff
@@ -76,7 +76,6 @@ const BannerPageWrapper = styled.div`
 `
 
 const BannerPageHeader = styled.div`
-  margin-bottom: 20px;
   ${imageWrapperHeight};
 `
```
ChristopherChudzicki, line 80: See https://github.com/mitodl/open-discussions/pull/3619/commits/241ed7c555378d9d0a602c6211c9959ee0d6798b 

> This change should have no visible effect on open-discussions and simplies the css there, and makes BannerPageHeader easier to reuse.
> 
> In open-discussions, BannerPageHeader is used in 6 places.
>
> - in 5 of the 6 places, the margin-bottom: 20px was doing nothing because the sibling element had class `.main-content`, which specifies margin-top: 20px. And the adjacent 20px margins were being collapsed.
>
>    For info about collapsing margins, see https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing or or https://www.joshwcomeau.com/css/rules-of-margin-collapse/ for a deep dive.
> - In ChannelHeader, the 20px was doing something, but only because `.main-content.channel-page` specified 0px margin. So just remove that override, and now 6/6 of the bottom-margins were unncessary.

frontends/infinite-corridor/src/scss/button.scss
```diff
@@ -0,0 +1,20 @@
+@use "sass:color";
+
+.ol-btn {
```
ChristopherChudzicki, line 3: For the moment this button style (and `one-column`) below are mostly copied from open-discussions.

I was initially reluctant to move lots of styling to shared areas because I wanted to include only what we actually needed in infinite-corridor... I'd like to work on that a bit tomorrow (in a different PR), but IMO we should review this PR as-is for now, since I think it has stuff that will be helpful to other people.

frontends/ol-network-interface/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+class NetworkError extends Error {
+  status: number
+
+  constructor(status: number, msg: string) {
+    super(msg)
+    this.status = status
+  }
+}
+
+const makeRequest: NetworkInterface["makeRequest"] = async (
```
ChristopherChudzicki, line 51: This implements the `NetworkInterface` above with native fetch.

I have not, for the moment, included any tests for this. I do think we should use `react-query`, and I do think we should interact with it through some standard adapter like this `NetworkInterface`. I am a bit torn about whether
- we should implement network interface with library like axios or superagent
- or keep this minimal fetch wrapper.

I lean toward keeping the minimal fetch implementation: axios and superagent both have a fair amount of stuff we don't need. (Some of the "stuff we don't need" is because react-query manages for us, e.g., request retries.) *Open-discussions is already using superagent via `redux-query-interface-superagent`. But redux-query-interface-superagent specifies superagent v5, which is deprecated: https://www.npmjs.com/package/superagent/v/5.3.1*

Anyway, if people are ok keeping this fetch implementation, i could add simple tests for it 🤷

frontends/infinite-corridor/src/pages/HomePage.tsx
```diff
@@ -0,0 +1,73 @@
...
+            </button>
+          </div>
+        </LearningResourceSearchbox>
+      </BannerPageHeader>
+      <div className="mdc-layout-grid one-column">
+        <h2>MIT Fields</h2>
+        <ul>
+          {fieldsList.data?.results.map((field) => (
+            <li key={field.name}>
+              <Link to={`${currentPath}/fields/${field.name}`}>
```
abeglova, line 62: The links take me to http://localhost:8063/infinite//fields/{name} (extra /)

abeglova (approved):
LGTM - one small comment about fixing the url

frontends/infinite-corridor/src/pages/HomePage.tsx
ChristopherChudzicki: Good catch.

The way I had it set up, routing was too sensitive to trailing slashes... it worked from `http://localhost:8063/infinite` but not from `http://localhost:8063/infinite/`.

New commit should fix that.

PR #3629, opened 2022-07-14
Cc/cards

ChristopherChudzicki:
#### Pre-Flight checklist

- [ ] Testing
  - [x] Code is tested

Related to #3602

#### What's this PR do?
Extracts `Card` and styling to to `old-util

Check that cards still work/look the same.

I would suggest reviewing commit by commit. 

This PR extracts Card to ol-util AND replaces its usages in open-discussions with the ol-util version. I don't anticipate doing much "replace existing usages" moving forward, but since this was already done, seems ok to keep it.


frontends/ol-util/assets/card.scss
```diff
@@ -0,0 +1,44 @@
+@use "variables";
```
ChristopherChudzicki, line 1: `@use` is the Sass module system version of `@important`, see https://sass-lang.com/blog/the-module-system-is-launched for its announcement.

Whereas `@import` imports stuff into the global scope, `@use` provides variables onto a namespace (e.g., `variables.$card-background` below).

Sort of related... if you `@import` the same thing many times, it gets included in the output css multiple time. With `@use`, it just gets included once. (Which make sense, since you'd need to `@use` the variables thing a bunch more). 

Apparently `@import` is being deprecated, but not super soon. https://github.com/sass/sass/blob/main/accepted/module-system.md#timeline

One big difference is that `@use` declarations have to be at the top of a sass file.

frontends/ol-util/assets/card.scss
```diff
@@ -0,0 +1,44 @@
...
+    &:hover {
+      box-shadow: 3px 4px 12px 0 rgba(0, 0, 0, 0.36);
+    }
+  
+    .card-contents {
+      margin: 0;
+      padding: 20px;
+    }
+  
+    .card-title {
```
ChristopherChudzicki, line 18: This class was originally called `.title`. After taking a pretty thorough look at our CSS rules for `.title`, it's was clear to me that almost all of our `.title` are completely unrelated to card titles. For example, `channels.scss` declares a rule targeting `.title-and-headline .title`, but the only usage of `title-and-headline` is outside of cards, so that rule never affects cards.

frontends/ol-util/src/components/Card.tsx
```diff
@@ -0,0 +1,38 @@
+import React from "react"
+import classnames from "classnames"
+
+interface CardProps {
```
ChristopherChudzicki, line 4: The original version of `Card` included an `onClick` prop. It was never used.

frontends/ol-util/src/components/Card.tsx
```diff
@@ -0,0 +1,38 @@
...
+  borderless,
+  title,
+  children,
+}) => {
+  return (
+    <div
+      className={classnames("card", className, {
+        borderless,
+        "persistent-shadow": persistentShadow,
+      })}
```
ChristopherChudzicki, line 28: We have a whole bunch of custom `getClassName` logic throughout open-discussions. This is [`classnames`](https://www.npmjs.com/package/classnames), it's great.

PR #3634, opened 2022-07-15
Change behavior of course_ui_enabled flag

### LINKED ISSUE #3631: Make /learn available on prod but hide from home page
Modify the behavior of the `course_ui_enabled` flag to only hide the "Learn at MIT" section on the home page.  The `/learn` urls should be available and work regardless of the flag.  The purpose of this is to enable authorized users (members of `settings.STAFF_MOIRA_LISTS` moira lists) to create public lists that will be added to infinite corridor field pages.

mbertrand:
#### Pre-Flight checklist
- [ ] Testing
  - [ ] Code is tested

Closed #3631 

#### What's this PR do?
Changes the behavior of `course_ui_enabled` so that if `False`, it only hides the "Learn at MIT" section on the home page.  The `/learn` urls should always be accessible now regardless of the setting.

- Set `FEATURE_COURSE_UI=True` in your .env:
  - Go to the discussions home page, there should be a `Learn at MIT` section on the right.
  - Go to `../learn` and `../learn/search`, they should be reachable and have the expected content.
- Set `FEATURE_COURSE_UI=False` in your .env:
  - Go to the discussions home page, there should not be a `Learn at MIT` section on the right.
  - Go to `../learn` and `../learn/search`, they should be reachable and have the expected content.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,8 @@ COPY apt.txt /tmp/apt.txt
 RUN apt-get update
 RUN apt-get install -y $(grep -vE "^\s*#" apt.txt  | tr "\n" " ")
 RUN apt-get update && apt-get install libpq-dev postgresql-client -y
+# The following is required for moira requests to work within docker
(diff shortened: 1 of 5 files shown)


ChristopherChudzicki (approved):
👍

Issue #3595, opened 2022-06-23
Infinite Corridor: Search 


pdpinch:
Like https://discussions-rc.odl.mit.edu/learn/search but without any facets. 

We are still in discussion with users about facets, so it would be best to start with none. We plan to add some facets, like "field" after the first iteration. 

#### Designs and Mockups

something like this: 

<img width="1394" alt="Screen Shot 2022-06-23 at 9 57 34 AM" src="https://user-images.githubusercontent.com/430126/175317298-cffbf6e9-e19f-4b56-a869-e43f9f1ef756.png">

#### Acceptance criteria

- [ ] Only show courses in search results, no videos, podcasts, etc. 
- [ ] Only courses from mitx , ocw, OLL, and MITx online. No xPRO, MIT PE, etc. 

#### Out of scope

- add mitxonline courses to the index

pdpinch:
this can be closed, right?

PR #3645, opened 2022-07-20
infinite search

abeglova:
#### Pre-Flight checklist

Screen Shot 2022-07-20 at 11 06 26 AM
- [ ] Testing
  - [ ] Code is tested

https://github.com/mitodl/open-discussions/issues/3595

#### What's this PR do?
This PR creates the first iteration of infinite corridor search. Right now it's just a text search that filters for programs and courses. Styling and facets will come in a followup pr

Go to /infinite/search. Verify that you can use the text search


frontends/infinite-corridor/src/pages/SearchPage.tsx
```diff
@@ -1,11 +1,152 @@
...
-      Search params: <code>{searchParams.toString()}</code>
+    <div>
+      <BannerPageWrapper>
+        <BannerPageHeader tall compactOnMobile>
+          <Searchbox
+            className="course-searchbox"
+            onChange={updateText}
+            value={text || ""}
+            onClear={clearText}
+            //@ts-expect-error - I'm not sure how to fix this
```
abeglova, line 110: I was having a lot of trouble getting the types for the onSubmit in SearchInput.tsx to match the event type from course-search-util. I would appreciate help fixing this error

frontends/infinite-corridor/src/pages/SearchPage.tsx
```diff
@@ -1,11 +1,152 @@
...
+    [
+      setSearchResults,
+      setTotal,
+      setCompletedInitialLoad,
+      setRequestInFlight,
+      setSearchApiFailed
+    ]
+  )
+
+  const search = async (params: SearchQueryParams) => {
```
ChristopherChudzicki, line 81: This could be moved outside the component definition.

frontends/infinite-corridor/src/pages/SearchPage.tsx
ChristopherChudzicki: The types in `course-search-util` are wrong. I've been meaning to make a PR.

`course-search-util` *says* that `onSubmit` should be a function `(e: React.FormEvent<Element>) => void`, but in reality onSubmit is triggered from a few different ways, and none of them are FormEvents. (They're click events; they previously were Keyboard events and click events).

Within course-search-utils itself, the only things that are used on the submit event are `event.target.value` and `event.preventDefault`, and the preventDefault is called unnecessarily.

frontends/ol-search-ui/src/lib/search.tsx
```diff
@@ -0,0 +1,131 @@
+export type LearningResourceSummary = {
```
ChristopherChudzicki, line 1: This file should just be `ts`, not `tsx`

frontends/ol-search-ui/src/lib/search.tsx
```diff
@@ -0,0 +1,131 @@
...
+  id: number
+  title: string
+  image_src: string | null
+  platform: string | null | undefined
+  topics: Array<CourseTopic>
```
ChristopherChudzicki, line 6: **Suggestion:** Here and elsewhere, `CourseTopic[]` instead of `Array<CourseTopic>`.

Both work, the `CourseTopic[]` notation is more common in TS, I think. (Possibly in early versions of TS only the Array notation was supported.)

frontends/ol-search-ui/src/lib/search.tsx
```diff
@@ -0,0 +1,131 @@
...
+  overrideObject: overrideObjectProps = {}
+): LearningResourceSummary => ({
+  id:          result.id,
+  title:       result.title,
+  image_src:   result.image_src,
+  object_type: result.object_type,
+  lists:       result.lists,
+  is_favorite: result.is_favorite,
+  offered_by:
+    "offered_by" in result && result.offered_by ? result.offered_by : [],
```
ChristopherChudzicki, line 124: I think just `offered_by: result.offered_by ?? []` would work fine.

frontends/infinite-corridor/src/test-utils/setupJest.ts
```diff
@@ -9,9 +11,6 @@ jest.mock("axios", () => {
   }
 })
 
-// This should be the only top-level hook.
-// eslint-disable-next-line mocha/no-top-level-hooks
 afterEach(() => {
   jest.resetAllMocks()
-  resetApi()
```
ChristopherChudzicki, line 16: **Requested Change** Could you add this line back.

frontends/infinite-corridor/src/factories/search.ts
```diff
@@ -0,0 +1,148 @@
...
+        full_name:  casual.name
+      }
+    ],
+    prices: [{ mode: "audit", price: casual.integer(1, 1000) }]
+  }
+}
+
+export const makeSearchResponse = (
+  //@ts-expect-error SETTINGS throws an error
+  pageSize: number = SETTINGS.search_page_size,
```
ChristopherChudzicki, line 112: **Requested Change**

In `frontends/infinite-corridor/src/types/global.d.ts`, could you add
```ts
/* eslint-disable no-var */

/**
 * Settings injected by Django
 */
interface SETTINGS {
  search_page_size: number;
}
export declare global {
  declare var SETTINGS: SETTINGS
}
```

Then TS will know SETTINGS is globally available.

IMO we should make settings globally available in infinite-corridor, but not in say, `ol-search-ui`. We can pass the embedly key and ocw base url as props to LEarningResourceCard. But I don't have a strong opinion about this.

ChristopherChudzicki (changes requested):
Search is working for me 👍 

I made two small requested changes, and left a bunch of other comments.

As a broader comment: I really think we should avoid writing any tests for Infinite Corridor with Enzyme.

We should use React Testing Library instead. See https://github.com/mitodl/ocw-studio/issues/1275 and https://github.com/mitodl/hq/discussions/17. The tldr is:
- React Testing Library (RTL) is officially recommended by the react team
- RTL tests are more realistic since RTL tests that interact with the DOM (like a user would) rather than with React props
- Enzyme relies on react internals, and its official version does not support React 17 or 18.

If you haven't used RTL before it is a bit of a mindset shift from Enzyme since you can't access component props directly. But a result of that is that the tests are much more resilient to refactoring.

IMO, we should
- merge this (with the minor changes requested, and whatever other changes you want to make)
- not add any new enzyme tests to Infinite Corridor
- re-write the two enzyme tests in RTL and remove Enzyme from Infinite Corridor

frontends/infinite-corridor/src/components/LearningResourceCard.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { makeSearchResult } from "../factories/search"
+import {
+  searchResultToLearningResource,
+  readableLearningResources
+} from "ol-search-ui"
+
+describe("LearningResourceCard component", () => {
+  const render = object => mount(<LearningResourceCard object={object} />)
+  ;["course", "program"].forEach(resourceType => {
```
ChristopherChudzicki, line 13: We should use `test.each` for this (or `it.each`). There's a few supported syntaxes... I like:
```
it.each([
  { resourceType: "course" },
  { resourceType: "program" },
])("should render the things we expect for a $resoureType", () => {...})
```

abeglova:
Thanks @ChristopherChudzicki back to you

frontends/infinite-corridor/src/pages/SearchPage.tsx
```diff
@@ -1,11 +1,145 @@
...
+            autoFocus
+          />
+        </BannerPageHeader>
+      </BannerPageWrapper>
+      <InfiniteScroll
+        hasMore={from + pageSize < total}
+        loadMore={loadMore}
+        initialLoad={from === 0}
+        loader={
+          <div className="loader" key={0}>
```
ChristopherChudzicki, line 114:
```suggestion
          <div className="loader">
```

frontends/ol-search-ui/src/lib/searchCardComponents.tsx
```diff
@@ -0,0 +1,148 @@
+import moment from "moment"
```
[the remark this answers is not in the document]
abeglova: I think it's helpful to separate the helpers into separate files based on what they do. So I think lib is the /random_helpers/ folder. IDK if there is a better name for it

frontends/infinite-corridor/src/pages/SearchPage.tsx
abeglova: This is needed for the infinite scroll. I get
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `InfiniteScroll`. It was passed a child from SearchPage. See https://fb.me/react-warning-keys for more information.
    in div (created by SearchPage)
    in InfiniteScroll (created by SearchPage)
    in div (created by SearchPage)
    in SearchPage (created by App)
    in Route (created by App)
    in Router (created by App)
    in QueryClientProvider (created by App)
    in Fe (created by App)
    in div (created by App)
    in App
```
If I remove it

PR #3641, opened 2022-07-19
Cc/frontpage round 2

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [x] Mobile width screenshots
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
#3630 

#### What's this PR do?
This PR makes the infinite-corridor front page match https://projects.invisionapp.com/d/main#/console/22082155/468411241/preview, in particular:
- new header
- nicer search box
- field link list
- styling 

#### How should this be manually tested?
1. *If you want the thumbnail photos to actually work, get `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_STORAGE_BUCKET_NAME` from rc.
2. Create some fields via admin panel at http://cac.od.odl.local:8063/admin/channels_fields/fieldchannel/
3. visit http://cac.od.odl.local:8063/infinite ... you should see the homepage.
4. Check that clicking a link goes to the field page (which is a placeholder right now)
5. Check that the searchbox works as expected
6. **Check /learn:** Visit http://cac.od.odl.local:8063/learn and http://cac.od.odl.local:8063/learn/search and check that the searchbox still works correctly. (It's now using the new search input from).

#### Context
The biggest change here is related to the search input. Previously `ol-search-ui` had a `Searchbox` component which `/learn` was using. That component was our custom-built searchbox with icons at start/end. `Searchbox` was kind of annoying because it (A) some container divs and (B) also the search text input. So I broke it into `Searchbox` and `SearchInput`.

Putting icon buttons at the start/end of a text field is nontrivial, so seems best to leave this to a library....So the new SearchInput uses [`@mui`](https://www.npmjs.com/package/@mui/material) ... "Material UI". Open-discussions was already using uses Material UI under its old name, `@material`. (the name changed to `@mui` in  .... v5 iirc). 

#### Screenshots (if appropriate)

infinite_homepage_desktop infinite_homepage_mobile

Issue #3630, opened 2022-07-14
Infinite Corridor Front Page, round 2


ChristopherChudzicki:
#### Designs and Mockups

https://projects.invisionapp.com/share/EZ12VH612GP8#/screens

#### Acceptance Criteria:

- [ ] Header at top
- [ ] Styling, including: center input, horizontal link list, etc
- [ ] icon/thumbnail should show up in with the links to fields

#### Out of Scope

- Designing those neat icons.

pdpinch:
And this one?

PR #3647, opened 2022-07-20
Infinite Corridor Front Page Round 2

ChristopherChudzicki:
_Same as #3641, but rebased after linting changes #3644_

#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#3630 

#### What's this PR do?
This PR makes the infinite-corridor front page match https://projects.invisionapp.com/d/main#/console/22082155/468411241/preview, in particular:
- new header
- nicer search box
- field link list
- styling 

1. *If you want the thumbnail photos to actually work, get `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_STORAGE_BUCKET_NAME` from rc.
2. Create some fields via admin panel at http://cac.od.odl.local:8063/admin/channels_fields/fieldchannel/
3. visit http://cac.od.odl.local:8063/infinite ... you should see the homepage.
4. Check that clicking a link goes to the field page (which is a placeholder right now)
5. Check that the searchbox works as expected
6. **Check /learn:** Visit http://cac.od.odl.local:8063/learn and http://cac.od.odl.local:8063/learn/search and check that the searchbox still works correctly. (It's now using the new search input from).

#### Context
The biggest change here is related to the search input. Previously `ol-search-ui` had a `Searchbox` component which `/learn` was using. That component was our custom-built searchbox with icons at start/end. `Searchbox` was kind of annoying because it (A) some container divs and (B) also the search text input. So I broke it into `Searchbox` and `SearchInput`.

Putting icon buttons at the start/end of a text field is nontrivial, so seems best to leave this to a library....So the new SearchInput uses [`@mui`](https://www.npmjs.com/package/@mui/material) ... "Material UI". 

infinite_homepage_desktop infinite_homepage_mobile


frontends/infinite-corridor/src/scss/combined.scss
```diff
@@ -1,28 +1,28 @@
 @use "material-components-web/dist/material-components-web";
-@use "ol-util/assets/variables.scss";
-
 @use "ol-util/assets/card";
 
-@use "button";
-@use "columns";
+@use "theme";
+@use "button" with ( $color: theme.$color-primary );
```
ChristopherChudzicki, line 5: @abeglova Btw re styles... `@use` seems great. It replaces `@import` in Sass's new module system... `@use <modulename> with ( <overrides> )` seems like it could be very useful for us.

Like, LearningResourceCard can specify its own styles, but we can override the variables it uses if we need to.

frontends/infinite-corridor/package.json
```diff
@@ -7,22 +7,29 @@
     "url": "https://github.com/mitodl/mit-open.git"
   },
   "dependencies": {
+    "@emotion/react": "^11.9.3",
+    "@emotion/styled": "^11.9.3",
+    "@mui/icons-material": "^5.8.4",
+    "@mui/material": "^5.9.0",
```
ChristopherChudzicki, line 13: These are the packages required for [`@mui`](https://github.com/mui/material-ui), see https://github.com/mitodl/hq/discussions/170

I originally thought we were already using an old version of this library, else I would have written the above discussions post sooner. My confusion stems from the fact that
- we do use `@material`
- `@mui` used to be called `@material-ui`

So I thought we were using the old version of `@mui`.

abeglova (approved):
LGTM!

Issue #3633, opened 2022-07-15
Support field page featured courses, lists, description (backend)


mbertrand:
Update the backend to support additional attributes for field pages:
- description
- featured_list
- lists with ordering

Also modify the API to support returning all public lists

Issue #3632, opened 2022-07-15
Appearance configuration for field pages


mbertrand:
Add a page/tab to edit the appearance (header banner + cover image) of a field page, similar to that for discussion channels.

mbertrand:
Banner and avatar will be handles via django admin for now

PR #3648, opened 2022-07-20
Featured list, lists and subfields for FieldChannel

mbertrand:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Part of https://github.com/mitodl/open-discussions/issues/3633
Part of #3632 

#### What's this PR do?
- Adds a few new attributes to `FieldChannel` (`lists`, `featured_list`, `subfields`) as well as new models to support ordering of lists and subfields per `FieldChannel`
- Modifies serializers to support adding/modifying these new attributes
- Moves avatar/banner serializer fields into a separate `ChannelAppearanceMixin` used by `ChannelSerializer` and `FieldChannelWriteSerializer`
- Modifies the `UserListViewSet` to optionally show all public lists if a `public=True` request parameter is included (otherwise it only shows lists created by the user).
- Bases common channel fields off of the `discussions.models.Channel` field instead of `channels.models.Channel` (hadn't realized or maybe forgot this existed, see https://github.com/mitodl/open-discussions/pull/2758)

- Go to `http://localhost:8063/api/v0/fields/` while logged in as an admin user, if there are already some existing field channels then you should see them listed.
- Create several new public user lists from `http://localhost:8063/learn` (click the bookmark icon for any ocw course, click 'Create New List', fill out the form using 'Public' for privacy, and save).  Create a private list too.
- Create a new `FieldChannel` by posting the following JSON, you should get a valid response.
    ```
    {
        "name": "new_field_1",
        "title": "New Field 1",
        "description": "New Field 1",
    }
   ```

- Go to `http://localhost:8063/api/v0/fields/new_field_1`, the details on your new field should be listed.
- Create another new `FieldChannel` with more data by posting the following JSON, you should get a valid response.
    ```
    {
        "name": "new_field_2",
        "title": "New Field 2",
        "description": "New Field 2",
        "lists": [<public_list_1_id>, <public_list_2_id>],
        "featured_list": <public_list_1_id>,
        "subfields": ["new_field_1"]
    }
   ```

- Go to `http://localhost:8063/api/v0/fields/new_field_2`, the details on your new field should be listed, including lists and subfields.

- PATCH the following JSON to update the data including lists and subfields:
    ```
    {
        "title": "New Field 2 modified",
        "description": "New Field 2 new desc",
        "lists": [<public_list_2_id>],
        "featured_list": <public_list_2_id>,
        "subfields": []
    }
    ```
- Reload the detail page and verify that the fields have been correctly updated
- PATCH the following JSON to try updating the `lists` field with a private user list, you should get a validation error:
```
    {
        "lists": [<private_list_id>]
    }
```

- Log in as a different user and create some more private and public lists.
- Go to `http://localhost:8063/api/v0/userlists/`, you should see only that user's public and private lists
- Go to `http://localhost:8063/api/v0/userlists/?public=true`, you should see all users' public lists


channels_fields/migrations/0004_field_channel_refactor.py
```diff
@@ -0,0 +1,77 @@
...
+            field=imagekit.models.fields.ProcessedImageField(
+                max_length=2083, null=True, upload_to=profiles.utils.banner_uri
+            ),
+        ),
+        migrations.AlterField(
+            model_name="fieldchannel",
+            name="name",
+            field=models.CharField(
+                max_length=100,
+                validators=[
```
ChristopherChudzicki, line 53: **Question:** Does this append a new validator to some list of validators for the field, or does it overwrite the list of validators for the field?

In particular, migration 0001 says `name` should be unique, but I was able to create two fields with the same name, and I'm wondering if maybe this is why?.

channels_fields/serializers.py
```diff
@@ -1,40 +1,271 @@
...
+            "title",
+            "public_description",
+            "subfields",
+            "featured_list",
+            "lists",
+            "about",
+        )
+
+    def validate_lists(self, lists):
+        """Validator for lists"""
```
ChristopherChudzicki, line 113: This is allowing `lists` to either be `number[]` or `FieldList[]`, right? 

**Generic Question:** Is this a common pattern in our APIs, that they accept either an ID value or an object value? It doesn't necessarily seem worth the extra complexity to me.

**Suggestion:** Maybe add more of a docstring, or a type annotation like `List[Union[number, FieldList]]`. _Not sure I have that python typing syntax quite right._

ChristopherChudzicki (changes requested):
This all worked well except the issue with duplicate channel names.

I *think* that's a new issue, at least based on looking at migration 0001.

channels_fields/migrations/0004_field_channel_refactor.py
mbertrand: Good catch, the `discussions.models.Channel` model doesn't have `unique=True` for the `name` attribute but I think it should, will change that and update the migration.

channels_fields/serializers.py
mbertrand: I will change the code to expect a list of integers only instead of converting something else to that, it's what the frontend should send it once it gets written.  Ditto for subfields (list of strings ie field names).

ChristopherChudzicki (approved):
👍 Looks good!

Issue #3642, opened 2022-07-19
Field Pages basic layout


ChristopherChudzicki:
Focus is on the overall layout... Course cards / rows will be revisited.

#### Designs and Mockups

_Marked up from https://projects.invisionapp.com/d/main#/console/22082155/468413145/preview_

<img width="1865" alt="field_pages" src="https://user-images.githubusercontent.com/9010790/179837439-90913dfe-5bc1-4b88-b31b-1b3a66b86899.png">

#### Acceptance Criteria:

- [ ] Fields page should match mocks above

#### Out of Scope

- Course cards / course rows ... *These can probably be brought over from open-discussions, or perhaps have been already by @abeglova ... either way, I'll revisit them.*
- Any new API stuff

ChristopherChudzicki:
@pdpinch I left a few questions on the mocks above... Are the answers known yet? We may have discussed (2)... I can't remember exactly and the google docs notes are a bit sparse.

pdpinch:
- Let's drop the search icon for now. That's a holdover from the channel UI, where it allows searching within the channel. I don't think we have much use for that idea at the moment. 
- From a user perspective, "Ethics" is a subfield. You need to check with @mbertrand to see where that comes from in the data model.

ChristopherChudzicki:
Closed by #3654

PR #3654, opened 2022-07-26
Cc/field pages v0

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Toward https://github.com/mitodl/open-discussions/issues/3642

#### What's this PR do?
This PR adds the field pages, and creates a `<FieldPageSkeleton />` component to help create other field pages (e.g., editing fields).

See individual comments for more.

1. Make some fields. Add images if you want.
2. Check homepage is still responsive. Visit `http://cac.od.odl.local:8063/infinite` and make screen wide / narrow. For very wide pages, the content should still be centered with a max width of ~1000px.
3. Visit  http://cac.od.odl.local:8063/infinite/fields/dogs You should see a field page and tabs should work. I'll fill in more details later.

Screen Shot 2022-07-26 at 12 41 34 PM


frontends/infinite-corridor/src/pages/HomePage.tsx
```diff
@@ -1,13 +1,12 @@
 import React, { useCallback, useState } from "react"
```
ChristopherChudzicki, line 1: Main changes in this file:
- using mui's responsive Container for layout
- changed a `div` to a `section`, and moved the h2 into the `section`.

frontends/ol-util/src/components/PageBanner.tsx
```diff
@@ -9,10 +9,22 @@ const tallBannerHeight = "252px"
 const shortMobileBannerHeight = "115px"
```
ChristopherChudzicki, line 9: The main change in this file is a new component `BannerPage`... Previously this file exported a bunch of components `BannerPageWrapper`, `BannerPageHeader`, etc.

It still exports those components. But most of the time we want all those components in a specific layout. So if that's what you want, now you can just use `BannerPage`.

frontends/ol-util/src/hooks/index.ts
```diff
@@ -1,2 +1 @@
 export { default as useSearchParams } from "./useSearchParams"
-export { default as useId } from "./useId"
```
ChristopherChudzicki, line 2: I deleted this hook because I found a better way to do what I was after.

Previously I was using this for
```tsx
<h2 id={elementId}>Fields of Study</h2>
<ul aria-labelledby={elementId} className="field-list">
```

Now those two elements are both children of a `<section />`, which seems like a more semantic way to associate the title with the list.

package.json
```diff
@@ -28,9 +28,9 @@
         "global:lint-check": "cd $INIT_CWD && eslint src/ --ext .js,.jsx,.ts,.tsx",
         "global:scss-lint": "echo 'TODO: Set this up.'",
         "global:lint-fix": "cd $INIT_CWD && eslint --fix src/ --ext .js,.jsx,.ts,.tsx",
-        "global:test": "cd $INIT_CWD && jest",
+        "global:test": "cd $INIT_CWD && NODE_ENV=test jest",
```
ChristopherChudzicki, line 31: Jest normally does this by default, but our docker container already specifies NODE_ENV=development.

This generally seems better, but the Immediate motivation is that MUI logs some console errors that are silenced (correctly) during tests. For example, MUI logs an error when tab panels have zero height / width. But in JSDOM-based test environments (like jest) the height will always be zero since JSDOM does not implement any layout logic. (In particular, it returns a bunch of zeros for `getBoundingClientRect`).

frontends/infinite-corridor/src/pages/FieldPage.tsx
```diff
@@ -1,21 +1,73 @@
...
   return (
-    <div className="page-content">
-      <Card>
-        <h2>{name}</h2>
-        Welcome to the field page for: <code>{name}</code>!
-      </Card>
-    </div>
+    <FieldPageSkeleton name={name}>
+      <TabContext value={value}>
+        <Container>
```
ChristopherChudzicki, line 40: We do breakpoints ourselves in open-discussions.

But breakpoints are complicated...which I came to appreciate (even more) after looking at https://material.io/design/layout/responsive-layout-grid.html#breakpoints

Let's let MUI handle it. We *can* customize the breakpoint stuff if we want to. (e.g., to make it more consistent with open-discussions).

frontends/infinite-corridor/package.json
```diff
@@ -10,6 +10,7 @@
     "@emotion/react": "^11.9.3",
     "@emotion/styled": "^11.9.3",
     "@mui/icons-material": "^5.8.4",
+    "@mui/lab": "^5.0.0-alpha.91",
```
ChristopherChudzicki, line 13: This is where MUI puts their not-quite-ready APIs.

Still, they advertise it in their docs, and it has 500k+ npm downloads per month, so I am comfortable using it.

We do not **need** it (and ditching it wouldn't be very hard), but it significantly simplifies using MUI's tabs. See https://mui.com/material-ui/react-tabs/#experimental-api

frontends/infinite-corridor/src/pages/FieldPageSkeleton.tsx
```diff
@@ -0,0 +1,39 @@
...
+ */
+const FieldSkeletonProps: React.FC<FieldSkeletonProps> = ({
+  children,
+  name
+}) => {
+  const field = useFieldDetails(name)
+
+  return (
+    <BannerPage
+      compactOnMobile
```
mbertrand, line 24: Add this here (or something like it) so the banner appears:

```
src={field.data?.banner ?? ""}
```

Screen Shot 2022-07-26 at 3 34 43 PM

frontends/infinite-corridor/src/pages/FieldPageSkeleton.tsx
```diff
@@ -0,0 +1,39 @@
...
+      bannerContent={
+        <Container className="field-title-container">
+          <div className="field-title-row">
+            <img alt={field.data?.title} src={field.data?.avatar_small ?? ""} />
```
mbertrand, line 28: The medium avatar is used for discussion channel avatars in the header ([example](https://discussions-rc.odl.mit.edu/c/mbchannel02)), while the small avatar is used for the left side vertical channel list.  So I'm guessing  the medium avatar should be used for field channel banners too.

Screen Shot 2022-07-26 at 3 47 17 PM

mbertrand (changes requested):
A couple initial comments, will take a closer look tomorrow AM.

frontends/infinite-corridor/src/pages/FieldPageSkeleton.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <BannerPage
+      src={field.data?.banner ?? ""}
+      alt=""
+      compactOnMobile
+      bannerContent={
+        <Container className="field-title-container">
+          <div className="field-title-row">
+            <img alt="" src={field.data?.avatar_medium ?? ""} />
```
mbertrand, line 30: At some point (this or a future PR) this should handle null avatar images better, right now it shows a broken image.  Discussion channels show initials based on title if there is no avatar image, maybe we could do that here too.  Could use a populated alt tag too.

from `open-discussions/src/components/ChannelAvatar`:
```python
          {isDefault ? (
            <div className="avatar-initials">{initials(channel.title)}</div>
          ) : (
            <img
              src={imageUrl}
              alt={`Channel avatar for ${channel.name}`}
              className={`avatar-image`}
            />
          )}
```

ChristopherChudzicki: @mbertrand Good point about the channel title initials. I'll do that in a separate PR.

Re alt tags: I actually removed that on purpose. My understanding is that unless the banner & avatar alt actually add something, they are detrimental. I need to get a functional screenreader, but I believe that on open-discussions, a screen reader would basically read

> Banner image for My Cool Channel. Channel avatar for My Cool Channel. My Cool Channel

banner alt + avatar alt + channel title.... Right now, just reading the channel title seems better.

PR #3659, opened 2022-07-28
Cc/lr cards

ChristopherChudzicki:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

(Required)

#### What's this PR do?
(Required)

(Required)

Screen Shot 2022-07-28 at 5 52 07 PM

Screen Shot 2022-07-28 at 5 51 05 PM


frontends/ol-search-ui/src/util.ts
```diff
@@ -0,0 +1,119 @@
...
+  // The most recent run that "ended"
+  const [bestRecentRun] = dated
+    .filter(run => runStartDate(run).isSameOrBefore())
+    .sort(compareRuns)
+    .reverse()
+  if (bestRecentRun) return bestRecentRun
+  return null
+}
+
+const readableLearningResources: Record<LR, string> = {
```
ChristopherChudzicki, line 90: One of the reasons I re-defined `LearningResourceType` as an enum rather than a bunch of constants was so it would be easy to do something like this... now we can be sure that `readableLearningResources` includes all the LearningResourceTypes. (if someone adds a new LearningResourceType without editing adding a value to `readableLearningResources`, ts will error since records with enum keys need to use all of the enum).

frontends/ol-search-ui/src/components/LearningResourceCard.tsx
```diff
@@ -0,0 +1,152 @@
+import moment from "moment"
```
ChristopherChudzicki, line 1: @abeglova Continuing discussion from Slack... I think `LearningResourceCard` is self-contained enough that it can be moved to `ol-search-ui`.

Re differences between Infinite Corridor and Learn cards, namely:
- different certificate icons
- Learn shows more detail, like subjects and bookmark

I think it would be relatively easy to pass those details in, like:
```tsx
<LearningResourceCard
    resource={resource}
    extraFooterContent={<Bookmark onClick={...} />}            {/*This prop doesn't actually exist, but it could*/}
    certificate={<CustomCertificate />}      {/*Although, maybe better to just use the same cert in Learn and IC*/}
/>
```
Or we could have a `SimpleResourceCard` (learn) and `DetailedResourceCard` (Both in ol-search-ui)... but there's enough here, and it is self contained enough to move to ol-search-ui that I think we should.

frontends/ol-util/src/components/PageBanner.tsx
```diff
@@ -51,7 +51,6 @@ const imageWrapperHeight = css<ImgProps>`
 const BannerContainer = styled.div`
   position: absolute;
   width: 100%;
-  z-index: -1;
```
ChristopherChudzicki, line 54: This was causing issues with click events in the field page and search page (after my change to use BannerPage).

I checked the existing usages, and removing the z-index didn't seem to have any negative affects.

frontends/infinite-corridor/src/pages/SearchPage.tsx
```diff
@@ -92,51 +107,66 @@ const SearchPage: React.FC = () => {
     )
 
   return (
-    <div>
-      <BannerPageWrapper>
```
ChristopherChudzicki, line 96: I swapped this out for [`BannerPage`](https://github.com/mitodl/open-discussions/blob/3dfa43edf54606f96b0a69a9c2aaa02b9e237f32/frontends/ol-util/src/components/PageBanner.tsx#L130), which does most of this automatically.

frontends/infinite-corridor/src/pages/SearchPage.tsx
```diff
@@ -92,51 +107,66 @@ const SearchPage: React.FC = () => {
...
+          onClear={clearText}
+          //@ts-expect-error - types need to be fixed in course-search-utils
+          onSubmit={onSubmit}
+          autoFocus
+        />
+      }
+    >
+      <Container disableGutters>
+        <Grid container>
+          <Grid item xs={3}>
```
ChristopherChudzicki, line 126: `<Grid container />` and `<Grid item />` are like our old Grid and Cell components, but not custom.

`xs={3}` says "Use 3 of the 12 columns on any screen xs or larger" (so, all screens)

frontends/infinite-corridor/src/pages/SearchPage.tsx
```diff
@@ -92,51 +107,66 @@ const SearchPage: React.FC = () => {
...
+            {lipsum}
+          </Grid>
+          <Grid item xs={9} component="section">
+            <InfiniteScroll
+              className="ic-searchpage-list"
+              hasMore={from + pageSize < total}
+              loadMore={loadMore}
+              initialLoad={from === 0}
+              loader={
+                <div key="loader" className="loader">
```
ChristopherChudzicki, line 137: I see what you mean about the key and InfiniteScroll erroring without it. That's really weird. I've never seen hard-coded keys be necessary before.

frontends/ol-search-ui/assets/learning-resource-card.scss
```diff
@@ -0,0 +1,75 @@
+$lightText: #8c8c8c !default;
+$spacer: 0.75rem !default;
+$boldWeight: 500 !default;
+$smallFontSize: 0.75em !default;
```
ChristopherChudzicki, line 4: I put *most* of the css for LearningResourceCard in ol-search-ui. These 4 values are customizable.

The component also accepts a class:
```ts
<LearningResourceCard resource={resource} className="your-favorite-class" />
```
so you can pass a classname to customize it more. 

I added the hover shadow by passing a class in Infinite Corridor.

frontends/ol-search-ui/src/components/LearningResourceCard.tsx
```diff
@@ -0,0 +1,152 @@
...
+}
+
+const Offerers: React.FC<OffererProps> = ({ offerers }) => {
+  return (
+    <>
+      {offerers.map((offerer, i) => {
+        const isLast = i === offerers.length - 1
+        return (
+          <React.Fragment key={`${offerer}-${i}`}>
+            <a href={`/infinite/search${toQueryString({ o: offerer })}`}>
```
abeglova, line 74: This should be  ```<a href={`/infinite/search?${toQueryString({ o: offerer })}`}>``` (missing the ?)

abeglova (approved):
LGTM except for one small issue with the offered bys

PR #3661, opened 2022-07-28
Field edit page with preliminary appearance tab

### LINKED ISSUE #3632: Appearance configuration for field pages

mbertrand:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Partially addresses #3632 

#### What's this PR do?
Adds a form for editing field title and description, but not banner and avatar.  These will be added in a separate PR.
Adds a tabbed edit page with 3 tabs (basic, appearance, moderators).  All but appearance tab are just placeholders right now.
Creates a separate component for field avatar
Adds `is_moderator` to the API response to determine if the current user can access the field edit page/forms.

- As an admin, go to the url `/infinite/fields/<your_field_name>/manage/` and verify that you see 3 tabs.
- Go to the "Appearance" tab and verify the title and description input fields are prepopulated with current values.
- Make both blank and verify you see validation errors
- Enter new values and submit.  You should be redirected back to the field page and see your changes there.
- As a non-admin/non-moderator your field, return to the edit page.  You should see a message that you're not allowed to access the page.

Screen Shot 2022-08-02 at 10 37 44 AM


frontends/infinite-corridor/src/libs/react-query.ts
```diff
@@ -20,4 +25,9 @@ const createQueryClient = (): QueryClient =>
     }
   })
 
-export { createQueryClient }
+const editFieldChannel = async (name: string, data: FieldChannelForm) => {
+  const { data: response } = await axios.patch(`${fieldDetails(name)}/`, data)
```
ChristopherChudzicki, line 29: **Question:** Should `fieldDetails(name)` just append the slash itself?

frontends/infinite-corridor/src/libs/react-query.ts
```diff
@@ -1,5 +1,10 @@
 import { QueryClient } from "react-query"
 import axios from "./axios"
+import { fieldDetails } from "../api/fields/urls"
+import { FieldChannelForm } from "../api/fields/interfaces"
+
+axios.defaults.xsrfCookieName = "csrftoken"
+axios.defaults.xsrfHeaderName = "X-CSRFToken"
```
ChristopherChudzicki, line 7: This can go in `frontends/infinite-corridor/src/libs/axios.ts`:
```ts
/**
 * Our axios instance with default baseURL, heads, etc.
 */
const instance = axios.create({
  baseURL:        "/api/v0",
  xsrfCookieName: "csrftoken",
  xsrfHeaderName: "X-CSRFToken"
})
```

ChristopherChudzicki (commented):
I was just skimming this, two small comments.

frontends/infinite-corridor/src/components/forms/EditChannelAppearanceForm.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    // an error will occur without act:
+    // "Warning: An update to Formik inside a test was not wrapped in act(...)."
+    // eslint-disable-next-line testing-library/no-unnecessary-act
+    await act(async () => {
+      fireEvent.submit(submitBtn)
+    })
+    ;[newTitle, newDesc].forEach(newText => {
+      const textHTML = screen.getByText(newText)
+      expect(textHTML).toBeInTheDocument()
+    })
```
ChristopherChudzicki, line 71: The `act` warnings are annoying and can be hard to figure out the cause.

In general, testing-library wraps its helpers in `act`, so you don't usuall need them. But it can be hard to figure out what's going on. This is probably the aspect of react testing that I find most frustrating.

The original code was:
```ts
fireEvent.submit(submitBtn)                    // 1. synchronously fire event
;[newTitle, newDesc].forEach(newText => {      
  const textHTML = screen.getByText(newText)   // 2. synchronously find element
  expect(textHTML).toBeInTheDocument()
})
```
Which errors with "Unable to find element". I **think** what was happening was:
1. synchronously fire event. But event handlers are asynchronous, so they won't respond right away
2. synchronously find element. But the event handlers haven't responded yet, so the test fails.

Instead, I would suggest using `findByText`. The `findBy` queries are asynchronous and timeout after a certain time. **So if an element might not exist yet, but should exist soon, use `findBy`.**

This works:
```ts
fireEvent.submit(submitBtn)
await screen.findByText(newTitle)
await screen.findByText(newDesc)
```

frontends/infinite-corridor/src/components/forms/EditChannelAppearanceForm.tsx
```diff
@@ -0,0 +1,131 @@
+import React from "react"
```
ChristopherChudzicki, line 1: *Rushed comment since I have to run:*

Would it make sense to put this component in `pages/admin`? I do not have a strong feeling.

ChristopherChudzicki (changes requested):
As mentioned: I have to run but this looks good! I left some comments/suggestions.

frontends/infinite-corridor/src/components/forms/EditChannelAppearanceForm.tsx
mbertrand: `forms/admin`?

mbertrand: But I guess all the forms for infinite will be admin forms

frontends/infinite-corridor/src/components/FieldAvatar.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { render, screen } from "@testing-library/react"
+
+import { makeField } from "../api/fields/factories"
+import FieldAvatar from "./FieldAvatar"
+
+describe("Avatar", () => {
+  it("Displays a small avatar image for the field", async () => {
+    const field = makeField()
+    render(<FieldAvatar field={field} imageSize="small" />)
+    const img = screen.getByAltText(`Channel avatar for ${field.title}`)
```
ChristopherChudzicki, line 11: **Suggestion:** I think we should get rid of the alt text, and replace this with `getByRole("img")`.

I am by no means an accessibility expert, but my understanding is that alt text is better left blank if it does not add meaningful context. Like on the field page, a screenreader would currently read, I believe

> "Channel avatar for physics. Physics".

Better (I think?) just to say "Physics".

Related: https://silktide.com/blog/is-alt-text-making-your-site-less-accessible/

_Actually: Probably best to make this component accept `alt` as a prop. Then if we ever use the avatar as an unlabeled link, we can use pass an alt tag for it. But... we don't need that right now._

frontends/infinite-corridor/src/components/forms/EditChannelAppearanceForm.tsx
ChristopherChudzicki: My thought was along the lines of co-locating the admin-editing code: `pages/admin` directory would contain the significant components that it uses, would import some stuff from ol-util or ol-forms. To make `pages/admin` more self-contained.

I.e., along the lines of this article Ferdi linked a while ago: https://alexmngn.medium.com/how-to-better-organize-your-react-applications-2fd3ea1920f1 *Note: If you do read that, its examples are a bit outdated. Like, people don't really have a `containers/` folder any more for redux-store connected components.*

What do you think? We can also keep it where it is for now reconsider in future prs.

ChristopherChudzicki (approved):

frontends/infinite-corridor/src/components/forms/EditFieldAppearanceForm.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const field = factory.makeField({ is_moderator: true })
+    setMockResponse.get(urls.fieldDetails(field.name), field)
+    renderTestApp({ url: `${urls.fieldDetails(field.name)}manage/#appearance` })
+    const titleInput = await screen.findByLabelText("Title")
+    const titleError = screen.queryByText("Title is required")
+    expect(titleError).toBeNull()
+    fireEvent.change(titleInput, {
+      target: { value: "" }
+    })
+    fireEvent.blur(titleInput)
```
ChristopherChudzicki, line 36: @mbertrand One more "by the way" comment: I usually trigger events via `user`:

```ts
import user from "@testing-library/userEvent" // test-utils also exports this as `{ user }` for convenience.

user.click(htmlElement)
user.type("blah blah blah{Enter}")
```
 
The `user` interface is a higher-level API that more closely resembles real user interaction. (`@testing-library`'s general philosophy is "write tests that resemble user interaction".)

For example, `user.type` types the text (character by character, I think... which is still plenty fast). And `user.click` fires most (all?) the events that would be fired by a click: pointerdown, pointermove, pointerup, click.

Anyway, that's why I generally use those methods though I suspect it doesn't usually matter.

**BTW:** `user.type` can be slightly annoying because it's like typing...you have to clear the text first:

```ts
input.setSelectionRange(0, input.value.length) // select all the text
user.type(input, 'whatever new thing you want') // typing clears the selected text
```
https://github.com/testing-library/user-event/issues/232 

cc @abeglova

ChristopherChudzicki: Oh! I remember why `user.click` is nice for clicking... it does not trigger events if the element is disabled, whereas fireEvent just triggers the event no matter what.

Related...
- https://github.com/jsdom/jsdom/issues/2665 _Closed because apparently this follows some spec, and browsers do not follow the spec. 🤷 _
- https://github.com/testing-library/dom-testing-library/issues/92

mbertrand: Thanks, I noticed that `user.type(input, text)` throws an error if `text` is blank ("") - `Expected key descriptor but found "" in ""` - so I left that test as is but changed the others.

frontends/infinite-corridor/src/components/forms/EditChannelAppearanceForm.tsx
mbertrand: Moving to `/pages/admin`

frontends/infinite-corridor/src/components/forms/EditChannelAppearanceForm.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+import {
+  renderTestApp,
+  screen,
+  setMockResponse,
+  fireEvent
+} from "../../test-utils"
+import * as factory from "../../api/fields/factories"
+import { urls } from "../../api/fields"
+
+describe("EditChannelAppearanceForm", () => {
```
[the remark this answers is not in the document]
mbertrand: Changed to `EditFieldAppearanceForm`

PR #3669, opened 2022-08-02
rtl tests for search

abeglova:
#### Pre-Flight checklist
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested

None

#### What's this PR do?
This PR upgrades the search page api to use axios and the search page tests to use RTL to match the rest of the infinite-corridor project.

Go to infinite/search. The functionality should remain unchanged


frontends/infinite-corridor/src/test-utils/testing_init.ts
```diff
@@ -15,7 +10,6 @@ global.SETTINGS = _createSettings()
 afterEach(function() {
   const node = document.querySelector("#integration_test_div")
   if (node) {
-    // @ts-expect-error we need to set SETTINGS for all tests
     ReactDOM.unmountComponentAtNode(node)
   }
   document.body.innerHTML = ""
```
ChristopherChudzicki, line 15: I think we can get rid of this all together. RTL cleans up after itself automatically in testing environments that support `afterEach`: https://testing-library.com/docs/react-testing-library/api/#cleanup

I also think it would be good to combine this file and `setupJest` into one.

ChristopherChudzicki (approved):
👍 Left a suggestion, but it can be done at any point separately if want.

ChristopherChudzicki (approved):

Issue #3672, opened 2022-08-03
Infinite Corridor, styling changes


ChristopherChudzicki:
A list of styling tweaks / feedback on initial infinite-corridor implementation:

- [x] fix searchbar styling
- [x] (mostly done, see below) make colors match open discussions

Remaining Tweaks https://mitodl.slack.com/archives/C03K5HYGPT9/p1659728393670959
- [ ] Square next/previous buttons like `learn/`
- [ ] Use the dark blue font from open-discussions for cards (not the black font from learn + infinite designs)

ChristopherChudzicki:
@Ferdi or @pdpinch a few questions, some of which may be related to differences in styling between open-discussions and the infinite corridor initial designs:

1. Where should the Biology field title link? _second screenshot above_
2. Re the background annotation (_second screenshot above_)
    - Do we want a gray background like the rest of the page, similar to the initial designs (https://projects.invisionapp.com/d/main#/console/22082155/468413145/preview)
    - or a white background similar to channel pages? (https://open.mit.edu/c/ocwchalkradio)
3. Similarly, there are some differences in font between the Infinite Corridor designs and the Open Discussions [style guide](https://projects.invisionapp.com/share/9SN8J0LPTQ2#/screens/311486535_Home_Logged_Style_Guide):
    - The infinite corridor designs use a gray background vs the blueish background in open-discussions.
    - The "default font" in the open-discussions is also blue (#03152d) vs black in Infinite Corridor.

Re (3) the colors are of course easy to change (controlled by scss variables). For now, I'm going stick with the Infinite Corridor designs for colors but use the Open Discussions style guide that @ferdi linked me for font-sizes and font faces.

Ferdi:
@ChristopherChudzicki when in doubt , try to match open-discussion.   We wanted to keep stying the same. The differences are probably the designer missing some details.

PR #3673, opened 2022-08-03
Fix styling conflicts: remove some classless tag selectors

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

#3672 

#### What's this PR do?
(Required)

Check that these all look reasonable:
- [ ] homepage http://cac.od.odl.local:8063/infinite/
- [ ] field pagehttp://cac.od.odl.local:8063/infinite/fields/puppies/
- [ ] field management http://cac.od.odl.local:8063/infinite/fields/puppies/manage
- [ ] _Check the searchpage, too, but it is not very styled yet._

The cause of this issue was CSS selectors targeting elements without classes, things like:
```css
input[type="text"] {
  border: 1px solid red;
}
```
Which applies a border to all text inputs. This CSS also has reasonably height specificity (element + attribute type; that's more specific than a single class). The fact that this is more specific than a single class makes it reasonably likely to conflict with external libraries that do (some) styling on their own.

And it does, for example, this turns our MUI searchbar input into:
Screen Shot 2022-08-03 at 12 29 55 PM
(note: MUI's input + adornments, e.g., buttons / icons, is not actually a single input field).

Anyway.... This PR makes the form-related CSS selectors apply only to fields with class `.form-field` 

### Screenshots

**Before**:
Screen Shot 2022-08-03 at 1 38 23 PM

**Afer**:
Screen Shot 2022-08-03 at 2 09 19 PM

### DIFF
--- a/frontends/infinite-corridor/src/scss/combined.scss
+++ b/frontends/infinite-corridor/src/scss/combined.scss
@@ -1,5 +1,3 @@
-@use "material-components-web/dist/material-components-web";
-
 @use "theme";
 @use "button" with ( $color: theme.$color-primary );
 @use "main-search";
--- a/frontends/infinite-corridor/src/scss/form.scss
+++ b/frontends/infinite-corridor/src/scss/form.scss
@@ -1,9 +1,15 @@
 @use "theme";
 
+$fontColor: theme.$font-black;
+$fontColorLight: theme.$font-grey-light;
+
+$radius: theme.$std-border-radius;
+$borderColor: theme.$input-border-grey;
+
 .form {
   .form-item {
     margin: 0 0 28px 0;
-    color: theme.$font-black;
+    color: $fontColor;
 
     .label,
     .input {
@@ -15,8 +21,8 @@
     background: #f8f8f8;
     border: 1px solid #bbb;
     font-size: 15px;
-    color: rgba(0, 0, 0, 0.87);
-    border-radius: theme.$std-border-radius;
+    color: $fontColor;
+    border-radius: $radius;
     height: 39px;
     /* If you add too much padding here, the options won't show in IE */
     padding: 8px 20px;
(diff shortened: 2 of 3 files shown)


frontends/infinite-corridor/src/scss/combined.scss
```diff
@@ -1,5 +1,3 @@
-@use "material-components-web/dist/material-components-web";
```
ChristopherChudzicki, line 1: We aren't using this currently. (MUI instead). It's a holdover from very early infinite-corridor homepage.

frontends/infinite-corridor/src/scss/form.scss
```diff
@@ -129,35 +141,6 @@ label.radiolabel {
...
-
-.mdc-checkbox::before,
-.mdc-checkbox::after {
-  background-color: theme.$navy;
-}
-
-.mdc-checkbox__background {
-  background-color: theme.$navy !important;
-  border-color: theme.$navy !important;
-}
```
ChristopherChudzicki, line 146: This is from material-web-components, which we aren't currently using in infinite-corridor.

I removed some other unused code in this file, but not all of it. E.g., I left the styling for `select` elements since it seems reasonably likely that we will use it.

ChristopherChudzicki:
@mbertrand @abeglova See above... I think we should avoid using CSS selectors without a class, like `input[type=text] {...}`.

CSS conventions probably deserves a bigger discussions post at some point...it's not really something I'm expert in. But I do feel pretty confident that we should avoid most classless selectors for the conflict reason above.  The "downside" is just means we end up applying ".form-field" to a bunch of our inputs, similar to how bootstrap has you apply ".form-control" to all of its styled inputs. 

We still have some classless selectors, like header tags. That's probably less problematic. The `input[type="whatever"]` ones are **more** problematic: because they have higher specificity, they are much more likely to unintentionally override stuff.

mbertrand (approved):
👍

PR #3674, opened 2022-08-03
Set imagekit prefix to ""

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [ ] Code is tested
- [x] Settings

N/A

#### What's this PR do?
Fixes access to imagekit-generated avatar images via cloudfront by setting the imagekit cache prefix to "" instead of "CACHED/images"

- Set `AWS*` and `CLOUDFRONT_DIST` .env settings to the same as RC.
- Set `OPEN_DISCUSSIONS_USE_S3=True` in your .env
- In django admin, upload an avatar image for one of your fields.
- Go to the field page, you should see the avatar image, and it should have a cloudfront url.


ChristopherChudzicki (approved):
👍 I see the cloudfront images!

@mbertrand Just curious... Was this related to the python upgrade? 

The avatar on rc at https://discussions-rc.odl.mit.edu/c/mbtestcollection05 still works... is that because it was uploaded before the issue started?

mbertrand:
@ChristopherChudzicki The discussion channels are based on a different model - `channels.models.Channel` instead of `discussions.models.Channel` - and doesn't use the imagekit code to generate those avatar sizes.

Issue #3675, opened 2022-08-03
Channel headers are missing on production


ChristopherChudzicki:
See https://open.mit.edu/c/ocwchalkradio

<img width="1668" alt="Screen Shot 2022-08-03 at 4 44 13 PM" src="https://user-images.githubusercontent.com/9010790/182708210-8407b2fc-208d-42cb-b015-0f5ffe126096.png">

This is a regression introduced here in #3659 🤦

PR #3676, opened 2022-08-03
Fix open discussions channel header issue

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [ ] Code is tested

Fixes https://github.com/mitodl/open-discussions/issues/3675

#### What's this PR do?
This PR restores the original `z-index: -1` that was in `BannerContainer`, and updates `PageBanner` (which is only used in infinite-corridor) to achieve the goal of that change in a different way.

- check that channel titles are visible in open-discussions (not infinite-corridor)
- check that the searchbar at http://cac.od.odl.local:8063/learn has an "Explore" button below it
- Check that http://cac.od.odl.local:8063/infinite/fields/your_file_name has a title and that its text can be selected

### DIFF
--- a/frontends/ol-util/src/components/PageBanner.tsx
+++ b/frontends/ol-util/src/components/PageBanner.tsx
@@ -48,9 +48,13 @@ const imageWrapperHeight = css<ImgProps>`
   }
 `
 
+/**
+ * Prefer direct use of `BannerPage` component.
+ */
 const BannerContainer = styled.div`
   position: absolute;
   width: 100%;
+  z-index: -1;
 `
 
 const imageStylesheet = `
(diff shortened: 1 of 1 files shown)


frontends/ol-util/src/components/PageBanner.tsx
```diff
@@ -48,9 +48,13 @@ const imageWrapperHeight = css<ImgProps>`
   }
 `
 
+/**
+ * Prefer direct use of `BannerPage` component.
+ */
 const BannerContainer = styled.div`
   position: absolute;
   width: 100%;
+  z-index: -1;
```
ChristopherChudzicki, line 57: This restores the original value, which I inappropriately removed in https://github.com/mitodl/open-discussions/pull/3659/files#diff-81cd8ecef16a778b60f2e836d96769a87f5082a154eab9fc0f3d6f9bb577cbc6L54

mbertrand (approved):
👍

PR #3689, opened 2022-08-11
Move widget/group creation for fields to signal

### LINKED ISSUE #3687: IC: Refactor WidgetList creation for field channels
A `WidgetList` is created for a new `FieldChannel` only if the field is created via the API (the code for doing so is in the serializer).  A signal should be used instead so that this will also happen when a `FieldChannel` is created via Django Admin instead.

mbertrand:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #3687

#### What's this PR do?
Moves code for creating a field channel widget and group/role from the serializer to a signal.  Makes channel image attributes blankable to make editing easier from django admin.

- Create a new `FieldChannel` in django admin.  Verify that it has an associated `WidgetList`
- Create a new `FieldChannel` via the API at `/api/v0/fields` (post this data: `{"name": "newfield1", "title": "New Field 1"}`.  Verify that it has an associated `WidgetList` and there is a `Group` for it (named `field_<name>_moderators`)


ChristopherChudzicki (approved):
👍

Issue #3694, opened 2022-08-15
Text Widget for IC Fields


ChristopherChudzicki:
#### Designs and Mockups
<img width="369" alt="Screen Shot 2022-08-15 at 2 54 21 PM" src="https://user-images.githubusercontent.com/9010790/184700897-23c04d3d-d6ad-40d7-ba34-de95742bedda.png">
<img width="279" alt="Screen Shot 2022-08-15 at 3 06 32 PM" src="https://user-images.githubusercontent.com/9010790/184700911-e7d2e4f3-5ec3-459b-8bab-0dea960abfbb.png">

#### Acceptance Criteria:

- [ ] Field author can create, edit, and delete text widget with form similar to above, but **limited to plaintext**. *Rich text will be added subsequently.*
- [ ] Anyone can view text widget in widget sidebar. 

#### Out of Scope

- rich text editing
- re-ordering widgets

PR #3699, opened 2022-08-17
Viewing (plaintext) widgets

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Toward #3694 

#### What's this PR do?
This PR:
1. Renames `pages/admin` to `pages/field-details`, which now holds the routes for viewing + editing field details. I also moved `FieldAvatar`, `FieldMenu` into that directory.
2. Adds a new local package ("yarn workspace") `ol-widgets` for widget-related stuff My plan is to put individual componenets for rendering widgets here.
3. Adds codes for **viewing** a plaintext widget. (Which will become a richtext widget). I'll add editing / deleting / reordering in the next PR.

1. In admin panel, view the WidgetList for a field. (There's a link to it in the Field's details in admin panel)
2. Add a few Markdown widget instances with title + configuration `{ "source": "whatever plaintext you want" }`
3. Check that the widget gets rendered on the field page.

⚠️ When I was adding markdown widget instances in the admin panel, I couldn't edit the configuration field initially. But if you submit the form with an empty config (it will fail) the field becomes editable.

With the github comments

#### Screenshots

Screen Shot 2022-08-17 at 1 10 34 PM


frontends/infinite-corridor/src/scss/fieldpage.scss
```diff
@@ -131,7 +131,13 @@ $avatar-dimension-small: 22px;
 
 
 $minTabHeigt: 44px;
```
ChristopherChudzicki, line 133: Basically all the CSS changes except this file are moving card-styling so that the affects all cards (e.g., widgets) rather than just learning resource cards.

frontends/ol-widgets/src/interfaces.ts
```diff
@@ -0,0 +1,50 @@
...
+  available_widgets: WidgetSpec[]
+}
+
+interface RichTextWidgetConfig {
+  source: string
+}
+export type RichTextWidgetInstance = WidgetInstance<RichTextWidgetConfig>
+
+export enum WidgetTypes {
+  RichText = "Markdown"
```
ChristopherChudzicki, line 49: The rich text widget is, for the moment, the old Markdown widget.

We probably want to replace the markdown widget because it is built upon ProseMirror which is much more complicated than we need for a minimalistic WYSIWYG. See discussion at https://github.com/mitodl/hq/discussions/210.

For this initial iteration, though, the rich text widget will just be plain text, and using the Markdown widget for it means no backend changes necessary till the plan in  https://github.com/mitodl/hq/discussions/210 is more final.

frontends/infinite-corridor/src/App.tsx
```diff
@@ -23,30 +23,44 @@ interface AppProps {
   queryClient: QueryClient
 }
 
+/**
+ * Renders child with Router, QueryClientProvider, and other such context provides.
+ */
+const AppProviders: React.FC<AppProps & { children: React.ReactNode }> = ({
```
ChristopherChudzicki, line 29: Moved this out of `App` to support `renderWithProviders`.

ChristopherChudzicki:
@mbertrand re first point above, namely:
> Renames pages/admin to pages/field-details, which now holds the routes for viewing + editing field details. I also moved FieldAvatar, FieldMenu into that directory.

Let me know if you object, I wanted to do this but I don't feel too strongly about it... I think it's good to co-locate the components that have fairly limited usage with the routes that use them. A potential downside is lots of nesting, but I'm hoping we avoid that by extracting stuff to separate local packages when appropriate.

mbertrand:
@ChristopherChudzicki that's ok with me

pdpinch:
not a blocker, but I'm curious how this widget will handle URLs?

pdpinch (approved):
UI looks ok to me, so far

mbertrand (approved):
Looks good to me, just 1 question.

frontends/infinite-corridor/src/pages/field-details/EditFieldAppearanceForm.test.tsx
```diff
@@ -6,20 +6,26 @@ import {
...
+  setMockResponse.get(
+    widgetUrls.widgetList(field.widget_list),
+    makeWidgetListResponse({}, { count: 0 })
+  )
+  return field
+}
 
+describe("EditFieldAppearanceForm", () => {
   it("Displays the field title, appearance inputs with current field values", async () => {
+    const field = setupApis()
```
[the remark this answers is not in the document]
ChristopherChudzicki: Mostly: you can pass parameters to `setupApis`, whereas you can't pass parameters to beforeEach hooks. So, below I can call it as:
```ts
const field = setupApis({featured_list: null, lists: []})
```
which I wanted to do so I didn't have to set up some mock API calls.¹

---
¹"So I don't have to set up mock API calls" is maybe a separate issue that's worth thinking about. I like the tests that use `renderTestApp` because they render the whole app which is more realistic. But it is a bit of a pain to things like "Clicking a link on PageX sends you to PageY", because you end up having to provide mock API responses for both pages X and Y... One could say "So don't write tests that way, just mock PageY". I tend to think it would be better to just make the API mocking easier... e.g., via [msw](https://github.com/mswjs/msw). _Nathan pointed me toward msw a while ago—we don't use it anywhere, but I've been trying it out in a side project._

Issue #3697, opened 2022-08-16
Add a course drawer to infinite corridor


abeglova:
When i click on a course card in infinite corridor I want to see a course drawer with more information.

Design:
https://projects.invisionapp.com/d/main#/console/22082155/468407666/preview

PR #3708, opened 2022-08-24
infinite drawer

abeglova:
#### Pre-Flight checklist

Screen Shot 2022-08-25 at 10 19 34 PM
Screen Shot 2022-08-26 at 8 57 38 AM
Screen Shot 2022-08-25 at 10 20 21 PM
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

[(Required)
](https://github.com/mitodl/open-discussions/issues/3697)

#### What's this PR do?
This PR adds a course drawer to the infinite corridor

Go to the infinite search page or field page in infinite corridor. You should be able to click on the cards to open the infinite drawer. The drawer should look like the design https://projects.invisionapp.com/d/main#/console/22082155/468407666/preview in desktop


frontends/ol-search-ui/src/components/ShareTooltip.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { FacebookIcon, TwitterIcon, LinkedinIcon } from "react-share"
+import Tooltip from "rc-tooltip"
+
+type HelperProps = {
+  url: string
+  hideSocialButtons?: boolean | null | undefined
+  objectType?: string | null | undefined
+}
+
+export class ShareTooltipHelper extends React.Component<HelperProps> {
```
ChristopherChudzicki, line 16: **Request:** Change this to a functional component, with the ref created via
```tsx
const inputRef = useRef<HTMLInputElement>(null)
```

frontends/ol-search-ui/src/components/ShareTooltip.tsx
```diff
@@ -0,0 +1,87 @@
...
+  constructor(props: HelperProps) {
+    super(props)
+    this.input = React.createRef()
+  }
+
+  selectAndCopyLinkText = (e: React.MouseEvent<HTMLElement>) => {
+    e.preventDefault()
+    if (this.input.current) {
+      // @ts-expect-error not sure why typscript expects a type here
+      this.input.current.select()
```
ChristopherChudzicki, line 28: The reason you needed `ts-expect-error` is that the ref has type HTMLElement, but not all HTMLElments have a `select` method.

The ref is actually an `HTMLInputElement`, which does have select.

Also:
**Suggestion:** Try `navigator.clipboard.writeText(this.input.current.value)` (see [writeText](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText)). 

[execCommand is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand)

frontends/ol-search-ui/src/components/ShareTooltip.tsx
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+
+type Props = {
+  url: string
+  hideSocialButtons?: boolean | undefined
+  children: React.ReactElement<string> | undefined
+  placement?: string
+  objectType?: string | null | undefined
+}
```
ChristopherChudzicki, line 75: `?` and `| undefined` are redundant:
```ts
type Props = {
  url: string
  hideSocialButtons?: boolean
  children?: React.ReactElement<string>
  placement?: string
  objectType?: string | null
}
```

frontends/ol-search-ui/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,108 @@
...
+    ))}
+  </React.Fragment>
+))
+
+export default function TruncatedText(props: Props) {
+  const { text, lines, estCharsPerLine, className, showExpansionControls } =
+    props
+  const [expanded, setExpanded] = useState(false)
+  const [hasOverflow, setHasOverflow] = useState(false)
+  const [, forceReRender] = useState(0)
```
ChristopherChudzicki, line 44: This requires calling `forceReRender(val => val + 1)`.

A slightly nicer way to do it is
```
const [ignored, forceUpdate] = useReducer(x => x + 1, 0);
```
See https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate

frontends/ol-search-ui/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const [currentText, setCurrentText] = useState(text)
+  const dotRef = useRef(
+    null
+  ) as React.MutableRefObject<DotDotDotElementWithContainer | null>
```
ChristopherChudzicki, line 48: I suspect that
```
const dotRef = useRef< DotDotDotElementWithContainer>(null)
```
would work here.

frontends/ol-search-ui/src/factories.ts
```diff
@@ -39,7 +39,7 @@ export const makeSearchResult = (type: string | null) => {
 }
 
 export const makeCourseResult: Factory<LearningResourceResult> = overrides => ({
-  id:                casual.integer(1, 1000),
+  id:                casual.integer(1, 1000).toString(),
```
ChristopherChudzicki, line 42: A few times recently we've had CI fail because of id collisions... which seems unlikely given (1,1000) range.

But... I've been user `faker.unique(faker.datatypes.number)` for this sort of thing lately. (I think faker's default range is 1 to 10k)

frontends/ol-search-ui/src/util.tsx
```diff
@@ -144,3 +155,76 @@ export const CertificateIcon = () => (
...
+    return moment(objectRun.best_start_date).format("MMMM DD, YYYY")
+  }
+  return "Ongoing"
+}
+
+export const getInstructorName = (instructor: CourseInstructor) => {
+  if (instructor.full_name) {
+    return instructor.full_name
+  } else if (instructor.first_name && instructor.last_name) {
+    return `Prof. ${instructor.first_name} ${instructor.last_name}`
```
ChristopherChudzicki, line 193: Do we record whether or not the instructor is actually a professor? Some of them are lecturers.....Or graduate students...... 😆 
Screen Shot 2022-08-26 at 1 28 17 PM
.

frontends/ol-search-ui/src/util.tsx
```diff
@@ -144,3 +155,76 @@ export const CertificateIcon = () => (
...
+}
+
+export const absolutizeURL = (url: string) =>
+  new URL(url, window.location.origin).toString()
+
+// @ts-expect-error typescript complains about getting 0 arguments
+export const capitalize = R.converge(R.concat(), [
+  R.compose(R.toUpper, R.head),
+  R.tail
+]) as F.Curry<(text: string) => string>
```
ChristopherChudzicki, line 230: Does this just capitalize the first letter? And if it does, would
```ts
const capitalize = (s: string): string => s[0].toUpperCase() + s.slice(1)
```
work ?

package.json
```diff
@@ -61,15 +61,20 @@
...
         "jest-fail-on-console": "^2.4.2",
         "jest-when": "^3.5.1",
+        "locale-code": "^2.0.2",
         "prettier": "^2.7.1",
         "prettier-eslint-cli": "^6.0.1",
         "query-string": "^7.1.1",
+        "rc-tooltip": "^5.2.2",
+        "react-share": "^4.4.0",
         "rimraf": "^3.0.2",
+        "swc-loader": "^0.2.3",
```
ChristopherChudzicki, line 77: curious... what did this get added for?

ChristopherChudzicki (changes requested):
The drawer looks nice!

I left a bunch of comments, some important, some less so. Plus the stuff we talked about in slack: the Share button never dismisses.

I am **not** requesting this change, but I want to leave it as a comment here from our slack discussion since it might be easier to find here in the future if we ever want it....
> [regarding drawer] I was initially imagining:
>
> 1.drawer component in infinite-corridor
> 2. "LearningResourceDetails" or something in ol-search-ui ... that's the drawer content
> 3. LearningResourceCards have an onClick prop that we use to set the drawer parameters
> 4.drawer makes API call and passes relevant info the the details

frontends/ol-search-ui/src/util.tsx
ChristopherChudzicki: (I realize this is how Learn does it right now, so no requested change)

frontends/ol-search-ui/src/components/LearningResourceDrawer.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { useQuery } from "react-query"
+import { useSearchParams } from "ol-util"
+
+export type drawerResource = {
+  type: string
+  id: number | string
+  runId?: number | string | undefined
+}
+
+export const GetDrawerResourceFromUrl = () => {
```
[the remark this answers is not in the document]
abeglova: This is needed for the share urls to work

frontends/ol-search-ui/src/util.tsx
abeglova: The above breaks for empty strings

frontends/ol-search-ui/src/components/LearningResourceCard.tsx
```diff
@@ -121,9 +114,26 @@ const LearningResourceCard: React.FC<LearningResourceCardProps> = ({
...
+                type:  resource.object_type,
+                id:    resource.id,
+                runId: bestAvailableRun?.id
+              })
+            }
+          >
+            <Dotdotdot className="ol-lrc-title" tagName="h3" clamp={3}>
+              {resource.title}
+            </Dotdotdot>
+          </div>
```
ChristopherChudzicki, line 131: This should be clickable/focusable with the keyboard. Although we might want to check with Ferdi/peter whether the whole card should be clickable, or just the title.

If we do decide that just the title should be clickable, it would be nice to use a more semantic component

frontends/ol-search-ui/src/components/ShareTooltip.tsx
```diff
@@ -0,0 +1,77 @@
...
+    if (inputRef.current) {
+      navigator.clipboard.writeText(inputRef.current.value)
+    }
+  }
+
+  return (
+    <div className="tooltip">
+      <div className="tooltip-text">
+        {`Share a link to this ${objectType}:`}
+      </div>
```
ChristopherChudzicki, line 34: This should probably be an actual button, although I'm not sure the tooltip is keyboard focusable at all right now and that would probably take a little more effort. (Possibly best done by finding a different library that handles it for us.)

frontends/ol-search-ui/src/components/ShareTooltip.tsx
```diff
@@ -0,0 +1,77 @@
...
+  url,
+  hideSocialButtons,
+  objectType
+}) => {
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  const selectAndCopyLinkText = (e: React.MouseEvent<HTMLElement>) => {
+    e.preventDefault()
+    if (inputRef.current) {
+      navigator.clipboard.writeText(inputRef.current.value)
```
ChristopherChudzicki, line 26: Perhaps you discovered this, but apparently `navigator.clipboard` is only available on https websites, so this didn't work for me locally. Do you think that's a problem?

Can go back to the old execCommand way if need be.

ChristopherChudzicki (changes requested):
Working well! Noticed a few issues and some unused stuff.

More broadly, we might want to improve the drawer routing a bit. For example, it seems weird to me that:
```
// click card to open drawer does not change route:
http://cac.od.odl.local:8063/infinite/fields/astronomy/
// sharing drawer: route-powered drawer
http://cac.od.odl.local:8063/infinite/fields/astronomy/?resourceId=8987&resourceType=course
```

So if you share me a course, I end up at `?resourceId=8987&resourceType=course`. Then I close the drawer, and I am still at that route.

What do you think? Could come back to it later?

frontends/ol-search-ui/src/components/ShareTooltip.tsx
abeglova: Weird, it worked fine for me at http://localhost:8063/

ChristopherChudzicki: Ah, that makes sense. [SO says](https://stackoverflow.com/a/51823007/2747370)
> This requires a secure origin — either HTTPS or localhost 

I'm running at http://cac.od.odl.local:8063/infinite/fields/philosophy/ ... I guess I don't need to do that, or maybe i can figure out how to tell chrome that cac.odl.local is a secure context

frontends/ol-search-ui/src/components/ShareTooltip.tsx
abeglova: Good catch that the tooltip isn't tab operable. I'm going to fix this in a followup pr

Issue #3714, opened 2022-08-31
Bundle size is unnecessarily big (open discussions, infinite corridor)


ChristopherChudzicki:
The bundle sizes for infinite-corridor and open-discussions have both jumped considerably recently. For example, `[@fakerjs/faker](https://www.npmjs.com/package/@faker-js/faker)` is bloating both bundles by `2.6M`  even though we only use it for tests.

Shinigami92:
I had a look into your repo via https://github.com/mitodl/open-discussions/search?p=1&q=faker
Looks like `lorem` is the only API that calls localized values, but even there you take the default faker instance without changing the locale to non-"en"
So I would highly recommend for now to use `import { faker } from "@faker-js/faker/locale/en"` to reduce the faker instance from around 4mb to 600kb

For later things please subscribe to: https://github.com/faker-js/faker/issues/1340 and https://github.com/faker-js/faker/issues/1351
You might be specially interested in the `BaseFaker` we want to introduce

ChristopherChudzicki:
@Shinigami92 Thanks for taking a look, and your work on faker! Interested to follow those issue, especially new module structure. Since it's a dependency we only use in development, the size of faker doesn't really worry me, it's the fact that it was included in our production build at all. (Which is easy for us to avoid with some simple changes to our local modules.)

ChristopherChudzicki:
Closing this for now ; #3725 did a lot. More could be done, of course.

PR #3725, opened 2022-09-06
Cc/smaller bundle

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings

#3714 

#### What's this PR do?
1.  This PR reduces the bundle size of open discussions and infinite corridor:
    ```
                            master      this branch
    ------------------------------------------------                        
    Open Discussions:       7.5M        4.2M
    Infinite Corridor:      5.8M        2.5M
    ```
2. Additionally, added [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer) to both webpack configs so that we can more easily see what is contributing to our bundle size.

**How does this reduce bundle size?**
- Primarily, this reduces the bundle size by removing the factory functions from our local modules's top-level exports. So:
```ts
// Before:
import { makeLearningResource} from "ol-search-ui"
// Now:
import { makeLearningResource} from "ol-search-ui/build/factories"
``` 
See https://github.com/mitodl/open-discussions/issues/3714 for more.

1. In the `watch` container, run `NODE_ENV=production yarn run build`.
2. Run `ls -lh frontends/infinite-corridor/build` and `ls -lh frontends/open-discussions/build`.
    - Check that the JS bundles are smaller like I said. Check that there is
    - Check that there is no report.html file
3. In the container again, run `WEBPACK_ANALYZE=True NODE_ENV=production yarn run build`.
4. Open `frontends/infinite-corridor/build/report.html` in your browser to see the bundle analysis.

#### Screenshots 

This is the infinite corridor bundle report... There's still some work we could do, e.g., CKEditor is only used by moderators, I think it could be loaded dynamically. But that can be looked into separately.

Screen Shot 2022-09-07 at 10 26 30 AM


frontends/infinite-corridor/package.json
```diff
@@ -31,7 +31,6 @@
     "ol-util": "workspace:*",
     "ol-widgets": "workspace:*",
     "postcss-loader": "^7.0.1",
-    "ramda": "^0.28.0",
```
ChristopherChudzicki, line 34: I also removed all usage of ramda in this PR because:
- it was easy to remove. All of our usages have simple native JS alternatives or analogous lodash functions
- we were including it twice in our bundle accidentally.

If we need it for something, we can definitely add it back.

frontends/infinite-corridor/src/api/fields/factories.ts
```diff
@@ -1,7 +1,8 @@
 import { faker } from "@faker-js/faker"
-import * as R from "ramda"
-import { makePaginatedFactory, Factory } from "ol-util"
-import { LearningResourceType, factories } from "ol-search-ui"
+import { makePaginatedFactory, Factory } from "ol-util/build/factories"
```
ChristopherChudzicki, line 2: I would very much have preferred an import from `ol-util/factories` instead of `ol-util/build/factories`. It's fairly easy to configure that with Node [subpath exports](https://nodejs.org/api/packages.html#subpath-exports).

But as far as I can tell, making that work with Typescript means moving to ESM-style module resolution, which is a fairly big change. So I went for the uglier `ol-util/build/factories` imports. From https://github.com/microsoft/TypeScript/issues/33079#issuecomment-1216866007 (emphasis mine):

> TypeScript doesn’t yet have a resolution mode suitable for most bundlers, and **package.json exports are not yet configurable independently of the module resolution mode on the whole**. This is discussed at length in https://github.com/microsoft/TypeScript/issues/50152.

Related: https://www.typescriptlang.org/docs/handbook/esm-node.html

frontends/infinite-corridor/src/scss/fieldpage.scss
```diff
@@ -31,8 +31,8 @@ $avatarSize: 60px;
 $carouselSpacing: 24px;
 .ic-carousel-card {
   height:100%;
-  margin-left: $carouselSpacing/2;
-  margin-right: $carouselSpacing/2;
+  margin-left: $carouselSpacing*0.5;
+  margin-right: $carouselSpacing*0.5;
```
ChristopherChudzicki, line 35: This is from running `npx sass-migrator division **/*.scss`, which fixed some deprecated Sass syntax.

(*At one point on this branch I had a failing build that I thought might be related to a sass issue, so I was trying to clean up some warnings.*)

frontends/ol-util/src/lib/utils.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  it("Returns true for empty objects, strings, sets, and arrays, and maps", () => {
+    expect(u.emptyOrNil("")).toBe(true)
+    expect(u.emptyOrNil([])).toBe(true)
+    expect(u.emptyOrNil(new Set())).toBe(true)
+    expect(u.emptyOrNil({})).toBe(true)
+    expect(u.emptyOrNil(new Map())).toBe(true)
+  })
+
+  it("Returns true for numbers", () => {
+    // _.isEmpty(5) returns true; this is different from ramda.
```
ChristopherChudzicki, line 40: If the different ever matters to us, I would be surprised.

frontends/ol-widgets/src/factories.ts
```diff
@@ -1,5 +1,6 @@
 import { faker } from "@faker-js/faker"
```
Shinigami92, line 1:
```suggestion
import { faker } from "@faker-js/faker/locale/en"
```

As you are not changing the locale **at runtime**, just use the `en` locale

frontends/infinite-corridor/src/api/fields/factories.ts
```diff
@@ -1,7 +1,8 @@
 import { faker } from "@faker-js/faker"
```
Shinigami92, line 1:
```suggestion
import { faker } from "@faker-js/faker/locale/en"
```

frontends/ol-search-ui/src/factories.ts
```diff
@@ -1,9 +1,8 @@
 //@ts-expect-error casual-browserify does not have typescript types
 import casual from "casual-browserify"
 import { faker } from "@faker-js/faker"
```
Shinigami92, line 3:
```suggestion
import { faker } from "@faker-js/faker/locale/en"
```

frontends/infinite-corridor/src/api/fields/factories.ts
ChristopherChudzicki: _Aside for other mitodl folks: Shinigami92 is one of the faker-js maintainers [previously noted](https://github.com/mitodl/open-discussions/issues/3714#issuecomment-1239102062) that using specific imports like this would reduce faker's footprint._

Personally, I'm somewhat inclined to keep the faker imports as is because:
- faker should not influence the overall bundle size at all. The fact that faker was influencing our bundle size prior to this PR was a mistake.
- the subpath import is a little more verbose. I'm not really confident that we'd use it consistently, and per ☝️ I do not understand the benefit.

Shinigami92: If I've misunderstood and there is some benefit I'm not seeing, feel free to let me know.

Shinigami92: No you are right :+1: 
I would suggest to upgrade to v8 as soon as it's available

abeglova (approved):
LGTM

PR #3728, opened 2022-09-07
support videos in drawer

abeglova:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

(Required)

#### What's this PR do?
(Required)

(Required)

(Optional)

(Optional)

(Optional)

(Optional)


frontends/ol-search-ui/src/components/ExpandedLearningResourceDisplay.tsx
```diff
@@ -30,6 +32,40 @@ type Props = {
   runId: string | number | undefined
   setShowResourceDrawer: (params: ResourceIdentifiers | null) => void
 }
+export const formatDurationClockTime = (value: string) => {
```
ChristopherChudzicki, line 35: **Requested Change:** Could you move the comment to a JSDoc docstring:. Something like

```ts
/**
 * Format an ISO-8601 duration string so to a readable format
 * The logic here ensures that if there is a colon (:) to the left
 * of a time component (minutes, seconds) it is zero-padded
 * hours are not included if they are zero
 * this follows what most humans would consider a reasonable "clock display" format
 *
 * Example return values:
 *  - 3:00:01
 *  - 43:07
 *  - 3:09
 *  - 0:47
 */
export const formatDurationClockTime = (value: string): string => { /* */ }
```
Then it shows up on hover:
Screen Shot 2022-09-08 at 10 01 17 AM

**Other Suggestions:**
- I would be inclined to extract it to `ol-utils`. Not necessarily for reuse, but because it is a self contained chunk of code that makes perfect sense in isolation.
- IMO, it would also be worth writing a few simple tests for this.
- Use `padStart` instead of ``0${duration.minutes().toString()}`.slice(-2)`:
    - the lodash [padStart](https://lodash.com/docs/4.17.15#padStart) function
    - or the [builtin String.prototype.padStart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart), though [caniuse](https://caniuse.com/?search=padstart). says its browser support is only 95%

frontends/ol-search-ui/src/components/ExpandedLearningResourceDisplay.test.tsx
```diff
@@ -45,6 +48,35 @@ describe("ExpandedLearningResourceDisplay", () => {
...
+    expect(screen.getByText("Date Posted:")).toBeInTheDocument()
+
+    const a = screen.getByRole("link")
+    assertInstanceOf(a, HTMLAnchorElement)
+    expect(a.href).toBe(resource.url.toLowerCase())
+    expect(a.dataset.cardChrome).toBe("0")
+    expect(a.dataset.cardControls).toBe("0")
+    expect(a.dataset.cardKey).toBe("fake")
+    expect(a).toHaveClass("embedly-card")
+    expect(typeof a.dataset.cardKey).toBe("string")
```
ChristopherChudzicki, line 77: No ned for this, redundant with line 75.

ChristopherChudzicki (changes requested):
Works well and changes make sense to me 👍 

A few minor requests and some suggestions.

ChristopherChudzicki (approved):

Issue #3691, opened 2022-08-12
Webpack builds often take a long time or fail


mbertrand:
### Steps to Reproduce

Run `docker-compose up`

### Expected Behavior

Within a minute or two, webpack is done building the front-end

### Actual Behavior

Occasionally, webpack takes >15 minutes or fails to complete altogether, and `docker-compose up` has to be restarted.

ChristopherChudzicki:
Possibly related... I have noticed that `docker compose up` fails to build infinite corridor (because `ol-util` has not been built) but then if I edit an infinite corridor file (triggering a rebuild via watch) then everything is fine.  The prod build is set up with `yarn workspace foreach --topological`, where the `topological` bit ensures that dependent workspaces are build first, but I don't think the ordering is guaranteed the way I set up watch mode for local dev

ChristopherChudzicki:
@mbertrand I believe this was cleared up by https://github.com/mitodl/open-discussions/pull/3874. If I'm wrong and it's still an issue, please re-open.

PR #3736, opened 2022-09-09
Remove transient errors in start-dev

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing

https://github.com/mitodl/open-discussions/issues/3691

#### What's this PR do?
This PR ensures that all projects have been build before the dev servers starts.

For context:
- `yarn run build` would build the all the local packages in using yarn's `--topological` flag, which says to build dependencies before dependents
- Previously, `yarn run start-dev` (which `docker compose up` uses) would start watching all the projects concurrently. Since they all start concurrently, sometimes `ol-search-ui` would try to build before its dependents and fail. 

Run `docker compose up`. Webpack should eventually finish building and you'll be able to visit infinite corridor.

You can use
```
docker compose logs --follow --no-log-prefix  --tail 1000 watch
``` 
to monitor the build with less clutter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "scripts": {
         "foreach-with-fallback": "yarn workspaces foreach --verbose --topological exec node ${PROJECT_CWD}/scripts/yarn-run-with-fallback.js $0 $1",
         "postinstall": "yarn run build",
-        "start-dev": "yarn workspaces foreach --verbose --parallel --jobs unlimited --interlaced exec node ${PROJECT_CWD}/scripts/yarn-run-with-fallback.js start-dev global:start-dev",
+        "start-dev": "yarn run build && yarn workspaces foreach --verbose --parallel --jobs unlimited --interlaced exec node ${PROJECT_CWD}/scripts/yarn-run-with-fallback.js start-dev global:start-dev",
         "build": "run foreach-with-fallback build global:build",
         "clean": "run foreach-with-fallback clean global:clean",
         "lint-check": "run foreach-with-fallback lint-check global:lint-check",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
     "scripts": {
         "foreach-with-fallback": "yarn workspaces foreach --verbose --topological exec node ${PROJECT_CWD}/scripts/yarn-run-with-fallback.js $0 $1",
         "postinstall": "yarn run build",
-        "start-dev": "yarn workspaces foreach --verbose --parallel --jobs unlimited --interlaced exec node ${PROJECT_CWD}/scripts/yarn-run-with-fallback.js start-dev global:start-dev",
+        "start-dev": "yarn run build && yarn workspaces foreach --verbose --parallel --jobs unlimited --interlaced exec node ${PROJECT_CWD}/scripts/yarn-run-with-fallback.js start-dev global:start-dev",
```
ChristopherChudzicki, line 10: **Downside:** One downside to this approach is that if `yarn run build` fails (e.g., due to a typescript error) then dev mode won't successfully start. 

If errors happen after the initial build during development, everything is fine since the dev servers don't mind that.

mbertrand (commented):
watch exited with an error (but not sure what the error is):

```
watch_1    | ➤ YN0000: ┌ Link step
watch_1    | ➤ YN0008: │ open-discussions-root@workspace:. must be rebuilt because its dependency tree changed
watch_1    | ➤ YN0009: │ open-discussions-root@workspace:. couldn't be built successfully (exit code 1, logs can be found here: /tmp/xfs-a596fe1c/build.log)
watch_1    | ➤ YN0000: └ Completed in 1m 38s
watch_1    | ➤ YN0000: Failed with errors in 2m 7s
open-discussions_watch_1 exited with code 1
```

Ran again after deleting all `node_modules` folders but got the same result.  Here is full output:
https://gist.github.com/mbertrand/fc3c581515cd5aeefada3eb2b2a6d8dc

mbertrand (commented):
Made progress, the webpack build now succeeds, but there are issues with the open-discussions frontend urls.  Both the home page and `/learn` urls show a blank page and throw this error:

Screen Shot 2022-09-12 at 10 23 05 AM

However I am getting that error on the master branch too.

mbertrand (approved):
After clearing out the root `node_modules` folder again, it works.

Issue #3737, opened 2022-09-12
Make field page widget column wider


ChristopherChudzicki:
Widgets are too narrow. Especially youtube videos, which can overflow the widget column on field pages

- [ ] make widget column wider,  like on Open Discussions
- [ ] ensure widget column looks ok on different device widths, and make mobile breakpoints more consistent throughout.

PR #3742, opened 2022-09-13
Wider widget sidebar; more consistent breakpoints

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#3737

#### What's this PR do?
This PR:
- Fixes the vertical centering of app header
- Makes the widget column in field pages a bit wider, and displays 3, 2, or 1 courses in the carousel depending on the current  [MUI Breakpoint](https://mui.com/material-ui/customization/breakpoints/)
- Additionally, I tried to make our breakpoint usage a bit more consistent within Infinite Corridor. (*There are some behavior changes here, but they seem OK to me. (E.g., the search page used to "go mobile" at 998px; now it "goes mobile" at 839px.*)

"Breakpoints" tell the website to respond different at screen widgets. Usually, breakpoints are implemented with CSS [media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries). Our use of breakpoints has been a bit inconsistent:
- In Open Discussions, we use three breakpoint systems:
    1. material web, via classes like `mdc-layout-grid__cell--span-8`
    2. our own breakpoints, via sass mixin, e.g.,  `breakpoint(desktop)`
    3. a JS help `useDeviceCategory` which has somewhat different breakpoint values than our CSS
- In Infinite Corridor, we haven't been entirely consistent, but IMO we should try and stick with MUI's breakpoint system (the values of which we can customize if we like).
     1. We are already using MUI's Grid component instead of stuff like `mdc-layout-grid__cell--span-8` anyway
     2. We may still want a sass mixin, which I've added here.
     3. We can use MUI's `useMediaQuery` hook instead of `useDeviceWidth`.

Load a field page (with widgets) and the search page and check that stuff looks reasonable at different widths.

https://user-images.githubusercontent.com/9010790/190007675-f360afed-f843-4ec8-9996-1b0a52fb33f6.mov

Screen Shot 2022-09-13 at 4 57 44 PM

**Search Page:** Minor changes, really just using MUI's breakpoints

Screen Shot 2022-09-13 at 4 32 06 PM

Screen Shot 2022-09-15 at 4 45 07 PM Screen Shot 2022-09-15 at 5 03 25 PM


ChristopherChudzicki:
@Ferdi Screenshots and such for you (mostly the video)

frontends/infinite-corridor/src/scss/learning-resources.scss
```diff
@@ -1,17 +1,16 @@
...
 @use "ol-search-ui/assets/learning-resource-drawer";
 
-@import "./breakpoints";
-
 $cardListSpacer: 1.5em;
 
 .ic-resource-card {
 
-  @include breakpoint(phone) {
+  @include breakpoints.down(sm) {
```
ChristopherChudzicki, line 13: This is slightly different (old: 580px, new: 600px) but the general behavior is unchanged

frontends/infinite-corridor/src/scss/breakpoints.scss
```diff
@@ -0,0 +1,76 @@
+/*
```
ChristopherChudzicki, line 1: This replaces `frontends/infinite-corridor/src/scss/_breakpoints.scss`. This version is a bit different from Open Disucssions, but is consistent with MUI. And Open Discussions had ...a variety ... of breakpoint sets. See PR description.

frontends/infinite-corridor/src/test-utils/setupJest.ts
```diff
@@ -35,3 +35,8 @@ afterEach(() => {
   global.SETTINGS = _createSettings()
   resetApi()
 })
+
+const DEFAULT_DEVICE_WIDTH = "1200px"
+beforeAll(() => {
+  window.matchMedia = createMatchMediaForJsDom({ width: DEFAULT_DEVICE_WIDTH })
+})
```
ChristopherChudzicki, line 42: This lets us use MUI's `useMatchMedia` hook in JSDom tests. See docstring for `createMatchMediaForJsDom`

frontends/ol-search-ui/src/components/SearchFilterDrawer.tsx
```diff
@@ -1,7 +1,9 @@
 import React, { useCallback, useState } from "react"
```
ChristopherChudzicki, line 1: General changes here:

- replaced `useDeviceCategory` with an `alwaysOpen` prop. The value of `alwaysOpen` (which SearchPage sets based on `useBreakpoint`.
- fixed issue where "close" button was obscured by header
- Some styling changes (see screenshots below) and replaced some divs with actual buttons

**OLD / NEW**
Screen Shot 2022-09-15 at 5 25 55 PM / Screen Shot 2022-09-15 at 5 03 25 PM

frontends/ol-util/src/lib/utils.ts
```diff
@@ -17,41 +19,12 @@ export const initials = (title: string): string => {
...
-    window.addEventListener("resize", cb)
-    return () => {
-      window.removeEventListener("resize", cb)
-    }
-  }, [])
-
-  return width
-}
-
-export const useDeviceCategory = () => {
```
ChristopherChudzicki, line 45: In Open discussions, the `PHONE_WIDTH` and `TABLET_WIDTH` values are different from our breakpoint values used in CSS.

`useMuiBreakpoint` will help us be more consistent.

mbertrand (commented):
Code looks good to me.  Only issue I noticed is that at mobile width, if I click on "Manage widgets", nothing seems to happen at first glance.  The widget editor section does appear, but at the bottom of the page and I have to scroll down to find it.  Would it be possible to have it appear at the top of the page instead at narrow widths?

ChristopherChudzicki:
> Code looks good to me. Only issue I noticed is that at mobile width, if I click on "Manage widgets", nothing seems to happen at first glance. The widget editor section does appear, but at the bottom of the page and I have to scroll down to find it. Would it be possible to have it appear at the top of the page instead at narrow widths?

Yeah, that's not the best. Still, I'm inclined to leave this as-is for now because: it's simplest to keep the editable widget list in the same position as the read-only widget list, and  mobile editing doesn't seem like a too important use-case, so I hesitate to add extra code for that behavior.

Also, before adding any handling for this, I'd like to double check where we want the widgets for mobile view.

PR #3802, opened 2023-01-05
Simplify the docker/compose setup

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing

N/A

#### What's this PR do?
I'd like to try out some features in newer elasticsearch/opensearch, but before I do that I'd like to have a cleaner foundation to work on with docker - this cleans up the configuration to be in line with how mitx online is setup by using compose profiles and multistage docker builds.

You should be able to run `docker-compose up` as normal and not have the notebook container start. Running `docker-compose --profile notebook up` should give you the notebook.


mbertrand (changes requested):
I'm getting 404's on this branch for the following:

http://od.odl.local:8062/style.js
http://od.odl.local:8062/root.js

After switching back to the master branch and rebuilding, the files were available.

Switched back to this branch again, rebuilt, and the files are missing again.

PR #3821, opened 2023-02-01
Fix mgmt commands for importing content files

### LINKED ISSUE #3820: Mgmt commands for syncing content files should default to chunk size specified in settings, -c option needs to be integer
### Steps to Reproduce

Run any of the `backpopulate_<platform>_files` management commands with and without a chunk size parameter

### Expected Behavior

Without the chunk size parameter, chunk size should be equal to `settings.LEARNING_COURSE_ITERATOR_CHUNK_SIZE` (xpro, mitx, mitxonline) or `settings.OCW_ITERATOR_CHUNK_SIZE` (OCW)

With the `--chunk-size` parameter, the command should run without error using that size.

### Actual Behavior

The default chunk size parameter is hardcoded to 1000

Using the `--chunk-size` parameter throws an exception because it is treated as a string instead of an integer.

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #3820 

#### What's this PR do?
Adds `type=int,` to the optional `chunk_size` argument specification and sets the default to the same as defined in settings.

Run any or all of the modied mgmt commands, with and without a `--chunk-size` parameter.  They should all start without errors (you don't need to wait for them to finish), assuming all the required .env settings are configured with the same values as on RC: 

```
XPRO_LEARNING_COURSE_ACCESS_KEY=
XPRO_LEARNING_COURSE_SECRET_ACCESS_KEY=
XPRO_LEARNING_COURSE_BUCKET_NAME=
XPRO_CATALOG_API_URL=https://xpro.mit.edu/api/programs/
XPRO_COURSES_API_URL=https://xpro.mit.edu/api/courses/

OCW_CONTENT_ACCESS_KEY=
OCW_CONTENT_SECRET_ACCESS_KEY=
OCW_CONTENT_BUCKET_NAME=
OCW_LEARNING_COURSE_BUCKET_NAME=

MITX_ONLINE_BASE_URL=https://mitxonline.mit.edu/
MITX_ONLINE_COURSES_API_URL=https://mitxonline.mit.edu/api/courses/
MITX_ONLINE_PROGRAMS_API_URL=https://mitxonline.mit.edu/api/programs/
MITX_ONLINE_LEARNING_COURSE_BUCKET_NAME=

EDX_API_URL=
EDX_API_CLIENT_ID=
EDX_API_CLIENT_SECRET=
EDX_API_ACCESS_TOKEN_URL=
EDX_LEARNING_COURSE_BUCKET_NAME=
```

### DIFF
--- a/course_catalog/management/commands/backpopulate_edx_files.py
+++ b/course_catalog/management/commands/backpopulate_edx_files.py
@@ -1,6 +1,7 @@
 """Management command for populating MITx course run file data"""
 
 from django.core.management import BaseCommand
+from django.conf import settings
 
 from course_catalog.tasks import import_all_mitx_files
 from open_discussions.utils import now_in_utc
@@ -16,7 +17,8 @@ class Command(BaseCommand):
             "-c",
             "--chunk-size",
             dest="chunk_size",
-            default=1000,
+            default=settings.LEARNING_COURSE_ITERATOR_CHUNK_SIZE,
+            type=int,
             help="Chunk size for batch import task",
         )
 
--- a/course_catalog/management/commands/backpopulate_mitxonline_files.py
+++ b/course_catalog/management/commands/backpopulate_mitxonline_files.py
@@ -3,6 +3,7 @@
 from django.core.management import BaseCommand
 
 from course_catalog.tasks import import_all_mitxonline_files
+from open_discussions import settings
 from open_discussions.utils import now_in_utc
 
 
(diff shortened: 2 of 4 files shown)


JenniWhitman (approved):
I was able to run mgmt commands I found, changing chunk size, and I like moving the default out to settings rather than hard coding a value here.  As far as I can tell, works as expected wtihout any negative impact.

PR #3826, opened 2023-02-09
Documentation and .env changes for initial setup

JenniWhitman:
#### Pre-Flight checklist

- [ X ] Testing
  - [ X ] Changes have been manually tested

N/A

#### What's this PR do?
While setting up I found that there was a missing `.env` value and instructions to add it as well as a bad link on the readme. To save someone else the time I lost, I am putting the small doc and `.env.example` changes up for review/merge.

You can check that the link works in the readme and that the value I added to example `.env` file makes sense.  This is to allow the webpack to build. Without adding this value, my image was failing to fully build.

I already said this, but I lost a good amount of time trying to figure out why this wasn't set as it wasn't super obvious what was feeding $HOST in the docker build. Also had to track down the generic web app setup to the new github.io page :)


README.md
```diff
@@ -38,6 +38,11 @@ The following settings must be configured before running the app:
     is not needed. It's recommended that you eventually configure the site to be able
     to send emails. Those configuration steps can be found [below](#enabling-email).
 
+- `OPEN_DISCUSSIONS_HOSTNAME`
+    
+    Sets the hostname for DOCKER_HOST in setup. Should likely be whatever you set the host to in your /etc/hosts
```
mbertrand, line 43: Minor nitpick - this is a separate a .env variable from `DOCKER_HOST`, only seems to be used by webpack for building the front end.  "Sets the hostname required by webpack for building the front end"?

mbertrand (approved):
Looks great! Just 1 minor suggestion, feel free to take it or leave it.

README.md
JenniWhitman: You're not wrong - I like it! Will update before doing the thing :)

mbertrand (approved):
👍

Issue #3819, opened 2023-01-31
Userlist functionality in infinite


ChristopherChudzicki:
We want to port the list creation functionality from open-discussions to infinite-corridor:

### Acceptance Criteria
- [ ] list creation should function similarly (open a modal, fill out the form)
- [ ] view lists at `infinite/lists`, where can edit list info / delete lists
- [ ] view individual list at `infinite/lists/:id`
- [ ] add items to lists via card action
- [ ] Public lists should NOT show "Edit" buttons to non-owners

### Out of scope
- Drag re-ordering learning paths.
- 403 handling for private lists

ChristopherChudzicki:
Small update on this:

@Ferdi @mbilalmughal Some screenshots of migrated (partial) userlist functionality. Still working on the bookmark / add-to-list flow. I'll open a PR later today for the work so far.

Notable changes from existing existing version (@mbilalmughal existing version is at https://discussions-rc.odl.mit.edu/learn/lists/)
- default choices for list type and privacy level in creation dialog
- Using Material UI (MUI) default styles for forms (see creation dialog). If we prefer the [old look](https://user-images.githubusercontent.com/9010790/217878082-70e05ed1-f69c-4ed9-8e7e-c522858f158e.png), we can go back to that.

### Userlist Listing
<img width="1865" alt="Screenshot 2023-02-09 at 9 26 07 AM" src="https://user-images.githubusercontent.com/9010790/217876790-585330c6-7832-498f-8f0c-c1b575a755c4.png">

### Individual Userlist

<img width="1858" alt="Screenshot 2023-02-09 at 9 25 45 AM" src="https://user-images.githubusercontent.com/9010790/217876889-3da31ac5-4340-4364-ab36-5e79254e9d7d.png">

### Create/Edit userlist
<img width="1855" alt="Screenshot 2023-02-09 at 10 47 42 AM" src="https://user-images.githubusercontent.com/9010790/217876962-59f5e6c1-4634-42b7-aa1c-fdb8f91a1865.png">

ChristopherChudzicki:
@Ferdi @pdpinch: Nathan's review of #3835 has prompted me to clean up a few things related to users (e.g., #3864, #3865). One question for you about userlists in particular.

**Question:** Should we show the userlist listings page to anonymous users? I.e., the analog of https://discussions-rc.odl.mit.edu/learn/lists ? Learn currently does show this to anonymous users (including a "Create" button...) but I do not think this makes sense:

[...]

PR #3831, opened 2023-02-10
Move userlist types out of fields, update TS

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Toward #3819 

#### What's this PR do?
While working on #3819 I came across some issues with our `LearningResource` TS declarations that motivated me to rename a bunch of stuff. This PR includes those renames and a few small other changes. There are no behavioral changes in this PR.

There are no intentional behavioral changes in this PR. The only things that are possibly affected are in Infinite Corridor. You could `docker compose up` and check that:
- http://od.odl.local:8063/infinite still looks OK
- the individual linked field pages still look OK
- http://od.odl.local:8063/infinite/search still looks OK

**To construct field data:** I would do it this way
1. Create some (2+) public userlists at http://od.odl.local:8063/learn (bookmark icon on cards)
2. Create a field via admin panel http://od.odl.local:8063/admin/channels_fields/fieldchannel/
3. Associate userlists with your field via PATCH request at http://od.odl.local:8063/api/v0/fields/YOUR_FIELD_NAME/
    - `featured_list: <list_id>`
    - `subfields: <array of list ids>`

See my comments on files.


frontends/ol-widgets/src/components/editing/ManageWidgetDialog.test.tsx
```diff
@@ -72,11 +72,12 @@ describe("ManageWidgetDialog (Editing)", () => {
   })
 
   it("passes classes to relevant elements", () => {
+    const fakerClassName = () => faker.unique(faker.lorem.slug)
```
ChristopherChudzicki, line 75: I added this because a test as being flaky, I *believe* because of collisions between the randomly generated classnames.

frontends/infinite-corridor/src/api/fields/factories.ts
```diff
@@ -1,50 +1,16 @@
...
-import { LearningResourceType } from "ol-search-ui"
-import * as factories from "ol-search-ui/build/factories"
-import { times } from "lodash"
-import type { FieldChannel, UserList, UserListItem } from "./interfaces"
+import { makeUserList } from "ol-search-ui/build/factories"
+import type { UserList } from "ol-search-ui"
+import { PrivacyLevel, LearningResourceType } from "ol-search-ui"
+import type { FieldChannel } from "./interfaces"
 
-export const makeUserList: Factory<UserList> = overrides => {
```
ChristopherChudzicki, line 8: UserLists are a "learning resource", returned from our the `course-catalog` django app. Some of its type definitons and its react-query hooks were in `api/fields` because fields use userlists.

But userlists should go with all the other learning resources.

**For now:** that means it goes in `ol-search-ui`. That package has too much stuff that is not directly search related. I'd like to split it out into two packages, `ol-search-ui` for search stuff (searchbox, facets) and `ol-learning-resources` for learning resources (typedefs, card, drawer).

frontends/infinite-corridor/src/api/learning-resources/hooks.ts
```diff
@@ -1,9 +1,34 @@
...
+} from "ol-search-ui"
+import type { PaginatedResult, PaginationSearchParams } from "ol-util"
+import axios from "../../libs/axios"
+import { useQuery, UseQueryResult } from "react-query"
+import { urls, keys, UserListOptions } from "./urls"
 
 const useResource = (type: string, id: number) => {
-  return useQuery<LearningResourceResult>(urls.resource(type, id))
+  const url = urls.resourceDetails(type, id)
+  const key = keys.resourceDetails(type, id)
```
ChristopherChudzicki, line 13: React query requires a query key for all queries: `useQuery(key, fetcher?)`. Previously we were using API urls as string-based keys and relying on a default fetcher specified in our overall react query configuration.

The alternative (which I've switched to here) is array (`string[]`) keys. _We could still use a default fetcher, but I decided its not worth the effort._

Array-based keys have two advantages:
- they make invalidating caches easier. E.g., if you have several queries with keys `["userlist", "details", 5]`, `["userlist", "listing"]`, `["userlist", "something else"]`, then you can invalidate all those keys by calling `queryClient.invalidateQueries(["userlist"])`. Ie, react-query treats array-based query keys as something like a directory structure.
- we use react query 3, but react-query 4 only supports array-based keys

frontends/infinite-corridor/src/pages/field-details/EditFieldAppearanceForm.tsx
```diff
@@ -65,7 +65,7 @@ const EditFieldAppearanceForm = (props: FormProps): JSX.Element => {
                 ) : null}
               </div>
             </div>
-            <div className="row form-item">
+            <div className="form-row form-item">
```
ChristopherChudzicki, line 68: Our `row` class was specifically for forms, and `row` is hard to search for. So I changed it to `form-row`, which is a little easier.

package.json
```diff
@@ -76,6 +76,6 @@
         "swc-loader": "^0.2.3",
         "ts-jest": "^28.0.5",
         "ts-node": "^10.8.1",
-        "typescript": "^4.8.2"
+        "typescript": "^4.9.5"
```
ChristopherChudzicki, line 79: Bump typescript. In #3833 it was slightly convenient to have the [`satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator, which as included in TS 4.9.

Corresponding bumps to swc and prettier to allow using `satisfies`.

mbertrand (approved):
👍

PR #3835, opened 2023-02-13
CRUD for userlists in TS

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#3819 

#### What's this PR do?
This PR adds userlist CRUD functionality in `/infinite`, similar to what can currently be done at https://discussions-rc.odl.mit.edu/learn/lists/, but in the new typescript codebase. Cannot add items to lists yet.

Assuming Open Discussions is set up with at least some course data, and you have logged in (for now via  http://od.odl.local:8063)

1. Visit http://od.odl.local:8063/infinite/lists. Try things like:
    - creating new lists *Add a few learning lists and learning paths. The list type should be reflected on the card*
    - submitting an invalid form for list creation (e.g., missing title)
    - editing lists after you've created them
    - deleting a list.

2. Once you've created a few lists, add some courses to them. The new TS codebase does not support this yet. Do it via http://od.odl.local:8063/learn (click the bookmark, then select a list to add to). Also add some things to the special "favorites" list.
3. View http://od.odl.local:8063/infinite/lists again. Your lists should show the course count.
4. At http://od.odl.local:8063/infinite/lists, click a list title to go to the details page for that list. At the list details page (something like `http://od.odl.local:8063/infinite/lists/<id here>`, you should be able to:
    - click individual resources to open a details drawer
    - edit the list data (e.g., title)
    - ⚠️ re-ordering learning paths not yet supported. 

In terms of code review, I would start with `UserListsListingPage` and the other files in `frontends/infinite-corridor/src/pages/user-lists`.

listing_desktop listing_mobile

form_desktopform_mobile

details_desktopdetails_mobile


frontends/infinite-corridor/src/api/learning-resources/hooks.ts
```diff
@@ -20,15 +26,102 @@ const useUserList = (id: number) => {
...
+const useUpdateUserList = () => {
+  const queryClient = useQueryClient()
+  return useMutation(updateUserList, {
+    onSuccess: (_data, variables) => {
+      queryClient.invalidateQueries({
+        queryKey: keys.userList.id(variables.id).details
+      })
+      queryClient.invalidateQueries({
+        queryKey: keys.userList.listing.all
+      })
```
ChristopherChudzicki, line 74: The `onSuccess` logic is important and has no direct tests in this PR. It's responsible for invalidating the cached queries. I did add some direct testings for the onSuccess logic in [cc/populate-userlists](https://github.com/mitodl/open-discussions/blob/cc/populate-userlists/frontends/infinite-corridor/src/api/learning-resources/hooks.test.tsx).

We could also add an `onMutate` callback here for optimistic updates.

frontends/ol-util/src/lib/utils.ts
```diff
@@ -63,3 +63,14 @@ export const formatDurationClockTime = (value: string) => {
 
   return values.join(":")
 }
+
+/**
+ * Append an 's' to the end of a string if the count is not 1. Optionally,
+ * provide a custom plural string.
+ */
+export const pluralize = (singular: string, count: number, plural?: string) => {
```
rhysyngsun, line 71: Could you write some tests for this?

frontends/infinite-corridor/src/scss/learning-resources.scss
```diff
@@ -18,12 +18,20 @@ $cardListSpacer: 1.5em;
 }
 
 .ic-card-row-list {
-  padding-left: 0;
+  padding-left: 0pt;
+  margin-left:0pt;
+  margin-right:0pt;
```
rhysyngsun, line 23: I'm seeing some formatting inconsistencies here, did we not get scss linting/formatting setup when we switched over to the new project setup?

frontends/ol-search-ui/src/components/LearningResourceCard.tsx
```diff
@@ -1,7 +1,7 @@
 import moment from "moment"
```
ChristopherChudzicki, line 1: LearningResourceCard is not really search-specific; at some point I would like to move this file (and related typedefs) to a separate package, perhaps `ol-learning-resource-ui`, so that `ol-search-ui` goes back to being more search-specific (searchbox, facets).

frontends/infinite-corridor/src/api/learning-resources/urls.ts
```diff
@@ -1,54 +1,174 @@
...
+      return podcastEpisodeUrls.details(id)
+    case LRT.Userlist:
+      return userListUrls.details(id)
+    case LRT.LearningPath:
+      return userListUrls.details(id) // LearningPaths are handled by UserList api
+    default:
+      /**
+         * This should not happen and will likely result in a 404.
+         */
+      return `/${type}s/${id}/`
```
rhysyngsun, line 98: Should this throw an error instead so we can catch a problem before it when it happens?

frontends/infinite-corridor/src/pages/user-lists/UserListDetailsPage.tsx
```diff
@@ -0,0 +1,85 @@
...
+                          className="ic-resource-card"
+                          resource={list.content_data}
+                          imgConfig={imgConfigs["row-reverse"]}
+                          onActivate={activateResource}
+                        />
+                      </li>
+                    )
+                  })}
+                </ul>
+              ))}
```
rhysyngsun, line 73: You mentioned this and `FavoritesPage` don't make sense to be the same component, but I think you could probably pull this bit out into its own component since it's the same on both pages except for the `empty-message` text.

rhysyngsun (changes requested):
A relatively quick pass over this so you have some stuff to start with on Monday, I'll do a more thorough pass over it then.

frontends/infinite-corridor/src/scss/learning-resources.scss
ChristopherChudzicki: Apparently we (I, I guess) never did (`"global:scss-lint": "echo 'TODO: Set this up.'"`). Good catch.

I just created #3848 for this. I would suggest not adding it as part of this PR. I can make a separate PR against master and, if that's merged first, rebase this.

frontends/infinite-corridor/src/api/learning-resources/urls.ts
ChristopherChudzicki: That was my initial thought, too. On the other hand, the queries are sometimes constructed dynamically from URLs, e.g., the learning resource drawer uses this query: https://discussions-rc.odl.mit.edu/infinite/fields/philosophy/?resource_id=10973&resource_type=course

So if a user copy-pastes the URL wrong  (https://discussions-rc.odl.mit.edu/infinite/fields/philosophy/?resource_id=10973&resource_type=cours), probably the correct behavior is a 404. How should that come about?

**Option 1:** The drawer page validates the URL, and shows a custom error message if it's wrong.

**Option 2:** Just try and make the request, and if it's a 404, show a "resource not found" error message.

What do you think is better? I'm not sure, but I leaned toward Option 2.

**Now:**
- none of that reasoning is clear from the function, and even if we do prefer Option 2, some comment is probably worthwhile
- At the moment, Option 2 doesn't really behave right. The resource drawer has no logic to handle failed queries. It should have that logic.

ChristopherChudzicki:
Weird. A Python test just failed. https://github.com/mitodl/open-discussions/actions/runs/4297330246/jobs/7490167955

frontends/infinite-corridor/src/api/learning-resources/urls.ts
rhysyngsun: It looks like we only use this in 2 places: tests and the `useResource` hook. It's been a while since I did JS, but if we did throw an error, would this work?

```ts
const useResource = (type: string, id: number) => {
  const key = keys.resource(type).id(id).details
  return useQuery<LearningResource>(key, async () =>{
    const url = urls.resource.details(type, id)
    const result = await axios.get(url)
    return result.data
  })
}
```
Moving `resource.details()` into the callback and making that callback an `async` would mean if the error gets thrown, it'll be a rejected promise and then `isError` on the return value of `useQuery` would reflect that. While the component doesn't handle errors at least the interface would behave correctly.

frontends/infinite-corridor/src/pages/user-lists/ManageListDialogs.tsx
```diff
@@ -0,0 +1,115 @@
...
+      title="Edit list"
+    />
+  )
+}
+
+type DeletionDialogProps = {
+  onClose: () => void
+  resource: UserList | null
+}
+const DeletionDialog: React.FC<DeletionDialogProps> = ({
```
rhysyngsun, line 41: Nit: should this and the props be `DeleteListDialog` to align to the other two dialogs' naming convention?

rhysyngsun (changes requested):

frontends/infinite-corridor/src/api/learning-resources/urls.ts
ChristopherChudzicki: Great idea! Works well. I added that to this PR, but did NOT add a test for it. I added tests for these react-query hooks in the followup branch, so I just added the test to that: https://github.com/mitodl/open-discussions/blob/cc/populate-userlists/frontends/infinite-corridor/src/api/learning-resources/hooks.test.tsx#L53

It turned out this change broke a few tests (exposed an issue, I guess): The `makeLearningResource` factory was creating resources with type `"favorite"`, which was then triggering requests like `/favorites/1234`, which is invalid. "favorites" is fairly separate from the other resource types, so I [separated it out](https://github.com/mitodl/open-discussions/pull/3835/files#diff-dd3d4732baaaff09891b00335b33b3ac776c3ccf3a72d60c06e094e81c9d60d8R23).

rhysyngsun (changes requested):
Generally looking good, although if I try to view a private list when not logged in as that user the UI gets stuck in a "Loading..." state rather than providing an error message.

ChristopherChudzicki:
@rhysyngsun Good points. Since this PR is already a bit large, I'd like to handle some of these routing issues elsewhere:

- 403s in https://github.com/mitodl/open-discussions/issues/3864
- 404s in https://github.com/mitodl/open-discussions/issues/3865

Additionally, your comment made me realize that the Edit and Create buttons are currently visible to anonymous users, which I think it best to address as part of this PR.

ChristopherChudzicki:
@rhysyngsun Following up on my previous comment...

> Additionally, your comment made me realize that the Edit and Create buttons are currently visible to anonymous users, which I think it best to address as part of this PR.

I have updated this PR to only show the edit button if it's your list. I did **not** address the Create button on the lists listing page. _Rationale: I don't think it even makes sense to show the lists listing page to anonymous users, though I'm double checking with Ferdi and Peter ([here](https://github.com/mitodl/open-discussions/issues/3819#issuecomment-1462860274)... let me know if you have thoughts). IMO, Probably lists listing should redirect to login, so I would handle it in #3864 _

There is one additional change since your last review. You noted 
>  the UI gets stuck in a "Loading..." state 

Turns out it wasn't really stuck, it was just (slowly) retrying some queries. Apparently react query's default behavior is to retry all failed queries 3 times with an exponential backoff. There's not really any point in retrying 403, etc, so I changed the config a little to be less aggressive in retries. https://github.com/mitodl/open-discussions/pull/3835/files#diff-b83757bb09981196e33ccda995b461855cda33356a60412e624422c5b80285e9R27

frontends/infinite-corridor/src/pages/user-lists/ManageListDialogs.test.tsx
```diff
@@ -0,0 +1,340 @@
...
+ *
+ */
+const inputs = {
+  list_type: {
+    [LearningResourceType.LearningPath]: () =>
+      screen.getByLabelText("Learning Path", { exact: false }),
+    [LearningResourceType.Userlist]: () =>
+      screen.getByLabelText("Learning List", { exact: false })
+  },
+  pivacy_level: {
```
rhysyngsun, line 46: Typo `privacy_level`

frontends/infinite-corridor/src/pages/user-lists/UpsertListDialog.tsx
```diff
@@ -0,0 +1,226 @@
...
+      />
+      <RadioChoiceField
+        className="form-row"
+        name="privacy_level"
+        label="Privacy"
+        choices={priaceChoices}
+        value={formik.values.privacy_level}
+        row
+        onChange={formik.handleChange}
+      />
```
rhysyngsun, line 170: There's a typo in `priaceChoices`, but when I thought on it more, is there even any value to show this option to non-staff because they can only make private lists and they'll only ever have one option? Should we just do something where it's a hidden input with `PRIVATE` if they're non-staff, otherwise they get the radio and you don't need to filter the options?

rhysyngsun (changes requested):

frontends/infinite-corridor/src/pages/user-lists/UpsertListDialog.tsx
ChristopherChudzicki: I guess showing it communicates that the list is private, but not much else. Showing nothing seems better so I've hidden it for now. I'll solicit some more feedback from Bilal and Ferdi on the followup PR that has "add to list" functionality.

Here is the current version (for non staff)

Screenshot 2023-03-10 at 2 59 26 PM

rhysyngsun (approved):
LGTM

PR #3837, opened 2023-02-14
Assign department to MITX Online, MITx, and OLL courses

### LINKED ISSUE #3827: Assign department to MITX Online, MITx, and OLL courses
Assign a `department` value for each `Course` imported from MITx Online (`platform='mitxonline'`), Open Learning Library (`platform='oll'`, and MITx (`platform='mitx'`).  The department number can be deduced from the course id:

`MITx+7.03.2x` -> Department number 7
`course-v1:MITxT+21A.819.2x` - Department number 21A

The full list of valid course ids is included in [this json file](https://github.com/mitodl/ocw-studio/blob/6379dc014c34dc41bfdff2fb7ce9a38cabd10983/static/js/resources/departments.json) from OCW.

Sometimes the department number won't be in the course id:
`MITx+CITE101x`

In that case, leave the field blank?

#### Designs and Mockups
The ETL `transform_course` functions for each platform are here:

[MITx Online](https://github.com/mitodl/open-discussions/blob/bd51b108b9170c7d854e57473cd3991255cf62df/course_catalog/etl/mitxonline.py#L120-L143)

[MITx, OLL](https://github.com/mitodl/open-discussions/blob/bd51b108b9170c7d854e57473cd3991255cf62df/course_catalog/etl/openedx.py#L222-L259)

Something like the following line will need to be added to each `transform_course` function:

```python
"department":  extract_department_from_id(course.get("key"))
```

#### Acceptance Criteria:

- [ ] Most MITx, OLL, and MITx Online courses will have non-null `department` values.

JenniWhitman:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested
  - [X] Changes have been manually tested

Closes #3827 

#### What's this PR do?
Adds a utility function taking the department number/identifier from the passed course id and assigning it to the `department` key in the course dict. 
The logic for this looks for the first instance of a `+` and the following `.`, takes the string between them, and compares it to the list of departments. If the string is there, that identifier is put in a list and passed back. If there is no matching string OR that matching string is not in the list of valid departments, 'None' is passed back.

I'll offer my process, but there is potentially(probably) a better way, but this let me walk through piece by piece.

1. I took the json from. `test_json.mitxonline_courses.json` and loaded it from file to a dict (basically replaced the loader function with a file open and json.loads, which returns the same result as the API call). 
2. I then ran that data through `transform_courses` which will run each of the courses in that list through the modified function `_transform_course`
3. Then ran `course_catalog.etl.loaders.load_courses` with the parameters `"mitxonline"` and the transformed data to success and verified that the new department identifiers saved.

### DIFF
--- a/course_catalog/etl/mitxonline.py
+++ b/course_catalog/etl/mitxonline.py
@@ -10,6 +10,7 @@ import requests
 from django.conf import settings
 
 from course_catalog.constants import OfferedBy, PlatformType
+from course_catalog.utils import extract_valid_department_from_id
 from course_catalog.etl.utils import transform_topics
 
 log = logging.getLogger(__name__)
@@ -130,6 +131,7 @@ def _transform_course(course):
     return {
         "course_id": course["readable_id"],
         "platform": PlatformType.mitxonline.value,
+        "department": extract_valid_department_from_id(course.get("key")),
         "title": course["title"],
         "offered_by": copy.deepcopy(OFFERED_BY),
         "topics": transform_topics(course.get("topics", [])),
--- a/course_catalog/etl/mitxonline.py
+++ b/course_catalog/etl/mitxonline.py
@@ -10,8 +10,7 @@ import requests
 from django.conf import settings
 
 from course_catalog.constants import OfferedBy, PlatformType
-from course_catalog.utils import extract_valid_department_from_id
-from course_catalog.etl.utils import transform_topics
+from course_catalog.etl.utils import transform_topics, extract_valid_department_from_id
 
 log = logging.getLogger(__name__)
 
@@ -131,7 +130,7 @@ def _transform_course(course):
     return {
         "course_id": course["readable_id"],
         "platform": PlatformType.mitxonline.value,
-        "department": extract_valid_department_from_id(course.get("key")),
+        "department": extract_valid_department_from_id(course["readable_id"]),
         "title": course["title"],
         "offered_by": copy.deepcopy(OFFERED_BY),
         "topics": transform_topics(course.get("topics", [])),
--- a/course_catalog/etl/mitxonline_test.py
+++ b/course_catalog/etl/mitxonline_test.py
@@ -11,7 +11,10 @@ import pytest
 from course_catalog.constants import PlatformType
 from course_catalog.etl import mitxonline
 from course_catalog.etl.mitxonline import _parse_datetime, parse_page_attribute
-from course_catalog.etl.utils import UCC_TOPIC_MAPPINGS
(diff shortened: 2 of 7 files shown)


mbertrand (approved):
Looks great! Ran the following and they all imported without error and populated `department` for most courses:

`manage.py backpopulate_mitxonline_data`
`manage.py backpopulate_edx_data`
`manage.py backpopulate_oll_data`

👍

pdpinch:
Is it possible to tell me which courses weren't assigned departments?

pdpinch:
I reviewed the department names and numbers. The only thing that gives me pause is `RES`. Did we get any courses that match `RES` ?

https://github.com/mitodl/open-discussions/blob/c969b81912fff836b2d8b535318c6d42aef413f4/course_catalog/constants.py#L228

JenniWhitman:
> Is it possible to tell me which courses weren't assigned departments?

I used a django queryset to grab all of them and then printed them out with their department to see if it had one (super scientific, I know, but quick for me)

JenniWhitman:
> I reviewed the department names and numbers. The only thing that gives me pause is `RES`. Did we get any courses that match `RES` ?
> 
> https://github.com/mitodl/open-discussions/blob/c969b81912fff836b2d8b535318c6d42aef413f4/course_catalog/constants.py#L228

I don't think I've tried one. I could - why does that one in particular cause worry? It was on the [json that is in OCW](https://github.com/mitodl/ocw-studio/blob/6379dc014c34dc41bfdff2fb7ce9a38cabd10983/static/js/resources/departments.json) as well.

pdpinch:
”Supplemental Resources”  is not a department. It's a catch-all for OCW things that aren't courses. On Feb 14, 2023, at 6:35 PM, Jenni ***@***.***> wrote:﻿

I reviewed the department names and numbers. The only thing that gives me pause is RES. Did we get any courses that match RES ?
https://github.com/mitodl/open-discussions/blob/c969b81912fff836b2d8b535318c6d42aef413f4/course_catalog/constants.py#L228

I don't think I've tried one. I could - why does that one in particular cause worry? It was on the json that is in OCW as well.

JenniWhitman:
ok that makes sense @pdpinch - I can easily pop it out of the list once we decide? That list pre-dates this, was just identical to the one we were using, so I reused it and won't remove it form the dict itself.

pdpinch:
@JenniWhitman if there are any `RES` courses, it's not worth blocking your PR. Let's proceed.

pdpinch (approved):

PR #3843, opened 2023-02-22
Open discussions Elastic/Opensearch performance updates

Ardiea:
Added a configuration setting for specifying the number of index replicas to create when making a new elasticsearch index.

#### Pre-Flight checklist
- [ ] Testing
  - [ ] Code is tested
- [X] Settings
  - [X] New settings are documented and present in `app.json`
  - [X] New settings have reasonable development defaults, if applicable

https://github.com/mitodl/ol-infrastructure/issues/1199

#### What's this PR do?
This PR will add a configuration setting that allows us to specify the number of index replicas when the application creates a new index in OpenSearch. Our focus is on search performance rather than ingest performance. For improved search performance we should have a replica of all the data on each node in the cluster, which will allow any node that receives a search request to service that request rather than forwarding it to another node 33% of the time (in the case of a 3 node cluster with 1 replica, the current config). 

In confusing ES terminology "1 replica" == 2 copies of the data, a primary + replica. So, "2 replicas" == 3 copies of the data, a primary and 2 replicas. 🤷 

We set the default replica count to 2, which is appropriate for a 3 node cluster. DevStack implementations will not break for this, the cluster will simply be unable to allocate the additional replicas (which is already what happens because the default setting is 1 -> 2 copies of the data but only one node). 

Additionally, we are going to set the refresh interval on new indices to 60 seconds. The default value is 1 second. This change will prevent OS/ES from doing constant "refreshes" during our reindex process that are of no value to us with our given usage pattern. We execute all of our searches against the alias for an index, rather than the explicit index name. Once our reindex process is completed and the new index is fully populated, we swap the alias over and being executing searches against the new index. A "refresh" updates an index to a usable state for search with all the new data that has been added since the last "refresh" took place. Because never execute a search against an index that is being actively updated, there is no value provided by doing this every second. 
[
More information on refreshes.](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html)

Additionally-Addtionally, we're going to add an ELASTICSEARCH_CONNECTIONS_PER_NODE setting to allow us to override the default size of the connection pool(s) created by ElasticSearch-py in urllib3. The default set by elasticsearch-py is 10, and that is our default as well, but we have an option to adjust it per environment based on performance issues observed. 

There isn't much point to testing this in a devstack. We will want to check on the CI/RC environment to confirm that new indices are created with the appropriate setting populated. It might not be possible to asses the performance impact of these changes in CI/RC though, given then have low search load. 

(Optional)

(Optional)

(Optional)

(Optional)


blarghmatey (approved):
Great work!

pdpinch:
Should we plan to do some load tests?

Ardiea:
> Do you think this will address errors like 

Yes the hope is to help out in circumstances where errors like those are generated. Either from indexing impacts or external search spam. 

> Should we plan to do some load tests?

We can I don't anticipate performance degrading from any of this though.

Ardiea:
@pdpinch I have verified this change in the CI and there are no performance concerns. 

`ELASTICSEARCH_CONNECTIONS_PER_NODE` is just explicitly setting the default value and now gives us the option to tweak as we see fit. 

`ELASTICSEARCH_REPLICA_COUNT` increases the default replica count from 1 to 2, which introduces a negligible increase in indexing latency. Replica updates/inserts are all done concurrently, so there isn't much difference between having 1 and having 2 of them. A one-time expense at indexing latency to gain a continuous reward in search latency, which is more important in this use case. 

I do have some thoughts about how to further improve performance and reduce situations where searches fail for users and 5XX are registered in sentry. 

1. Introduce a configurable random wait between indexing batches. Celery is basically doing everything as quickly as possible right now. If we just slow it down a bit, we will not tax the ingestion capacity of the cluster so much. It could be very small, 50-100ms between could have a huge impact and it is a background task without many expectations about performance at the moment. 
2. This one is more involved, but basically rate limiting the api search endpoint for users. I'd have to go look for an example, but from time-to-time there are bots of some sort that will hit OCW search continuously with hundreds of searches per minute. I don't think that is necessarily something we should prohibit but we also shouldn't let it impact the availablity of the application to other users. I will try to collect some better data around these instances and share it with you this week if I can.

PR #3849, opened 2023-02-27
Add stylelint for scss linting

### LINKED ISSUE #3848: Add SCSS Linting
As @rhysyngsun noticed in #3835, we never set up SCSS linting for infinite corridor or any of the new (non-`open-discussions`) the packages under `frontends/`.

Most of our projects use `sass-lint`. However, this is not suitable tool for infinite corridor:
-  It has been long unmaintained, see [github readme](https://github.com/sasstools/sass-lint) Its last publish was 2019.
- It does not support Sass's module system, which was released in late 2019 and which we have been using in Infinite Corridor. *Sass modules, `@use` has some significant advantages over older `@import`; see for more https://css-tricks.com/introducing-sass-modules/*

### Acceptance Criteria
- [ ] Set up linting for infinite-corridor and other local packages, using some tool that is maintained and supports modern (Dart-Sass) syntax

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #3848 

#### What's this PR do?
This PR adds linting for `.scss` files in "new" open. Previous `.scss`-linting was only happening in `frontends/open-discussions`. Now it happens in all the `frontends/` packages:
- `frontends/open-discussions` continues to use the [`sass-lint`](https://github.com/sasstools/sass-lint) for scss linting
- All other packages in `frontends/` are using [`stylelint`](https://www.npmjs.com/package/stylelint)

The reason to use a new tool in the newer packages are primarily:
- [`sass-lint`](https://github.com/sasstools/sass-lint) is no longer maintained and hasn't been updated in 4 years
- **and in particular**, `sass-lint` does not support Sass module syntax (`@use`), which we have been using in infinite-corridor

On this branch:
1. Enter `watch` container terminal via something like `docker compose exec watch bash`. (Or prefix commands below with `docker compose run --rm watch <cmd>`)
2. Run `yarn scss-lint`. You should see each frontend package (the directories under `frontends/`) be linted. Linting should pass.
3. Introduce a linting error in any of the frontend packages other than `open-discussions`. E.g., add some extra newlines in `frontends/infinite-corridor/src/scss/learning-resources.scss`, or change a variable name from kebab-case to camelCase. Run `yarn scss-lint` again. It should fail with a helpful error message.
4. (Optional) If you use VS Code, try the [stylehint extension](https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint). You'll need to add
    ```json
    {
            "stylelint.validate": ["css", "scss"]
    }
    ```
    to your `settings.json`. If you do, you should be able to see scss linting errors in the editing.


mbertrand (commented):
Looks good, one question, why not use `stylelint` on the original open-discussions frontend too instead of retaining sass-lint just for that?

ChristopherChudzicki:
> Looks good, one question, why not use stylelint on the original open-discussions frontend too instead of retaining sass-lint just for that?

We could, but `stylelint`'s default preferences are a bit different than sass-lint, so we'd have to either
1. resolve/suppress the errors
2. or configure stylelint so that its preferences are similar to sass-lint's

My understanding of the plan for "New Open" is that we'll generally not develop in `open-discussions`, and instead move functionality to the typescript packages instead. In light of that, leaving `open-discussions` as-is seems reasonable to me.

mbertrand (approved):
👍

Issue #3842, opened 2023-02-17
Learn search results list dates are different than default drawer details dates


mbertrand:
### Steps to Reproduce

Search for xpro courses at https://open.mit.edu/learn/search

### Expected Behavior

The date shown on each search result card will match the default date shown in the details drawer for that course.

### Actual Behavior

The dates differ.

### Screenshot or Screencast

![Screenshot 2023-02-17 at 12 25 06 PM](https://user-images.githubusercontent.com/187676/219728472-1b2c595c-7258-42c1-9e72-9c604708be58.png)

PR #3804, opened 2023-01-09
Add all prolearn imports to professional offerings, handle multiple runs

mbertrand:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes #3795 
Closes #3796

#### What's this PR do?
- Adds all courses/programs imported via the prolearn API to "Professional Offerings"
- If a course/programs has multiple start/end dates, create a run for each.

#### How should this be manually tested?
- Set `PROLEARN_CATALOG_API_URL=https://prolearn.mit.edu/graphql` in your .env file
- If you've already imported these before, delete them:
    ```
     manage.py backpopulate_prolearn_data --delete
    ```
- Otherwise or afterward, run the above without `--delete` to import them.
- Go to `../learn/search` and check the `Professional Offerings` checkbox.  Under "Offered By" you should see the following, plus XPro if you've imported anything from there:

Screen Shot 2023-01-09 at 10 05 25 AM

- Search for "accelerating digital", click on the "Accelerating Digital Transformation with Algorithmic Business Thinking" course to bring up the drawer.  At the top right there should be a dropdown of run start dates.

Screen Shot 2023-01-09 at 10 09 05 AM

PR #3850, opened 2023-03-01
Learn search results list dates are different than default drawer details dates

JenniWhitman:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes [3842](https://github.com/mitodl/open-discussions/issues/3842)

#### What's this PR do?
Swaps both the ol-search-ui and open-discussions front end repos off using best_start_date and onto start_date as best_start_date is often enrollment date and a mismatch... and misnomer.

search for courses in open-discussions/learn/search and compare against the database to find preferably courses with mismatched best_start_date and start_date - I would recommend:

- 1 that is currently running with past dates
- 1 that is currently running with future dates
- 1 that is currently running only
- 1 that ran in the past without future dates
- 1 that ran in the past but has future dates
- 1 that runs in the future only

The selected "best run" should be the one expected which would be:

- Currently running (today's date is between the start and end dates)
- the next run (start date is after today's date, there is no current run)
- the last run that ran if neither of the previous exist.

The date on the search results should match the date in the drawer that opens when you click on the course
The dates in the date drop down (if there are multiple runs) should be in order. This can be hard to tell as sometimes there are wacky end dates. To fix this, I modified line 207 of expandedLearningResourceDisplay.js to `{getStartDate(object, run)} - {run.end_date}` so I could see this while testing.

Start with the js changes as those are what is on learn/search and what is referenced in the ticket.

You know that dog flying hte plane meme? That's me. I'm not certain that best_start_date doesn't have a purpose, only that that date causes the mismatch and because of inconsistencies with saved data vs what we display elsewhere. I also made the changes to both the js and ts versions since this would be incorrect in search-ui going forward if change isn't made there as well, although it should work the same in both.

### DIFF
--- a/frontends/ol-search-ui/src/util.tsx
+++ b/frontends/ol-search-ui/src/util.tsx
@@ -62,10 +62,10 @@ const resourceThumbnailSrc = (
 export const DATE_FORMAT = "YYYY-MM-DD[T]HH:mm:ss[Z]"
 
 const runStartDate = (objectRun: LearningResourceRun): moment.Moment =>
-  moment(objectRun.best_start_date, DATE_FORMAT)
+  moment(objectRun.start_date, DATE_FORMAT)
 
 const runEndDate = (objectRun: LearningResourceRun): moment.Moment =>
-  moment(objectRun.best_end_date, DATE_FORMAT)
+  moment(objectRun.end_date, DATE_FORMAT)
 
 const compareRuns = (
   firstRun: LearningResourceRun,
@@ -75,7 +75,7 @@ const compareRuns = (
 const findBestRun = (
   runs: LearningResourceRun[]
 ): LearningResourceRun | undefined => {
-  const dated = runs.filter(run => run.best_start_date && run.best_end_date)
+  const dated = runs.filter(run => run.start_date && run.end_date)
 
   // Runs that are running right now
   const [bestCurrentRun] = dated.filter(
--- a/frontends/open-discussions/src/components/LearningResourceCard.js
+++ b/frontends/open-discussions/src/components/LearningResourceCard.js
@@ -10,13 +10,12 @@ import LoginTooltip from "./LoginTooltip"
 import LearningResourceIcon from "./LearningResourceIcon"
 
 import { setDialogData } from "../actions/ui"
-import { bestRun } from "../lib/learning_resources"
+import { bestRun, getStartDate } from "../lib/learning_resources"
 import { defaultResourceImageURL, embedlyThumbnail } from "../lib/url"
 import {
   CAROUSEL_IMG_WIDTH,
   CAROUSEL_IMG_HEIGHT,
   LR_TYPE_VIDEO,
-  DISPLAY_DATE_FORMAT,
   readableLearningResources
 } from "../lib/constants"
 import { SEARCH_GRID_UI, SEARCH_LIST_UI } from "../lib/search"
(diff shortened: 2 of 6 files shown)


ChristopherChudzicki:
@JenniWhitman Re `best_start_date` I have found this confusing in the past. Matt clarified for me:

> best_start_date is specific to each run, and was meant to handle cases from other sources where there were multiple start dates, some of which might be blank, for example:
> 
> ```
> course_run.best_start_date = (
>     course_run.enrollment_start
>     or course_run.start_date
>     or semester_year_to_date(course_run.semester, course_run.year)
> )
> ```
> 
> But for prolearn there’s only 1 start date, so that’s what best_start_date should be too.



https://github.com/mitodl/open-discussions/pull/3804#discussion_r1068540384

JenniWhitman:
> @JenniWhitman Re `best_start_date` I have found this confusing in the past. Matt clarified for me:
> 
> > best_start_date is specific to each run, and was meant to handle cases from other sources where there were multiple start dates, some of which might be blank, for example:
> > ```
> > course_run.best_start_date = (
> >     course_run.enrollment_start
> >     or course_run.start_date
> >     or semester_year_to_date(course_run.semester, course_run.year)
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But for prolearn there’s only 1 start date, so that’s what best_start_date should be too.
> 
> [#3804 (comment)](https://github.com/mitodl/open-discussions/pull/3804#discussion_r1068540384)

The problem is, the code actually prefers enrollment start date to best start date - I COULD make it so it grabs another if it's not there, but quite often I'm seeing it use enrollment date even if there is a start date which is the reason for the discrepency.

frontends/ol-search-ui/src/util.tsx
```diff
@@ -75,7 +75,7 @@ const compareRuns = (
 const findBestRun = (
   runs: LearningResourceRun[]
 ): LearningResourceRun | undefined => {
-  const dated = runs.filter(run => run.best_start_date && run.best_end_date)
+  const dated = runs.filter(run => run.start_date && run.end_date)
```
mbertrand, line 78: I think doing this instead would make it more consistent with `ExpandedLearningResourceDisplay.tsx`:

```javascript
  const startDate =
    hasCertificate && bestAvailableRun ?
      getStartDate(resource.platform ?? "", bestAvailableRun).format(DISPLAY_DATE_FORMAT) :
      null
```

frontends/open-discussions/src/components/LearningResourceCard.js
```diff
@@ -148,7 +148,7 @@ export function LearningResourceDisplay(props: Props) {
   const hasCertificate = object.certification && object.certification.length > 0
   const startDate =
     hasCertificate && bestAvailableRun
-      ? moment(bestAvailableRun.best_start_date).format(DISPLAY_DATE_FORMAT)
+      ? moment(bestAvailableRun.start_date).format(DISPLAY_DATE_FORMAT)
```
mbertrand, line 151: Similar to above:

```javascript
  const startDate =
    hasCertificate && bestAvailableRun
      ? getStartDate(object, bestAvailableRun)
      : null
```

mbertrand (changes requested):

frontends/ol-search-ui/src/util.tsx
JenniWhitman: you're not wrong - that better solves the "and what if there's none" problem

mbertrand (approved):
Looks good! 👍

PR #3851, opened 2023-03-02
Remove yarn postinstall script

ChristopherChudzicki:
#### Pre-Flight checklist

- [ ] Testing *Needs to be tested on Heroku*.

Should help with #3691 

#### What's this PR do?

Post install is a script yarn runs automatically after install. I believe we do not need this and it is contributing to our overbuilding.
- We don't need it in dev because run-watch-dev.sh runs build.
- I believe we don't need it on Heroku because the buildpack runs "build" directly if found. (https://github.com/heroku/heroku-buildpack-nodejs/blob/5917461b195feea3b67571f39bfdcdf2c0e5c25d/lib/dependencies.sh#L108)

This really needs to be tested on Heroku. But you could check locally that:

1. `docker compose up` builds the app.

On `docker compose up`, we build too often:
1. Once after `yarn install` if there have been changes since your last install
2. Twice with `run-watch-dev.sh`:
    1. Once at the start,
    2. then again when entering watch mode

This should eliminate (1). 2i is also undesirable, but not addressed here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,6 +127,9 @@ jobs:
       - name: Install dependencies
         run: yarn install --immutable
 
+      - name: Webpack build
+        run: yarn run build
+
       - name: Lints
         run: yarn run lint-check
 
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
Works fine locally, 👍

PR #3861, opened 2023-03-08
More efficient download and processing of edx content archives

### LINKED ISSUE #3860: EdX contentFile import tasks cause excessive memory consumption on heroku
### Steps to Reproduce

The reguarly scheduled tasks for importing content files from MITx, MITx Online, and xPro cause memory usage to increase to the point where heroku restarts the dynos, causing the tasks to silently fail.

### Expected Behavior

Run any of the following:
backpopulate_edx_files
backpopulate_xpro_files
backpopulate_mitxonline_files

Each should start a task that eventually completes.

### Actual Behavior

Eventually the task will silently fail when heroku kills the dynos it is running on due to memory exceeding limits.

### Screenshot or Screencast

Screenshot 2023-03-08 at 8 31 39 AM

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #3860

#### What's this PR do?
- Uses boto's `Bucket.download_file()` function to save edx archives to a temporary file, instead of reading the entire file into memory before saving to disk.
- Uses `yield` to return file content one by one instead of all at once in a single list
- Fixes the omission of mitx and mitxonline from file content index recreation
- Places a try-catch block around the parsing of "verticals" content via `XBundle`.  A handful of mitxonline tar archives raise an exception here due to "external entities" in an html file, that was preventing the rest of the archive from being processed.

- Set these to the same as production in your .env file:
```
MITX_ONLINE_BASE_URL
MITX_ONLINE_COURSES_API_URL
MITX_ONLINE_PROGRAMS_API_URL
MITX_ONLINE_LEARNING_COURSE_BUCKET_NAME
```

- Delete any existing content files for mitxonline you might have, in a shell:
```python
ContentFile.objects.filter(run__platform="mitxonline").delete()
```

- Run the following
```python
manage.py recreate_index --courses
```

- Go to http://localhost:8063/learn/search?q=%22tau%20prime%20and%20this%20curly%20tau%22 - there should be 1 course result

- Run `manage.py recreate_index --courses` again.  The above url should still return 1 result afterward.

### DIFF
--- a/search/tasks.py
+++ b/search/tasks.py
@@ -764,6 +764,8 @@ def start_recreate_index(self, indexes):
                             platform__in=(
                                 PlatformType.ocw.value,
                                 PlatformType.xpro.value,
+                                PlatformType.mitx.value,
+                                PlatformType.mitxonline.value,
                             )
                         )
                         .exclude(course_id__in=blocklisted_ids)
--- a/course_catalog/etl/edx_shared.py
+++ b/course_catalog/etl/edx_shared.py
@@ -12,7 +12,6 @@ from course_catalog.constants import PlatformType
 from course_catalog.etl.loaders import load_content_files
 from course_catalog.etl.utils import get_learning_course_bucket, transform_content_files
 from course_catalog.models import Course, LearningResourceRun
-from course_catalog.utils import get_s3_object_and_read
 
 log = logging.getLogger()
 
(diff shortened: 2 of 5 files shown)


search/tasks.py
```diff
@@ -764,6 +764,8 @@ def start_recreate_index(self, indexes):
                             platform__in=(
                                 PlatformType.ocw.value,
                                 PlatformType.xpro.value,
+                                PlatformType.mitx.value,
+                                PlatformType.mitxonline.value,
```
JenniWhitman, line 768: For my own knowledge - do we have/should we have plans to add any other platforms here? Not sure what resources the others have, but I noticed, for example, mm isn't here.  Just curious.

JenniWhitman (approved):
Looks good and behaves well locally (better than it did before even). I left one question for my own knowledge, but try as I might, couldn't find a reason not to say LGTM👍

search/tasks.py
mbertrand: Eventually, "Open Learning Library" courses (platform type "oll") will need to be added, once its content archives are available.  These are all edx-based platforms.

Issue #3852, opened 2023-03-02
Stop/Undo sharing code between new & old open


ChristopherChudzicki:
tldr: Initially we wanted to "share" code between the old/new frontend sections. Subsequently we decided to **move** features from old to new. Some code is stilled shared, and this makes changing things harder. Let's stop sharing.

---

The frontend portion of the MIT Open codebase is split between "new" Typescript and "old" flow files. The new TS files are organized as several local node packages.

When we began working on new features for Open, our initial plan was to migrate code as needed from the old flow js files to separate local packages and new TS version in both "old" and "new" open. An example of this is the Searchbar component: It is written in Typescript in `ol-search-ui` package; the compiled JS is used to provide the searchbar both at https://discussions-rc.odl.mit.edu/learn/search and https://discussions-rc.odl.mit.edu/infinite/search. 

Sharing code in this way between new features and old features proved difficult[^1]. We decided to stop sharing: new TS code should only be used in "new" open, and we will gradually move code from the JS to TS codebases. 

[^1]: Example: old codebase uses a fair amount of redux, largely for storing api requests, but also for some other things. New codebase does not. Writing a shared component that worked well in both codebases would be hard. More generally, a stricter separation allows us to make separate decisions about library-usage in the new codebase.

We should remove the old shared code. Two benefits:
1. We can iterate on new open without worrying about breaking old open features
2. We will no longer need to compile individual local packages from TS to JS, simplifying development.

pdpinch:
What's the scope of this? Can it be broken down into smaller chunks?

ChristopherChudzicki:
@pdpinch The shared components are:
1. The searchbox from `ol-search-ui` package
2. ~7 simple utility layout components from `ol-util` (e.g, Card, the "hero area" banner of pages)
3. A simple component for validation errors from `ol-forms`

I would expect the changes to maybe 2 days overall. I think doing each of 1, 2, 3 as a separate PR makes sense for easier reviews.

A followup chunk of work would be "Stop compiling individual packages from TS to JS".

pdpinch:
Can you say more about the follow-up chunk? How big of a lift is that, and what are the benefits?

ChristopherChudzicki:
[...]

PR #3862, opened 2023-03-08
Stop sharing new code with old open

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#3852 

#### What's this PR do?
This PR finishes removing the new TS-based packages from the old codebase. In particular, this
- adds a JS copy of SearchInput and Searchbox components to `frontends/open-discussions`

There should be no visible changes, but it is worth checking that:
1. search on `/learn` and `/learn/search` still work as expected.


ChristopherChudzicki:
The test that failed is a difference in formatting between (`MMMM DD, YYYY` vs `MMMM D, YYYY`) on the learning resource cards that only matters when day is in 1-9. I'm a little confused why we haven't run into this more often, though. Maybe because start date wasn't showing very often until #3842 was resolved

abeglova (approved):
LGTM

Issue #3863, opened 2023-03-09
Flaky frontend test


ChristopherChudzicki:
Sometimes an old `/learn` test fails ([example](https://github.com/mitodl/open-discussions/actions/runs/4375294331/jobs/7655831329#step:13:3016)).

> AssertionError: expected 'September 06, 2000' to equal 'September 6, 2000'

But only when the random test data has a date with day 1-9 and only when the `start_date` actually shows on resource cards. (I suspect the start date rarely showed prior to https://github.com/mitodl/open-discussions/pull/3850, which is why we hadn't noticed this much.)

JenniWhitman:
@ChristopherChudzicki is this failing the same test: https://github.com/mitodl/open-discussions/actions/runs/4747252982/attempts/1

ChristopherChudzicki:
@JenniWhitman Good reminder though, I meant to close this after https://github.com/mitodl/open-discussions/pull/3869

PR #3869, opened 2023-03-09
fix flaky test: use standard date format for cards

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#3863 

#### What's this PR do?
~This PR changes the date format on learning resource cards to the same format we use most places—`MMMM D, YYYY`. Previously, the format had `DD` so single-digit days would display as `04`, etc. See screenshots.~

This PR updates the `DISPLAY_DATE_FORMAT` constant. **We were previously using this constant only in tests.** Now we use it also as the format for generating dates. 

View a learning resource card on `/learn` that has a date. It should look the same as it did before.

### DIFF
--- a/frontends/open-discussions/src/lib/constants.js
+++ b/frontends/open-discussions/src/lib/constants.js
@@ -128,7 +128,7 @@ export const readableLearningResources = {
 }
 
 export const DATE_FORMAT = "YYYY-MM-DD[T]HH:mm:ss[Z]"
-export const DISPLAY_DATE_FORMAT = "MMMM D, YYYY"
+export const DISPLAY_DATE_FORMAT = "MMMM DD, YYYY"
 
 export const PHONE = "PHONE"
 export const TABLET = "TABLET"
--- a/frontends/open-discussions/src/lib/learning_resources.js
+++ b/frontends/open-discussions/src/lib/learning_resources.js
@@ -12,7 +12,8 @@ import {
   LR_TYPE_LEARNINGPATH,
   LR_TYPE_PROGRAM,
   LR_TYPE_COURSE,
-  platforms
+  platforms,
+  DISPLAY_DATE_FORMAT
 } from "./constants"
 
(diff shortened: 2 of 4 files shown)


frontends/open-discussions/src/lib/learning_resources.js
```diff
@@ -116,9 +117,9 @@ export const getStartDate = (
   if (object.platform === platforms.OCW) {
     return `${capitalize(objectRun.semester || "")} ${objectRun.year || ""}`
   } else if (objectRun.start_date) {
-    return moment(objectRun.start_date).format("MMMM DD, YYYY")
+    return moment(objectRun.start_date).format(DISPLAY_DATE_FORMAT)
```
ChristopherChudzicki, line 120: This is the main change in this PR... `DISPLAY_DATE_FORMAT = "MMMM D, YYYY"`. (D, not DD)

mbertrand (approved):
👍 assuming Peter & Ferdi are oikay with the date format change

frontends/open-discussions/src/lib/constants.js
```diff
@@ -128,7 +128,7 @@ export const readableLearningResources = {
 }
 
 export const DATE_FORMAT = "YYYY-MM-DD[T]HH:mm:ss[Z]"
-export const DISPLAY_DATE_FORMAT = "MMMM D, YYYY"
+export const DISPLAY_DATE_FORMAT = "MMMM DD, YYYY"
```
ChristopherChudzicki, line 131: Prior to this PR, this constant was only used in tests. Changing it does not affect runtime behavior.

ChristopherChudzicki:
> 👍 assuming Peter & Ferdi are oikay with the date format change

For posterity: this PR now leaves the format unchanged, just updating the test.

mbertrand (approved):
👍

PR #3874, opened 2023-03-10
stop building individual ts packages

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#3691 

#### What's this PR do?
This PR changes how we build Open Discussions.
- **Previously:** We had several local typescript packages (`ol-widgets`, etc) that were used by the apps (`infinite-corridor` aka "new open" and `open-discussions` aka "old open"). These individual TS packages were built from TS to JS for consumption in old open. This had some downsides:
    - We had been running 1 TS watch process per package. Doesn't seem great.
    - It complicated the webpack build... We needed to ensure that all TS packages were build prior to building webpack. This is fairly simple when actually building (just pass `--topological` to yarn's foreach command). But it was trickier for dev mode, where the webpack dev servers would try to build, but the inidviudal TS builds might not have completed yet. (Or one `ol-widgets` might try to build itself before its dependency, `ol-util`, was done building). We had previously "solved" this building **before** entering dev mode. But that's extra work.  
  
- **Now:** We do not build Typescript directly: `ol-widgets`, etc, stay as typescript files. The consuming app (just `infinite-corridor`) compiles the typescript to JS as part of the webpack build. _This has become simpler after #3862 _

1. Run `docker compose up` and separately watch the `watch` logs: `docker compose logs --follow --no-log-prefix  --tail 1000 watch`. You should see `open-discussions` and `infinite-corridor` enter dev mode, and they should be done building fairly quickly (For me, about 10s for infinite and 30s for open).
2. Make a change to one of the TS packages (e.g., add a `console.log("Hi")` to some file. Webpack should rebuild itself. should build.


frontends/ol-widgets/src/index.ts
```diff
@@ -1,2 +1,14 @@
...
+ * A TS project consuming this TS source code will try to typecheck this
+ * package. Since CKEditor doesn't have great typing, it will fail unless it has
+ * the ambients declarations in ./types/ckeditor.d.ts (or declares its own
+ * CKEditor types).
+ *
+ * Eslint wants this to be an import, but there is no type-only version of a
+ * sideffect import. So use old /// ts syntax. See https://github.com/microsoft/TypeScript/issues/36812
+ */
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference
+/// <reference path="./types/ckeditor.d.ts" />
```
ChristopherChudzicki, line 11: This is annoying, and maybe the one downside of not compiling TS to JS for each package. But still very much worth it overall.

_Also, CKEditor types are getting **much** better. They have re-written a lot of the core functionality in TS, so we're starting to get builtin typings for it anyway._

rhysyngsun (approved):
LGTM

Issue #3868, opened 2023-03-09
API updates for new open


JenniWhitman:
As a developer, I'd like to be able to search using API with new filters (split out from #3847 ):

- upcoming courses
- courses with certificate
- micromasters courses
- professional ed courses

#### Designs and Mockups

Updating the current course_catalog api to include those filters makes the most sense for a forward facing API & consistent calls that are going to be made often and without needing the power of the /search api to pull results (aka infinite). Adding the filters through django-filters which should translate nicely into openAPI spec.

Will have 4 new available filters:
- **Certificate** - which uses the same logic from the certificates property, to filter down by courses offering certificates (and those that are not). My plan here is for this to be a boolean field, question in comments for discussion.
- **Upcoming** - Currently this is handled in a separate view which overwrites any filtering because of how it calls the queryset - meaning it can not be combined with other filters (such as the ones we are adding). It is also a different format than the other filters and will not make as much sense in documentation unless added to the filterset with the rest. I'm proposing we add it as a keyword argument here, and leave the other views out of any documentation or public facing API as they're both confusing and not all filters (new, for example, is a sort). This just means adding it to the filterset using the existing logic.
- **Audience** - filter courses by whether they are professional or open, matching the[ current model property](https://github.com/mitodl/open-discussions/blob/72cbb98c8dfb0d7cec98004d4b2ad382414c5307/course_catalog/models.py#L312)
- **offered_by** - filter courses by who/what they are offered by

#### Acceptance Criteria:

- [ ] filterable by upcoming courses (can easily also translate this into past and ongoing filters as well)
- [ ] filterable by courses with certificate
- [x] filterable by micromasters courses
- [x] filterable by professional ed courses

#### Out of Scope

- Not going to add the docs aspect now (unless we want a demo, it's only a few lines to generate a swagger or redoc example quickly)
- Not going to add other filters for now

JenniWhitman:
- micromasters and profesional ed are being solved by #3877 
- upcoming (also new and featured) and certificate will be in the next PR

JenniWhitman:
@Ferdi @pdpinch  there's a couple points of discussion clarification here - let me know if i can make them more clear, but tried to italicize the actual questions/plan themselves. 

[...]

PR #3877, opened 2023-03-14
2 New API filters for new open - offered_by and audience

JenniWhitman:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

updates https://github.com/mitodl/open-discussions/issues/3868 

- one more PR will work out the last 2 filters:  certificates and upcoming/new/featured

#### What's this PR do?
This adds a filterset to courses, as a start, which will allow the use of django-filters on the `/api/v0/course/` endpoint using kwargs for:

- `offered_by` (to satisfy the need to filter by "micromasters" courses as in the original issue) - the available values of which can be found in the constants [here](https://github.com/mitodl/open-discussions/blob/bc117402290d48930b70fbadf0068b966f8d85dd/course_catalog/constants.py#L7).
- audience (to satisfy the need to filter by professional courses) - this adds a filter method to filter down platform by professional or open offerings which align with the `PROFESSIONAL_COURSE_PLATFORMS` found in our [models](https://github.com/mitodl/open-discussions/blob/bc117402290d48930b70fbadf0068b966f8d85dd/course_catalog/models.py#L28), which is how the property is calculated.

Using the backpopulate management commands, populate data for at least one professional platform and one open platform (I have xpro and mitxonline locally, but you can choose your own -more is better in this case as it will show better filtering). Suggested commands are here:
```
manage.py recreate_index --courses
manage.py backpopulate_xpro_data
manage.py backpopulate_xpro_files
```

With those in place, can use your browser to go to the following URLs and test that you have the correct data:
`http://od.odl.local:8063/api/v0/courses/` - Should show ALL courses imported
`http://od.odl.local:8063/api/v0/courses/?audience=professional` - Should show ONLY xpro courses and no mitxonline courses
`http://od.odl.local:8063/api/v0/courses/?platform=MITx` - should show ONLY mitx courses, and no xpro courses

You can also use the filters button in the web interface to change around filters.

(Optional)

To be clear - the issue has 4 filters listed. This PR only covers 2 and does not cover the combination of those filters. Mostly this one is geared towards the simpler filters + a new method of using filters in our API - to allow for a little discussion/review around that before applying the two larger filters.

### DIFF
--- a/course_catalog/views.py
+++ b/course_catalog/views.py
@@ -198,7 +198,6 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet, FavoriteViewMixin):
         """
         Get upcoming courses
         """
-        kwargs = request.body
         page = self.paginate_queryset(
             self._get_base_queryset(runs__start_date__gt=timezone.now()).order_by(
                 "runs__start_date"
--- a/course_catalog/views.py
+++ b/course_catalog/views.py
@@ -25,6 +25,7 @@ from authentication.decorators import blocked_ip_exempt
 from course_catalog.constants import PlatformType, PrivacyLevel, ResourceType
 from course_catalog.etl.podcast import generate_aggregate_podcast_rss
 from course_catalog.exceptions import WebhookException
+from course_catalog.filters import CourseFilter
 from course_catalog.models import (
     Course,
     CourseTopic,
@@ -145,6 +146,8 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet, FavoriteViewMixin):
     serializer_class = CourseSerializer
     pagination_class = DefaultPagination
     permission_classes = (AnonymousAccessReadonlyPermission,)
+    filter_backends = [CourseFilter]
+    filterset_fields = ["upcoming", "platform", "program_type", "certificate"]
 
     def _get_base_queryset(self, *args, **kwargs):
         """Return the base queryset for all actions"""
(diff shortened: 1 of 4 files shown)


course_catalog/views.py
```diff
@@ -194,6 +198,7 @@ def upcoming(self, request):
         """
         Get upcoming courses
         """
+        kwargs = request.body
```
abeglova, line 201: I don't think this line is needed

abeglova (changes requested):
Doesn't look like the filters work when combined with the existing /new and /featured filters. Otherwise, this looks great

abeglova (approved):
LGTM

abeglova:
Oh yeah - I should have caught that sorry. I missed that the platform filter was to get micromasters. That information is in the "offered_by" field

abeglova:
Yeah it's confusing that micromasters is in the PlatformType array in
https://github.com/mitodl/open-discussions/blob/master/course_catalog/constants.py#L28/. I don't think anything actually has micromasters as a platform

JenniWhitman:
TBH looks like it's just for that mapping?

abeglova:
Yeah - we probably don't use it much but conceptually platform is useful information about a course - the site on which the course is offered vs the organization at mit that is offering it

JenniWhitman:
The code is up for re-review and the description has been updated.

PR #3882, opened 2023-03-16
new open API part 2: withcertificate

JenniWhitman:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

updates https://github.com/mitodl/open-discussions/issues/3868

#### What's this PR do?
This PR adds  a new filter to the filterset for courses: withcertificate

The certificated filter is based on the [logic from the certification property](https://github.com/mitodl/open-discussions/blob/master/course_catalog/models.py#L320).

populate data using the backpopulate management commands. I would suggest mitx, a professional, and an open platform at the very least to test all three branches

```manage.py recreate_index --courses
manage.py backpopulate_xpro_data
manage.py backpopulate_xpro_files
manage.py backpopulate_ocw_data
manage.py backpopulate_ocw_files
```

Once those complete, you can use your method of choice to hit the following URLS:
`http://od.odl.local:8063/api/v0/courses/?withcertificate =True` - should show courses that show "Certificates" in the certification field
`http://od.odl.local:8063/api/v0/courses/?withcertificate =False` - should not have any courses that show "Certificates" in the certification field
`http://od.odl.local:8063/api/v0/courses/` - should show all courses

(Optional)

There's a slight addition to another test to cover a missing case


course_catalog/filters.py
```diff
@@ -39,3 +42,30 @@ def filter_audience(self, queryset, _, value):
...
+            )
+            qs2 = queryset.filter(
+                runs__availability__in=[
+                    AvailabilityType.current.value,
+                    AvailabilityType.upcoming.value,
+                    AvailabilityType.starting_soon.value,
+                ],
+                platform__in=PROFESSIONAL_COURSE_PLATFORMS,
+            )
+            union_queryset = qs1.union(qs2)
```
abeglova, line 67: maybe has_certificate_queryset?

course_catalog/filters.py
```diff
@@ -39,3 +42,30 @@ def filter_audience(self, queryset, _, value):
...
+        else:
+            qs1 = queryset.filter(
+                runs__availability__in=[
+                    AvailabilityType.current.value,
+                    AvailabilityType.upcoming.value,
+                    AvailabilityType.starting_soon.value,
+                ],
+                platform=PlatformType.mitx.value,
+            )
+            qs2 = queryset.filter(
```
abeglova, line 59: This logic longer matches the existing logic for determining the certificate property here
https://github.com/mitodl/open-discussions/blob/master/course_catalog/models.py#L320. We should either update that logic to match or change the query here. I think the original makes sense since archived mitx courses are available to audit but professional courses are not. Did you discuss this change with Peter?

course_catalog/filters.py
```diff
@@ -39,3 +42,30 @@ def filter_audience(self, queryset, _, value):
...
+                runs__availability__in=[
+                    AvailabilityType.current.value,
+                    AvailabilityType.upcoming.value,
+                    AvailabilityType.starting_soon.value,
+                ],
+                platform__in=PROFESSIONAL_COURSE_PLATFORMS,
```
abeglova, line 65: If we are making the proposed change to show only professional courses with non-archived availability, you don't need q1 and q2 you can just have one query with  `platform__in=PROFESSIONAL_COURSE_PLATFORMS.append(PlatformType.mitx.value)`

abeglova (changes requested):

course_catalog/filters.py
JenniWhitman: That wasn't the intent, I think I just muddled logic trying to make it match that logic which is - if it's in one or the other, and not archived anywhere - I believe i updated it correctly now

course_catalog/filters.py
JenniWhitman: good call

course_catalog/filters.py
JenniWhitman: yeah, unintentional, BUT I was able to drop it into a Q or I think and simplify it a ton

JenniWhitman:
After yesterday's conversation, this is looking to be changing to a field in the database instead, rather than a property as it is currently

JenniWhitman:
Holding off on merging/finalizing this until we resolve the discussion around https://github.com/mitodl/open-discussions/issues/3894

PR #3891, opened 2023-03-22
Revert "Set concurrency=4 (2x CPU) for the extra_worker_performance d…

mbertrand:
This reverts the modified concurrency value, which apparently is ignored if it's greater than the # of CPU's.

- [x] Testing
  - [x] Code is tested

Related to #3889 , #3860 

#### What's this PR do?
Reverts the previous commit

Should not see any change on RC


JenniWhitman (approved):
approved to go back to the way it was :)

Issue #3870, opened 2023-03-09
Make staff lists searchable, disable search/indexing of user lists


mbertrand:
Related to https://github.com/mitodl/open-discussions/issues/3846

Staff lists should be indexed and searchable.  
User lists should not be (but currently are).

#### Designs and Mockups
  - Add an index named "stafflist" to elasticsearch
  - Write code to add/updates/delete staff list documents in the new index
  - Change the current "Learning lists" facet type from "userlists" to "stafflists" (or "learninglists") - there will be backend and frontend changes involved for this- so that it only searches the new stafflist index.  Rename title to "Staff Lists" or leave as is?
  - Remove any existing userlists from the search index and disable indexing for userlists via a feature flag.

PR #3893, opened 2023-03-22
Search index for staff lists

mbertrand:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

Closes https://github.com/mitodl/open-discussions/issues/3870

#### What's this PR do?
Add public staff lists to the search index
Hides user list search behind a feature flag

Create a mix of public and private staff lists with at least 1 item each, as described in this PR.
With your browser's network tab open, go to http://od.odl.local:8063/learn/search?type=stafflist
Find the call in the network tab to http://localhost:8063/learn/search?type=stafflist, and check the response.  Your public staff lists should be included in the `hits` section of the JSON.

Search for user lists (create some first if necessary), you should get 0 results back: ttp://localhost:8063/learn/search?type=userlist
Set `FEATURE_USER_LIST_SEARCH=True` in your .env file, restart containers, then search again.  Now you should see your userlists.

Run `manage.py recreate_index --stafflists`, then repeat the search for staff lists above, the same results should be returned.


search/serializers_bulk_test.py
```diff
@@ -0,0 +1,333 @@
+"""Tests for elasticsearch bulk serializers"""
```
mbertrand, line 1: Most of these are pre-existing tests moved over to a new file because the original was getting too long (>1000 lines)

course_catalog/serializers.py
```diff
@@ -776,6 +781,13 @@ def update(self, instance, validated_data):
                 stafflist = super().update(instance, validated_data)
                 if topics_data is not None:
                     stafflist.topics.set(CourseTopic.objects.filter(id__in=topics_data))
+                if (
+                    instance.items.exists()
+                    and instance.privacy_level == PrivacyLevel.public.value
+                ):
+                    upsert_staff_list(stafflist.id)
+                else:
+                    delete_staff_list(stafflist)
```
ChristopherChudzicki, line 790: This initially confused me...I did not pay enough attention to the imports, so my I thought "Why are we deleting stafflists on update if they are not public?". Then from imports I saw this isn't deleting the stafflist record, just removing it from search.

Do you think it's worth importing the `search.task_helpers` so the usage is `search.task_helpers. delete_staff_list()` and more clearly search related? Or some other solution.

Anyway...not a big deal!t...the IDE hover-docstrings alleviated my confusion pretty quickly. But if you think of some easy resolution, 👍

ChristopherChudzicki (approved):
👍 Looks good to me.

Thanks for help testing... To get `FEATURE_USER_LIST_SEARCH` to work locally, I had to manually edit the docker-compose file since that specifies `OPEN_DISCUSSIONS_FEATURES_DEFAULT='True'` as part of "environment", which takes precedence over "env_file".

course_catalog/serializers.py
mbertrand: Good idea, I'll make a new issue to rename the `search.task_helpers` functions to be more obvious in general about what they do (like `delete_userlist` -> `userlist_index_delete` and `upsert_userlist` -> `userlist_index_upsert`)

PR #3899, opened 2023-03-24
Run edx content file tasks weekly in a separate queue

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to #3860 

#### What's this PR do?
Reduces the celery schedule of edx contenfile ingestion tasks from daily to weekly, and runs them in a separate queue, which is only processed by the performance worker or 2x workers.

Run the following, it should complete successfully:
```
manage.py backpopulate_edx_files
```

Edx course archives are saved to S3 every day with distinct daily keys, even if nothing changes.  Ideally, only courses with new/updated content should be archived to S3, and then subsequently processed by open-discussions.  For now, this PR reduces the frequency of those processing tasks.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": null,
     "lines": null
   },
-  "generated_at": "2020-04-27T20:25:55Z",
+  "generated_at": "2023-03-14T22:27:41Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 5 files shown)


JenniWhitman (approved):
Even without the fix, I can see things aren't getting stuck behind the content tasks while they spin out.. but once I fixed those also, things went more smoothly.

Issue #3761, opened 2022-10-04
Implement docs for public New Open API


mbertrand:
Generate a basic set of documentation for our current course catalog API using drf-spectacular

#### Acceptance Criteria:

- [ ] TBD

#### Out of Scope

- Adding to said documentation much
- styling said documentation

PR #3900, opened 2023-03-27
Jw/drf spectacular

JenniWhitman:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
- [x] Settings
  - [x] New settings are documented and present in `settings.py`

closes https://github.com/mitodl/open-discussions/issues/3761

#### What's this PR do?
This PR implements, at a basic level, openAPI through the use of the drf-spectacular. 

**UI**
This package includes "plug-n-play-esque" routes for two UI options - Swagger UI and Redoc. Both are included in this PR, both are safe to have on together. We can turn one or the other on or off with a quick PR once we decide which we prefer.  I have included screenshots at the bottom.

**OpenAPI spec**
This package will also generate, on the fly, a `.yaml` representation of our API in the OpenAPI specification which would allow for both documentation and code generation of our API.

##### Implementation
This PR installs the `drf-spectacular` package, applies the most basic of settings, gets routes in order to display the schema ONLY on our public, course catalog API, and offers the two available UI routes as well.

do a full docker-compose build, then up. navigate to /api/v0/schema and your browser should download the openAPI yaml file - this is the openAPI definition of our API. 
navigate to /api/v0/schema/swagger-ui/ and you should see a swagger-ui representation of our API
navigate to /api/v0/schema/redoc/ and you should see a redoc representation of our API

Test that userlists CRUD functionality has not been affected by the serializer change.

There are two concerns here!

**Concern 1 - package installation**
There is a non-essential package (looks like it's essential to their testing, but not to our running of the package) that is misbehaving. I have removed it from requirements manually and the install works now (thanks @mbertrand ). `jsonschema` was requiring functools32 which is the python 2 version of `functools` for some odd reason. Removal of the package works just fine, just calling this out as it's a non-standard installation TECHNICALLY although I don't see any issues.

**Concern 2 - bad serializer**
There was a serializer, [GenericForeignKeyFieldSerializer](https://github.com/mitodl/open-discussions/blob/99b693614dab1969189866d1920caec3d9605caa/course_catalog/serializers.py#L54) which was marked as a model serializer, but has no model Meta property.  After looking for a reason, I found none, and tested that lists still work with this modified - and they do. All tests still pass as well, so I have made it a vanilla serializer as I'm not sure what it gained from being a model serializer as it wasn't even using the model property.

This PR does a little, but allows a lot. I'll highlight a few things here, but as we develop, I can answer/talk about more.

**What if there's not enough information or the information looks off?**
drf-spectacular has an [extend-schema decorator ](https://drf-spectacular.readthedocs.io/en/latest/customization.html#step-2-extend-schema)we can use to add more - or if it's picking up on the wrong information, we can send it in a better direction.

**But what about all the words?**
This will bring over what's in docstrings. Typical docs are out of date the second code changes - this couples our docs and code changes and halts a lot of that.

**But what if we want to write more and don't want it all in our docstrings?**
EZ PZ - we can easily write functions in other places that are then inherited into the docs. There are a number of ways to do this through the use of that same decorator. For now, I've left all this out. 

Swagger UI
Swagger UI

Redoc
Redoc UI


JenniWhitman:
> Looks great! Just a couple initial thoughts for now.
> 
> * If this is meant for public consumption, might be good to exclude routes that are only intended for internal use by the front-end via authenticated users.  Like favoriting/unfavoriting, user/staff list CRUD, and also webhooks.  Seems like [this should be possible](https://stackoverflow.com/questions/68943065/django-drf-spectacular-can-you-exclude-specific-paths).  But it's probably fine to add that in a later PR?

So, it's meant for public consumption, but it also shows that these are behind auth. I think, seeing as the whole thing is open source - people will be able to tell those endpoints are there. I'd rather include them with the information that they're unavailable, then not include them and have people wonder why teh docs for the API are incomplete and ... try to fix or investigate something that's not broken because we omitted it. If that makes sense?

> * The requirements issue might pop up again whenever we make a change to `requirements.in` and need to rerun `pip-compile`.  But hopefully it will get fixed in a future drf-spectacular release.  Not sure what else to do at this point.
Yeah I meant to ask yesterday but I think I just found how to exclude packages from `pip-compile` so I'm going to do that and test that.

mbertrand (approved):
👍

PR #3901, opened 2023-03-27
Podcast frontpage notifications

### LINKED ISSUE #3876: Podcast frontpage notifications are inaccurate
We've seen some email notifications go out that include really old (10+ years) podcast episodes in them. It looks like the current logic sorts by `created_on` rather than `last_modified` (which corresponds to `pubDate` in the feed xml). Additionally, it is returning _all_ podcasts, including ones that aren't published anymore.

- [ ] Update the logic to sort by `last_modifed` instead of `created_on`
- [ ] Update logic to filter to `published=True`

JenniWhitman:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #3876

#### What's this PR do?
This PR changes that podcast episode logic to use the `last_modified` datetime instead of `created_on`. `last_modfied` is set during import of the podcast data to the date of `pubDate` which should be a more accurate representation of podcast releases rather than using the date we import it, `created_on`. In the current implementation, the user gets a larger amount of podcast notifications than expected.

Run daily notification task without having run one prior, should pick up all episodes. Then, you can make some new podcast episodes, make some with modified date before that notification went out (so... that datetime - 1day) and some after that went out (that datetime + any amount of time). Rerun. Should only see the second set of podcast episodes in the notification.

Really, just the logic on what gets picked up changed - from a date of creation to the date handed over by the incoming data.

### DIFF
--- a/notifications/notifiers/frontpage.py
+++ b/notifications/notifiers/frontpage.py
@@ -108,10 +108,10 @@ def _podcast_episodes_since_notification(notification):
         list of PodcastEpisode: list of podcast episodes
     """
 
-    episodes = PodcastEpisode.objects.order_by("-created_on")
+    episodes = PodcastEpisode.objects.filter(published=True).order_by("-last_modified")
 
     if notification:
-        episodes = episodes.filter(created_on__gt=notification.created_on)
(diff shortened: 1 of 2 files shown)


abeglova (approved):
LGTM

Issue #3824, opened 2023-02-06
Comment count includes spam comments


pdpinch:
### Steps to Reproduce

1. Make a post
2. Make a comment on the post
3. Mark the post as spam (requires django admin user or user with moderator role) 
4. View the post as an anonymous user

For example, see https://open.mit.edu/c/maca/56q/webinar-by-prof-mark-z-jacobson-stanford

### Expected Behavior

- The count of comments below the post should not include the count of spam comments.

### Screenshot or Screencast

![image](https://user-images.githubusercontent.com/430126/217017626-735fce37-582c-4d10-b5f8-8a878fa0b8fb.png)

rhysyngsun:
Fixing this would be best done as part of #3043 since the inclusion of spam comments in the value comes from there.

JenniWhitman:
Circling back to that idea - that would fix future spam comment counts, but all current counts would still be askew. Glad I got sidetracked by API before doing so.

To modify this using the current method of showing comment count, wouldn't allow for the fix of past comments wtihout some big script to re-count all comments.

The other option would be to change how we display comment count from a counter (current) to a .count() of not deleted or removed comments.

Currently, spam comments aren't the only ones counted. I also counts deleted comments :)

JenniWhitman:
@rhysyngsun 
Do you think it's worth putting together a PR annotating posts with something like (pardon the terrible pseudo code) count('comment__not_removed')+count('comment__not_deleted') and using that (potentially very costly) query to generate comment counts rather than the very incorrect counter?

rhysyngsun:
@JenniWhitman no, I don't think we should do anything that'll worsen performance. I think this is best left until we make the time to address the whole reddit dependency because it just patches over that pain with more pain (worse performance).

Ferdi:
We can hide the comments counter for now

PR #3904, opened 2023-03-31
Remove comment count from discussions

JenniWhitman:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3824

#### What's this PR do?
This removes the comment count from post and front pages since the number is inaccurate due to spam removed and deleted comments

Open those pages and be sure the comment count is gone and the styling isn't completely broken.

[local open](http://od.odl.local:8063/)

The issue gives context - there is a fix, but it would likely cause load issues at scale. For now, we're removing the comment count until reddit is removed.


abeglova (approved):
lgtm

PR #3914, opened 2023-04-05
More stafflist index code

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to #3893 

#### What's this PR do?
Adds some missing calls to update the stafflist index when appropriate

Similar to https://github.com/mitodl/open-discussions/pull/3893, but try these additional tests:
- Create a new public list, no items.  
- Go to http://localhost:8063/learn/search?type=stafflist and check the API search response in your network tab, the new list should not show up because it has no items
- Add an item to the list
- Wait a minute or two, search again, this time it should show up in API results
- Delete the item from the list
- Wait a bit, search again, this time it should not show up again because it's back to 0 items.


course_catalog/admin.py
```diff
@@ -122,6 +134,16 @@ class ProgramAdmin(admin.ModelAdmin):
     inlines = [ProgramItemInline, LearningResourceRunInline]
 
 
+class StaffListAdmin(admin.ModelAdmin):
```
ChristopherChudzicki, line 137: Thanks for adding this. I think it's good since we have it for the other catalog models, too.

ChristopherChudzicki (approved):
👍 Looks good to me!

PR #3923, opened 2023-04-06
Unpublish runs for edx sources if they are no longer included in API results

### LINKED ISSUE #3920: Old course runs are not unpublished
### Steps to Reproduce

Run any of the edx-related course ETL import functions, like `backpopulate_mitxonline_data`

### Expected Behavior

Runs included in the mitxonline API response will be added/updated.  Previously imported runs that are no longer included in the API response will be unpublished.

### Actual Behavior

Runs included in the mitxonline API response are added/updated, but previously imported runs that are no longer included in the API response are not unpublished.

mbertrand:
…results

#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #3920

#### What's this PR do?
- If a course/run is no longer included in the API results used to ingest edx sources (xpro, mitx, mitxonline, open learning library), then unpublish it and its content.
- Does not run `load_content_files` from within `load_run` for edx platforms because this should be done in a separate task

- Set `MIT_ONLINE_*` values to the same as RC in your .env file
- Run `backpopulate_mitxonline_data` followed by `backpopulate_mitxonline_files` if you haven't already
- Run the following query to verify you have published content files:
```python
from course_catalog.models import ContentFile
ContentFile.objects.filter(run__platform="mitxonline", published=True).count()
```

- Create a fake `mitxonline` course:
```
fake_course = Course.objects.create(platform="mitxonline", course_id="fake_course", published=True) #fake course
LearningResourceRun.objects.create(platform="mitxonline", run_id="fake_run_1", published=True, object_id=fake_course.id) # Fake run for fake course
LearningResourceRun.objects.create(platform="mitxonline", run_id="fake_run_2", published=True, object_id=<real_course_id>) #fake run for real course
```
- Run `backpopulate_mitxonline_data`.  When it is done, the above course and runs should all be `published=False`
- Run the same content query again, you should have the same number of published content files.

### DIFF
--- a/course_catalog/etl/edx_shared.py
+++ b/course_catalog/etl/edx_shared.py
@@ -81,6 +81,7 @@ def sync_edx_course_files(
             platform=platform,
             content_type=ContentType.objects.get_for_model(Course),
             object_id__in=ids,
+            published=True,
         )
         if platform == PlatformType.mitx.value:
             # Additional processing of run ids and tarfile names, because MITx data is a mess of id/file formats
--- a/search/indexing_api.py
+++ b/search/indexing_api.py
@@ -756,6 +756,8 @@ def delete_run_content_files(run_id, unpublished_only=False):
         serialize_content_file_for_bulk_deletion(content_file)
         for content_file in content_files
     )
+    if not documents:
+        return
     course = run.content_object
     delete_items(
         documents,
--- a/search/indexing_api.py
+++ b/search/indexing_api.py
@@ -752,12 +752,14 @@ def delete_run_content_files(run_id, unpublished_only=False):
     else:
         content_files = run.content_files.only("key")
 
+    if not content_files.exists():
+        return
+
     documents = (
         serialize_content_file_for_bulk_deletion(content_file)
         for content_file in content_files
     )
-    if not documents:
-        return
+
     course = run.content_object
     delete_items(
         documents,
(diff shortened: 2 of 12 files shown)


course_catalog/etl/edx_shared.py
```diff
@@ -81,6 +81,7 @@ def sync_edx_course_files(
             platform=platform,
             content_type=ContentType.objects.get_for_model(Course),
             object_id__in=ids,
+            published=True,
```
mbertrand, line 84: No point in doing this for unpublished runs

search/indexing_api.py
```diff
@@ -731,6 +731,8 @@ def delete_run_content_files(run_id, unpublished_only=False):
         serialize_content_file_for_bulk_deletion(content_file)
         for content_file in content_files
     )
+    if not documents:
+        return
```
mbertrand, line 735: Nothing to see here, move along

search/indexing_api.py
```diff
@@ -727,10 +727,14 @@ def delete_run_content_files(run_id, unpublished_only=False):
     else:
         content_files = ContentFile.objects.filter(run=run)
 
+    if not content_files.exists():
+        return
```
mbertrand, line 731: No point in continuing

JenniWhitman:
looks good, works good.

JenniWhitman (approved):
👍

PR #3936, opened 2023-04-11
Use react query for search page; fix bookmarking bug

### LINKED ISSUE #3908: Bookmark status is not updated until reload when adding items to lists
_Matt noticed this in https://github.com/mitodl/open-discussions/pull/3883#discussion_r1156312724. Because that PR is already large and a fix requires, I believe, nontrivial changes to the /search API calls, it seemed best to fix this in a separate PR._ 

### Steps to Reproduce

1. Go to `/infinite/search` as a logged-in user.
2. Pick any resource card and click the bookmark icon. Add the card to a list.

### Expected Behavior
Bookmark should turn solid.

### Actual Behavior

Bookmark stays unfilled. (But will be solid if page is refreshed.)

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#3908 

#### What's this PR do?
This PR changes `SearchPage` to use `react-query` to manage its infinite scroll. This is accomplished largely by upgrading  [`@mitodl/course-search-utils`](https://github.com/mitodl/course-search-utils)
 - Previously, we were using `useCourseSearch` course-search-utils. It managed search UI + some aspects of API requests.
 - Now we use `useSearchInputs` and `useSyncUrlAndSearch` (new exports of course-search-utils introduced in https://github.com/mitodl/course-search-utils/pull/51) to manage UI and URL syncing, and use react-query's `useInfiniteScroll` for managing the API queries.

Using `react-query` for managing the API calls has a few advantages: We get its builtin support for status info (loading, error, successful); we can let it handle storing the API results; if we ever wanted to switch to a paginated UI (as opposed to infinite scroll) that would be easy.

It also means that updates to lists / favorite status can appropriately patch the search results, which fixes #3908 

Because this includes a fairly large refactor of the search pag, we should test (1) all aspects of the search page, and (2) the bugfix for #3908.

**Test the search page:** Visit http://od.odl.local:8063/infinite/search and test things like:
- search results appear to be correct, and update when you scroll
- toggling a facet on/off automatically re-submits a query
- edit the search text and click submit; query should be correct
- check that the URL updates correctly when doing all of the above
- try the back button in your browser. Search UI (facets, text) & and search results should update
- check the network tab in dev console...Are extra search requests being made? _Should be "no"._

**Test bookmark status is updated appropriately:** (bug #3908)
As an authenticated user who has some userlists:
1. visit http://od.odl.local:8063/infinite/search
2. Click the bookmark button to add/remove result from a list. The UI should update quickly and without a reload.
3. Also add/remove item from "Favorites" (which is in the list modal, but not really a list). The UI should update quickly and without a reload.

Start by looking at the new `SearchPage.tsx` file to see how the updated course-search-utils is being used. 

### Related
https://github.com/mitodl/course-search-utils/pull/51


frontends/infinite-corridor/src/pages/SearchPage.tsx
```diff
@@ -32,94 +24,43 @@ const facetMap: FacetManifest = [
...
-  }
-}
-
 const SearchPage: React.FC = () => {
-  const [results, setSearchResults] = useState<Result[]>([])
-  const [total, setTotal] = useState(0)
-  const [completedInitialLoad, setCompletedInitialLoad] = useState(false)
-  const [requestInFlight, setRequestInFlight] = useState(false)
-  const [searchApiFailed, setSearchApiFailed] = useState(false)
-  const [aggregations, setAggregations] = useState<Aggregations>(new Map())
```
ChristopherChudzicki, line 57: None of these are necessary anymore. React query handles things like `completedInitialLoad`, `requestInFlight`, `searchApiFailed` itself. The other three are individual pieces of the search response. React query stores the responses, and we'll compute what we care about with `useMemo`.

frontends/infinite-corridor/src/pages/SearchPage.tsx
```diff
@@ -32,94 +24,43 @@ const facetMap: FacetManifest = [
...
+    activeFacets,
+    size:         pageSize,
+    allowedTypes: ALLOWED_TYPES
+  })
+  useSyncUrlAndSearch(history, search)
 
-      setAggregations(new Map(Object.entries(newResults.aggregations ?? {})))
+  const aggregations = useMemo(() => {
+    return new Map(
+      Object.entries(searchQuery.data?.pages[0].aggregations ?? {})
```
ChristopherChudzicki, line 43: **Question: ** @abeglova if two queries have same facets / text, just different `from` parameters, then they have the same `aggregations`, right? I'm just using the first page's aggregations.

ChristopherChudzicki:
I've tagged this "do not merge" just because of the search demo today and need to release course-search-utils first.

abeglova (approved):
LGTM

PR #3941, opened 2023-04-12
Checksums for content archives and resources

### LINKED ISSUE #3930: Impove efficiency of content file ETL pipelines for edx
Currently, content archives for edx are uploaded to S3 every day, even if nothing has changed, and every file within those archives is processed by the content file ETL pipelines, again even if they have not changed since the last pipeline run.

The frequency of S3 uploads are out of the scope of the open-discussions app, that is handled separately by devops via dagster.  But there may be ways to detect if an archive/archive file has changed since the last run and process it only if a change is detected.

#### Designs and Mockups
- Add a nullable `content_archive_size` field to `LearningResourceRun`.  If a matching archive for the run is found on S3, retrieve boto3's `Bucket.object.size` property.  Compare to the saved value for this field.  If the same, skip it.  Otherwise, process the archive and update the `content_archive_size`  value.

- Add a nullable `checksum` field to `ContentFile`.  After a modified archive has been unzipped, calculate the checksum of each file.  If equal to the current `checksum` value, skip it.  Otherwise, process the file and update the value.

#### Acceptance Criteria:

- [ ] Archives and archive files that have not changed in size/checksum since the last ETL pipeline run are skipped.  Those that have changed are processed.

#### Out of Scope

- Frequency of archive uploads to S3

mbertrand:
Based on https://github.com/mitodl/open-discussions/pull/3923 which should be reviewed & merged first

#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #3930 

#### What's this PR do?
- Calculates and saves checksums of edx content archives and ContentFiles within the archives.  Does not process them further if checksums match.

This shows how to test for MITx archives, but you can do MITx Online or xPro if you prefer, just swap out the relevant .env variables and management command.

- Use the RC values for `EDX_*`  settings in your .env
- If you haven't done so already, run `backpopulate_edx_data`
- Run `backpopulate_edx_files`.  It will take awhile, note the total runtime when finished.
- Run it again.  This time it should be about 10x faster.

For content archives, my initial idea was to use the S3 `object.size` property as the comparison metric to determine if anything had changed, which would avoid having to actually download the archive to figure out the checksum.  But since there's a chance, however small, that a changes in the archive might not lead to a change in size, I decided to go with a real checksum instead.  However i can be easily persuaded to go back to size instead of checksum by whoever reviews this. 😄


search/constants.py
```diff
@@ -154,3 +154,232 @@
     "Video Lectures and Slides": OCW_TYPE_LECTURE_VIDEOS,
     "Video Lectures": OCW_TYPE_LECTURE_VIDEOS,
 }
+
```
mbertrand, line 157: Moved these constants from `search/indexing_api.py` because that file exceeded 1000 lines (and this seems like a more fitting place anyway).

course_catalog/etl/utils.py
```diff
@@ -372,22 +378,24 @@ def transform_content_files(course_tarpath: str) -> Generator[dict, None, None]:
...
-            )
 
-            if tika_output is None:
-                log.info("No tika response for %s", key)
-                continue
+            existing_content = ContentFile.objects.filter(key=key, run=run).first()
+            if not existing_content or existing_content.checksum != metadata.get(
+                "checksum"
+            ):
+                tika_output = extract_text_metadata(
```
mbertrand, line 386: Only process an existing ContentFile with tika if the checksum changed.

course_catalog/etl/edx_shared.py
```diff
@@ -95,12 +100,16 @@ def sync_edx_course_files(
...
 
         if not run:
-            log.info("No %s course runs matched tarfile %s", platform, key)
             continue
         with TemporaryDirectory() as export_tempdir:
             course_tarpath = os.path.join(export_tempdir, key.split("/")[-1])
             bucket.download_file(key, course_tarpath)
+            checksum = calc_checksum(course_tarpath)
+            if run.checksum == checksum:
+                continue
```
mbertrand, line 109: Skip processing the archive if the checksum is the same as last time.

ChristopherChudzicki (approved):
Looks good to me 👍 

IMO... The checksums are better than filesize. To me it seems fairly likely that someone might change a date or fix a typo without affecting the filesize.

Issue #3922, opened 2023-04-06
offered_by shows int instead of string for API documentation


JenniWhitman:
As a user of the docs, I'd like to get an accurate representation of endpoints with custom filters in the documentation.

#### Designs and Mockups

The `offered_by` filter is displaying as int instead of string because the database field is an int that maps to an enum for the string value.

`@extend_schema` is drf-spectacular's way of over-writing default functionality or schema definitions when they've been modified by other code, as has `offered_by`.

To fix the `offered_by view, we should add `@extend_schema` as a decorator and use it to modify the schema to fit the actual functionality of the endpoint.

#### Acceptance Criteria:

- [ ] `offered_by` matches functionality and works in the "Try it" portion of documentation

#### Out of Scope

- Additional write-up on the endpoint

JenniWhitman:
After some research, I've found that drf-spectacular [doesn't always correctly identify filters of certain types](https://drf-spectacular.readthedocs.io/en/latest/drf_spectacular.html#drf_spectacular.contrib.django_filters.DjangoFilterExtension), namely the `ChoiceField` that we are using on `offered_by`. 

I've been able to rectify the type discrepancy by using `extend_schema_field`, but lost the rest of the documentation, so I am rebuilding the parameters and will get the PR out for that.

PR #3942, opened 2023-04-12
Correct issues with documentation for offered_by filter on courses endpoint

JenniWhitman:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3922

#### What's this PR do?
This PR fixes the erroneously marked `offered_by` endpoint in the `drf-spectacular` docs so that it shows the correct type and works in the "try it now" section.

This is done by using `extend_schema_field` which removes the other documentation, so I have put them back to be consistent with other filters, such as audience.

Go to `http://od.odl.local:8063/api/v0/schema/swagger-ui/#/courses/courses_list` and you should see that the type is `string` and the enum & choices should be correct and match the[ OfferedBy constant](https://github.com/mitodl/open-discussions/blob/51909e81d663e9ef6f6c517f2176c90b12704b6f/course_catalog/constants.py#L7).

Click the `Try it now` button and you should be able to filter by `offeredby` successfully now.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,11 @@
 Release Notes
 =============
 
+Version 0.212.1 (Released April 11, 2023)
+---------------
+
+- demo ui css (#3933)
+
 Version 0.212.0 (Released April 11, 2023)
 ---------------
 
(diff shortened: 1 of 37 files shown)


abeglova (approved):
lgtm

PR #3950, opened 2023-04-19
Fix profile bulk indexing

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/open-discussions/issues/3949

#### What's this PR do?
When calculating the length of a document for indexing, uses the `default=str` kwarg to force anything that can't be serialized into a string (like datetime objects) - this happens for some profiles who are channel members.

- Tests should pass
- If you remove the `default=str` option added to `json.dumps()` in this PR, the new test will fail.
- If you have users who are members of a discussion channel (requires reddit), running `manage.py recreate_index --profiles` should work on this branch (and fail on master).


abeglova (approved):
lgtm

PR #3951, opened 2023-04-19
Use single modal for "Add to list..." -> fix premature close

### LINKED ISSUE #3938: On list pages, add/remove item modal closes too early when removing items
### Steps to Reproduce
As an authenticated user:

1. At https://discussions-rc.odl.mit.edu/infinite/lists, create a list and add some items to it. (Items can be added at https://discussions-rc.odl.mit.edu/infinite/search via bookmark icon)
2. From https://discussions-rc.odl.mit.edu/infinite/lists, visit the details page for your list.
3. click bookmark icon and remove an item

### Expected Behavior

When removing the item, modal stays open. *This is the behavior on old /learn site, and seems very nice in case you accidentally removed the item.*

### Actual Behavior

When removing item, modal disappears.

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#3938 

#### What's this PR do?
This PR adds [`@ebay/nice-modal-react`](https://github.com/eBay/nice-modal-react) (see below) and uses it to manage the "Add to list" dialog. This fixes #3938 

1. Go to any list page http://od.odl.local:8063/infinite/lists/LIST_ID or favorites page `http://cac.od.odl.local:8063/infinite/lists/345` 
2. Click bookmark on any list item. Remove it from the current list. Modal should stay open. If you re-add it to the list, you should see it appear in the background (if your list is short enough).

**tldr:** I am proposing to use a new library, [`@ebay/nice-modal-react`](https://github.com/eBay/nice-modal-react), to help manage modals, especially modals that can be opened from several places in the app. It's small, simple and has no dependencies. It seems reasonably maintained. 

**Problem::** Sometimes we want to open the same modal from many different places within the application. One example of this is the "Add to list" dialog that is openable from most resource cards on several pages. Two general ways to achieve this:

1. Have a single instance of the modal that reads from some global state. When we want to open the modal, update the associated global state. *The **"Learning Resource Drawer"** is handled this way, with the relevant global state in the URL bar. But it's not always natural to put the relevant info in the URL.*
3. Have many instances of the modal, co-located with where they are openable from, so that the parent component can pass the appropriate data. *This is how the **"Add to list" dialog** is currently handled. Every resource card has its own copy. (Only one ever exists in the DOM at a time, though).* 

A downside of (2) is if the modal causes its parent to unmount, then the modal will automatically close. Which might not be desirable. That's what happening in #3938.

I'd like to propose moving some of the modals (in particular, "Add to list...") to use global state managed by [`@ebay/nice-modal-react`](https://github.com/eBay/nice-modal-react).

Things I like about this library:
1. easy to have a single modal instance opened from several places
2. modal code is still isolated (We do declare a root `<NiceModal.Provider>` once, but declaring new modals does not need to change that root declaration at all, like you would have to if each modal were its own piece of redux)
4. Good typescript integration.
5. Modals manage their own visibility (including things like "don't unmount until after the transition ends", which is a small thing but easy to mess up). 
6. UI layer agnostic... can use with any modal UI (bootstrap, mui, ant design...) and comes with convenient helpers for those three popular libraries.

The flipside of (4) is that is that the library is more imperative than most react components. But I don't think that's bad... everything besides visibility is still declartive. 

See [their readme](https://github.com/eBay/nice-modal-react#nice-modal) or code in this PR for more. 

**Re global state:** In the new portions of Open we have generally avoided "global" state, state that can be read / updated by any components. Less global state => more encapsulation of components. But some things are global: the URL state is an obvious one, and controlling modal visibility seems reasonable to me, too.

**Before:**

https://user-images.githubusercontent.com/9010790/233194711-8164b67c-3c5c-4c5d-9e48-f15d107dec12.mov

**After:**

https://user-images.githubusercontent.com/9010790/233195018-ec229340-95ae-4e45-9941-705910612cb1.mov

### DIFF
--- a/frontends/infinite-corridor/src/App.tsx
+++ b/frontends/infinite-corridor/src/App.tsx
@@ -17,6 +17,7 @@ import { QueryClientProvider, QueryClient } from "react-query"
 import Header from "./components/Header"
 import { ThemeProvider as MuiThemeProvider } from "@mui/material/styles"
 import { muiTheme } from "./libs/mui"
+import { Provider as NiceModalProvider } from "@ebay/nice-modal-react"
 
 export const BASE_URL = "/infinite"
 
@@ -42,7 +43,9 @@ const AppProviders: React.FC<AppProps & { children: React.ReactNode }> = ({
     <MuiThemeProvider theme={muiTheme}>
       <QueryClientProvider client={queryClient}>
         <HelmetProvider>
-          <Router history={history}>{children}</Router>
+          <Router history={history}>
+            <NiceModalProvider>{children}</NiceModalProvider>
+          </Router>
         </HelmetProvider>
       </QueryClientProvider>
     </MuiThemeProvider>
--- a/frontends/infinite-corridor/src/components/LearningResourceCard.spec.tsx
+++ b/frontends/infinite-corridor/src/components/LearningResourceCard.spec.tsx
@@ -1,30 +1,23 @@
 import React from "react"
+import * as NiceModal from "@ebay/nice-modal-react"
 import {
   makeLearningResource,
   makeListItemMember,
   makeSearchResult
 } from "ol-search-ui/src/factories"
-import {
-  renderWithProviders,
-  user,
-  screen,
-  expectProps,
-  within
-} from "../test-utils"
+import { renderWithProviders, user, screen, within } from "../test-utils"
 import LearningResourceCard from "./LearningResourceCard"
 import type { LearningResourceCardProps } from "./LearningResourceCard"
 import AddToListDialog from "../pages/user-lists/AddToListDialog"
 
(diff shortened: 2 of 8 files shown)


mbertrand (approved):
Works well with the new library.

mbertrand:
Just need to resolve a yarn.lock conflict

ChristopherChudzicki:
@rhysyngsun Matt reviewed this, but do you have any thoughts re using [@ebay/nice-modal-react](https://github.com/eBay/nice-modal-react) ?

rhysyngsun:
@ChristopherChudzicki that library looks good, but I think we should capture the design decisions made here somewhere other than a PR. Could you start a google document for frontend technology/design decisions and put that in there?

Issue #3957, opened 2023-04-20
Do not default to "English" when courses specify no language.


ChristopherChudzicki:
### Steps to Reproduce

1. View the resource drawer for any non-english course on Prolearn. For example, https://discussions-rc.odl.mit.edu/infinite/search?q=Innovaci%C3%B3n&resource_id=17023&resource_type=course

### Expected Behavior

Language should either be correct or not specified.

### Actual Behavior

Language is "English"

### Screenshot or Screencast

<img width="1722" alt="Screenshot 2023-04-20 at 4 25 22 PM" src="https://user-images.githubusercontent.com/9010790/233479997-2ffec27d-f8b0-4dc8-a4e1-08370fd1f5e1.png">

### Comments
When `language` is null, the frontend is defaulting to English https://github.com/mitodl/open-discussions/blob/13380978c121af6abd1d55b8acff22080c29cbd4/frontends/ol-search-ui/src/components/ExpandedLearningResourceDisplay.tsx#L221

ChristopherChudzicki:
@pdpinch @mbertrand  cc @Ferdi : We discussed fixing this by not defaulting to English and displaying nothing if no language is specified—a simple change to make.

This would affect **a lot** of resource objects. Basically only mitx courses specify a language (See table below).

IMO this is still good to do—defaulting to English can be incorrect and looks silly when the description text is clearly not English. But I wanted to give a heads up that this would change how many things are displayed.

*In table below, note that basically only `mitx` platform has non-null language. OCW has a few, mostly unpublished.*
```
total_published ,total_published_runs ,total ,platform     ,language ,model
           8190 ,                8190 , 8190 ,youtube      ,         ,video
           2687 ,                2687 , 3334 ,ocw          ,         ,course
           1295 ,                1295 , 1295 ,podcast      ,         ,podcastepisode
            180 ,                 191 , 1106 ,mitx         ,en-us    ,course
            112 ,                 186 ,  262 ,see          ,         ,course
            177 ,                 177 ,  290 ,mitpe        ,         ,course
             56 ,                 141 ,  372 ,xpro         ,         ,course
             62 ,                  63 ,   63 ,oll          ,         ,course
             18 ,                  45 ,  198 ,mitx         ,         ,course
             34 ,                  34 ,   62 ,mitxonline   ,         ,course
             30 ,                  30 ,   30 ,podcast      ,         ,podcast
             22 ,                  22 ,   22 ,xpro         ,         ,program
[...]

PR #3961, opened 2023-04-24
Do not display "English" as default language on resources

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#3957 *A course clearly in spanish displaying "English" as its language, because the course did not specify a language and "English" was our default.*

#### What's this PR do?
This PR changes our Learning Resource displays (card, drawer) to not show "English" as the default course language. If a run has no language specified, then no language row is displayed in course info section.

1. **no language specified**:  At http://od.odl.local:8063/infinite or http://od.odl.local:8063/infinite/search, click on a card title. Most resources have no language specified, so clicking any of them should be fine.
    - In network tab, check the API response for the resource details. In "runs"  property, check that no language is specified.
    - Check that no language is displayed in the learning resource drawer
2. **Language is displayed if exists on resource runs:** Only `mitx` resources have languages defined, and most of the time it's English.
    - Find some MITx courses that have language specified (see below), and check that it displays correctly in the learning resource drawer. 

To find MITx course that have a language specified, you can run `docker compose run --rm web ./manage.py shell` and then run
```python
from course_catalog.models import LearningResourceRun

languages = ["tr", "ko", "en-us", "es-es", None]
base_url = "http://cac.od.odl.local:8063/infinite"
for lang in languages:
    try:
        filter_kwrags = {
            "language": lang,
        } if lang is not None else { "language__isnull": True }
        course = next(
            run for run in
            LearningResourceRun.objects.filter(platform='mitx', **filter_kwrags, published=True)
            if getattr(run.content_object, "published")
        ).content_object
        print(f"{lang} course:")
        print(f"{base_url}?resource_id={course.id}&resource_type=course")
    except StopIteration:
        print(f"Could not find a published course with language {lang}")
        continue
```

Which should print you the URL of some resources with non-null langauges.

Screenshot 2023-04-25 at 4 38 31 PM
Screenshot 2023-04-25 at 4 38 19 PM

### DIFF
--- a/frontends/ol-search-ui/src/util.test.ts
+++ b/frontends/ol-search-ui/src/util.test.ts
@@ -1,0 +1,34 @@
+import { languageName } from "./util"
+
+describe("languageName", () => {
+  it("Returns english name for 2-digit ISO-639-1 code", () => {
+    expect(languageName("en")).toBe("English")
+    expect(languageName("fi")).toBe("Finnish")
+    expect(languageName("zh")).toBe("Chinese")
+  })
+
+  it("allows case-insensitive input", () => {
+    expect(languageName("EN")).toBe("English")
+    expect(languageName("eN")).toBe("English")
+    expect(languageName("En")).toBe("English")
+    expect(languageName("ZH")).toBe("Chinese")
+    expect(languageName("zH")).toBe("Chinese")
+  })
+
+  it("Ignores trailing country codes", () => {
+    expect(languageName("en-DE")).toBe("English")
+    expect(languageName("fi-US")).toBe("Finnish")
+    expect(languageName("zh-US")).toBe("Chinese")
+  })
+
+  it("Returns null if input is null or undefined", () => {
+    expect(languageName(undefined)).toBe(null)
+    expect(languageName(null)).toBe(null)
+  })
+
+  it("Returns null if input is not a 2-digit ISO-639-1 code", () => {
+    expect(languageName("zz")).toBe(null)
+    expect(languageName("")).toBe(null)
+    expect(languageName("eng")).toBe(null)
+  })
+})
--- a/frontends/ol-search-ui/src/util.tsx
+++ b/frontends/ol-search-ui/src/util.tsx
@@ -8,7 +8,7 @@ import {
 } from "./interfaces"
 import React, { useState, useEffect } from "react"
 import { capitalize, emptyOrNil } from "ol-util"
-import LocaleCode from "locale-code"
+import ISO6391 from "iso-639-1"
 import Decimal from "decimal.js-light"
 
 export const getImageSrc = (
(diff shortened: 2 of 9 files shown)


frontends/ol-search-ui/src/util.tsx
```diff
@@ -8,7 +8,7 @@ import {
 } from "./interfaces"
 import React, { useState, useEffect } from "react"
 import { capitalize, emptyOrNil } from "ol-util"
-import LocaleCode from "locale-code"
+import ISO6391 from "iso-639-1"
```
ChristopherChudzicki, line 11: We had been using [locale-code](https://www.npmjs.com/package/locale-code) for language names. Swap to [iso-639-1](https://www.npmjs.com/package/iso-639-1?activeTab=readme). These packages are by the same author & `language-code depends` on `iso-639-1`, so we already had it as a dependency. 

The difference is:
- `iso-639-1` expects a language code... 'en', 'fr', etc
- `locale-code` expects a locale-code, `en-US`, `en-GB`, `zh-CN`, etc

**The reason to swap is:**
- we had been appending a country code 'US' to **all** language codes. That was arbitrary and confusing, but was required to `locale-code`.

frontends/ol-search-ui/src/util.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  it("Returns null if input is null or undefined", () => {
+    expect(languageName(undefined)).toBe(null)
+    expect(languageName(null)).toBe(null)
+  })
+
+  it("Returns null if input is not a 2-digit ISO-639-1 code", () => {
+    expect(languageName("zz")).toBe(null)
+    expect(languageName("")).toBe(null)
+    expect(languageName("eng")).toBe(null)
+  })
```
ChristopherChudzicki, line 33: This whole test suite is new, but these two tests are the changed behavior. Previously they would have returned "en-US".

mbertrand (changes requested):
Works for  `../infinite/search`, but English is still showing up as the default under `../learn/search` via [frontends/open_discussions/src/components/ExpandedLearningResourceDisplay.js](https://github.com/mitodl/open-discussions/blob/f0d300bd6b480836f291d8ff5f860bfc6faeaec8/frontends/open-discussions/src/components/ExpandedLearningResourceDisplay.js#L331)

ChristopherChudzicki:
> Works for ../infinite/search, but English is still showing up as the default under ../learn/search via [frontends/open_discussions/src/components/ExpandedLearningResourceDisplay.js](https://github.com/mitodl/open-discussions/blob/f0d300bd6b480836f291d8ff5f860bfc6faeaec8/frontends/open-discussions/src/components/ExpandedLearningResourceDisplay.js#L331)

That was deliberate, though in this case the bug fix is super simple so probably worthfixing both cases.

If a less trivial issue comes up, maybe worth more of a discussion.

mbertrand (approved):
👍

Issue #3494, opened 2022-01-24
Replace Open Discussions Elasticsearch with OpenSearch


pdpinch:
Elasticsearch 6.5 has passed it's end of life. Instead of upgrading Elasticsearch, we would like to switch to a hosted OpenSearch service.

#### Plan:

- [x] investigate what changes are necessary to switch from Elasticsearch to OpenSearch
- [x] create a branch for OpenSearch testing 
- [x] upgrade from 6.3 -> 6.8
- [x] Upgrade client from 6.8 -> 7.10
    - [x] Upgrade python packages
    - [x] Test local
    - [x] Merge and test RC
    - [x] Merge and test production
- [x] Upgrade service from 6.8 -> 7.10
    - [x] Update docker-compose settings and test locally
    - [x] Once PR is approved, will merge which won't affect RC at all
    - [x] Mike will update RC, we test RC
    - [x] Once RC is proven good, update prod
- [x] Swap OpenSearch client in for Elasticsearch client
    - [x] Update python packages & test local. PR
    - [x] Upgrade on RC
    - [x] Test - this is OS client with ES service
- [x] Swap to OpenSearch service
  - [x] Change docker-compose settings and test locally
  - [x] Once PR is approved, will merge which won't affect RC at all
  - [x] Mike will update RC, we test RC
  - [x] Once RC is proven good, update prod

#### AC:

- [x] Get Production client to OpenSearch 1.0.0
- [x] Get production service to OpenSearcch 1.0.0
- [x] Pass all tests
- [x] All previous search endpoints work
  - [x] Results do not break for OCW
  - [x] Results do not break for /learn/search
  - [x] Results do not break for posts/open discussions
  - [x] Suggested results work

pdpinch:
@asadiqbal08 has done the research for this for micromasters and opened a PR for the upgrade, https://github.com/mitodl/micromasters/pull/5166. I'm not sure if he found any issues that would affect this project.

JenniWhitman:
The deploy of the es7 client to RC has shown that there are compatibility issues between the es7 client and 6 service in regards to mapping structure that were not seen in prior testing. I have a plan for moving forward:
1. I'm pulling back the 7.10 branch and modifying it with a local 6.8 client and the changes needed for type. (this morning)
2. We'll test this branch locally and push again to RC (Today/Monday)
3. Mike can then test the upgrade to 7.10 service once the 7.10 client and 6.8 service are stable. (Early next week)
[...]

PR #3974, opened 2023-05-01
Elasticsearch 7.10 upgrade

JenniWhitman:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

updates https://github.com/mitodl/open-discussions/issues/3494

#### What's this PR do?
This PR upgrades from ES 6.8 to ES 7.10 in preparation for the OpenSearch swap.

Pull the branch locally and run `docker build` and `docker up` to run the new version of Elasticsearch. 

I would then:
- run tests
- check that you can use the search API manually using postman or fetch or some tool of your choice - specifically as OCW might make requests
- make sure that pages within Open that are using the search API function (new open demo, infinite, learn, open discussions)

Pull the branch and play with it, honestly. I can run all the tests in the world and I can try to break it, but ultimately it's best if we have many people poke it from all sides.

ImprobableFarflungHapuka-max-1mb


docker-compose.yml
```diff
@@ -39,16 +39,18 @@ services:
...
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.10.0
+    container_name: elastic
     environment:
       - network.host=0.0.0.0
       - http.cors.enabled=true
       - http.cors.allow-origin=*
       - http.max_content_length=10mb
       - rest.action.multi.allow_explicit_index=false
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+      - discovery.type=single-node
```
mbertrand, line 51: I noticed that elastic shuts down without this line locally, but my guess is RC/prod isn't single-node.  Probably doesn't matter in terms of how the code works though, right?  I googled this setting and it doesn't seem to require any other changes.

mbertrand (commented):
Looks like you need to add this to `course_catalog.etl.loaders_test.mock_upsert_tasks` in order to fix the failing test (no idea why it wasn't failing before):

```
index_run_content_files=mocker.patch("search.task_helpers.index_run_content_files"),
```

docker-compose.yml
JenniWhitman: This was suggested for local (and as you noticed it's unhappy without), but prod is multiple nodes - I'm pretty sure @Ardiea handles that elsewhere? Or do I need to create yet another yaml file?

JenniWhitman:
> index_run_content_files=mocker.patch("search.task_helpers.index_run_content_files"),

Thank you - I knew there was a missing mock somewhere but was more confused how it ran prior

docker-compose.yml
Ardiea: No this won't affect anything outside of the local dev envs.

abeglova:
I also tested out the search with 7.10 seems to work fine!

mbertrand (approved):
Works great! 👍

PR #3871, opened 2023-03-09
Staff lists: backend models and API

mbertrand:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [ ] Migrations
  - [ ] Migration is backwards-compatible with current production code
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested
- [ ] Settings
  - [ ] New settings are documented and present in `app.json`
  - [ ] New settings have reasonable development defaults, if applicable

#### What are the relevant tickets?
Closes #3856 

#### What's this PR do?
(Required)

#### How should this be manually tested?
(Required)

#### Where should the reviewer start?
(Optional)

#### Any background context you want to provide?
(Optional)

#### Screenshots (if appropriate)
(Optional)

#### What GIF best describes this PR or how it makes you feel?
(Optional)

PR #3980, opened 2023-05-03
Mb/stafflist items serializer

mbertrand:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

Related to #3871 

#### What's this PR do?
Adds a `stafflists` property to learning resource objects indicating which stafflists they belong to.

Create some stafflists and add the same course to each of them (see PR # for details on how to do so).
Go to `../api/v0/courses/<course_id>` and check that there is a `stafflists` property with a list of items, one for each stafflist the course is in (assuming you are logged in as admin or a stafflist editor.  If not, the list should be empty).


course_catalog/models.py
```diff
@@ -55,6 +56,29 @@ def prefetch_list_items_for_user(self, user=None):
...
         )
 
+    def prefetch_stafflist_items_for_user(self, user=None):
+        """Prefetch stafflist_items based on the current user"""
+        if (
+            user
+            and user.is_authenticated
+            and (
+                user.is_staff
+                or user.is_superuser
```
ChristopherChudzicki, line 66: **Question:** Do you think it's worth moving this computation to some util class? Similar to `is_public_list_editor(user)`. I think we'll need the same computation in the `infinite_example/views.py` file for the frontend SETTINGS object, too. *We should change the name of that file...*.

Also... https://github.com/mitodl/open-discussions/blob/8f92b2f9f9cf932c3df9a9494ce7064a8ad563c5/open_discussions/permissions.py#L28-L38 might be useful ?

ChristopherChudzicki (approved):
👍 Thanks matt! Left one comment/suggestion.

course_catalog/models.py
mbertrand: There are helper functions but like this one, they take `request` rather than `user` as input since they are always used in the context of a view (and there's no request here).  I think it would be useful to change the input from a HttpRequest to a User (pass in request.user instead of request) but that should probably be done in a separate PR since they are used pretty extensively in permissions.

Issue #3964, opened 2023-04-25
Learning resource type filter is broken


abeglova:
### Steps to Reproduce

When you select a resource type in either learn/search or infinite/search all other resource types disappear from the filter and aggregations so you can't select a second resource type

Expected behavior:

Other learning resource types should still be visible in the filter after you select one, similar to the other filters

PR #3987, opened 2023-05-08
fix resource type facet

abeglova:
This should be tested with https://github.com/mitodl/course-search-utils/pull/52

Once that PR is merges I will update this pr to point to course-search-utils on npm and not the branch

closes https://github.com/mitodl/open-discussions/issues/3964

#### What's this PR do?
This pr updates course-search-utils so that both the selected types and the types that should be shown in the aggregation are passed to buildSearchQuery 

Run https://github.com/mitodl/open-discussions/pull/3987/files#diff-b95387fcdedc9cb7b3c27edce0e5e71aeecf6b46f42278b7e4c61f05c313796b verify that the resource type facet on both learn search and infinites search now work the same as other facets and you can select more than one resource type

Verify that open discussions channel search still works. The dropdown to select "profile" "comment" or "post" is broken on prod but should also work in that branch.

Point ocw_hugo_themes to open running this branch and run ocw search locally. Verify that it is also unaffected

### DIFF
--- a/frontends/infinite-corridor/package.json
+++ b/frontends/infinite-corridor/package.json
@@ -13,7 +13,7 @@
     "@emotion/react": "^11.9.3",
     "@emotion/styled": "^11.9.3",
     "@faker-js/faker": "^7.3.0",
-    "@mitodl/course-search-utils": "2.1.0",
+    "@mitodl/course-search-utils": "^2.2.0",
     "@mui/icons-material": "^5.8.4",
     "@mui/lab": "^5.0.0-alpha.91",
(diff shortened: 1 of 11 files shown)


mbertrand (changes requested):
When I check a learning resource facet type, the other facet checks remain visible but the results are no longer filtered by that facet selection:

Screenshot 2023-05-10 at 10 38 48 AM

mbertrand (approved):
Works great within open-discussions and w/ocw-hugo-themes

Issue #3981, opened 2023-05-04
Update React Query


ChristopherChudzicki:
We should update `react-query` to v4 (new open frontend pages use v3). 
- The [breaking changes](https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4) are fairly small; primarily the package has changed names and query keys must now be arrays. (Previously they could be strings or arrays).
- There are some helpful new features, e.g., optimistic updates can return `undefined` to [bail out of the update](https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#bailing-out-of-updates-with-setquerydata).

ChristopherChudzicki:
I'm going to do this before #3956

PR #3989, opened 2023-05-10
Update react query

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#3981 

#### What's this PR do?
This PR updates react-query to v4. See code comments for specific changes.

There should be no change in behavior. I recommend testing...
- `/infinite/search`
- `/infinite/lists` ... try creating/editing a list. Updates should be visible.
- **Definitely try sorting items:**
    - Create a learning path at `/infinite/lists` if you don't have one already. Populate it with some items via the bookmark icon on search page / home page learning resource cards
    - View the details page for the learning path. _Details page navigable from `/infinite/lists`_
    - Click "Reorder" and try changing the item order. Re-ordering should be smooth. Items **should not snap back briefly** when released.

### DIFF
--- a/frontends/infinite-corridor/src/api/fields/hooks.ts
+++ b/frontends/infinite-corridor/src/api/fields/hooks.ts
@@ -1,4 +1,4 @@
-import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
+import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query"
 import {
   FieldChannel,
   FieldChannelAppearanceForm,
--- a/frontends/infinite-corridor/src/api/fields/hooks.ts
+++ b/frontends/infinite-corridor/src/api/fields/hooks.ts
@@ -10,11 +10,11 @@ import axios from "../../libs/axios"
 import { fieldDetails } from "./urls"
 
 const useFieldsList = () => {
-  return useQuery<FieldList>(urls.fieldsList)
+  return useQuery<FieldList>([urls.fieldsList])
 }
 
 const useFieldDetails = (name: string) => {
-  return useQuery<FieldChannel>(urls.fieldDetails(name))
+  return useQuery<FieldChannel>([urls.fieldDetails(name)])
 }
 
 const editFieldChannel = async (
@@ -33,7 +33,7 @@ const useMutateField = (field: FieldChannel) => {
     },
     {
       onSuccess: () => {
-        queryClient.invalidateQueries(fieldDetails(field.name))
+        queryClient.invalidateQueries([fieldDetails(field.name)])
       }
     }
   )
--- a/frontends/infinite-corridor/src/api/fields/hooks.ts
+++ b/frontends/infinite-corridor/src/api/fields/hooks.ts
@@ -1,4 +1,4 @@
-import { useMutation, useQuery, useQueryClient } from "react-query"
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import {
   FieldChannel,
   FieldChannelAppearanceForm,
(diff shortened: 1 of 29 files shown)


frontends/infinite-corridor/src/api/fields/hooks.ts
```diff
@@ -10,11 +10,11 @@ import axios from "../../libs/axios"
 import { fieldDetails } from "./urls"
 
 const useFieldsList = () => {
-  return useQuery<FieldList>(urls.fieldsList)
+  return useQuery<FieldList>([urls.fieldsList])
```
ChristopherChudzicki, line 13: In React Query v4, query keys **must** be arrays. Previously they could be arrays or strings.

frontends/infinite-corridor/src/api/learning-resources/hooks.test.tsx
```diff
@@ -75,68 +77,29 @@ test("useResource rejects with invalid resource type", async () => {
 })
 
 test("useCreateUserList invalidates userlist Listings", async () => {
```
ChristopherChudzicki, line 79: When we perform mutations (POST, PATCH, etc) we invalidate previously fetched queries, causing them to be refetched.

Here, we were calling `useCreateUserList` (a mutation) and some queries (e.g., `useUserListsListing`) and checking that the UserList listing was re-fetched after the mutation was made.

This approach proved to be a bit fragile and was also more effort to write, so I had previously re-written a bunch of other, similar tests to just check that `invalidateQueries` is called correctly by the mutation.

Here, this a similar update is made to this test.

frontends/infinite-corridor/src/api/learning-resources/hooks.test.tsx
```diff
@@ -198,7 +164,9 @@ test.each([
     expect(spies.queryClient.invalidateQueries).toHaveBeenCalledTimes(2)
 
     // The POST response result updated resource data
-    expect(resourceResult.current.data).toEqual(modifiedAddedResource)
+    await waitFor(() => {
+      expect(resourceResult.current.data).toEqual(modifiedAddedResource)
+    })
   }
```
ChristopherChudzicki, line 170: This line was checking that the POST result from `useAddToListItems` was used to update the List resource.

Previously, the update was happening syncrhonously after the POST request resolved. In v4, it seems to happen asyncrhonously (but very fast) after the request resolves.

frontends/infinite-corridor/src/api/learning-resources/hooks.ts
```diff
@@ -494,28 +481,6 @@ const useMoveListItem = (mode: "userlist" | "stafflist") => {
       keys.staffList.id(id).itemsListing.all
   return useMutation({
     mutationFn,
-    onMutate: vars => {
```
ChristopherChudzicki, line 497: This `onMutate` was performing an optimistic update to the item order prior to API request finishing. See `ItemsListing.tsx` for more about this change.

jest-shared-setup.ts
```diff
@@ -2,7 +2,15 @@ import failOnConsole from "jest-fail-on-console"
...
 
-failOnConsole()
+failOnConsole({
+  /**
+   * This warning is removed from newer versions of React because it is
+   * misleading (not usually a memory leak).
+   *
+   * See https://github.com/reactwg/react-18/discussions/82
+   */
+  silenceMessage: msg => /Can't perform a React state update on an unmounted component/.test(msg),
```
ChristopherChudzicki, line 12: This warning has been removed in React 18 because it is misunderstood and misleading (generally not a memory leak).

React Query [recommends](https://github.com/TanStack/query/issues/3804#issuecomment-1184212985) ignoring these warnings since they are removed from react anyway (in newer versions, as mentioned).

abeglova (approved):
LGTM

Issue #3958, opened 2023-04-20
Do not execute all searches on all indexes


abeglova:
We currently execute all searches against  all indexes including comment, post, stafflist, etc
https://github.com/mitodl/open-discussions/blob/master/search/api.py#L305

Then filter to just the relevant resource type with an object_type filter. All elements in each index have the same value for object_type.

Since we have an index per resource type we can probably improve search performance by only searching the relevant resource type indexes when we do searches.

PR #3991, opened 2023-05-11
Ab/dont search all indexes for all queries

abeglova:
closes https://github.com/mitodl/open-discussions/issues/3958

#### What's this PR do?
This PR updates the search to only search the indexes relevant for the query. This should improve OCW search performance since OCW searches only need the course index.

Run this file locally verify that open learn search, open channel search, infinite search work as expected

Run ocw-hugo-themes locally with search pointed to your local open discussions running this code. Verify that ocw search also works as expected

### DIFF
--- a/search/api.py
+++ b/search/api.py
@@ -32,6 +32,8 @@ from search.constants import (
     PODCAST_TYPE,
     USER_LIST_TYPE,
     USER_PATH_TYPE,
+    VALID_OBJECT_TYPES,
+    RESOURCE_FILE_TYPE,
 )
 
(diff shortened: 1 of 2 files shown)


mbertrand (commented):
Needs a rebase, I went ahead and did that locally & just about everything works great.  One issue I encountered though is that I get 0 results back for post/comment search.  However I do get results back if I make the query directly to ES via Postman.  It doesn't seem related to this PR though because the same thing happens on master branch.  Do you get post/comment results?

mbertrand (approved):
👍

PR #4000, opened 2023-05-18
Change search conditional to fix the 7.10-related error

JenniWhitman:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

#3999 

#### What's this PR do?
This PR brings over the conditional that was erroneously left in the service, rather than client, PR.

Have master locally, index. pull this, attempt to search using teh API (either on a page or using an API post).

1. pull master, reindex
2. pull this branch, attempt search locally (using API endpoint)
3. pull the branch with teh 7 service (which will have this branch merged into it)
4. reindex local again (verify can reindex)
5. test search on 7

Start indexing on master now lol

Why did this test ok? This conditional only kicks off on the es7 service because of the structure change of the RETURN which we don't get without integration testing, so in the testing and deploy of the 7 client, we did not see it until the SERVICE was updated.

### DIFF
--- a/search/api.py
+++ b/search/api.py
@@ -499,12 +499,26 @@ def transform_results(search_result, user, department_filters):
                     user, object_type, object_id
                 )
 
+    search_result["hits"]["total"] = _transform_search_result_total_es7(search_result)
     search_result = _transform_search_results_suggest(search_result)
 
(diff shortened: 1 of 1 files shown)


abeglova (approved):
LGTM

Issue #3990, opened 2023-05-10
Use React 18 in new frontend pages of Open


ChristopherChudzicki:
Currently, we use React 16.14 in all Open frontend bundles: both the old (js+flow) and new (typescript) sections. Because the "old" and "new" portions of Open are separate bundle with separate entrypoints, we can use different React versions.

I would like to suggest that we use React 18 (the current major version, released in March 2022) in the new portions of Open. 
There are no amazing new features that immediately benefit us. However:
- There are **some** improvements that might be beneficial. E.g., React now batches state updates that occur asynchronously (like as the result of an API call or from an event handler) => fewer unnecessary rerenders.
- Several improvements to server-side rendering with React[^1]. We do not currently pursue SSR, but if we ever go down that path (for example, by using NextJS) it would be useful to use React 18. *NextJS [suggests](https://nextjs.org/docs/messages/react-version) using React 18.2 at a minimum.*

I suggest updating React now because, in general, I think it's easier to update early (when the project is smaller) than to update later when it has grown.

[^1]: One example of improved SSR is: In React 16/17, an entire page had to be rendered on the server. Now you can selectively render portions of the page on the server. So, e.g., if `/search` page is slow to load, you could send the HTML for everything except search results immediately,  then send the data when it is ready

ChristopherChudzicki:
We decided not to use React 18 for Infinite Corridor initially because we were sharing packages like `ol-util` and `ol-forms` between the "new" and "old" frontends. We did not want to deal with ensuring the shared packages would work with both versions. Since we no longer share packages between old and new, this is no longer an issue.

rhysyngsun:
I think this is a good idea - the only other reason we've held back historically on React upgrades past 16 was `enzyme`'s nonexistent support for higher React versions and we don't use that library the new infinite code.

PR #4007, opened 2023-05-23
Use React 18 in New Pages

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#3990 

#### What's this PR do?
This PR updates React to version 18 **in infinite-corridor only**. I.e., "old open" is not affected. 

Now:
```
frontends:
|-- infinite-corridor      <-- React 18 as dependency
|-- ol-forms                      ... (React is peer dependency)
|-- ol-search-ui                  ... (React is peer dependency)
|-- ol-util                       ... (React is peer dependency)
|-- ol-widgets                    ... (React is peer dependency)
|-- open-discussions       <-- React 16 as dependency
```

1. Stop docker
2. `docker compose up`; wait for installs to finish.
3. view "old open" pages and "new open" pages (/infinite, /infinite/search, /infinite/lists). They should look and function the same.

I would recommend starting with the [comment on `hoistingLimits`](https://github.com/mitodl/open-discussions/pull/4007/files#r1204357894)


frontends/open-discussions/package.json
```diff
@@ -1,5 +1,8 @@
 {
   "name": "open-discussions",
+  "installConfig": {
+    "hoistingLimits": "workspaces"
+  },
```
ChristopherChudzicki, line 5: Hoisting is a mechanism for deduplicating dependencies in a dependency tree. See [hoisting](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/) for more background. *That article is from Yarn 1; we use Yarn 3, but the concepts are the same.*

Hoisting can be problematic, particularly with peer dependencies. For example:
- workspace `open-discussions` (old open) uses React 16 + Enzyme for testing, with React 16 as a peer dependency
- workspace `infinite-corridor` is using React 18 + React Testing Library (v14) for testing, with React 18 as a peer dependency

If `@testing-library/react` gets hoisted to root node_modules, it will use the version of React in root node_modules. But is that 16 (from open-discussions) or 18 (from infinite-corridor). Not super predictable. Bad.

```
git repo
|-- package.json
|-- node_modules                A (root)
|-- frontends
    |-- infinite_corridor
        |-- package.json        <-- specifies React 18
        |-- node_mudles         B      
    |-- open_discussions
        |-- package.json        <-- specifies React 16
        |-- node_modules        C
    |-- ...others

Does React 16 go in A or C?
           18 go in A or B?

Not easily predictable.
```

### Enter `hoistingLimits`
Since workspace `open-discussions` is separate from all the other workspaces, the simplest solution is to never hoist its dependencies to root. So in the tree above, all dependencies from `open-discussions/package.json` end up in `open-discussions/node_modules`. 

It's still possible for workspace `open-discussions` to use dependencies from root `node_modules` if open-discussions is using a dependency without specifying it in the package.json. This was the case with `immer`:
- `immer` was previously specified in the root package.json...
-  `immer` was being used by workspace `open-discussions`. *Because of Node's module resolution ("keep searching in higher and higher node_modules"), that was working fine.*

So to really isolate open-discussions, we want:
1. do not hoist its dependencies to root node_modules
2. make sure it's package.json specifies its dependencies

(2) is taken care of by a linting rule.

frontends/open-discussions/.eslintrc.js
ChristopherChudzicki, line 1: We previously had both `.eslintrc` and `.eslintrc.js`. They were identical, but we only need one.

frontends/open-discussions/.eslintrc
```diff
@@ -1,8 +1,10 @@
 {
   "root": true,
   "extends": ["eslint-config-mitodl", "eslint-config-mitodl/flow", "eslint-config-mitodl/mocha"],
+  "plugins": ["import"],
   "rules": {
-      "operator-linebreak": "off"
+      "operator-linebreak": "off",
+      "import/no-extraneous-dependencies": "error"
```
ChristopherChudzicki, line 7: > Forbid the import of external modules that are not declared in the package.json's dependencies, devDependencies, optionalDependencies, peerDependencies, or bundledDependencies. **The closest parent package.json will be used**.

(Emphasis mine)

frontends/open-discussions/package.json
```diff
@@ -31,8 +36,18 @@
...
     "@mitodl/mdl-react-components": "0.0.3",
+    "@mui/icons-material": "^5.8.4",
+    "@mui/material": "^5.9.0",
     "@reduxjs/toolkit": "1.4.0",
+    "@rmwc/checkbox": "^5.7.2",
+    "@rmwc/dialog": "^5.7.2",
+    "@rmwc/drawer": "^5.7.2",
+    "@rmwc/menu": "^5.7.2",
+    "@rmwc/snackbar": "^5.7.2",
+    "@rmwc/theme": "^5.7.2",
```
ChristopherChudzicki, line 47: Previously we were getting these (and `material-components-web`) from `rmwc`; since `rmwc` was never being used directly, installing these is equivalent, and makes `import/no-extraneous-dependencies` happy.

frontends/infinite-corridor/package.json
```diff
@@ -20,13 +20,14 @@
...
     "@types/lodash": "^4.14.182",
-    "@types/react": "^18.0.13",
-    "@types/react-dom": "^18.0.5",
+    "@types/react": "^18.2.6",
+    "@types/react-dom": "^18.2.4",
     "@types/react-router": "^5.1.18",
     "@types/react-router-dom": "^5.3.3",
+    "autoprefixer": "^10.4.14",
     "axios": "^0.27.2",
+    "bootstrap": "^5.2.3",
```
ChristopherChudzicki, line 30: We were using this before, it just wasn't explicitly listed as a dependency.

We aren't using it very much: only for utility classes, and only in a few places. We may want to remove it.

frontends/infinite-corridor/package.json
```diff
@@ -35,14 +36,17 @@
...
-    "react": "^16.14",
-    "react-dom": "^16.14",
+    "react": "^18.2",
+    "react-dom": "^18.2",
+    "react-infinite-scroller": "^1.2.6",
     "react-router": "^5.3.3",
     "react-router-dom": "^5.1.2",
-    "rmwc": "5.7.1",
+    "sass": "^1.62.1",
+    "sass-loader": "^13.3.0",
```
ChristopherChudzicki, line 45: `rmwc` was a mistake and is not used in infinite-corridor.

`sass`, `sass-loader`, and the other "new" dependencies, are not really new: we were using them before, but they were not explicitly listed by infinite-corridor. (Workspace `open-discussions` was using them, they got hoisted to root, which enabled their use in infinite-corridor. This PR stops hoisting of open-discussions's dependencies, so now they *need* to be explicitly listed.

frontends/infinite-corridor/src/api/learning-resources/favorites.test.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react"
-import { renderHook } from "@testing-library/react-hooks/dom"
+import { renderHook, waitFor } from "@testing-library/react"
```
ChristopherChudzicki, line 2: Updating react required updating `@testing-library/react`; The new version of `@testing-library/react` includes `renderHook` as part of the main library, so we don't need `@testing-library/react-hooks` anymore

The `renderHook` API changed very slightly—it no longer returns `waitFor` helper.

frontends/infinite-corridor/src/test-utils/index.tsx
```diff
@@ -58,34 +58,39 @@ const renderWithProviders = (
...
  * @param fc the mock or spied upon functional component
  * @param partialProps an object of props
- * @param nthCall Optional index. If specified, then asserts that particular
- *  call occurred with given props.
  */
 const expectProps = (
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   fc: jest.Mock<any, any>,
-  partialProps: unknown,
-  nthCall?: number
```
ChristopherChudzicki, line 72: I removed the `nthCall` argument. It doesn't make much sense when running in StrictMode (since StrictMode calls component render functions multiple times).

Additionally, we almost always used it with `nthCall = -1`. That value ("previous call") still makes sense, so now `expectLastProps` exists.

mbertrand (approved):
LGTM, everything still seems to work

Issue #4004, opened 2023-05-22
Don't create a separate task for elasticsearch upserts/deindexes


abeglova:
### Steps to Reproduce

After a resource is updated, open discussions updates the elasticsearch index with the changes.

### Expected Behavior

The index should be updated right away

### Actual Behavior

The a new celery job is created to update the index. The job is placed at the back of the celery queue. If the change is a resource being unpublished, the resource remains in the index but clicking it brings up an empty card until the job runs.

PR #4009, opened 2023-05-25
dont start extra tasks for index updates

abeglova:
closes https://github.com/mitodl/open-discussions/issues/4004

#### What's this PR do?
We had an issue where an ocw course was deleted and it took most of a day for the change to propagate to the search index.  Also for some of that time the course was already removed from the open database so it came up in /learn/search and /infinite/search but opened to a blank drawer. Not starting separate task for reindexing will improve the speed that things are indexed

Run some of the backpopulate_ tasks in course_catalog/management/ and verify that you don't see errors and the index is updating. 

Make a new post or comment in disscussions and verify that the new post is available in the  /search page

### DIFF
--- a/channels/api.py
+++ b/channels/api.py
@@ -84,8 +84,8 @@ from notifications.models import (
     FREQUENCY_IMMEDIATE,
 )
 from open_discussions.utils import now_in_utc
-from search import task_helpers as search_task_helpers
-from search.task_helpers import reddit_object_persist
+from search import search_index_helpers
+from search.search_index_helpers import reddit_object_persist
 from widgets.models import WidgetList
 
 USER_AGENT = "MIT-Open: {version}"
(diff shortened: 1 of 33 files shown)


mbertrand (commented):
Works great locally, one concern though is that by calling the task immediately, any task decorator kwargs are ignored, for example lots of the index tasks use these:

```
PARTIAL_UPDATE_TASK_SETTINGS = dict(
    autoretry_for=(NotFoundError,),
    retry_kwargs={"max_retries": 5},
    default_retry_delay=2,
)

@app.task(**PARTIAL_UPDATE_TASK_SETTINGS)
def upsert_content_file(file_id):
    ....
```

Not sure how much of an impact this might have in production.  What do you think?

abeglova:
So maybe we run it immediately, then start a task if something breaks?

abeglova:
I think I thought of another potential issue - this change will prevent a task updating a batch of learning resources from completing if open search throws an error for a single item

mbertrand (approved):
👍

PR #4014, opened 2023-05-31
OpenSeach Local Service

JenniWhitman:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

updates https://github.com/mitodl/open-discussions/issues/3494

#### What's this PR do?
This PR changes the local docker-compose and ci files to use OpenSearch rather than Elasticsearch to match production once Mike updates the service on prod.

Pull local, `docker compose build` - this will not be a migration or upgrade to your current container, but rather will build a new one. Once that is done, make sure you can populate and index data and files.

You CAN do a local export and import to protect your data if you would like to preserve it, but this is not necessary to test. 

This is the last (local) step in the migration off of Elasticsearch.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -45,7 +45,7 @@ services:
       - cluster.name=opensearch-cluster
       - node.name=opensearch-node1
       - bootstrap.memory_lock=true # along with the memlock settings below, disables swapping
-      - "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m" # Set min and max JVM heap sizes to at least 50% of system RAM
+      - "OPENSEARCH_JAVA_OPTS=-Xms1024m -Xmx1024m" # Set min and max JVM heap sizes to at least 50% of system RAM
       - "DISABLE_INSTALL_DEMO_CONFIG=true" # disables execution of install_demo_configuration.sh bundled with security plugin, which installs demo certificates and security configurations to OpenSearch
       - "DISABLE_SECURITY_PLUGIN=true" # disables security plugin entirely in OpenSearch by setting plugins.security.disabled: true in opensearch.yml
       - "discovery.type=single-node" # disables bootstrap checks that are enabled when network.host is set to a non-loopback address
@@ -60,7 +60,6 @@ services:
       - opensearch-data1:/usr/share/opensearch/data
     ports:
       - 9100:9200
-      - 9600:9600 # required for Performance Analyzer
 
   nginx:
     image: nginx:1.9.5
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -116,3 +127,6 @@ services:
     entrypoint: java -jar tika-server-1.23.jar -h 0.0.0.0 -spawnChild
     ports:
       - "9998:9998"
+
+volumes:
+  opensearch-data1:
```
abeglova, line 132: Should this be an empty key?

abeglova (changes requested):
Ok I was able to run the container and update the index. Just one question about the docker file

docker-compose.yml
JenniWhitman: Yes - the mapping is declared in the `opensearch-node1` definition:
```- opensearch-data1:/usr/share/opensearch/data```

The volumes here is referring back to that when docker compose builds the... build?

Comment (Ardiea) [reply docker-compose.yml:132]: It looks weird but it is okay. It is really:

```
volumes:
  opensearch-data1: {}
```

https://docs.docker.com/compose/compose-file/compose-file-v3/#volumes

File: docker-compose.yml [hunk 2]
```diff
@@ -38,18 +38,28 @@ services:
...
-      - http.max_content_length=10mb
-      - rest.action.multi.allow_explicit_index=false
-      - "ES_JAVA_OPTS=-Xms1024m -Xmx1024m"
-      - discovery.type=single-node
+      - cluster.name=opensearch-cluster
+      - node.name=opensearch-node1
+      - bootstrap.memory_lock=true # along with the memlock settings below, disables swapping
+      - "OPENSEARCH_JAVA_OPTS=-Xms1024m -Xmx1024m" # Set min and max JVM heap sizes to at least 50% of system RAM
+      - "DISABLE_INSTALL_DEMO_CONFIG=true" # disables execution of install_demo_configuration.sh bundled with security plugin, which installs demo certificates and security configurations to OpenSearch
+      - "DISABLE_SECURITY_PLUGIN=true" # disables security plugin entirely in OpenSearch by setting plugins.security.disabled: true in opensearch.yml
```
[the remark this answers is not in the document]
JenniWhitman: I meant to respond to this earlier, I was working on getting this, and will toss something in. Likely in a separate PR to avoid holding this up since prod and local are running difference services as of now.

docker-compose.yml
abeglova: Ok sounds good!

abeglova (approved):
LGTM

PR #4015, opened 2023-05-31
Update Typescript

### LINKED ISSUE #4011: Use Typescript 5
Typescript 5 came out in March. We are on 4.9.x.

It would be nice to keep that up-to-date.

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #4011 

#### What's this PR do?
This PR:
- updates us to Typescript 5 (from 4); TS 5 was released a few months ago and includes potentially helpful improvements like support for `exports` fields in package.json (which specify what subset of the package's files are available for use by the outside world).
- start typecheck test files

1. Install new dependences (either re `up` your docker, or manually `docker compose exec watch yarn install`)
2. View various pages within `/infinite` and make sure they still function.

https://github.com/mitodl/open-discussions/pull/4015#discussion_r1211856905

### DIFF
--- a/frontends/ol-forms/src/components/SelectField.tsx
+++ b/frontends/ol-forms/src/components/SelectField.tsx
@@ -14,7 +14,7 @@ export interface Option {
 
 interface Props {
   name: string
-  value: null | undefined | string | string[]
+  value?: null | string | string[]
   onChange: (event: ChangeEvent<HTMLSelectElement>) => void
   multiple?: boolean
   options: Array<string | Option>
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -20,8 +20,4 @@
     "resolveJsonModule": true,
     "preserveWatchOutput": true,
   },
-  "exclude": [
-    "frontends/*/**/*.test.ts",
-    "frontends/*/**/*.test.tsx"
-  ]
 }
--- a/frontends/infinite-corridor/src/api/learning-resources/favorites.test.tsx
+++ b/frontends/infinite-corridor/src/api/learning-resources/favorites.test.tsx
@@ -27,7 +27,7 @@ const setup = () => {
     }
   })
 
-  const wrapper = ({ children }) => (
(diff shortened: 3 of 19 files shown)


frontends/ol-forms/src/components/SelectField.tsx
```diff
@@ -14,7 +14,7 @@ export interface Option {
 
 interface Props {
   name: string
-  value: null | undefined | string | string[]
+  value?: null | string | string[]
```
ChristopherChudzicki, line 17: `value: undefined ... ` means that `value` can be explicitly assigned `undefined`: `{ value: undefined }`, but does NOT accept missing properties (like `{}`).

So when a property is optional, we generally want `propName?: PropType`. 

Javascript... :sigh:

tsconfig.json
```diff
@@ -13,15 +13,11 @@
...
     "esModuleInterop": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
     "preserveWatchOutput": true,
   },
-  "exclude": [
-    "frontends/*/**/*.test.ts",
-    "frontends/*/**/*.test.tsx"
-  ]
```
ChristopherChudzicki, line 26: Originally, we compiled TS to JS use `tsc` (the typescript compiler) and stored the JS files on disk prior to building the app via webpack. We excluded test files intentionally since they did not need to be transpiled.

Now we do not actually transpile any TS to JS prior to building webpack, so I see no reason to exclude them from our normal tsconfig. Consequences:
- TS files are now typechecked
- test files use the same tsonfig as our main files; previously, I am not entirely sure what tsconfig it was using for test files. Possibly some default from `ts-node`.

    It's nice to use the same config for test+regular files because some settings (like `moduleResolution`) are important to have the same.


*Alternatively, we could have different configs for test & non-test files that extend the same base config; but since we are bundling the app, I don't think there's any reason to use two separate configs.*

PR #4022, opened 2023-06-02
Resolve Github Action Deprecation Warnings

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#4020 

#### What's this PR do?
This PR updates our github actions to resolve some [deprecation warnings](https://github.com/mitodl/open-discussions/actions/runs/5134607216) we have been getting.

- Observe that the tests are running and passing on CI, both JS and Python
- Observe that the cache is used... See the "pip install" and "yarn install" steps.

**NOTE:** The caches will only be used on the *second* run of github actions one this PR. From [About caching workflow dependencies](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#about-caching-workflow-dependencies):
> Workflow runs can restore caches created in either the current branch or the default branch (usually main)

Since the caching is different for this branch vs master, the first run of actions did not hit the cache. The second run of actions does hit the cache.

I expect that once this is part of the default branch, PRs will hit the new cache on their first run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,27 +104,14 @@ jobs:
   javascript-tests:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-
-      - name: Set up NodeJS
-        uses: actions/setup-node@v2-beta
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
-          node-version: 16.15.1
+          node-version: '16.15.1'
+          cache: 'yarn'
 
       - name: Setup environment
         run: sudo apt-get install libelf1
-
-      - name: Get yarn cache directory path
-        id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
-
-      - uses: actions/cache@v1
-        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -41,26 +41,21 @@ jobs:
...
         run: cat Aptfile | sudo xargs apt-get install
 
-      - name: Set up Python
-        uses: actions/setup-python@v2
+      - uses: actions/setup-python@v4
         with:
-          python-version: "3.9.13"
-
-      - id: cache
-        uses: actions/cache@v1
```
ChristopherChudzicki, line 58: We no longer need to use `actions/cache` directly. Both `actions/setup-python` and `actions/setup-node` support caching of pip and yarn dependencies.

In the case of NodeJS dependencies, caching means "no need to download the package". It still needs to be copied from the package manager's global cache into `node_modules`.

JenniWhitman (approved):
Approving per discovery and discussion in slack.

ChristopherChudzicki:
Re
> Approving per discovery and discussion in slack.

Summary of our slack discussion:
- Some of the CI runs on this branch took longer, but not consistently. The CI runs that did take longer often were not using the cache.
- Based on the [About caching workflow dependencies](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#about-caching-workflow-dependencies) docs, we expect cache to be used much more consistently once this is on master.
- We still have some deprecation warnings (mostly codecov) but many fewer

PR #4047, opened 2023-06-05
fix ocw course delete

abeglova:
None

#### What's this PR do?
Course run slugs for ocw courses always have format `courses/{course url substing}`. The --delete command of backpopulate_ocw_next_data currently only works if you set --course-url-substring to  `courses/{course url substing}` instead of just the course url substring. This fixes the issue.

Pick a ocw course published in your local database and search

Run 
docker compose run web ./manage.py backpopulate_ocw_next_data --course-url-substring {the url substring}  --delete

Where {the url substring} is the final url portion of the course on ocw. You can get it by clicking "Take Course" from the drawer or taking the part of the course's learning resource runs slug value that follows the  `courses/` prefix

Verify that the course is removed from search and set to unpublished


abeglova:
I checked and all ocw courses where both the run and course have published=True have slugs that start with courses/

course_catalog/management/commands/backpopulate_ocw_next_data.py
```diff
@@ -46,7 +46,8 @@ def handle(self, *args, **options):
                     )
                 )
                 runs = LearningResourceRun.objects.filter(
-                    slug=course_url_substring, platform=PlatformType.ocw.value
+                    slug="courses/{}".format(course_url_substring),
+                    platform=PlatformType.ocw.value,
```
ChristopherChudzicki, line 50: **Suggestion:** I think it would be worthwhile to update the parameter description, and possible the name.

IMO:
```diff
         parser.add_argument(
-            "--course-url-substring",
-            dest="course_url_substring",
+            "--course-name",
+            dest="course_name",
             required=False,
-            help="If set, backpopulate only courses whose urls match with this substring",
+            help="If set, backpopulate only the course with this name",
         )
```

- `name` is the OCW term for these things.
- The old text says "courses that match...". It's only ever at most 1 course, right?

ChristopherChudzicki (approved):
👍 One suggestion re docs

course_catalog/management/commands/backpopulate_ocw_next_data.py
abeglova: No it can be multiple names seperated by a comma

abeglova: I'll add that to the documentation

Issue #4010, opened 2023-05-25
Make creating new local JS packages easier


ChristopherChudzicki:
Our `frontends/` folder contains several local JS packages. But it's not easy to create a new one. Basically, the process is "copy and paste an existing package from `frontends/`, edit metadata, and remove stuff you don't need". This means that creating new frontend packages is confusing and takes a decent amount of effort. In my opinion, this has led to some bload—e.g., `ol-search-ui` is pretty big and not specifically related to the search ui.

It would be nice to have some templating command analogous to `./manage.py startapp` in django.

PR #4049, opened 2023-06-07
Local node package templates

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested *Linting changes are "tested"; I have not included any tests for the templating.*

#4010 

#### What's this PR do?
This PR adds a local node package `ol-template` that provides a template for new packages.

**Additionally:** This PR changes the linting a bit. 
- Previously, we ran linting / formatting on the `src/` directory in each package. This had two undesirable effects:
    - if there was no `src/` directory, pretter & eslint would throw an error. *That's the immediate motivation for making the change in this PR; the `ol-template` package has no `src/` directory.*
    - some configuration files (e.g., `jest.config.ts`) at package roots were not being linted / formatted.
- Now: We lint all js/ts files, except those in `build/` directories and in `node_modules/`. 

1. Enter the frontend container: `docker compose exec watch bash`.
2. Create a new local package with `yarn create-package`. After creation, you should be instructed to run `yarn install`. Go for it.
3. Add linting error (example: use `let` when `const` would do) in your local package, and save. *Depending on editor, may need to save without formatting. In VSCode, that's cmd+shift+P --> "Save without formatting".*
4. Run `yarn lint-check`. It should fail. *If the error was linting error rather than formatting. 
5. Add a formatting error (example: a semicolon) in your local package & save without formatting.
6. Run `yarn fmt-fix`. You should see the file reformatted.
7. Run `yarn workspace infinite-corridor add <your package name>`.
8. Try importing your package in `infinite-corridor`. It should work. 

This uses [`plop`](https://www.npmjs.com/package/plop) for the templating.

I considered using [copier](https://github.com/mitodl/hq/discussions/916) for this. I did experiment with Copier, but decided against using it for this because:

- the use case here is pretty simple; plop is good enough
- it seemed more natural to create frontend packages via the `watch` container, which does not have Python.
- Copier's main distinction is its ability to update files when the template and the copy are tracked in git[^1]. This doesn't seem super useful for the local node packages because:
    
    the template has 4 files (package.json, tsconfig, jest config, and index.ts). Updates to these files packages generally happen via updating the base tsconfig or jest config, OR by installing new dependencies via `yarn`. In the case of installing new dependencies via `yarn`, the package.json ended up getting git conflicts with Copier's template.  

The templates are pretty simple, so if we end up adopting copier and want to use it here, it would be easy to convert them.

[^1]: For example, Copier seems like it would be nice for setting up common Github Actions CI files, then when breaking changes happen to, say, `actions/setup-node` or `actions/setup-python`, we could update all our repos.


frontends/infinite-corridor/jest.config.ts
```diff
@@ -1,13 +1,12 @@
-import type { Config } from '@jest/types'
+import type { Config } from "@jest/types"
```
ChristopherChudzicki, line 1: This change, and all changes outside of `ol-template`, are because the list of files that is now being linted/formatted got bigger.

.eslintrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
-  extends:   ["eslint-config-mitodl", "eslint-config-mitodl/jest", "plugin:testing-library/react"],
-  plugins: ["testing-library"],
+  extends:        ["eslint-config-mitodl", "eslint-config-mitodl/jest", "plugin:testing-library/react"],
+  plugins:        ["testing-library"],
+  ignorePatterns: ["**/build/**"]
```
ChristopherChudzicki, line 4: Previously we were only linting files under `frontends/<package>/src/` (via CLI argument).

Now we lint/format all files under `frontends/<package>`, except those in node_modules and build directories.

tsconfig.json
```diff
@@ -13,7 +13,7 @@
     "declaration": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
ChristopherChudzicki, line 16: `moduleResolution: "bundler"` is new in Typescript 5. Roughly, it lets use the `"exports"` field in `package.json`, which gives us better control over what is exported from a package. 
- see https://nodejs.org/api/packages.html#package-entry-points for more on `exports`
- https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler for more. for more on `"moduleResolution": "bundler"`.

*Additionally, `"moduleResolution": "node"` is somewhat discouraged, and is a deprecation candidate for Typescript 6. Though...I'm sure it will be around for a couple more years.*

mbertrand (approved):
LGTM 👍

Issue #3864, opened 2023-03-09
Permissions in new Open frontend (handling 403s)


ChristopherChudzicki:
As a user, I'd like to be prompted to login if I need to

#### Acceptance Criteria:

- [ ] when an anonymous user visits a page they do not have access to, they should be prompted to login and subsequently re-directed to that page. (Example:  `/infinite/stafflists`, `/infinite/lists/123` where `123` is a private userlist)
- [ ] when an authenticated user visits a page they do not have access to, they should be directed to a Forbidden page. (Example: authenticated user visiting `/infinite/stafflists` who lacks `is_staff_list_editor` permission. 

#### Out of scope
- Any particular design for the "Forbidden" page. For now, something very simple will do.

ChristopherChudzicki:
In some cases, we do not know that the user lacks access to a page until after an API request responds 403. I've been thinking a bit about how we could handle this in general.

One idea:
- tell react-query to redirect to Forbidden page on any 403. see [React Query Error Handling](https://tkdodo.eu/blog/react-query-error-handling)
- forbidden page can redirect unauthenticated users to existing auth flow `https://discussions-rc.odl.mit.edu/login/?next=where-you-came-from`

PR #4054, opened 2023-06-13
New Open 403 handling

JenniWhitman:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3864

#### What's this PR do?
This PR modifies the `queryCache` of the `queryClient` that is used so that any return of a 403 response will redirect to a forbidden page if a user is logged in, or login with `?next=wheretheycamefrom` appended to the end to be handled by the existing auth flow.

Try to get to another user's userlist or log out and get to a private userlist - you should get a bad response which will trigger the flow.

The 403 page is largely unstyled as that will be handled separately.


frontends/infinite-corridor/src/libs/react-query.test.tsx
```diff
@@ -3,9 +3,11 @@ import { renderHook, waitFor } from "@testing-library/react"
 import { allowConsoleErrors } from "ol-util/src/test-utils"
 import { QueryClientProvider } from "@tanstack/react-query"
 import { useQuery } from "@tanstack/react-query"
+import { createBrowserHistory } from "history"
 import { createQueryClient } from "./react-query"
 
-const queryClient = createQueryClient()
+const browserHistory = createBrowserHistory()
+const queryClient = createQueryClient(browserHistory)
 const wrapper: React.FC<{ children: React.ReactNode }> = ({ children }) => (
```
ChristopherChudzicki, line 11: This test file probably shouldn't use a global wrapper. Right now I think history and queryClient are shared between tests in this file. _My bad..._

If you want to abstract the wrapper, my rec would be:
```ts
const getWrapper = (history: MemoryHistory) => {
  const queryClient = createQueryClient(history)
  const wrapper: React.FC<{ children: React.ReactNode }> = ({ children }) => (
  <QueryClientProvider client={queryClient}>
    <Router history={history}>
      {children}
    </Router>
  </QueryClientProvider>
}
```

Some notes:

Need to use `MemoryHistory` for tests, not BrowserHistory. Tests run in JSDOM which does not implement some browser APIs required by BrowserHistory

MemoryHistory accepts an `initialEntries` object that lets you specify initial URLs if needed.

frontends/infinite-corridor/src/pages/ForbiddenPage.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react"
+import Container from "@mui/material/Container"
+import Button from "@mui/material/Button"
+
+const ForbiddenPage: React.FC = () => {
```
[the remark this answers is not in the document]
JenniWhitman: Thank you! I Swapped over to this - I'll definitely keep this in mind. One question, though. One that might ignite a war: What is our "preferred" method of spacing. You'll see I used em, but I see us using... em, rem, px, etc all over the place and I'm not sure if we've "standardized" or not - more than willing to swap to whatever we SHOULD be using. And even more willing to read style guide if we... have one?

JenniWhitman:
@ChristopherChudzicki I swapped over. Took me a bit yesterday afternoon to wrap my brain around it for whatever reason, but once it clicked, I actually like it. I did read up on two ways of catching 404s so I am going to toss that in the other PR as well.

frontends/infinite-corridor/src/pages/ForbiddenPage.tsx
```diff
@@ -0,0 +1,26 @@
...
+import Container from "@mui/material/Container"
+import Card from "@mui/material/Card"
+import CardContent from "@mui/material/CardContent"
+import CardActions from "@mui/material/CardActions"
+import Button from "@mui/material/Button"
+
+const ForbiddenPage: React.FC = () => {
+  return (
+    <Container maxWidth="sm">
+      <Card style={{ marginTop: "1em" }}>
```
ChristopherChudzicki, line 11: It's not really relevant here, but fyi there is also `sx`: https://mui.com/system/getting-started/the-sx-prop/

`sx` is mui specific, whereas `style` is... just `style`. The main differences are (1) sx uses classes internally to apply the css, which means it can be overriden and it can use things like pseudoselectors; (2) `sx` has access to MUI's theme.

ChristopherChudzicki (commented):
Looks good other than the `useEffect` thing causing failing tests we were just discussing.

frontends/infinite-corridor/src/App.tsx
```diff
@@ -65,43 +68,70 @@ const AppProviders: React.FC<AppProps & { children: React.ReactNode }> = ({
...
+      delete state.forbidden
+    }
+
+    history.replace({ ...history.location, state: state })
+  }, [history])
+  if (
+    location.state !== undefined &&
+    "forbidden" in location.state &&
+    location.state.forbidden
+  ) {
```
ChristopherChudzicki, line 87:
```suggestion
  if (location.state?.forbidden) {
```

frontends/infinite-corridor/src/App.tsx
```diff
@@ -65,43 +68,70 @@ const AppProviders: React.FC<AppProps & { children: React.ReactNode }> = ({
...
 
+const AppRoutes: React.FC = () => {
+  const history = useHistory<AppLocationState>()
+  const location = useLocation<AppLocationState>()
+
+  useEffect(() => {
+    const state = history.location.state
+    if (state !== undefined && "forbidden" in state && state.forbidden) {
+      delete state.forbidden
+    }
```
ChristopherChudzicki, line 79: I don't think the `in` check here does anything?

Aside: In general, I avoid `in`, because it searches the prototype chain, which is not usually what one wants. (E.g., `"valueOf" in {}` is true.

ChristopherChudzicki (approved):

ChristopherChudzicki:
Looks good, two small suggestions. 👍

ChristopherChudzicki (approved):

Issue #3865, opened 2023-03-09
404s in new Open


ChristopherChudzicki:
As a user, I want to be informed when a page does not exist.

#### Acceptance Criteria:

- [ ] when visiting a page that does not exist, users should see a suitable 404 page. (This could be either an invalid regex pattern, or a valid regex pattern with a 404-failed request.)

ChristopherChudzicki:
One aspect of SPAs like Open is that the backend generally responds to all URLs with the same HTML page (200 OK), then routing and data requests happen on the frontend.

I've seen a few recommendations for how 404s (user types `lists/bad-id`) should be handled in this scenario: 

1. Do a "hard" redirect (new HTTP request) to a specific 404 page (say, `/not-found`) served by the backend. *Good: Correct HTTP response code; Bad: poor UX for user; their URL bar now says `/not-found`, rather than the page that was not found.
2. Sever-side validation of the URLs: have a backend router mirroring the frontend router.
    - We currently do this, but we don't validate URL params, so things like `open.mit.edu/dog` gets a 404 status response, but `open.mit.edu/c/fake-channel` gets a 200 OK showing 404 page. 
    - harder the more URL paths you have, but we don't have that many. 
3. Have the frontend display a 404 UI, even though the server response was 200 OK.
    - We also do this on MIT open. E.g., https://open.mit.edu/c/ocwchalkrad. Although sometimes we forget to display the 404 UI (https://open.mit.edu/c/ocwchalkrad/about)
    - I've seen recommendations to include `<meta name="robots" content="noindex" />` on such pages to search engines from indexing them. (Which would only happen if people posted bad links?). Open does not currently do this, but perhaps we should. *Example: This is how the (modern) Angular site handles its 404s, https://angular.io/fakepage*.
4. Another recommended approach is full-blown server-side rendering of the react app. This doesn't seem like something we would do any time soon.

My inclination for 'new open' is to continue with (2) and (3), but add the `<meta name="robots" content="noindex" />` to things akin to  https://open.mit.edu/c/ocwchalkrad

pdpinch:
Is there any evidence of the bad URL https://open.mit.edu/c/ocwchalkrad getting indexed by public search engines?

ChristopherChudzicki:
@pdpinch Not that I know of, though I don't really know how to look.

ChristopherChudzicki:
IMO, we probably don't want to show a 404 page **whenever** an API responds 404. For example:
[...]

PR #4089, opened 2023-06-21
New Open 404 Handling

JenniWhitman:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

closes https://github.com/mitodl/open-discussions/issues/3865

#### What's this PR do?
This PR adds a default 404 behavior to the app which will display a NotFound component or page when the resource is not found.  The `QueryCache` will accept a flag which will allow this default behavior to be overwritten or skipped on a case-by-case basis.

Attempt to go to:
- A page that does not exist (i.e. `/fake/route/that/we/do/not/have`)
- A result for a resource that does not exist locally (i.e. `/list/12345678/`)
- You can also manipulate API calls back from the API to simulate a missing item/update the js to request a missing item

#### What does this PR not do?
As of this, the default handling only is on - so you might find somewhere that has multiple API calls, only one returns a 404, but the whole page redirects. This is somewhere we should be setting the `querty.meta.hasCustomNotFoundHandler` value to true.


frontends/infinite-corridor/src/pages/NotFoundPage.tsx
```diff
@@ -0,0 +1,26 @@
...
+const NotFoundPage: React.FC = () => {
+  return (
+    <Container maxWidth="sm">
+      <Card sx={{ marginTop: "1rem" }}>
+        <CardContent>
+          <h1>404 Not Found Error</h1>
+          Resource Not Found
+        </CardContent>
+        <CardActions>
+          <Button variant="outlined" href="/infinite">
```
ChristopherChudzicki, line 17: I think there's a `HOME` const in `urls.ts`.

frontends/infinite-corridor/src/libs/react-query.test.tsx
```diff
@@ -119,3 +119,44 @@ test.each([
...
+      .mockRejectedValue({ response: { status: status } })
+
+    const { result } = renderHook(() => useQuery(["test"], { queryFn }), {
+      wrapper
+    })
+    await waitFor(() => {
+      expect(result.current.isError).toBe(true)
+    })
+    expect(history.location.pathname).toBe(startingLocation)
+    if (notFound) {
```
ChristopherChudzicki, line 152: I like to avoid conditionals in tests. If you feel like doing that here, you could parameterize the test by `state: undefined` and `state: { notFound: true }`.

ChristopherChudzicki (changes requested):

ChristopherChudzicki:
@JenniWhitman One additional comment: Most people rebase against `master` / `main` rather than merging it in, to keep the PR history fairly clean. 

It doesn't matter a whole lot since we squash PRs upon merge, but still useful.

frontends/infinite-corridor/src/pages/NotFoundPage.tsx
JenniWhitman: Added to both

frontends/infinite-corridor/src/libs/react-query.test.tsx
JenniWhitman: Got it. Thank you. I updated the test to reflect this

JenniWhitman:
> Most people rebase against `master` / `main` rather than merging it in, to keep the PR history fairly clean.

@ChristopherChudzicki Thank you - I come from the opposite for some reason, because of the squash and merge at the end. I'll swap to the other (although I might still have some workflow set up somewhere that does a merge, but hopefully I'll catch em all).

frontends/infinite-corridor/src/pages/NotFoundPage.tsx
```diff
@@ -0,0 +1,26 @@
...
+import Container from "@mui/material/Container"
+import Card from "@mui/material/Card"
+import CardContent from "@mui/material/CardContent"
+import CardActions from "@mui/material/CardActions"
+import Button from "@mui/material/Button"
+
+const NotFoundPage: React.FC = () => {
+  return (
+    <Container maxWidth="sm">
+      <Card sx={{ marginTop: "1rem" }}>
```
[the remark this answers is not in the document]
JenniWhitman: Yes yes. I just pushed what i hope is the final version of this. If it fails again will you take a look at what I broke? I finally understand why doc.qs and not find on the meta bit, though. Added tests for the other important parts as well.

frontends/infinite-corridor/src/pages/NotFoundPage.tsx
```diff
@@ -0,0 +1,32 @@
...
+          <meta name="robots" content="noindex,noarchive" />
+        </MetaTags>
+        <CardContent>
+          <h1>404 Not Found Error</h1>
+          Resource Not Found
+        </CardContent>
+        <CardActions>
+          <Button aria-label="Return Home" variant="outlined" href={HOME}>
+            Home
+          </Button>
```
ChristopherChudzicki, line 25: _Not requesting a change!_

**Comment:** The `aria-label` is fine, but IMO not necessary.
- with aria-label, screenreaders will read this as something like "Link: Return Home"
- without, "Link: Home"

"Link: Home" seems pretty intelligible to me.

frontends/infinite-corridor/src/pages/NotFoundPage.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+})
+
+test("The NotFoundPage loads with Correct Title", () => {
+  renderWithProviders(<NotFoundPage />, {})
+  screen.getByRole("heading", { name: "404 Not Found Error" })
+})
+
+test("The NotFoundPage loads with a link that directs to HomePage", () => {
+  renderWithProviders(<NotFoundPage />, {})
+  screen.getByRole("link", { name: "Return Home" })
```
ChristopherChudzicki, line 22: **Requested change:** Can you add an assertion about the `href`? (Probably in ForbiddenPage, too).

ChristopherChudzicki (approved):
Looks good and behaves as advertised.

I requested one change in a test, but assuming everything is passing, I don't think it warrants a re-review. 👍

frontends/infinite-corridor/src/pages/NotFoundPage.test.tsx
JenniWhitman: ah, fair. yes.

Issue #2292, opened 2019-10-16
Switch from tox to 


noisecapella:
As a developer, I'd like to run tests the same way on xPRO and Open. Since we don't have multiple versions of python which need support and we are running tests inside a docker container, we can use just use pytest like we do on xPRO.

#### Designs and Mockups

See https://github.com/mitodl/mitxpro/pull/44. We should do something similar here.

#### Acceptance Criteria:

 - [ ] Travis runs the tests like it did before
 - [ ] Developers can run tests with `docker-compose run web pytest` instead of `docker-compose run web tox`
 - [ ] Environment variables in `.env` do not leak into the tests. This is important to prevent accidental S3 uploads or other similar things.
- [ ] Update docs to refer to pytest

#### Out of Scope

- Not going to do

aliceriot:
will this make it so that each python related thing is under a different command? personally I am not a fan of the current setup where running `tox` does the linter, the formatter, etc before the tests run, and also prints out a whole lot of coverage-related stuff to the terminal. if we could have separate commands for each thing that would be :100: imo

aliceriot:
related issue: #1564

noisecapella:
Yes, separate commands. The general idea is to make it the same as what we do in xPRO.

PR #4119, opened 2023-06-26
Use pytest directly without tox

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #4055, #2292

#### What's this PR do?
This PR removes tox and runs pytest directly.

1. Run `docker compose exec web pytest` locally. It should pass.

The local testing environment is now a bit less clean / less consistent between developers: Running tests locally will use your local `web` container environment variables, plus some overrides in `pytest.ini`. IMO, this is fine (and is what we do in other projects). The CI env is still very consistent.


scripts/test/test_suite.sh
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 set -euf -o pipefail
 
-docker-compose run web tox
-docker-compose run watch npm run-script lint
-docker-compose run watch npm run-script flow
-docker-compose run watch npm run-script coverage
```
ChristopherChudzicki, line 7: There is not currently a coverage command.

scripts/test/python_tests.sh
ChristopherChudzicki, line 1: This is what we use in other non-tox projects, e.g., https://github.com/mitodl/ocw-studio/blob/master/scripts/test/python_tests.sh

_`black` runs separately on CI, so I didn't include it in this._

tox.ini
```diff
@@ -1,41 +0,0 @@
...
-    OPEN_DISCUSSIONS_DEFAULT_SITE_KEY=mm_test
-    OPEN_DISCUSSIONS_SECURE_SSL_REDIRECT=False
-    OPEN_DISCUSSIONS_NOTIFICATION_EMAIL_BACKEND=anymail.backends.test.EmailBackend
-    OPEN_DISCUSSIONS_BASE_URL=http://localhost:8063/
-    MAILGUN_SENDER_DOMAIN=other.fake.site
-    MAILGUN_KEY=fake_mailgun_key
-    OPENSEARCH_URL=http://localhost:9300/
-    OPENSEARCH_INDEX=testindex
-    INDEXING_API_USERNAME=mitodl
-    OCW_WEBHOOK_KEY=fake_key
```
mbertrand, line 41: Some of these aren't in the ci.yml file and/or pytest.ini file, might be good to make them all consistent?  For example `OPEN_DISCUSSIONS_NOTIFICATION_EMAIL_BACKEND` (xpro, mitxonline have the equivalent env in both ci.yml and pytest.ini)

mbertrand (changes requested):
Works great, just a couple suggestions, but since tests are passing on CI anyway they probably aren't critical

.github/workflows/ci.yml
```diff
@@ -68,14 +68,11 @@ jobs:
...
         run: ./scripts/test/stub-data.sh
-
       - name: Tests
         run: |
           export MEDIA_ROOT="$(mktemp -d)"
-          tox
+          ./scripts/test/python_tests.sh
         env:
-          DEBUG: 'False'
-          NODE_ENV: 'production'
```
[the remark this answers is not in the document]
ChristopherChudzicki: - `NODE_ENV` shouldn't affect the python tests at all, so I don't see any point in setting it.
- `DEBUG=false` was previously being set in both `tox.ini` and `ci.yaml`, with same value. Now it's set in `pytest.ini`. I don't think there's any reason to set it in both places.

ChristopherChudzicki:
@mbertrand I did start a discussion post about the `pytest-dotenv` idea I mentioned in slack: https://github.com/mitodl/hq/discussions/1836

mbertrand (approved):
👍

Issue #4120, opened 2023-06-27
Convert to use Poetry for Python dependency management


mbertrand:
Use poetry instead of relying on separate requirements.in, requirements.txt, and test_requirements.txt files

#### Designs and Mockups

See https://github.com/mitodl/odl-video-service/pull/1060

Related to https://github.com/mitodl/ol-infrastructure/issues/1578

PR #4151, opened 2023-07-05
Use poetry instead of requirements files

mbertrand:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/open-discussions/issues/4120
Includes changes from #2701 

#### What's this PR do?
Removes requirements files, adds poetry configuration, modified a couple tests that were failing after the switch (how they managed to pass before is a mystery to me).

- CI tests should pass
- Everything should work locally after running `docker-compose --no-cache build web celery` and `docker-compose up`


channels/views/posts_test.py
```diff
@@ -445,6 +446,7 @@ def test_get_post_forbidden(client, logged_in_profile):
 def test_get_post_not_found(client, logged_in_profile):
     """Get a post the user doesn't have permission to"""
     post_id = "missing"
+    PostFactory.create(post_id=post_id)
```
mbertrand, line 449: I'm surprised these two tests passed before without the additional lines.

mbertrand:
@blarghmatey Will this also require changes to the heroku configuration in order to work properly there?

blarghmatey:
It looks like we just need to add [this](https://elements.heroku.com/buildpacks/moneymeets/python-poetry-buildpack) buildpack to the app config and remove the requirements.txt and runtime.txt files.

ChristopherChudzicki (changes requested):
Looks good to me other than Tobias's comment (which is the requested change).

pyproject.toml
```diff
@@ -20,3 +20,117 @@ include_trailing_comma = true
...
+authors  = ["MIT ODL"]
+
+
+[tool.poetry.dependencies]
+python = "^3.11.4"
+akismet = "^1.1"
+attrs = "^23.1.0"
+base36 = "^0.1.1"
+beautifulsoup4 = "^4.8.2"
+betamax = "^0.8.0"
```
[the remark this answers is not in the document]
mbertrand: These are also in the requirements.txt file being replaced, instead of test_requirements.txt.  I'll move them to the dev section

app.json
```diff
@@ -19,6 +19,9 @@
     },
     {
       "url": "https://github.com/heroku/heroku-buildpack-nginx"
+    },
+    {
+      "url": "https://github.com/moneymeets/python-poetry-buildpack"
```
ChristopherChudzicki, line 24: I'm pretty certain that this needs to go before the python buildpack:
> The Python Poetry Buildpack prepares the build to be processed by a Python buildpack such as heroku/python by generating requirements.txt and runtime.txt from poetry.lock. With that said, your repo cannot have a requirements.txt, it will be exported from Poetry (for runtime.txt see below).

https://elements.heroku.com/buildpacks/moneymeets/python-poetry-buildpack

ChristopherChudzicki (changes requested):
Looks good other than the buildpack order!

Issue #4227, opened 2023-09-19
Configure Keycloak user migration plug-in


collinpreston:
# User Story
Configure the https://github.com/daniel-frak/keycloak-user-migration Keycloak plug-in to work with MIT Open Discussions.

# Description/Context
This plug-in allows for creating accounts in Keycloak when the following occurs:
1. The user enters an email and password that does not match any existing Keycloak account.
2. The email and password combination is sent from Keycloak to a client application (such as Open Discussions) in order to be validated.
3. If the client application confirms that the email and password combination matches an existing user record, an account is created for the user in Keycloak using the information from the client application's user record.

By being able to create Keycloak accounts after a user's email and password combination is verified with the client application, via the plugin, we can create a Keycloak account that matches the client application account.  This will allow users to login to the client application after we've enabled Keycloak as the authentication provider, but before we've had the opportunity to migrate their profile from the client application to Keycloak.

# Acceptance Criteria
- [ ] Functional testing ensures this works with Open Discussions.
- [ ] Issues are created for any subsequent configurations or customizations required to configure the plug-in with Open Discussions.
- [ ] Determine what happens if the client application's user record contains information, such as first name, which does not conform with a validation in Keycloak.  We must know what happens since we will be adding user info validations into Keycloak as more applications are added.  It is possible that a client application's user record contains info that is not valid for other applications.
- [ ] Documentation (https://docs.google.com/document/d/17tJ-C2EwWoSpJWZKjuhMVgsqGtyPH0IN9KakXvSKU0M/edit#heading=h.b3abnhpo8xea) is updated.

collinpreston:
Added to Keycloak via https://github.com/mitodl/ol-infrastructure/pull/1781

PR #4231, opened 2023-09-22
4227: configure keycloak user migration plug in - REST API

collinpreston:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/open-discussions/issues/4227

#### What's this PR do?
Adds a REST API endpoint (`http://od.odl.local:8063/api/v0/auth/<email>/` which can receive the GET and POST calls expected from the Keycloak plug-in described here: https://github.com/daniel-frak/keycloak-user-migration#prerequisites---rest-endpoints-in-the-legacy-system

Adds an override to `TokenAuthentication` in order to allow for the `Authorization` header to use "Bearer <YOUR_TOKEN_KEY>" instead of the OOTB "Token <YOUR_TOKEN_KEY>".

This has been tested with the plugin using the instructions provided here: https://github.com/daniel-frak/keycloak-user-migration#launching-and-configuring-the-example

1. Using this branch, run ./manage.py migrate in order to add the Auth Token models.
2. Create a token for your staff user by going to http://od.odl.local:8063/admin/authtoken/tokenproxy/ and clicking "ADD TOKEN" in the top right.
3. I completed the following steps using the Postman application.
4. This new endpoint requires authorization for a user with staff permissions in Django.  This is done by adding a bearer token header to your request that is formatted like similar to `Authorization: Bearer 956e252a-513c-48c5-92dd-bfddc364e812` but adding your token.
5. When making a GET request, such as (http://od.odl.local:8063/api/v0/auth/collinp%40mit.edu/), you should expect the following response:
```
{
    "email": "collinp@mit.edu",
    "firstName": "Collin",
    "lastName": "Preston",
    "enabled": true,
    "emailVerified": true,
    "attributes": {},
    "roles": [
        "default-roles-olapps"
    ],
    "groups": [],
    "requiredActions": []
}
```
6. When making a POST request, you will need to have the body of:
```
{
  "password": "your_users_password"
}
```
7. If `your_users_password` matches your Django user's password, then you should expect a 200 response.

### DIFF
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -266,8 +266,7 @@ def get_user_details_for_keycloak(request, email):
     if user:
         if request.method == "POST":
             body = json.loads(request.body)
-            password = body["password"]
-            if password and user[0].check_password(password):
+            if "password" in body and user[0].check_password(body["password"]):
                 return Response({}, status=status.HTTP_200_OK)
             else:
                 return Response({}, status=status.HTTP_403_FORBIDDEN)
@@ -278,6 +277,10 @@ def get_user_details_for_keycloak(request, email):
                 "lastName": user[0].last_name,
                 "enabled": True,
                 "emailVerified": True,
+                "attributes": {},
+                "roles": ["default-roles-olapps"],
+                "groups": [],
+                "requiredActions": [],
             }
             return Response(response, status=status.HTTP_200_OK)
     else:
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -36,6 +36,7 @@ from authentication.serializers import (
 from authentication.utils import load_drf_strategy
 from mail.api import render_email_templates, send_messages
 from open_discussions.authentication import BearerAuthentication
+from open_discussions.permissions import IsStaffPermission
 
 User = get_user_model()
 
@@ -242,7 +243,7 @@ class CustomDjoserAPIView(UserViewSet, ActionViewMixin):
 
 
 @api_view(["GET", "POST"])
-@permission_classes([IsAuthenticated])
+@permission_classes([IsAuthenticated, IsStaffPermission])
 @authentication_classes([BearerAuthentication])
 def get_user_details_for_keycloak(request, email):
     """
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -1,6 +1,7 @@
 """Authentication views"""
 import json
 from urllib.parse import quote
+from django.http import Http404
 
 import requests
 from anymail.message import AnymailMessage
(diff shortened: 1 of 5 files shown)


authentication/views.py
```diff
@@ -232,3 +240,50 @@ def set_password(self, request, *args, **kwargs):
...
+                "enabled": True,
+                "emailVerified": True,
+                "attributes": {},
+                "roles": ["default-roles-olapps"],
+                "groups": [],
+                "requiredActions": [],
+            }
+            return Response(response, status=status.HTTP_200_OK)
+    else:
+        return Response({}, status=status.HTTP_404_NOT_FOUND)
```
ChristopherChudzicki, line 289: **Suggestion:** 
```suggestion
        raise Http404("User not found")
```

Seems to be what we do elsewhere, e.g., https://github.com/mitodl/open-discussions/blob/c8d922415e61e707780ddc8000b062b6cd96e475/open_discussions/views.py#L128

authentication/views.py
```diff
@@ -232,3 +240,50 @@ def set_password(self, request, *args, **kwargs):
...
+            and email address, along with some additional flags specific for
+            Keycloak.
+            POST requests will respond with a 200 status if the password,
+            contained in the body, matches the user's record.  If the
+            password does not match the user's record, a 403 response is
+            provided.
+            Both requests will respond with a 404 if no Open Discussion's user
+            has an email matching the email argument.
+    """
+    user = User.objects.filter(email=email).all()
```
ChristopherChudzicki, line 266: **Suggestion:** `.first()` instead of `.all()` (and no `[0]` below).

ChristopherChudzicki (changes requested):
For `email` params, this is working good and works as expected.

Left a question about whether we need to support username.

authentication/views.py
```diff
@@ -232,3 +240,50 @@ def set_password(self, request, *args, **kwargs):
...
+
+@api_view(["GET", "POST"])
+@permission_classes([IsAuthenticated, IsStaffPermission])
+@authentication_classes([BearerAuthentication])
+def get_user_details_for_keycloak(request, email):
+    """
+    Endpoint for the Keycloak plug-in: https://github.com/daniel-frak/keycloak-user-migration.
+
+    Args:
+        email (string): The email of a user record in Open Discussions.
```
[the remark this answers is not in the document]
collinpreston: Yeah this will only support email coming from Keycloak.  The reasoning being is that each SSO client (Open Discussions, MIT Open, MITx Online) we intend to connect with this plug-in could have user's with the same username.  We don't have any way to verify that they are indeed the same user.  

So I think our best course of action is to keep the management of usernames to within their respective client applications, and just set the user's Keycloak username to their email.

PR #4238, opened 2023-10-13
Logout user from Keycloak without confirm page - Client logout flow

collinpreston:
#### Pre-Flight checklist
- [x] Testing

https://github.com/mitodl/ol-infrastructure/issues/1825

#### What's this PR do?
When you visit http://od.odl.local:8063/logout or click "Sign out" from the user menu, you will be redirected to Keycloak and automatically logged out without having to perform any additional actions.

http://od.odl.local:8063/logout should resolve to a 404 page if the user is not authenticated.

This PR requires:
1. Keycloak configured with your local environment.
2. An Open Discussions account which was registered with a Keycloak account.
3. `FEATURE_KEYCLOAK_ENABLED=True` in your .env file.

1. Log into Open Discussions.
2. Navigate to http://od.odl.local:8063/logout
3. You should expect to be redirected to Keycloak and shown a screen indicating that you have been logged out.
4. Verify that you are logged out of Keycloak and Open Discussions by visiting: https://sso-qa.odl.mit.edu/realms/olapps/account/ and http://od.odl.local:8063/admin.  You should see a login screen at both locations.  The Django Admin screen should NOT show that any message that you are currently logged.

In order to avoid the logout confirmation page in Keycloak, I needed to add the `id_token_hint` parameter to the logout redirect.  This is described here: https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout.  Specifically: "At the Logout Endpoint, the OP SHOULD ask the End-User whether to log out of the OP as well. Furthermore, the OP MUST ask the End-User this question if an id_token_hint was not provided or if the supplied ID Token does not belong to the current OP session with the RP and/or currently logged in End-User. If the End-User says "yes", then the OP MUST log out the End-User."

https://docs.google.com/document/d/17tJ-C2EwWoSpJWZKjuhMVgsqGtyPH0IN9KakXvSKU0M/edit#heading=h.9y6s0jvyqjsw

### DIFF
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -25,8 +25,9 @@ from rest_framework.views import APIView
 from rest_framework_jwt.settings import api_settings
 from social_core.backends.email import EmailAuth
 from social_django.models import UserSocialAuth
-from social_django.utils import load_backend
+from social_django.utils import load_backend, load_strategy
 
+from authentication.backends.ol_open_id_connect import OlOpenIdConnectAuth
 from authentication.serializers import (
     LoginEmailSerializer,
     LoginPasswordSerializer,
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
👍

PR #4260, opened 2024-04-10
Add OCW SP department

pt2302:
Part of https://github.com/mitodl/hq/issues/3793.

#### What's this PR do?
This PR adds a new  SP (Special Programs) constant to OCW departments, allowing for users to search by that department.

This will primarily be tested on RC; it adds a new constant value for an additional OCW department and does not impact functionality in any other way.


abeglova (approved):
👍

PR #4362, opened 2024-11-12
Validate users' email address against mailgun

rhysyngsun:
Closes https://github.com/mitodl/hq/issues/5570

#### What's this PR do?
This adds management commands to upload a list of all users to mailgun to check for their validity. If an address is invalid that user is marked as inactive.

- Copy the `MAILGUN_*` settings from CI 
- Have some normal users created and at least one user with an invalid email (e.g. `example@invalid.domainname`).
- Run `./manage.py validate_user_emails start` to upload the data to be validated
- Run `./manage.py validate_user_emails check` and you should see that it has been uploaded. It might be in progress of if you only have a few users most likely already complete.
- Run `./manage.py validate_user_emails process` and then verify that the `EmailValidation` records have been updated with the results and invalid users have been marked inactive.

NOTE: the testing libraries around `requests` don't support mocking binary file downloads so I wasn't able to write tests for that portion of the code here. Since we're really only going to run this once and it isn't user facing, manual testing should be ok.


authentication/management/commands/validate_user_emails.py
```diff
@@ -0,0 +1,123 @@
...
+class Command(BaseCommand):
+    """Verify user emails"""
+
+    help = "Verify user emails"
+
+    def add_arguments(self, parser):
+        subparsers = parser.add_subparsers(dest="subcommand", required=True)
+        start_parser = subparsers.add_parser("start")
+        start_parser.add_argument(
+            "--db-chunk-size",
```
abeglova, line 34: I tried running `./manage.py validate_user_emails start --db-chunk-size 1` and got '>' not supported between instances of 'str' and 'int'

authentication/management/commands/validate_user_emails.py
```diff
@@ -0,0 +1,123 @@
...
+            dest="db_chunk_size",
+            default=1000,
+            help="The number of record to fetch frmo the db at a time",
```
abeglova, line 37: typo

authentication/email_validation_api.py
```diff
@@ -0,0 +1,232 @@
...
+
+        batch_idx += 1
+
+
+def start_user_email_validation(
+    users,
+    list_name_base: str,
+    *,
+    db_chunk_size: int = 1000,
+    upload_batch_size: int = 20000,
```
abeglova, line 149: the default batch size is different in different functions. Can you remove the default everywhere except the management command?

authentication/email_validation_api.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def fetch_result_zip_file(url: str, temp_dir: str) -> str:
+    """Fetch the result file and write it out locally"""
+    filename = os.path.join(temp_dir, "results.json.zip")
+
+    with requests.get(url, stream=True, timeout=60 * 5) as resp:
+        resp.raise_for_status()
+        with open(filename, mode="wb") as zip_file:
+            for chunk in resp.iter_content(chunk_size=8192):
```
abeglova, line 185: Why 8192?

abeglova (changes requested):
Back to you

authentication/management/commands/validate_user_emails.py
rhysyngsun: I removed the option

rhysyngsun:
@abeglova should be good to review again

authentication/management/commands/validate_user_emails.py
```diff
@@ -0,0 +1,116 @@
...
+class Command(BaseCommand):
+    """Verify user emails"""
+
+    help = "Verify user emails"
+
+    def add_arguments(self, parser):
+        subparsers = parser.add_subparsers(dest="subcommand", required=True)
+        start_parser = subparsers.add_parser("start")
+        start_parser.add_argument(
+            "--upload-catch-size",
```
abeglova, line 34: Typo. Also you should just remove this argument

abeglova (approved):
lgtm

PR #4366, opened 2024-12-12
Only validate active user emails

rhysyngsun:
N/A

#### What's this PR do?
The command was validating inactive accounts, all of which didn't have an email address to validate since we scrubbed spammer accoutns, so we should just skip those accounts because it can be a lot to process.

Disable a `User` Account, remove any `EmailValidation` records for them, and then verify no new `EmailValidation` record is created for them when you run `./manage.py. validate_user_emails start`


abeglova (approved):
Approved!

PR #4448, opened 2026-04-21
make ocw file descriptions the way they were previously

abeglova:
None 

https://github.com/mitodl/open-discussions/pull/4445 updated the the ocw etl to use the new format for data.json files. As part of that i replaced the description field with the contents of the content field which had more detailed resource descriptions and is more frequently non-empty. This is what we plan to use for learn. Unfortunately this does not work for ocw search because 1) the content has embedded html which ocw does not display correctly and 2) long descriptions do not look good with the small cards in the ocw resource search.  

For example these are the updated search cards for 2-003sc-engineering-dynamics-fall-2011 
https://live-qa.ocw.mit.edu/search/?q=Lecture%2021%3A%20Vibration%20Isolation&type=resourcefile

And this is what it looks like on prod:
https://ocw.mit.edu/search/?q=Lecture%2021%3A%20Vibration%20Isolation&r=Lecture%20Videos&type=resourcefile

This undoes that change and uses the original description field which is populated in both the old and new data format. 

### Screenshots (if appropriate):
Set aws credentials to the values from either learn or discussions rc 
Set OCW_NEXT_LIVE_BUCKET=ocw-content-live-qa

Run `docker-compose run web ./manage.py backpopulate_ocw_next_data --course-name 2-003sc-engineering-dynamics-fall-2011  --overwrite`

From the shell run
```
ContentFile.objects.filter(title="Lecture 21: Vibration Isolation").last().description
```
should be ''
```
ContentFile.objects.filter(title="Recitation 9 Notes: Generalized Forces with Double Pendulum Example").last().description
```
should be

`This file contains information regarding recitation 9 notes: generalized forces with double pendulum example.'

Any other 2.003SC ContentFIle should have the description that is on prod currently - you can find the files at 
https://ocw.mit.edu/search/?q=2.003SC&type=resourcefile (or search the prod discussions database)

### DIFF
--- a/course_catalog/etl/ocw_next.py
+++ b/course_catalog/etl/ocw_next.py
@@ -256,7 +256,7 @@ def transform_resource(
                 )
 
     resource_data = {
-        "description": resource_data.get("content"),
+        "description": resource_data.get("description"),
         "file_type": file_type,
         "content_type": content_type,
         "url": "../" + s3_path,
--- a/course_catalog/etl/ocw_next_test.py
+++ b/course_catalog/etl/ocw_next_test.py
@@ -61,7 +61,7 @@ def test_transform_ocw_next_content_files(settings, mocker):
     assert content_data[2] == {
         "content": "TEXT",
         "content_type": "pdf",
-        "description": "This resource contains problem set 1",
+        "description": "Description of problem set 1",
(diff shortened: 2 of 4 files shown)


mbertrand (approved):
👍

PR #4457, opened 2026-07-23
Add department 21T (Theater Arts)

zawan-ila:
Part of https://github.com/mitodl/hq/issues/5616

Adds `21T` to `OCW_DEPARTMENTS` in `course_catalog/constants.py`:

- name: `Theater Arts`
- slug: `theater-arts`

Department name/slug and the department search facet are derived from this map, so 21T courses would otherwise show a blank department and be absent from the facet.

### How can this be tested?
- Will be tested on RC.
- Ingest an OCW course whose `department_number` is `21T`. Its `department_name` / `department_slug` resolve to **Theater Arts** / `theater-arts`, and it appears under the Theater Arts department facet in search.


mbertrand (approved):
👍

PR #4458, opened 2026-07-23
Rename department 21M from Music and Theater Arts to Music

zawan-ila:
Part of https://github.com/mitodl/hq/issues/5616

Renames department `21M` in `OCW_DEPARTMENTS` (`course_catalog/constants.py`):

- name: "Music and Theater Arts" → **Music**
- slug: `music-and-theater-arts` → `music`

### How can this be tested?
- Will be tested on RC
- Re-ingest/reindex an OCW course in department 21M — its `department_name` / `department_slug` and the search facet now show **Music** / `music`.

### DIFF
--- a/course_catalog/constants.py
+++ b/course_catalog/constants.py
@@ -217,7 +217,7 @@ OCW_DEPARTMENTS = {
     },
     "21H": {"slug": "history", "name": "History"},
     "21L": {"slug": "literature", "name": "Literature"},
-    "21M": {"slug": "music-and-theater-arts", "name": "Music and Theater Arts"},
+    "21M": {"slug": "music", "name": "Music"},
     "22": {"slug": "nuclear-engineering", "name": "Nuclear Science and Engineering"},
     "24": {"slug": "linguistics-and-philosophy", "name": "Linguistics and Philosophy"},
     "CC": {"slug": "concourse", "name": "Concourse"},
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
👍

pdpinch:
@zawan-ila @mbertrand all this futzing with open-discussions is getting distracting. What happens if we don't merge this? I'm willing to tell the OCW team that they will only get this department change in Learn.

zawan-ila:
> What happens if we don't merge this?

@pdpinch This change is needed to ensure that the department search facet flips from `Music and Theater Arts -> Music` in the open discussions search response. Not deploying this would mean that the discussions search response still has the `Music and Theater Arts` facet in place of the intended "Music" facet. That may be acceptable. However, since ocw-hugo-themes performs sanitization of facets where it "removes" "unknown" facets in that while they show up in the UI, clicking on them has no effect. Once https://github.com/mitodl/ocw-hugo-themes/pull/1838 is merged, "Music and Theater Arts" will become "unknown" to ocw-hugo-themes. Thus, the end result would be that we'll have "Music" courses counted in the "Music and Theater Arts" facet but clicking on that facet doesn't do anything, which is **bad**.

Note that we also have https://github.com/mitodl/open-discussions/pull/4457, which adds the 21T department in open-discussions. What happens if we don't deploy that? In that case, to-be-21T courses would not have a facet associated with them in the search UI. However, they will show up in the search results.

Note that we also have to recreate indexes in open-discussions. Otherwise, there will be duplicates in the search results after we migrate some of the 21M courses to 21T.

Overall, I think it is worth patching up the open-discussions deployment issues. It is probably nothing major, just some small issue that has cropped up due to general neglect of that project.

pdpinch:
@zawan-ila any news about this?

zawan-ila:
> any news about this?

Not yet. I'll try to wrap it up this week.

